PR #163, opened 2014-11-24
deposit: adaptation to checkbox handling change

kneczaj:
adaptation to https://github.com/inspirehep/invenio/pull/151 as `test` parameter format changed.

There is a problem with setting a custom message. I wonder how it was working before. Probably there will be a need to add a json serializer for LazyString sooner or later as described in the FIXME line above.


jmartinm:
Not applicable any more.

PR #183, opened 2015-01-14
deposit: fix non production mode RT creation

pedrogaudencio:
Fixes RT creation on non-production mode.

### DIFF
--- a/inspire/modules/workflows/tasks/submission.py
+++ b/inspire/modules/workflows/tasks/submission.py
@@ -93,9 +93,10 @@ def create_ticket(template, queue="Test"):
 
         # Trick to prepare ticket body
         body = "\n ".join([line.strip() for line in body.split("\n")])
-        rt = get_instance()
+        rt = get_instance() if cfg.get("PRODUCTION_MODE") else None
         if not rt:
             obj.log.error("No RT instance available. Skipping!")
+            obj.log.info("Ticket ignored.")
         else:
             ticket_id = rt.create_ticket(
                 Queue=rt_queue,


inspire/modules/workflows/tasks/submission.py
```diff
@@ -121,7 +124,7 @@ def _reply_ticket(obj, eng):
         from flask import render_template
 
         ticket_id = obj.extra_data.get("ticket_id", "")
-        if not ticket_id:
+        if not ticket_id or not cfg.get("PRODUCTION_MODE"):
```
jmartinm: would it be more understandable:

```python
if not ticket_id and cfg.get("PRODUCTION_MODE"):
```
?

inspire/modules/workflows/tasks/submission.py
```diff
@@ -93,9 +93,12 @@ def _create_ticket(obj, eng):
 
         # Trick to prepare ticket body
         body = "\n ".join([line.strip() for line in body.split("\n")])
-        rt = get_instance()
-        if not rt:
+        rt = None
+        if cfg.get("PRODUCTION_MODE"):
+            rt = get_instance()
+        if not rt or not cfg.get("PRODUCTION_MODE"):
```
jmartinm: Don't think we need the

```python
or not cfg.get("PRODUCTION_MODE")
```

```rt = None``` will already make this condition be true

Comment (jmartinm) [inline inspire/modules/workflows/tasks/submission.py]: Actually, shouldn't it be:

```
if not rt and cfg.get("PRODUCTION_MODE")
```

when we should get an error?

pedrogaudencio: True, I added the `or not cfg.get("PRODUCTION_MODE")` before the `rt = None` and then forgot to take it. ^^

jmartinm:
@jalavik can you have a look?

inspire/modules/workflows/tasks/submission.py
pedrogaudencio: I believe it should be `not cfg.get("PRODUCTION_MODE")`, because it should only continue if we're in production mode.

jalavik: On second thought, I am not sure we need the production mode check in `reply_ticket` as `ticket_id` would in principle be empty as `create_ticket` will not set `ticket_id` if not in production mode.

pedrogaudencio: Just checked it, you're correct. I'll amend it and it should be fine.

jalavik:
Looks good! Ready to merge and deploy.

PR #184, opened 2015-01-15
deposit: show form on refresh if not new

pedrogaudencio:
* If the form is not a fresh create instance and it was filled before, it's fully shown.

### DIFF
--- a/inspire/modules/deposit/static/js/deposit/literature_submission_form.js
+++ b/inspire/modules/deposit/static/js/deposit/literature_submission_form.js
@@ -207,6 +207,10 @@ define(function(require, exports, module) {
         .done(function(response_kb) {
           that.subject_kb = response_kb;
         });
+
+      if(!this.isFormBlank()){
+        this.showForm();
+      }
     },
 
     /*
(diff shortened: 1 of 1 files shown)


inspire/modules/deposit/static/js/deposit/literature_submission_form.js
```diff
@@ -780,7 +784,31 @@ define(function(require, exports, module) {
...
+        } else if ((type === 'checkbox' || type === 'radio') && this.checked) {
+          isBlank = false;
+          return false;
+        } else if (type === 'select-multiple' && $('#subject').val() !== null) {
+          isBlank = false;
+          return false;
+        }
+      });
+
+      return isBlank;
```
jmartinm: You can get rid of the ``isBlank`` variable as you are already returning ``false`` in the case when there is some content.

And replace this line by:

```javascript
return true;
```

pedrogaudencio: The `return false` is just to break out of the `$.each()` so it doesn't check more fields, whereas the `isBlank` is changed accordingly and by default is set to `true`.

jmartinm: I see, ok. You can also use ```return;``` to not confuse people like me ;)

pedrogaudencio: The thing is that `return` something other than `false` it breaks the iteration but it continues the each cycle, only returning `false` breaks the **each**. ;)

jmartinm:
merged 1b4f6ca

PR #186, opened 2015-01-15
deposit: allow multiple report numbers

pedrogaudencio:


inspire/modules/deposit/workflows/literature.py
```diff
@@ -376,11 +376,11 @@ def remove_exclusive_fields(fieldlist):
             metadata['collections'].extend([{'primary': "arXiv"}, {'primary': "Citeable"}])
 
         if user_report_number:
+            report_numbers = [{'primary': v['report_number']} for v in user_report_number]
             if 'report_number' in metadata and metadata['report_number']:
```
jmartinm: is ```'report_number'``` key possible?

Since you changed and now you export the key as:

```python
export_key='report_numbers',
```

If I understand correctly only the key ```'report_numbers'``` will be available?

jmartinm:
Apart from the comment, I tested it and works ok.

inspire/modules/deposit/workflows/literature.py
pedrogaudencio: I re-map it in post processing so every report number generates its MARC correctly. ;)

jmartinm: Ok, should be good.

jmartinm:
merged f31a8d3

PR #187, opened 2015-01-22
deposit: import information text improvements

pedrogaudencio:
* Corrects 'the DOI' to 'DOI'.

* Flashes import messages right above the import fields.


jmartinm:
Merged

PR #188, opened 2015-01-23
deposit: import DOI from arXiv

pedrogaudencio:


jalavik:
What would be the correct procedure to test this functionality and what is the expected behaviour?

pedrogaudencio:
1. fill the arXiv id field with an arXiv id that contains a DOI
2. the DOI field should be filled
3. it should import the data from arXiv
4. it should import the data from CrossRef
5. it should merge it and apply the priority mapping and fill the correspondent fields

jalavik:
@pedrogaudencio If you rebase on latest master you will not see the double The the problem. 

@jmartinm Otherwise this seem to work fine. :+1:

inspire/modules/deposit/static/js/deposit/literature_submission_form.js
```diff
@@ -555,31 +556,74 @@ define(function(require, exports, module) {
...
+      if (importFields.arxiv === "" && importedFields.isbn !== undefined){
+        this.$isbn_field.val(importedFields.isbn);
+        repeats = true;
       }
 
+      return repeats;
+    },
+
+    importData: function importData() {
+      var ImportTask = require("js/deposit/import_task");
```
jmartinm: this is not being used in this function?

jmartinm:
Just tried this out and seems to work nicely :)

jmartinm:
Merged

b9ca49116266432aa35b4adbdb9ea48c10667836
6ab24c652da46b0f933fde35fddd202057f6c134

PR #189, opened 2015-01-26
deposit: adapt userEXT change in literature

pedrogaudencio:


inspire/modules/deposit/workflows/literature.py
```diff
@@ -506,9 +506,11 @@ def restructure_ffts(fft):
         userid = deposition.user_id
         user = UserInfo(userid)
         email = user.info.get('email', '')
-        external_ids = UserEXT.query.filter_by(id_user=userid).all()
+        external_ids = UserEXT.query.filter_by(id=userid).all()
         sources = ["{0}{1}".format('inspire:uid:', userid)]
-        sources.extend(["{0}".format(e_id.id) for e_id in external_ids])
+        import ipdb; ipdb.set_trace()
```
jmartinm: I would prefer to not debug production

pedrogaudencio: Oops ^^

inspire/modules/deposit/workflows/literature.py
```diff
@@ -506,9 +506,10 @@ def restructure_ffts(fft):
         userid = deposition.user_id
         user = UserInfo(userid)
         email = user.info.get('email', '')
-        external_ids = UserEXT.query.filter_by(id_user=userid).all()
+        external_ids = UserEXT.query.filter_by(id=userid).all()
```
jmartinm:
```python
UserEXT.query.filter_by(id_user=userid).all()
```

inspire/modules/deposit/workflows/literature.py
```diff
@@ -506,9 +506,10 @@ def restructure_ffts(fft):
...
         sources = ["{0}{1}".format('inspire:uid:', userid)]
-        sources.extend(["{0}".format(e_id.id) for e_id in external_ids])
+        sources.extend(["{0}:{1}".format(e_id.method,
+                                        e_id.id_user) for e_id in external_ids])
```
jmartinm:
```python
sources.extend(["{0}:{1}".format(e_id.method,
                                        e_id.id) for e_id in external_ids])
```

inspire/modules/deposit/workflows/literature.py
pedrogaudencio: Hum but the keys are
```
+---------+------------------+------+-----+---------+-------+
| Field   | Type             | Null | Key | Default | Extra |
+---------+------------------+------+-----+---------+-------+
| id      | varbinary(255)   | NO   | PRI | NULL    |       |
| method  | varchar(50)      | NO   | PRI | NULL    |       |
| id_user | int(15) unsigned | NO   | MUL | NULL    |       |
+---------+------------------+------+-----+---------+-------+
```
Shouldn't we query the `id` key?

We'd have something like
```
mysql> select * from userEXT;
+----+--------+------------------------+
| id | method | id_user                |
+----+--------+------------------------+
| 1  | orcid  |    1234-1234-1234-1234 |
+----+--------+------------------------+
```
Right?

jmartinm: mysql> select * from userEXT;
+---------------------+--------+---------+
| id                  | method | id_user |
```
+---------------------+--------+---------+
| 0000-0002-3598-8536 | orcid  |      87 |
+---------------------+--------+---------+
1 row in set (0.00 sec)
```

https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/accounts/models.py#L325

you can see how the Foreign Key is defined.

jmartinm:
Merged 93bb106

PR #191, opened 2015-01-29
deposit: split publication note

pedrogaudencio:
* After the journal reference field is fetched from arXiv and the user accepts it, it's split in title, year, volume and page range.

### DIFF
--- a/inspire/modules/deposit/forms.py
+++ b/inspire/modules/deposit/forms.py
@@ -463,24 +463,28 @@ class LiteratureForm(WebDepositForm):
                                              mapper=journal_title_kb_mapper)
     )
 
-    page_range_article_id = fields.TextField(
+    journal_page = fields.TextField(
         label=_('Page Range/Article ID'),
         description=_('e.g. 1-100'),
+        export_key='page_artid',
         widget_classes="form-control" + ARTICLE_CLASS
     )
 
-    volume = fields.TextField(
+    journal_volume = fields.TextField(
         label=_('Volume'),
+        export_key='volume',
         widget_classes="form-control" + ARTICLE_CLASS
     )
 
-    year = fields.TextField(
+    journal_year = fields.TextField(
         label=_('Year'),
+        export_key='year',
         widget_classes="form-control" + ARTICLE_CLASS
     )
 
-    issue = fields.TextField(
+    journal_issue = fields.TextField(
         label=_('Issue'),
+        export_key='issue',
         widget_classes="form-control" + ARTICLE_CLASS
     )
 
@@ -601,8 +605,8 @@ class LiteratureForm(WebDepositForm):
         # ('Licenses and copyright',
         #     ['license', 'license_url'], {'classes': 'collapse'}),
         ('Journal Information',
-            ['journal_title', 'volume', 'issue', 'year',
-             'page_range_article_id']),
+            ['journal_title', 'journal_volume', 'journal_issue', 'journal_year',
+             'journal_page']),
         ('Conference Information',
             ['conf_name', 'conference_id'], {'classes': 'collapse'}),
         ('Proceedings Information (if not published in a journal)',
--- a/inspire/modules/deposit/static/js/deposit/literature_submission_form.js
+++ b/inspire/modules/deposit/static/js/deposit/literature_submission_form.js
@@ -1,6 +1,6 @@
 /*
  * This file is part of INSPIRE.
- * Copyright (C) 2014 CERN.
+ * Copyright (C) 2014, 2015 CERN.
  *
  * INSPIRE is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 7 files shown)


jalavik:
Looks good. Just missing some sort of "loading" notification for the user and it can be merged.

pedrogaudencio:
Amended ;)

jalavik:
@jmartinm Tested locally and works nicely, leaving to you to put final approval stamp :-)

inspire/modules/deposit/forms.py
```diff
@@ -463,24 +463,28 @@ class LiteratureForm(WebDepositForm):
                                              mapper=journal_title_kb_mapper)
     )
 
-    page_range_article_id = fields.TextField(
+    journal_page = fields.TextField(
         label=_('Page Range/Article ID'),
         description=_('e.g. 1-100'),
+        export_key='page_range_article_id',
         widget_classes="form-control" + ARTICLE_CLASS
```
jmartinm: I am wondering about the renaming?

For example, for this field now we would have 3 different names, the form id will be ``journal_page``, then we export that field with the key name ``page_range_article_id`` and then we have to map it to the name ``page_artid`` [here](https://github.com/inspirehep/inspire-next/blob/master/inspire/modules/deposit/workflows/literature.py#L237)

Can we directly use the name in our [inspire.cfg](https://github.com/inspirehep/inspire-next/blob/master/inspire/base/recordext/fields/inspire.cfg) ? Or at least ``export_key`` with that name. And then clean up the ``process_sip_metadata`` function to avoid confusion with so many names for the same thing?

inspire/modules/deposit/static/js/deposit/literature_submission_form.js
```diff
@@ -796,6 +804,48 @@ define(function(require, exports, module) {
...
+      this.toggleTemporaryMessage(options.blockFields[0], options.message);
+      this.toggleDisableFields(options.blockFields, true);
+
+      var that = this;
+
+      $.ajax({
+        url: options.url,
+        data: options.params
+      }).done(function(result) {
+        that.fillForm(result);
```
jmartinm: If the ``result`` here is just the journal information fields, will calling ``fillForm`` not have undesired side effects?

That function does [many things](https://github.com/inspirehep/inspire-next/blob/master/inspire/modules/deposit/static/js/deposit/literature_submission_form.js#L723) including recreating the author fields (removing them first).

So I guess this second call to fillForm will delete the authors?

inspire/modules/deposit/static/js/deposit/templates/loading_message.mustache
```diff
@@ -0,0 +1,21 @@
...
+  WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+  General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+  59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+}}
+<div class="alert">
+    <img src="/img/loading.gif"/>{{{message}}}
```
jmartinm: what does ``{{{``` do? I haven't seen it before, in Jinja you usually would put 2 curly brackets. maybe something I am not aware of?

jmartinm: Use ``` {{ url_for('static', filename=```

See example:

https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/communities/templates/communities/helpers_base.html#L50

pedrogaudencio: `{{{ }}}` is to receive the value and handle it as unescaped html, it's [mustache](http://mustache.github.io/mustache.5.html#Variables) templates syntax, actually. ;)

inspire/modules/deposit/forms.py
pedrogaudencio: It's just so they start with the same name and then I map them more easily [here](https://github.com/pedrogaudencio/inspire-next/blob/arxiv-split-pub-note/inspire/modules/deposit/views.py#L47). I can change the `export_key` to `page_artid` since it makes sense. ;)

inspire/modules/deposit/static/js/deposit/templates/loading_message.mustache
jmartinm: oh good. Thanks for the clarification, did not realise it was a mustache template. Then you can ignore the comments.

inspire/modules/deposit/forms.py
jmartinm: OK, great. Please, review the rest of the ``export_key`` so that we can directly map with the names in ``inspire.cfg`` and get rid of some code in ``process_sip_metadata``.

For the mapping [there](https://github.com/pedrogaudencio/inspire-next/blob/arxiv-split-pub-note/inspire/modules/deposit/views.py#L47) hopefully the API will be stable and we will receive always the same names, so should be ok.

inspire/modules/deposit/static/js/deposit/literature_submission_form.js
jmartinm: Any news on this? Did you check if this is a problem?

pedrogaudencio: It was removing them, it's fixed now. ;)

jmartinm:
@pedrogaudencio can you give some arXiv id examples to test the functionality?

pedrogaudencio:
astro-ph/0603450
astro-ph/9805202
astro-ph/9805245

jmartinm:
Thanks. 

I see if I use astro-ph/9805202 it works the first time, but if I can click on Import a second time I get:

import-bad-label

can you have a look?

pedrogaudencio:
Fixed, it was because I changed the form names and forgot to change them on arxiv.

jmartinm:
This is pending being revived once the legacy refextract gets ported as it is used here https://github.com/inspirehep/inspire-next/pull/191/files#diff-a3d2133d24d1186f17d91ec1b59269a8R44 or we can use grobid API instead.

jmartinm:
@jacquerie @jalavik Is there already an API available for Grobid that can be used to pass some string with a publication note so that it returns the individual fields? title, year, volume and page range

jacquerie:
Almost. We can use the `processCitation` endpoint in GROBID's REST API, which is a superset of a publication note (it also contains authors). We could implement an appropriate endpoint in `invenio-grobid` that discards any information about authors.

jalavik:
@jmartinm In principle yes, but the reference processing is not yet fully up to par with INSPIRE standards in many cases. E.g. when there is arXiv numbers etc. It is on the todo list to improve reference extraction but you can play already with the API on `http://inspire-grobid.cern.ch:8080`

kaplun:
Ehi, is there anything useful that can be taken from this PR or shall we simply close it as outdated?

PR #192, opened 2015-01-29
deposit: comments and submitter in 595 MARC field

pedrogaudencio:


jalavik:
This does not work in its current state because in the task before `process_sip_metadata`, the keys like `extra_comments` are deleted (see task `add_submission_extra_data`).

It's unclear if there is a specific reason why `add_submission_extra_data` happens before `process_sip_metadata`, but a possible work-around is too switch the order of these tasks.

I quickly tried this, and it seem to work:
```python
b = BibWorkflowObject.query.get(958)
b.get_data()["sips"][0]["metadata"]["hidden_note"]
Out[17]: {'submitter': 1L, 'value': u'This might work\r\n\r\nLALALAL'}
```

pedrogaudencio:
Yeah, when I wrote it I only tested it with the `literature_simple` workflow... I was wondering why it didn't showed up in `literature`.

Isn't querying the `BibWorkflowObject` kinda hackish?... If it doesn't make a difference, I suggest switching the order of the tasks, this way we have the correct mapping on the metadata already set before any other task runs on it.

jalavik:
@pedrogaudencio Yeah, let's switch the order. PS: The `BibWorkflowObject` code snippet was not meant to be used - just illustrated that by switching the task the data was there. Sorry for the confusion.

jalavik:
This is looking good. Creates:

```xml
<datafield tag="595" ind1="" ind2="">
        <subfield code="a">THIS IS A COMMENT</subfield>
        <subfield code="9">1</subfield>
</datafield>
```

@jmartinm

inspire/base/recordext/fields/inspire.cfg
```diff
@@ -236,10 +236,11 @@ funding_info:
     creator:
         @legacy((("595", "595__", "595__%"), ""),
                 ("595__a", "value"),
-                ("595__b", "cern_reference"))
-        marc, "595__", {'value': value['a'], 'cern_reference':value['b']}
+                ("595__b", "cern_reference"),
+                ("595__9", "submitter"))
+        marc, "595__", {'value': value['a'], 'cern_reference': value['b'], 'submitter': value['9']}
```
jmartinm: Should be ``'source': value['9']`` and above ``("595__9", "source")``

inspire/base/recordext/fields/inspire.cfg
```diff
@@ -236,10 +236,11 @@ funding_info:
...
     producer:
-        json_for_marc(), {"595__a": "", "595__b": "cern_reference"}
+        json_for_marc(), {"595__a": "value", "595__b": "cern_reference", "595__9": "submitter"}
```
jmartinm: ``"595__9": "source"``

inspire/modules/deposit/workflows/literature.py
```diff
@@ -527,6 +527,14 @@ def restructure_ffts(fft):
             submission_number=deposition.id,
         )
 
+        # ==============
+        # Extra comments
+        # ==============
+        if 'extra_comments' in metadata and metadata['extra_comments']:
+            metadata['hidden_note'] = {'value': metadata['extra_comments'],
+                                       'submitter': userid}
```
jmartinm: should be:

```python
 {'value': metadata['extra_comments'], 'source': 'submitter'}
```

inspire/modules/deposit/workflows/literature.py
```diff
@@ -527,6 +527,14 @@ def restructure_ffts(fft):
...
+            delete_keys.append('extra_comments')
```
jmartinm: if you delete the key ``extra_comments`` then they will be not available to be shown in the Holding Pen detailed page:

https://github.com/inspirehep/inspire-next/blob/master/inspire/modules/workflows/templates/workflows/details.html#L45

jmartinm:
Merged

68f63a828e0afb1c16f167f1b86402154e0b5638

PR #194, opened 2015-02-02
deposit: add user comment to RT ticket

pedrogaudencio:
Addresses [this task](https://app.asana.com/0/25354430037957/23837154840527).

Depends on https://github.com/inspirehep/inspire-next/pull/192.

@jalavik

### DIFF
--- a/inspire/modules/workflows/tasks/submission.py
+++ b/inspire/modules/workflows/tasks/submission.py
@@ -97,6 +97,7 @@ def create_ticket(template, queue="Test", ticket_id_key="ticket_id"):
             title=d.title,
             identifier=sip.metadata.get("system_number_external", {}).get("value", ""),
             user_comment=user_comment,
+            references = obj.extra_data.get("submission_data", {}).get("references"),
             object=obj,
         ).strip()
 
--- a/inspire/modules/workflows/tasks/submission.py
+++ b/inspire/modules/workflows/tasks/submission.py
@@ -86,11 +86,17 @@ def create_ticket(template, queue="Test", ticket_id_key="ticket_id"):
         rt_queue = cfg.get("CFG_BIBCATALOG_QUEUES") or queue
         sip = d.get_latest_sip(sealed=False)
         subject = u"Your suggestion to INSPIRE: {0}".format(d.title)
+        try:
+            user_comment = filter(lambda k: 'submitter' == k.get('source', ''),
+                                  sip.metadata.get("hidden_note", []))[0].get('value', None)
+        except IndexError:
+            user_comment = None
         body = render_template(
             template,
             email=email,
             title=d.title,
             identifier=sip.metadata.get("system_number_external", {}).get("value", ""),
+            user_comment=user_comment,
             object=obj,
         ).strip()
 
--- a/inspire/base/recordext/fields/inspire.cfg
+++ b/inspire/base/recordext/fields/inspire.cfg
@@ -233,6 +233,8 @@ funding_info:
 
 
  hidden_note:
+    schema:
+        {'hidden_note': {'type': 'list', 'force': True}}
     creator:
         @legacy((("595", "595__", "595__%"), ""),
                 ("595__a", "value"),
--- a/inspire/modules/deposit/templates/deposit/tickets/curator_submitted.html
+++ b/inspire/modules/deposit/templates/deposit/tickets/curator_submitted.html
@@ -18,3 +18,7 @@
 #}
  New submission from {{ email }}
  Approve/reject it here: {{ url_for('holdingpen.details', objectid=object.id, _external=True) }}
+
+{% if user_comment %}
+ User comment: {{ user_comment|safe }}
+{% endif %}
--- a/inspire/modules/deposit/templates/deposit/tickets/curator_submitted.html
+++ b/inspire/modules/deposit/templates/deposit/tickets/curator_submitted.html
@@ -22,3 +22,7 @@
 {% if user_comment %}
  User comment: {{ user_comment|safe }}
 {% endif %}
+{% if references %}
+ References:
+ {{ references|safe }}
+{% endif %}
--- a/inspire/modules/deposit/workflows/literature.py
+++ b/inspire/modules/deposit/workflows/literature.py
@@ -535,8 +535,8 @@ class literature(SimpleRecordDeposition, WorkflowBase):
         # Extra comments
         # ==============
         if 'extra_comments' in metadata and metadata['extra_comments']:
-            metadata['hidden_note'] = {'value': metadata['extra_comments'],
-                                       'source': 'submitter'}
+            metadata['hidden_note'] = [{'value': metadata['extra_comments'],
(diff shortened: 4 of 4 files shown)


pedrogaudencio:
I'm [escaping](http://jinja.pocoo.org/docs/dev/templates/#working-with-automatic-escaping) the characters in the [template](https://github.com/pedrogaudencio/inspire-next/blob/rt-ticket-add-user-comment/inspire/modules/deposit/templates/deposit/tickets/curator_submitted.html#L23) assuming this will not harm inspirert...

Escaping:
```
User comment:
uhuh now there&#39;s a comment! :)
```

No escaping:
```
User comment:
uhuh now there's a comment with escaped characters because we trust people! :D
```

@jalavik @kaplun, do you think this might be an issue?

jalavik:
I am not sure. You could try adding some JS `<script>alert("foo")</script>` and see if it works, but I believe we should be okay.

pedrogaudencio:
Tested it, it's safe and we can trust people. :)

jalavik:
@jmartinm Working when on top of #192 PR :+1:

jmartinm:
Please, have a look at the comments in https://github.com/inspirehep/inspire-next/pull/192

inspire/modules/workflows/tasks/submission.py
```diff
@@ -91,6 +91,7 @@ def _create_ticket(obj, eng):
             email=email,
             title=d.title,
             identifier=sip.metadata.get("system_number_external", {}).get("value", ""),
+            user_comment=sip.metadata.get("hidden_note", {}).get("value", ""),
```
jmartinm: what happens if we add a couple of hidden notes? Then:

```python
sip.metadata.get("hidden_note", {})
```

will be a list right?

We should probably take that into account and also just get the hidden note that has ``source: submitter`` - see https://github.com/inspirehep/inspire-next/pull/192

pedrogaudencio: Then we should make it a list on inspire.cfg, right?
```
url:
    schema:
        {'url': {'type': 'list', 'force': True}}
...
```

Anyway something like this should work
`filter(lambda k: 'submitter' == k['source'], sip.metadata.get("hidden_note", []))`

jmartinm: Yes to both. In that way the field will always be a list and ```sip.metadata.get("hidden_note", [])``` will be the right thing to do.

pedrogaudencio: I'd actually do it like
```
user_comment=filter(lambda k: 'submitter' == k['source'],
                    sip.metadata.get("hidden_note", []))
```

So this way if there isn't a user comment (a dict value in the `metadata['hidden_note']` list with the `source` key as `submitter`), then `user_comment=[]` and in the jinja template will skip it.

jmartinm: That's fine.

```python
hidden_note:
    schema:
        {'hidden_note': {'type': 'list', 'force': True}}
```

should be added as well as far as I can see.

By the way,

```python
user_comment=filter(lambda k: 'submitter' == k['source'],
                    sip.metadata.get("hidden_note", []))
```

will throw ```KeyError: 'source'``` if one of the elements in the ``hidden_note`` does not contain the ``source`` key. I guess that might be the case in the future if we add some extra hidden notes?

pedrogaudencio: True that, so `k.get('source', None)` should prevent it.

inspire/modules/workflows/tasks/submission.py
```diff
@@ -86,11 +87,17 @@ def _create_ticket(obj, eng):
         rt_queue = cfg.get("CFG_BIBCATALOG_QUEUES") or queue
         sip = d.get_latest_sip(sealed=False)
         subject = u"Your suggestion to INSPIRE: {0}".format(d.title)
+        try:
```
pedrogaudencio: `ifilter` returns an iterator, `next()` gets only the first element because having more than one comment with the `'submitter'` source would break it... Is this possible to happen? Should we prevent it from breaking the task?

@jmartinm @jalavik, WDYT?

jmartinm: I think this is fine as it is now. It should not be the case to have more than one ``hidden_note`` with ``source:submitter``

inspire/modules/workflows/tasks/submission.py
```diff
@@ -86,11 +86,17 @@ def _create_ticket(obj, eng):
...
+            user_comment = filter(lambda k: 'submitter' == k.get('source', ''),
+                                  sip.metadata.get("hidden_note", []))[0].get('extra_comments', None)
```
jmartinm: will this work? According to https://github.com/inspirehep/inspire-next/pull/192/files#diff-3b59804856efabb4ecf3b800ac348ee9R533 there is no key ``extra_comments`` inside of ``hidden_note`` ?

jmartinm:
BTW, maybe it was not clarified in Asana, but as we mentioned in person both this ticket and the one here https://github.com/inspirehep/inspire-next/pull/198 should contain the references as well (see ``add_submission_extra_data()``)

inspire/modules/workflows/tasks/submission.py
pedrogaudencio: I guess I meant `value`... But I believe I tested it like that, let me check it again!

pedrogaudencio: You were right, I forgot to amend the metadata to a list after adding the list type to the inspire.cfg. Now it works correctly.

pedrogaudencio:
As for the references, you mean just the metadata form field or some other stuff after being processed?

jmartinm:
Just the metadata form field exactly as the user inputs it in the references box.

pedrogaudencio:
Ok, added it now in the last commit.

pedrogaudencio:
RT ticket example [here](https://inspirert.cern.ch/Ticket/Display.html?id=469317), @jmartinm.

inspire/modules/workflows/tasks/submission.py
```diff
@@ -86,11 +86,18 @@ def _create_ticket(obj, eng):
         rt_queue = cfg.get("CFG_BIBCATALOG_QUEUES") or queue
         sip = d.get_latest_sip(sealed=False)
         subject = u"Your suggestion to INSPIRE: {0}".format(d.title)
+        try:
+            user_comment = filter(lambda k: 'submitter' == k.get('source', ''),
+                                  sip.metadata.get("hidden_note", []))[0].get('value', None)
```
jmartinm:
```python
sip.metadata.get("hidden_note", [])[0]
```

will throw ``IndexError`` if there is no ``hidden_note`` key.

jmartinm:
Thanks, merged

c7c92cbb51dc842d187c63b84afb77d9e9db4d05
a7f29ec4e9dbb027c19c03b705b60e8e8974ff6a

PR #195, opened 2015-02-02
deposit: automatic ticket status handling

jalavik:
* Automatically closes the submission ticket upon finished workflow.

* Sets the ticket status back to "new" after replying to it the first
  time.


kaplun:
I Jan, I guess all this ticketing with RT has to be moved back at some point to the future (bib)catalog module.

jalavik:
Yeah indeed it can be inspiration when porting `BibCatalog` to `tickets` module. 

There was too little time to do that from beginning for INSPIRE, but at some quiet point in time we should do this.

PR #196, opened 2015-02-03
deposit: add note entry on approval

pedrogaudencio:


inspire/modules/workflows/tasks/submission.py
```diff
@@ -256,3 +256,15 @@ def add_files_to_task_results(obj, eng):
         obj.add_task_result(file_obj.name,
                             fileinfo,
                             "workflows/results/files.html")
+
+
+def add_note_entry(obj, eng):
+    """Add metadata to sip on approval."""
+    if was_approved(obj, eng):
```
jalavik: I guess you do not need the was_approved bit here strictly speaking?

inspire/modules/workflows/tasks/submission.py
```diff
@@ -256,3 +256,15 @@ def add_files_to_task_results(obj, eng):
...
+        rt = get_instance() if cfg.get("PRODUCTION_MODE") else None
```
jalavik: I guess this `rt` thing is here as it is `wip`? I don't think we need rt in this context.

If you wanted to create a RT curation ticket tho, I'd suggest creating a new task for it - just to separate it from the note manipulation.

Come to think it of, since there is so much CORE record checking, we could evaluate if we should add another `workflow_if` etc branch in the workflow itself where we do in one case all the core business and `workflow_else` for all non-core business. Or is it overkill? WDYT?

pedrogaudencio: Since the task was
```
Core papers:
500__ $$a*Temporary entry*  <--- RT ticket generated for these
```

I understood it as `this has to be core AND has to have a RT ticket` statement.

Are there that many tasks core [exclusive] related to open a new `workflow_if/else` or you're worried about the future? ^^

inspire/modules/workflows/tasks/submission.py
pedrogaudencio: I added it before adding the task to the `workflow_if(was_approved)` and then I forgot to amend it. ;)

inspire/modules/deposit/workflows/literature.py
```diff
@@ -327,10 +335,11 @@ def remove_exclusive_fields(fieldlist):
             delete_keys.extend(thesis_fields)
 
         if 'defense_date' in metadata and metadata['defense_date']:
-            if 'note' in metadata and metadata['note']:
-                metadata['note'] += ', presented on ' + metadata['defense_date']
+            defense_note = {'value': 'Presented on ' + metadata['defense_date']}
+            if metadata.get('note', None):
+                metadata['note'].append(defense_note)
```
jalavik: There is a problem here in that it is not backwards compatible with existing submission data. We would probably need to add an ugly check if the value is a string.

Old submission data:
```
In [6]: m.get("note")
Out[6]: u'4 pages, 2 figures'
```
will get an error when trying to `list.append`: 
```
AttributeError: 'unicode' object has no attribute 'append'
```

jmartinm:
Amended and merged as 3eb55e5618f27c73a235d2dae8056330d26a95ee

I assume this PR was not tested as it was not generating the note on approval.

It was missing https://github.com/inspirehep/inspire-next/commit/3eb55e5618f27c73a235d2dae8056330d26a95ee#diff-f1428c95fafa85e7b6b6919cc933133eR302

PR #202, opened 2015-02-06
docs: update INSPIRE installation instructions

jacquerie:
* Mixer is already required, and the database has to be dropped
  before being recreated.

Signed-off-by: Jacopo Notarstefano <jacopo.notarstefano@cern.ch>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -40,31 +40,27 @@ Next step is to install assets (js, css etc.):
     plupload
     ...
 
-Now you can follow the standard invenio installation and development procedures, such as running `inveniomanage collect`.
+Now you can follow the standard Invenio installation and development procedures, such as running `inveniomanage collect`.
 
 ---------------------
 Demosite installation
 ---------------------
-As INSPIRE uses `Mixer <https://github.com/klen/mixer>`_ to load database fixtures, you should install the library:
+Since the database was already initialized while installing Invenio, you run:
 
 .. code-block:: bash
 
-     $ pip install -e .[mixer]
+     $ inveniomanage database recreate
 
-Before running:
+to drop the previous database structure and create a new one.
 
-.. code-block:: bash
-
-     $ inveniomanage database create
-
-Populate demo records and enable demo-site:
+You can then populate it with demo records and enable the demo site:
 
 .. code-block:: bash
 
     (inspire)$ cdvirtualenv src/inspire-next
     (inspire)$ inveniomanage demosite populate -p inspire.base -f inspire/demosite/data/demo-records.xml
 
-Start the server:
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -40,31 +40,27 @@ Next step is to install assets (js, css etc.):
     plupload
     ...
 
-Now you can follow the standard invenio installation and development procedures, such as running `inveniomanage collect`.
+Now you can follow the standard Invenio installation and development procedures, such as running `inveniomanage collect`.
 
 ---------------------
 Demosite installation
 ---------------------
-As INSPIRE uses `Mixer <https://github.com/klen/mixer>`_ to load database fixtures, you should install the library:
```
jmartinm: Mixer is in ``extras_require`` https://github.com/inveniosoftware/invenio/blob/pu/setup.py#L161 so still for INSPIRE we should tell people to install it.

README.rst
```diff
@@ -40,31 +40,27 @@ Next step is to install assets (js, css etc.):
...
+Since the database was already initialized while installing Invenio, you run:
 
 .. code-block:: bash
 
-     $ pip install -e .[mixer]
+     $ inveniomanage database recreate
```
jmartinm: I am not sure about this. In the instructions https://github.com/inveniosoftware/invenio/blob/pu/INSTALL.rst#33-development the

``` $ inveniomanage database create```

happens once both Invenio and the demosite have been installed. Since INSPIRE replaces the demosite but the instructions are the same I assume the database should not be created at this stage and ```$ inveniomanage database create``` is the right command?

README.rst
jacquerie: I might be totally wrong, but isn't it installed by INSPIRE? See: https://github.com/inspirehep/inspire-next/blob/a3a56cce87c98bb02146d85679e62aeec0d70419/setup.py#L60

In any case, running that command inside the `inspire-next` folder resulted in an obscure error, since INSPIRE's `setup.py` has no `extras_require` section.

jalavik: Since it is part of the `setup.py` it should indeed come automatically when installing the overlay.

We could still have a note in the readme that we use mixer for pre-loading the database without the specific installation instruction.

README.rst
jalavik: Indeed `create` is part of those instructions which probably even can be removed from here, or turned into a "tip" on how to reload the database and load demo records.

README.rst
jmartinm: Yes, right, I did not look into INSPIRE's ``setup.py``. If it gets correctly installed from there we should be good and you can remove those lines.

PR #193, opened 2015-01-30
deposit: url and additional_url in 8564 MARC field

pedrogaudencio:

PR #204, opened 2015-02-10
deposit: fix lost pdf url

pedrogaudencio:
@jalavik


jalavik:
ping @jmartinm . Was the idea to merge the two perhaps?

jalavik:
See also #193

pedrogaudencio:
Added it to #193.

PR #207, opened 2015-02-12
feedback: rewrite without feedback.js

### LINKED ISSUE #199: Possible UI improvements on small screen resolutions
* I'm currently testing INSPIRE on my Dell Inspiron 1010, which has a 1024x600 screen resolution. When I am on http://labs.inspirehep.net and I click on the "Send Feedback" link I get this:
screenshot from 2015-02-05 10 00 45
and I can't dismiss the feedback window except by reloading the page. Several other resolutions suffer from similar problems: http://www.responsinator.com/?url=labs.inspirehep.net (sorry for all the ads).

* http://labs.inspirehep.net has a very tall header which stays fixed on scrolling, cluttering the view. A possible solution is to use [WickyNilliams/headroom.js](https://github.com/WickyNilliams/headroom.js) (demo [here](http://wicky.nillia.ms/headroom.js/)), which hides the header when scrolling.

(CC: @kaplun @jalavik)

jacquerie:
* Rewrites the feature without inspirehep/feedback.js. Builds the
    modal using Bootstrap classes, making it responsive.  (closes #199)

  * Adds some validation to avoid submitting an empty form.

(CC: @kaplun @jalavik @jmartinm)

### DIFF
--- a/inspire/base/static/js/settings.js
+++ b/inspire/base/static/js/settings.js
@@ -1,6 +1,6 @@
 /*
  * This file is part of INSPIRE.
- * Copyright (C) 2014 CERN.
+ * Copyright (C) 2014, 2015 CERN.
  *
  * INSPIRE is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -59,7 +59,7 @@ require.config({
     "bootstrap-multiselect": "vendors/bootstrap-multiselect/dist/js/bootstrap-multiselect",
     "readmore": "vendors/readmore/readmore",
     "buckets": "vendors/buckets/buckets",
-    "feedback": "vendors/feedback/examples/js/feedback",
+    "feedback": "js/feedback/feedback",
     "html2canvas": "vendors/html2canvas/build/html2canvas",
     "highcharts": "vendors/highcharts-release/highcharts"
   },
@@ -170,10 +170,6 @@ require.config({
     },
     "html2canvas" : {
       deps: ["jquery"]
-    },
-    "feedback" : {
-      deps: ["jquery", "html2canvas"],
-       exports: "Feedback"
     },
     "highcharts" : {
       deps: ["jquery"],
--- a/inspire/base/bundles.py
+++ b/inspire/base/bundles.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 ## This file is part of INSPIRE.
-## Copyright (C) 2014 CERN.
+## Copyright (C) 2014, 2015 CERN.
 ##
 ## INSPIRE is free software: you can redistribute it and/or modify
 ## it under the terms of the GNU General Public License as published by
@@ -28,10 +28,6 @@ js = Bundle(
     output='base.js',
     filters=RequireJSFilter(exclude=[_j, _i]),
     weight=20,
-    bower={
-        'feedback': 'git://github.com/inspirehep/feedback.js.git',
-        'html2canvas': '*',
-    },
 )
 
 # index_js = Bundle(
@@ -64,7 +60,8 @@ _base_styles.contents += (
     'less/base/core.less',
     'less/accounts/login.less',
     'less/search/index.less',
-    'vendors/feedback/examples/css/feedback.css',
+    'less/feedback/button.less',
+    'less/feedback/modal.less',
 )
 
 to_remove = ["less/base.less",
--- a/inspire/base/static/js/feedback/feedback.js
+++ b/inspire/base/static/js/feedback/feedback.js
@@ -1,0 +1,103 @@
+ /*
+ * This file is part of INSPIRE.
+ * Copyright (C) 2015 CERN.
+ *
+ * INSPIRE is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * INSPIRE is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * In applying this licence, CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+define([
+  'jquery',
+  'hgn!js/feedback/templates/button',
+  'hgn!js/feedback/templates/error',
+  'hgn!js/feedback/templates/modal',
+], function ($, buttonTemplate, errorTemplate, modalTemplate) {
+  return {
+    'Feedback': function (opts) {
+      var options = {};
+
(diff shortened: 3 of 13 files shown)


inspire/base/static/js/settings.js
```diff
@@ -172,7 +172,7 @@ require.config({
       deps: ["jquery"]
     },
     "feedback" : {
```
jmartinm: I am wondering, do we need the shim configuration? Or if ``js/feedback/feedback.js`` is already requirejs friendly (as it seems to be) we can omit it. You can see a discussion on this [here](http://stackoverflow.com/questions/10918063/how-to-make-a-jquery-plugin-loadable-with-requirejs)

jacquerie: It appears we don't. I was parroting the previous code : )

jmartinm: Ok great, always nice removing some code.

jacquerie:
Ok, now I want to write some tests for this feature.

Where should I put them? Parroting https://github.com/jacquerie/invenio/commit/16b13ba6e055e258428ecafb9fa80e106a368710#diff-e07a967ed6eedebaf07ff9f570546aaa I'd say that tests for some JavaScript in `inspire/base/static/js/feedback/feedback.js` go in `inspire/base/js/feedback` in a file called `feedback.spec.js`. Correct?

Then: how should I run them?

jmartinm:
For where tests should be located:
https://github.com/inveniosoftware/invenio/blob/pu/invenio/ext/jasmine/registry.py#L33

Instructions on how to run them:
https://github.com/inveniosoftware/invenio/blob/pu/invenio/ext/jasmine/__init__.py

jacquerie:
:+1:

jacquerie:
Uh, weird, this PR is breaking the serving of assets using `LESS_RUN_IN_DEBUG = False`. It's working on `master`. Will investigate.

(CC: @jmartinm)

jacquerie:
Ah, it was nothing. From RequireJS FAQ (http://requirejs.org/docs/errors.html#mismatch):

> If you use the loader plugins or anonymous modules (modules that call define() with no string ID) but do not use the RequireJS optimizer to combine files together, this error can occur.

In fact, I wasn't defining a string ID for `feedback`.

jacquerie:
Whoops, no, I simply wasn't testing it thoroughly: the setting `LESS_RUN_IN_DEBUG` is broken on `master` as well. Will open an issue.

jmartinm:
You should not need to add name after ``define`` https://github.com/inspirehep/inspire-next/pull/207/files#diff-4fe96ec05f2b163ba7b1dca6e2e798bbR23 as the compiler [r.js](http://requirejs.org/docs/optimization.html) is adding those names for you.

Did you try without the explicit name and with ``ASSETS_DEBUG=False``? It should work correctly, and if not it might be some other problem.

jacquerie:
In fact, that wasn't the issue. I can reproduce this bug even on the Invenio pu demosite!

jmartinm:
@jacquerie some small comments before integration:

- [ ] For better english the message ``Seems like...`` should be ``It seems like...``
- [ ] If you click on Send with empty content and receive the alert, when the modal is closed this message should be cleared, so that you don't have the red banner next time the Feedback is opened.
- [ ] What happened with the idea of writing some tests? Was there any problem with it? Would have been nice to have an example on the INSPIRE repo of jasmine test.

jacquerie:
I updated the PR to fix the first two points. Sorry for the tests! I got sidetracked trying to diagnose this issue: https://github.com/inveniosoftware/invenio/issues/2777. Will work on them this afternoon, before the forum.

jacquerie:
Mh, I can't seem to get those tests to run. Right now I have a simple `feedback.spec.js` in the `inspire/base/testsuite/js/feedback` folder:
```js
define([
  'js/feedback/feedback',
  'jasmine-boot',
  'jasmine-core',
], function(Feedback) {
  describe("A suite", function() {
    it("contains spec with an expectation", function() {
      expect(true).toBe(true);
    });
  });
});
```
But, after running `fab clean_assets && inveniomanage assets build`, I see no spec at `localhost:5000/jasmine/specrunner` (404). I'm not sure if what is failing is the registration (wrong filename/folder?), the RequireJS configuration, or something else.

jacquerie:
I finally managed to get those tests to work. First of all I was missing a `__init__.py` file in `inspire/base/testsuite` (which in retrospect is obvious, since Jasmine's autodiscovery discovers *modules*).

Then, instead of seeing the test results, I was being redirected to one of the fixtures. The culprit is [this line](https://github.com/inveniosoftware/invenio/blob/ffbb1071be36fdf4961b34de721c0eb27c3994dd/invenio/ext/jasmine/testsuite/js/initialization_checker.spec.js#L29) in Invenio (!). Commenting it out gives 3 errors in Jasmine (which are expected, since the fixture is no longer there) but finally I see my dummy spec among the specs that ran:
```
A suite
    contains spec with an expectation

RemoteAutocomplete suite
    RemoteAutocomplete input text is as expected

[...]
```

jacquerie:
@jmartinm: I think it's ready to review.

jmartinm:
Thanks @jacquerie . Could you have a look at how we can plug into the Travis.yml the configuration so that your Jasmine tests would run as well?

jacquerie:
Sure. I'm guessing this should be a different branch/PR, right?

jmartinm:
Yes, sure, please create a separate PR.

jmartinm:
I am getting 3 errors while running the tests:

jasmine spec runner

jmartinm:
BTW, for the new PR, in this docstring you can see that both CI integration and a CLI runner are missing and would be very nice to have: https://github.com/inveniosoftware/invenio/blob/pu/invenio/ext/jasmine/__init__.py#L39

jacquerie:
Well, it works on my machine™... Let me check what's wrong. DOM testing in Jasmine might be finnicky; in fact I see a modal left added to the page despite it should have been cleaned by the outer `afterEach`. Can you try running the tests again?

jacquerie:
And yes, yesterday I worked on understanding the Karma test runner, since it appears to integrate nicely with RequireJS and Jasmine. There's a chance I am duplicating some work @lnielsen had done. I'm also not sure if I should be working against INSPIRE or Invenio: a problem I have with the latter is that there's no `package.json` to declare dependencies, so I'm guessing that might require an RFC and some discussion.

jmartinm:
I see now. The problem the other day was ``ASSETS_DEBUG=False`` does not work. Needs to be set to ``True`` With that they work for me as well.

That part seems to be missing in:
https://github.com/inveniosoftware/invenio/blob/pu/invenio/ext/jasmine/__init__.py#L22

jmartinm:
> That part seems to be missing in:
https://github.com/inveniosoftware/invenio/blob/pu/invenio/ext/jasmine/__init__.py#L22

Actually, now that I think about it, only your tests were failing. So it might be that there is a problem with the ``spec`` file and it is not being correctly compiled (with ``ASSETS_DEBUG=False``).

jacquerie:
Sorry for being completely unhelpful, but the tests appear to work on my machine™ even with `ASSETS_DEBUG=False` and `TESTING=True`, after a `fab clean_assets && inveniomanage assets build && inveniomanage runserver`.

I'm thinking that testing the DOM with Jasmine wasn't a great idea (too many false negatives), and I should only keep the tests for the Ajax part.

jmartinm:
My bad, I was missing ``inveniomanage assets build ``

It works ok now. Ready for merge.

PR #215, opened 2015-03-04
deposit: literature form update thesis date input values

pedrogaudencio:


jmartinm:
Thanks. I see that with this commit a date like:

``2001`` is converted to ``2001-01-01``

and a date like `2001-01`` to ``2001-01-01``

I am not sure we want to guess a date if it is not given explicitly, so probably we would just leave exactly what the user inputted. Can you double check with Annette this point?

inspire/modules/deposit/fields/thesis_date.py
```diff
@@ -0,0 +1,39 @@
...
+            for date_format in self.format:
+                try:
+                    self.data = datetime.datetime.strptime(date_str, date_format).date()
+                except ValueError:
+                    pass
+                else:
+                    break
+            else:
+                self.data = None
+                raise ValueError(self.gettext('Not a valid date value'))
```
jmartinm: What is the advantage of using ``process_formdata`` over a validator function in the existing ``Date`` field?

PS. Cool use of for... else..

pedrogaudencio:
Ok we need a TextField instead and then do some validation for the `YYYY-MM-DD`, `YYYY-MM` and `YYYY` formats. I'll check with Annette anyway ;)

inspire/modules/deposit/fields/thesis_date.py
pedrogaudencio: `process_formdata()` is where `self.format` is checked, otherwise it would raise an exception for normal date validation and it wouldn't accept more than one format.

`else` is used only when `for` doesn't run normally (like when it breaks), reference [here](https://docs.python.org/3/reference/compound_stmts.html#for) ;)

jmartinm:
Yes, that might be fine. Just make sure in the validator that each of the components is valid (e.g YYYY is a number with 4 digits)

kaplun:
@jmartinm is this still valid PR?

jmartinm:
Valid PR, still need to revive it.

jmartinm:
@gartocern we can talk tomorrow about reviving this PR.

jmartinm:
Superseded by https://github.com/inspirehep/inspire-next/pull/506

PR #216, opened 2015-03-10
setup: mixer downgrade

MSusik:
* Downgrades mixer to 4.9.5 which doesn't cause problems with
  knwKB table.

Signed-off-by: Mateusz Susik <mateusz.susik@cern.ch>


MSusik:
@dset0x @jalavik

PR #220, opened 2015-03-12
deposit: arxiv fft get refactor

pedrogaudencio:


jalavik:
Looks good. Will give it a test.

pedrogaudencio:
It was just so we don't lose the helper function... I believe it's clearer now even though we're not harvesting the arXiv pdf. ;)

inspire/modules/deposit/tasks.py
```diff
@@ -56,21 +64,23 @@ def arxiv_fft_get(obj, eng):
             dir=CFG_TMPSHAREDDIR,
         )
 
-        os.write(arxiv_file, arxiv_pdf.content)
-        os.close(arxiv_file)
```
jalavik: @pedrogaudencio I think you still need to call `os.close` to properly close the `arxiv_file` descriptor. See https://docs.python.org/2/library/tempfile.html#tempfile.mkstemp

jalavik:
Apart from the above comment, it looks fine. I will merge after fix and rebase + re-push.

pedrogaudencio:
Fixed, rebased and re-pushed ;)

jalavik:
Thanks!

PR #224, opened 2015-04-02
base: requirements cleanup

jmartinm:
* Reorganizes requirements and setup.py dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ setup(
             'Flask-DebugToolbar>=0.9',
             'ipython',
             'ipdb',
-            'Invenio-Kwalitee'
+            'kwalitee'
         ],
     },
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -58,17 +58,13 @@ setup(
         "rt",
         "HarvestingKit>=0.3",
         "mixer==4.9.5",
-        "ipython",
-        "Babel>=1.3",
-        "Invenio",
-    ],
-    dependency_links=[
-        "git+https://github.com/inspirehep/invenio@labs#egg=Invenio"
     ],
     extras_require={
         'development': [
             'Flask-DebugToolbar>=0.9',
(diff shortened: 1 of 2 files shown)


jmartinm:
@jalavik I have cleaned up some dependencies, can you have a look?

Do you know what is ``setuptools-bower``? It does not seem to be used so I removed it.

Also instead of ``dependency_links`` (which was not working for me) I have added the dependencies that need a specific branch to ``requirements.txt``

jalavik:
@jmartinm Cool. `setuptools-bower` is an Invenio thing that would install bower dependencies during Invenio installation: https://github.com/inveniosoftware/setuptools-bower. However, I believe it is not being used anymore. @lnielsen

Seems like `Invenio-Kwalitee` installation is failing. Looks like its called `kwalitee` in PyPi: https://pypi.python.org/pypi/kwalitee

setup.py
```diff
@@ -58,17 +58,13 @@
         "rt",
         "HarvestingKit>=0.3",
         "mixer==4.9.5",
-        "ipython",
-        "Babel>=1.3",
```
jalavik: I suspect we may still need Babel from a clean installation. If you look at the the `travis.yml` it reveals a few more versioned dependencies we need to add here it seems..

jalavik: We could try to kill them and see if it is still needed.

jalavik:
@jmartinm Issue with collections. Perhaps this PR should be merged into `master` first?

jmartinm:
The PR was already merged to `master` I think? https://github.com/inspirehep/inspire-next/commits/master

what it the issue with collections?

jalavik:
Oh right. Forgot it was already merged :boom: 

I think we can just close this PR then because at some point we'll rebase next on master I guess?

jmartinm:
Closed.

I am rebasing at the moment so `next` will contain this commit very soon.

PR #230, opened 2015-04-23
authors: better handling of no name case

MSusik:
* Introduces util_merge function

Signed-off-by: Mateusz Susik <mateusz.susik@cern.ch>


glouppe:
merged as 6e67f49bb94e7dafe8ff05bd955a3385368df7b5

PR #231, opened 2015-04-24
classifier: automatic tagging of CORE records

glouppe:
CC @jalavik  @Lilykos 

This requires Numpy, Scipy, Scikit-Learn and Beard as additional dependencies.

### DIFF
--- a/inspire/modules/classifier/core.py
+++ b/inspire/modules/classifier/core.py
@@ -1,0 +1,236 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2015 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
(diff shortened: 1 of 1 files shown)


glouppe:
So the idea would be retrain a model on Labs, using a record file that I can provide and then save it into disk. Then within the holding pen interface, one could simply reload the model and use the `predict` function to predict the CORE tag of a new record.

jalavik:
@glouppe Thanks! I'll merge it and test it out. I will come by to get the record file.

glouppe:
Just to make it easier, to reuse to model you should do something like this:

```
# These need to be in the current module upon unpickling
def _get_title(r):
    return r["title"] if r["title"] else ""

def _get_journal(r):
    return r["journal"] if r["journal"] else ""

def _get_collaborations(r):
    return " ".join(r["collaborations"])

def _get_keywords(r):
    return " ".join(r["keywords"])

def _get_abstract(r):
    return r["abstract"] if r["abstract"] else ""

import pickle
pipeline = pickle.load(open("my-model.pickle"))

# Now you can do:
core, overall_score, top_words = predict(pipeline, my_record)
```

PR #237, opened 2015-04-28
demosite: orcid datafield addition to demo-records.xml

etzemis:
Addition of fake orcid's to a part of the AUTHOR records. 

Orcid is searchable using the following type of query in the search bar:
035__a:xxxx-xxxx-xxxx-xxxx  -> Get AUTHOR with corresponding orcid.
035__9:orcid -> Get all AUTHOR's that have an orcid.


jmartinm:
Merged a1df8b12343d1e4bcc24cb76631ba951426186f5

PR #241, opened 2015-05-18
Editable records

Lilykos:


inspire/modules/formatter/format_templates/Default_HTML_detailed.bft
```diff
@@ -3,8 +3,8 @@
 <div class="detailed_record_info">
 
 <div style="text-align:center; margin-bottom: 10px">
-<BFE_INSPIRE_TITLE_BRIEF prefix="<big><big><strong>"
-suffix="</strong></big></big>" force_title_case="yes"
+<BFE_INSPIRE_TITLE_BRIEF prefix="<big><big><strong><div class='editable' field-value='title.title'>"
```
jalavik: Can you change `field-value -> data-field` to conform with HTML standard?

Also remember to change any affected selectors in the Javascript.

inspire/modules/workflows/static/js/workflows/actions/editable_selectors.js
```diff
@@ -0,0 +1,134 @@
...
+      this.getEventField = function(ev) {
+        return ev.target.getAttribute('field-value');
+      };
+
+      this.getEventInnerHTML = function(ev) {
+        return ev.target.innerHTML;
+      };
+
+      this.getEventSelector = function(ev) {
+        return ".editable[field-value='"+ this.getEventField(ev) +"']"
```
jalavik: Here you should use the component attribute `editableClassSelector`.

```javascript
return this.attr.editableClassSelector + "[data-field='"+ this.getEventField(ev) +"']"
```

inspire/modules/workflows/workflows/process_record_arxiv.py
```diff
@@ -110,7 +110,7 @@ class process_record_arxiv(RecordWorkflow):
                     with_author_keywords=True,
                 ),
                 filter_core_keywords(filter_kb="antihep"),
-                guess_coreness(),
+                #guess_coreness(),
```
jalavik: You can remove this change. I made it so that if there is no model prediction file, it will not fail. You'll still need to install the dependencies though as written in the README: https://github.com/inspirehep/inspire-next/blob/master/README.rst#core-prediction

jalavik:
@Lilykos Can you also squash these commits into one? Check https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history

jalavik:
Thanks. Merged as 5cf317b

PR #253, opened 2015-06-19
Audit logs

Lilykos:


inspire/modules/audit/models.py
```diff
@@ -0,0 +1,56 @@
...
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Model for Audit."""
+
+from datetime import datetime
+
+from invenio.ext.sqlalchemy import db
+from invenio.ext.sqlalchemy.utils import session_manager
+
+try:
```
jalavik: This part is not necessary because (a) it is not used anywhere, and (b) it is not a nice way of importing as you rely on six.moves importing cPickle in its module - always import modules directly. (Perhaps this was PyCharm autoimport thing?)

inspire/modules/audit/models.py
```diff
@@ -0,0 +1,56 @@
...
+    from six.moves import cPickle as pickle
+except:
+    import pickle
+
+
+class Audit(db.Model):
+
+    __table_args__ = {"mysql_engine": "InnoDB"}
```
jalavik: I think it is okay not to force InnoDB in this case, so this line can be removed.

inspire/modules/audit/__init__.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+## This file is part of INSPIRE.
```
jalavik: '#' instead of '##'

jalavik:
I addressed the comments right away, amended the commit and merged as 6b2351d

PR #258, opened 2015-07-01
workflows: arxiv approval ajax call

Lilykos:
* Checks whether the selected ids list is empty
  and if so it does not allow for calls to the server.

Signed-off-by: Ilias Koutsakis <ilias.koutsakis@cern.ch>

### DIFF
--- a/inspire/modules/workflows/static/js/workflows/actions/arxiv_approval.js
+++ b/inspire/modules/workflows/static/js/workflows/actions/arxiv_approval.js
@@ -102,20 +102,22 @@ define(
         var $this = this;
         var objectids = payload.objectids;
 
-        jQuery.ajax({
-          type: "POST",
-          url: $this.attr.action_url,
-          data: payload,
-          success: function(data) {
-            $this.post_request(data);
-            $this.trigger(document, "removeSentElements", {"ids":objectids});
-          },
-          error: function(request, status, error) {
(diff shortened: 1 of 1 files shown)


inspire/modules/workflows/static/js/workflows/actions/arxiv_approval.js
```diff
@@ -102,20 +102,22 @@ define(
...
-          data: payload,
-          success: function(data) {
-            $this.post_request(data);
-            $this.trigger(document, "removeSentElements", {"ids":objectids});
-          },
-          error: function(request, status, error) {
-            console.log(error);
-          }
-        });
+        if (objectids) {
```
jalavik: Is `objectids` usually an empty array? If so, traditionally I think this check only returns `false` for (empty string), false, null, undefined, 0, NaN but not (empty array) [].

Probably safer to change to: `objectsids.length > 0` if this variable is guaranteed to be an array.

jalavik:
Can you squash the two together?

```git
git checkout master
git pull
git checkout arxiv-ajax
git rebase -i master
# Then choose "squash" for the `workflows: fixed the fix after consulting` commit, save and exit.
# Then repush
git push Lilykos arxiv-ajax -f
```

jalavik:
Merged

PR #261, opened 2015-07-08
dojson: restructuring per collection

jalavik:
* Adds new `hep` and `institutions` dojson conversion.

* Amends the MARCXML record processing to perform the correct
  conversion according to the collection.


jalavik:
It works @jmartinm @kaplun

inspire/dojson/processors.py
```diff
@@ -17,11 +17,29 @@
...
+            yield institutions.do(record)
+        else:
+            yield hep.do(record)
+
+
+def _collection_in_record(record, collection):
+    """Return a list of collections (lowercased) from 980__a."""
+    colls = record.get("980__", [])
+    if not isinstance(colls, list):
+        colls = [colls]
```
kaplun: You can use `dojson.utils:force_list()` for that. :-)

inspire/dojson/processors.py
```diff
@@ -17,11 +17,29 @@
...
     """Convert MARC XML to JSON."""
-    from inspire.dojson import inspiremarc
     from dojson.contrib.marc21.utils import create_record, split_blob
+    from inspire.dojson.hep import hep
+    from inspire.dojson.institutions import institutions
 
     for data in split_blob(source.read()):
-        yield inspiremarc.do(create_record(data))
+        record = create_record(data)
+        if _collection_in_record(record, 'institution'):
```
kaplun: I guess this we should move it in a CFG, so that we can actually propose such a change upstream

inspire/dojson/institutions/fields/bd90x99x.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@institutions.over('deleted_recid', '^981..')
+@utils.for_each_value
+@utils.filter_values
+def deleted_recid(self, key, value):
+    """Collection this record belongs to."""
+    return {
+        'deleted_recid': value.get('a'),
+    }
```
kaplun: `deleted_recid()`, `collections()`, `spyrise_sysno()`, `creation_modification_date()`, `oai_pmh()` should be all part of central rules that could be imported by all configurations.

inspire/dojson/processors.py
jalavik: Sure! I guess it wont be to expensive of a function call.

inspire/dojson/institutions/fields/bd90x99x.py
jalavik: Indeed!

inspire/dojson/processors.py
jalavik: So, this is like parameters to the functions specified in the `RECORD_PROCESSORS` config (suggested by Javier). Since `RECORD_PROCESSORS` is already meant to be overridden on the overlay, this `CFG_DOJSON_MAPPINGS ` could live in the overlay only and not needed to be upstream. And then it could even just be hard-coded in the function we have on the overlay since it is a customisation anyway, if that make sense. 

On the other hand, we could suggest a patch to `invenio-records` to make the default `convert_records` use `CFG_DOJSON_MAPPINGS` config. But I suggest we merge this as is now and see when `invenio-records` matures.

PR #271, opened 2015-07-15
dojson: hep2marc addition

jmartinm:
* Adds hep2marc model to allow generating recstruct from our
  HEP data model.

* Adds legacy_export_as_marc to generate marcxml from recstruct.

* Adds xm format and marcxml Jinja2 filter.

### DIFF
--- a/inspire/dojson/common/base.py
+++ b/inspire/dojson/common/base.py
@@ -24,7 +24,7 @@
 
 from dojson import utils
 
-from ..hep.model import hep
+from ..hep.model import hep, hep2marc
 from ..institutions.model import institutions
 from ..experiments.model import experiments
 from ..journals.model import journals
@@ -39,6 +39,12 @@ def control_number(self, key, value):
     return value[0]
 
 
+@hep2marc.over('001', 'control_number')
+def control_number2marc(self, key, value):
+    """Record Identifier."""
+    return [value]
+
+
 @institutions.over('agency_code', '^003')
 @hep.over('agency_code', '^003')
 @experiments.over('agency_code', '^003')
@@ -48,6 +54,12 @@ def agency_code(self, key, value):
     return value[0]
 
 
+@hep2marc.over('003', 'agency_code')
+def agency_code2marc(self, key, value):
+    """Control Number Identifier."""
+    return [value]
+
+
 @institutions.over('date_and_time_of_latest_transaction', '^005')
 @hep.over('date_and_time_of_latest_transaction', '^005')
 @experiments.over('date_and_time_of_latest_transaction', '^005')
@@ -55,6 +67,12 @@ def agency_code(self, key, value):
 def date_and_time_of_latest_transaction(self, key, value):
     """Date and Time of Latest Transaction."""
     return value[0]
+
+
+@hep2marc.over('005', 'date_and_time_of_latest_transaction')
+def date_and_time_of_latest_transaction2marc(self, key, value):
+    """Date and Time of Latest Transaction."""
+    return [value]
 
 
 @institutions.over('url', '^856.[10_28]')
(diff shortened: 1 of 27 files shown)


jmartinm:
If you want to have a look @kaplun @jalavik this allows to generate MarcXML from the JSON, for the moment done for HEP.

You can do /export?recid=123 to see the XML generated as at the moment the /record/123/export/xm view is broken.

jalavik:
Cool stuff!

I see that the hep2marc mapping is put together with the regular hep as well. I guess this is fine as the mapping back and forth are next to each other and then easier to maintain. However, I wanted to mention that if we add more exports with dojson such as `hep2oaidc` or `hep2harvmarc` in the future, it can be beneficial to keep them in separate folders under `dojson`.

inspire/dojson/common/base.py
```diff
@@ -39,6 +39,12 @@ def control_number(self, key, value):
     return value[0]
 
 
+@hep2marc.over('001', 'control_number')
+def control_number2marc(self, key, value):
+    """Record Identifier."""
+    return [value]
```
jirikuncar: do you really want a list?

jmartinm: Not really, what confused me was that:

```python
In [1]: from dojson.contrib.marc21.utils import create_record, split_blob

In [2]: for data in split_blob(open("/Users/jmartinm/Desktop/xm.xml").read()):
    record = create_record(data)
   ...:

In [3]: record
Out[3]:
{'001': ['1381023'],
 '005': ['20150714113421.0'],
 u'020__': [{'a': '978-3-319-18335-0', 'b': 'online'},
  {'a': '978-3-319-18334-3', 'b': 'print'}],
 u'035__': {'9': 'INSPIRETeX', 'a': 'Proceedings:2015hya'},
 u'100__': {'a': u'P\xfctzfeld, Dirk', 'e': 'ed.'},
 u'245__': {'a': 'Proceedings, 524th WE-Heraeus-Seminar: Equations of Motion in Relativistic Gravity (EOM 2013)',
  'b': 'Bad Honnef, Germany, February 17-23, 2013'},
```

Maybe https://github.com/inveniosoftware/dojson/blob/52c052717a175098331d1224e42b7747777eb764/dojson/contrib/marc21/utils.py#L27 should be amended to not produce a list for the controlfields or is intended?

kaplun: `controlfield` are per definition unique. So this shall really be reflected also in `create_record()`

jirikuncar: see https://github.com/inveniosoftware/dojson/blob/52c052717a175098331d1224e42b7747777eb764/dojson/contrib/marc21/fields/bd00x.py#L15

jmartinm: I guess the question here is if ```create_record``` should put the controlfields inside of a list as it is currently doing. (and avoid `return value[0]` doing just `return value`

jirikuncar: You move the logic from individual functions to `create_record` which might not be desirable for liberal MARC (https://github.com/inveniosoftware/dojson/issues/26).

kaplun: @jirikuncar what do you mean? Anyway repeatable `controlfield` is really not valid MARC even in the case of liberal. (I mean, even without attaching any semantic to fields and subfields, a controlfield MUST be unique in any circumstances), so is perfectly fine to have that logic in `create_record`. Any way it's just a convention. As long as all layers agree on what to expects (currently lists even for controlfields).

jirikuncar: Well you would need to teach `create_record` what are `controlfield`s and then you do additional test for all non-`controlfield`s. Moreover you loose flexibility if you would like to validate that any `controlfield` is not repeated. howgh

jmartinm:
Merged https://github.com/inspirehep/inspire-next/commit/ebe832e5686b2dddb1cfb7df46a26bbb516e85b1

PR #285, opened 2015-07-28
base: requirements.txt update

jmartinm:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 git+https://github.com/inspirehep/invenio@labs-master-elasticsearch#egg=Invenio
-git+https://github.com/inveniosoftware/invenio-query-parser@master#egg=invenio-query-parser
+git+https://github.com/inveniosoftware/invenio-query-parser@master#egg=invenio-query-parser==0.2.1.dev20141210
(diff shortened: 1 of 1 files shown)


jmartinm:
This seems to be the way to go to specify given version. Travis build passes, the right version of package  gets installed and we avoid using dependency links which [(might) disappear](https://github.com/pypa/pip/issues/2023)

jalavik:
Cool. So I see that they have a '+' in front. Shall we add it to the other similar lines?

jmartinm:
Note that all packages when a new release is done have a commit like: https://github.com/inveniosoftware/invenio-records/commit/571617218fd0430c551100f1d171c0dde38dd812

so we can then specify the new version if we need it.

jmartinm:
I think that's just how the diff looks like? https://github.com/jmartinm/inspire-next/blob/master-elasticsearch-with-requirements/requirements.txt

There are no '+' in front.

jalavik:
Haha yeah, sorry my bad.

jmartinm:
Merged https://github.com/inspirehep/inspire-next/commit/104ee45df14578f16b03094ce825c59cda6a9f51

PR #288, opened 2015-08-04
Prediction astro update

jalavik:
Improvements to predictions and new arXiv prediction model for astro-ph.


glouppe:
Not sure "predicter" is a better name -- it is non standard, and I am not even sure this is a proper English word.

inspire/modules/workflows/workflows/process_record_arxiv.py
```diff
@@ -193,7 +193,7 @@ def get_sort_data(bwo, **kwargs):
             if decision == "CORE":
                 relevance_score += 10
             elif decision == "Rejected":
-                relevance_score -= 10
+                relevance_score = (max_score * -1) - 10
```
glouppe: Just out of curiosity, what is this relevance score?

jalavik:
@glouppe Yeah, predicter is not ideal. Needed to change from classifier as it was colliding with the Invenio classifier module. Feel free to suggest a name :+1:

inspire/modules/workflows/workflows/process_record_arxiv.py
jalavik: This is a arbitrary value is used for sorting, so that when sorting by the score in Holding Pen it allows to sort by most relevant records. Say confident CORE predictions on top, and most confident Rejection at bottom.

jalavik:
Maybe `predictor` is the more correct form of "predicter". http://dictionary.reference.com/browse/predictor

jalavik:
Merged for now as it is deployed. Can rename later.

PR #290, opened 2015-08-05
workflows: interface changes for editing

Lilykos:
* Adds edit buttons on all editable elements (title, subj, urls).

* Reworks the way accept/decline works, makes it more intuitive.

Signed-off-by: Ilias Koutsakis <ilias.koutsakis@cern.ch>


jalavik:
Thanks. Merged

PR #294, opened 2015-08-12
dojson: source and curated fields

kaplun:
* Introduces source and curated fields everywhere, so that it is
  possible to cleanly distinguish between arXiv and publisher
  information. Converte to list of objects when necessary.


kaplun:
@tsgit, @aw-bib, @ksachs, @annetteholtkamp this introduces the `source` and `curated` field, everywhere. Unfortunately it's not easy to see the diffed code, because of a reordering of the JSON.

kaplun:
This is no longer useful

PR #303, opened 2015-08-18
deposit: prediction in literature submission

jalavik:


glouppe:
would be curious to see that live :)

jalavik:
Yeah, I've tested with arXiv imports only which are already quite complete, so we'll see how it behaves with regular user submissions..

PR #312, opened 2015-08-24
template: fixes title problem

ioannistsanaktsidis:


jmartinm:
In the end we merged the updated version of the template into `master-rc`

PR #318, opened 2015-08-25
utils: implements latex(eu,us)

ioannistsanaktsidis:
* Implements Ajax call for latex_eu, latex_us.

* Refactors latex code.

### DIFF
--- a/inspire/utils/bibtex_booktitle.py
+++ b/inspire/utils/bibtex_booktitle.py
@@ -17,7 +17,7 @@
 ## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 #
 from invenio_records.api import get_record
-from invenio.legacy.search_engine import search_pattern
+from invenio_search.api import Query
 
 
 def traverse(o, tree_types=(list, tuple)):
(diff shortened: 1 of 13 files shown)


inspire/utils/bibtex_booktitle.py
```diff
@@ -43,8 +43,8 @@ def generate_booktitle(record):
                         if acronym:
                             booktitle = "%s: %s" % (rn, acronym, )
                         else:
-                            recids = search_pattern(p="reportnumber:%s"
-                                                    % (rn,))
+                            recids = Query(p="reportnumber:%s" % (rn,))\
```
jmartinm: I think if you want the same behaviour as before this should be:

```python
Query(p="reportnumber:%s" % (rn,)).search().recids
```

In order to get record ids back. Otherwise you might receive back the full json by default. Did you check?

ioannistsanaktsidis: I was getting back a search results object , I will double check to be sure.

2015-08-25 14:46 GMT+02:00 Javier Martin Montull <notifications@github.com>:

PR #321, opened 2015-08-26
predicter: classification without categories

glouppe:
This PR adds the following features:

* Train the classifier without using the "categories" field, by setting `use_categories=False` in `train`.
* Add the option of returning the top words explaining the classifier prediction in `predict`.
* Add the `closest` method for finding the closest records with respect to another given record.

CC: @jalavik @jacquerie


jalavik:
Thanks. Merged to master as fffe97e (PS: was already deployed on qa/production branches)

PR #329, opened 2015-08-31
workflows: use available data and extra_data

jacquerie:
Depends on https://github.com/inveniosoftware/invenio-workflows/pull/11


jacquerie:
Fixed.

jacquerie:
Failure appears related to https://github.com/inveniosoftware/invenio/pull/3468. Maybe we haven't fetched it from upstream?

PR #330, opened 2015-09-01
Production sync back on track

jalavik:
This puts us back on track with the INSPIRE production sync.

It adds a workflow task that can store records into the system with custom record IDs.


jmartinm:
Nice, for new records added in INSPIRE, will we keep the same record id in labs with this code?

jalavik:
@jmartinm Yeah, so records with an existing `control_number` will keep it upon the initial import. Any additional updates will also update the records.

jmartinm:
I see. Looking at the code seems good to me.

jalavik:
Merged. Note that for record updates to index properly, we depend on these fixes: https://github.com/inveniosoftware/invenio-collections/pull/5 and https://github.com/inveniosoftware/invenio-records/pull/28

PR #336, opened 2015-09-02
brief: fixes abstract problem

ioannistsanaktsidis:

### DIFF
--- a/inspire/base/templates/format/record/Inspire_Default_HTML_brief_macros.tpl
+++ b/inspire/base/templates/format/record/Inspire_Default_HTML_brief_macros.tpl
@@ -47,27 +47,41 @@
 {% macro record_abstract() %}
   {% set exists = [] %}
   {% if record.get('report_number') %}
-  {% if record.get('report_number') | is_list() %}
-  {% for i in record.get('report_number') %}
-  {% if i.get('source') == 'arXiv' %}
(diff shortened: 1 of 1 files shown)


jmartinm:
What was the problem exactly? Please, add a bit longer description in the commit

jmartinm:
ping @ioannistsanaktsidis was this needed?

ioannistsanaktsidis:
@jmartinm sorry for not replying, I just saw the messages. The problem was that sometimes the record had 2 abstracts e.g the arxiv one and another one. Till now both of them appeared. The commit ensures that when the arxiv one exists it will be the one displayed.
Cheers!

jmartinm:
Merged f4cfe202aa0397877734e3a270a0bd215389602a

annetteholtkamp:
Shouldn’t it be the other way around? Show publisher abstract with priority, arxiv as a clickable link?

- Annette

jmartinm:
You are right @annetteholtkamp 

@ioannistsanaktsidis if there are 2 abstracts, in the brief results the one from the publisher should be the one shown (see the current logic here https://github.com/inspirehep/inspire/blob/master/bibformat/format_elements/bfe_INSPIRE_abstract.py#L75)

And then in the detailed view we can show the second abstract ( that you can leave for @nikpap )

PR #337, opened 2015-09-02
dojson: institutions data model update

jalavik:
* Updates the dojson conversion, schema and index mappings.

### DIFF
--- a/inspire/dojson/institutions/schemas/insitutions-0.0.1.json
+++ b/inspire/dojson/institutions/schemas/insitutions-0.0.1.json
@@ -8,13 +8,53 @@
             "type": "string"
         },
         "address": {
-            "description": "Street etc, city with postal code + additions (native language) FIXME: isn't postal code redundant?",
-            "title": "Address",
-            "type": "string"
-        },
-        "city": {
-            "title": "City",
-            "type": "string"
+            "items": {
+                "properties": {
+                    "address": {
+                        "description": "Street etc, city with postal code + additions (native language) FIXME: isn't postal code redundant?",
+                        "title": "Address",
+                        "type": "string"
+                    },
+                    "city": {
+                        "title": "City",
+                        "type": "string"
+                    },
+                    "state_province": {
+                        "title": "State or Province",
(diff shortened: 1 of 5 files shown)


inspire/dojson/institutions/schemas/insitutions-0.0.1.json
```diff
@@ -8,13 +8,53 @@
...
+                        ],
+                        "title": "Country code",
+                        "type": "string"
+                    },
+                },
+                "title": "Address",
+                "type": "object"
+            },
+            "title": "Addresses",
+            "type": "array",
```
kaplun: Hi @jalavik, when I created the model, I though of keeping it as flat as possible in order to simplify coding and configuration. So e.g. if the city is only one I had decided to put it on the root level directly. Why have you decided to structure them in a deeper way?

inspire/dojson/institutions/schemas/insitutions-0.0.1.json
```diff
@@ -129,11 +168,6 @@
             "type": "array",
             "uniqueItems": true
         },
-        "non_public_note": {
-            "description": "FIXME: what is the difference with hidden_note!?",
-            "title": "Non public note",
-            "type": "string"
-        },
```
kaplun: This was corresponding to a different MARC field that the hidden note. Have you matched both to `hidden_note` in dojson?

inspire/dojson/institutions/schemas/insitutions-0.0.1.json
jalavik: I thought of it this way to prepare for possibly multiple addresses, so that you can then have coherent address structures (associated address, city, state, country in the same structure). We can always change it back if it's not feasible though. :-)

inspire/dojson/institutions/schemas/insitutions-0.0.1.json
jalavik: Yeah, I noticed. This takes it from the different MARC field. Apparently no jobs have 595 fields.

PR #338, opened 2015-09-02
converter: add encoding task

jalavik:
* Adds an encoding task for use in the INSPIRE production
  synchronization workflow as the data coming from INSPIRE
  is given as ISO-8859-1.

### DIFF
--- a/inspire/modules/converter/tasks.py
+++ b/inspire/modules/converter/tasks.py
@@ -1,3 +1,25 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2015 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


kaplun:
What!? :worried: Inspire production is sending metadata not in UTF8? How come?

jalavik:
@kaplun Yeah, was also surprised. Maybe it was only that case, but we'll have to see. The record I noticed it with is https://inspirehep.net/record/905829/export/hm.

I could not see anything obvious in the oairepository legacy code (although I did not look around for long..), but requests reported encoding being in ISO-8859-1.

kaplun:
To me it looks UTF8:
https://inspirehep.net/oai2d?verb=GetRecord&identifier=oai:inspirehep.net:905829&metadataPrefix=marcxml

If I convert it to UTF8 assuming it is ISO-8599-1 I obtain all wrong accents such as:
```xml
    <marc:subfield code="a">Ãcole polytechnique fÃ©dÃ©rale de Lausanne (EPFL)</marc:subfield>
```

Is it possible that requests simply assumes it's not UTF8 because it finds non ASCII chars and there is no encoding HTTP header?

Probably simply by adding such header on legacy would solve the issue.

jalavik:
So, something like this? https://github.com/inspirehep/invenio/pull/264

kaplun:
I think so.

PR #339, opened 2015-09-03
Jobs data model updates

jalavik:
Adds initial mapping for Jobs records from INSPIRE production.

Fixes some issues with Institution notes @kaplun 

Adds acquisition_source to hepnames mapping @jmartinm


inspire/base/searchext/mappings/jobshidden.json
```diff
@@ -0,0 +1,119 @@
+{
```
kaplun: Do we need to separate jobshidden and regular jobs?

Hidden jobs are simply expired jobs, but they have the same schema. So we could simply solve this by using native invenio collections? (we could simply restrict hidden jobs).

@jalavik, @jmartinm ?

kaplun: In particular, Jobs are moving from the Jobs collection to a hidden one, simply based on a check over the metadata. So we might want to preserve this easyness.

inspire/dojson/jobs/schemas/jobs-0.0.1.json
```diff
@@ -33,15 +33,15 @@
...
             "format": "date",
             "title": "Application deadline",
             "type": "string"
         },
         "experiment": {
             "description": "FIXME: we can simply store here the ID of the experiment.",
-            "title": "Experiment recid",
-            "type": "integer"
+            "title": "Experiment associated with Job",
+            "type": "string"
```
kaplun: We should rather use a $ref to experiment :-) maybe this could be done in one PR for all schemas.

inspire/dojson/jobs/schemas/jobs-0.0.1.json
```diff
@@ -102,18 +102,46 @@
...
+                    },
+                    "size": {
+                        "type": "integer",
+                        "title": "Size"
+                    }
+                },
+                "title": "URL"
+            },
+            "type": "array",
+            "title": "URLs"
```
kaplun: Do we need all this? In current usage we really have only 8564_u been filled, i.e. a URL pointing to the web page for the job.

jalavik: Maybe not. Currently the base.py under common has the `url` key mapped to this. So that's why jobs and inst was changed to follow the same schema. We could take it out of common and keep it different per "collection".

inspire/base/searchext/mappings/jobshidden.json
jalavik: So yeah, I dunno how this will work nowadays WRT restricted records and elasticsearch. I presume it will be done on the post-process of Invenio before displaying results, so we could then keep all jobs in the same index and have some custom check there to restrict "hidden" ones.

I cannot recall how Jobs move to Jobs hidden, but I thought it was a manual curator thing. If it is just a random script or cronjob, I guess we could re-implement that as a daily celery task to move around jobs. But we are sure as long as the deadline date is reached, we move them to hidden? I figured it's a manual notice from the submitter saying that "hey, we hired someone" or "hey, you can remove the notice now".

inspire/dojson/jobs/schemas/jobs-0.0.1.json
jalavik: Yeah, I dunno how that works, wanted to talk to you about it. That also goes for the aquisition_source specific json schema and how to embed it.

inspire/dojson/jobs/schemas/jobs-0.0.1.json
kaplun: I think in case of jobs (and maybe everything but HEP) this is necessary. The thing is that in the paper schema, the URL is needed to point to the actually PDF, while in all other types it is more to point to the official reference homepage (for the job, the institutution...). So they are slightly different usecases, and I'd keep them separate.

P.s. in case we want to exploit more and more the common fields, we should start using $ref and friends :-)

jalavik: Aight. So if we want to call it `url` across the board, I'll move it out of common and stick the advanced schema to hep only and fill the others as necessary.

jalavik:
@kaplun @jmartinm  See latest commit for URL mapping updates.

inspire/base/searchext/mappings/jobshidden.json
kaplun: Yep, currently it's simply based on a flag, controlled by a cataloguer:

980__a:JOB Vs. 980__a:JOBHIDDEN

I assume it will still stay like that, i.e. a curator will actively flip the switch. Maybe we can consider adding a regular task to alert cataloguers about expired job positions.

PR #345, opened 2015-09-07
brief: fixes abstract problem

ioannistsanaktsidis:
* Displays the publisher abstract when present.
  If not, displays the arxiv one.

* Displays only one abstract.


ioannistsanaktsidis:
@jmartinm this patch takes the arxiv one when no other abstract is present. In case there is another it takes the other.

jmartinm:
Merged 6c2b99769bf4c8ffa2c8e6602107e31844d8dc54

PR #347, opened 2015-09-08
WIP authors: uuid generator

jacenkow:
* New method, which generates uuid of authors in HEP, on record-insert.

### DIFF
--- a/inspire/modules/authors/receivers.py
+++ b/inspire/modules/authors/receivers.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2015 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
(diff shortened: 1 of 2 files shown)


inspire/modules/authors/receivers.py
```diff
@@ -0,0 +1,37 @@
...
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#
+
+import uuid
+
+from invenio_records.signals import before_record_insert
+
+
+@before_record_insert.connect
+def assignUuid(sender, *args, **kwargs):
```
jmartinm: please use `assignuuid` as name of the function as Python convention is not camel case.

jmartinm:
Looks good @jacenkow 

Can you have a look at this signal https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/signals.py#L67 so that we also add `uuids` when a record is updated.

jacenkow:
Fixed @jmartinm

jmartinm:
Tested and merged as 4cb98cce0f5988a22a59c27c4373fd7af4c3f682

jmartinm:
Note that I amended the commit message so that the description does not go over 70 lines

PR #352, opened 2015-09-14
demosite: brand new demo records

kaplun:
* Introduces new demo records by using the new script:
  inspirehep/inspire-scripts/utility/generate_demo_records.py.

* Contains the 10 most cited hep papers, 10 core books, theses,
  proceedings, notes, corresponding cited papers, 10 citing papers
  for each paper, 10 collaboration papers.

* Contains all the HepNames record referred across the papers.

* Expands the MARCXML of the publications, by adding 2 new subfields
  in the 100 and 700 tags:
  $x: containing the record ID of the author HepName
  $y: containing 1 if the given ID is claimed/checked, 0 otherwise


kaplun:
@jalavik , WDYT?
@jacenkow here are the records with additional $x and $y as agreed.

kaplun:
Note, due to the addition of involved HepNames and having followed references these demo set is made of 8709 records. It's a bit large, but should be digestible by the new Invenio.

jalavik:
Looks good. Indeed it's a bit big, but it should be fine.

$x and $y are only present in those linked to a HepNames correct? Those without are "unmatched"?

kaplun:
> $x and $y are only present in those linked to a HepNames correct? Those without are "unmatched"?

Exactly.

kaplun:
@Panos512 you can also use this demo records as I have built them to follow reference and citations. So you should have quite some good examples to test if the citation machinery works.

jalavik:
@kaplun As the `rc` branches got killed, would you mind reopening this for `master`? Sorry for the inconvenience.

kaplun:
Move do #357 :+1:

PR #357, opened 2015-09-15
demosite: brand new demo records

kaplun:
* Introduces new demo records by using the new script:
  inspirehep/inspire-scripts/utility/generate_demo_records.py.

* Contains the 10 most cited hep papers, 10 core books, theses,
  proceedings, notes, corresponding cited papers, 10 citing papers
  for each paper, 10 collaboration papers.

* Contains all the HepNames record referred across the papers.

* Expands the MARCXML of the publications, by adding 2 new subfields
  in the 100 and 700 tags:
  $x: containing the record ID of the author HepName
  $y: containing 1 if the given ID is claimed/checked, 0 otherwise

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -261,7 +261,7 @@ You can now load the INSPIRE demo records:
 .. code-block:: console
 
     (inspire)$ cdvirtualenv src/inspire-next
-    (inspire)$ inveniomanage records create inspire/demosite/data/demo-records.xml -t marcxml
+    (inspire)$ inveniomanage migrator populate -t marcxml -f inspire/demosite/data/demo-records.xml --force
 
 
 Now you should have a running INSPIRE demo site running at `http://localhost:4000 <http://localhost:4000>`_!
--- a/inspire/dojson/experiments/fields/bd1xx.py
+++ b/inspire/dojson/experiments/fields/bd1xx.py
@@ -90,7 +90,14 @@ def spokesperson(self, key, value):
 @experiments.over('collaboration', '^710..')
 def collaboration(self, key, value):
     """Collaboration of experiment."""
-    return value.get("g")
+    if isinstance(value, list):
+        collaborations = sorted((elem["g"] for elem in value if 'g' in elem), key=lambda x: len(x))
+        if len(collaborations) > 1:
+            self['collaboration_alternative_names'] = collaborations[1:]
+        if collaborations:
+            return collaborations[0]
+    else:
+        return value.get("g")
 
 
 @experiments.over('urls', '^856.[10_28]')
--- a/inspire/dojson/experiments/schemas/experiments-0.0.1.json
+++ b/inspire/dojson/experiments/schemas/experiments-0.0.1.json
@@ -22,6 +22,15 @@
             "title": "Collaboration",
             "type": "string"
         },
+        "collaboration_alternative_names": {
+            "title": "Collaboration alternative names",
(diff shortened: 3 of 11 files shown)


jmartinm:
So it seems the blocker for merging these records is being able to workaround hardcoding the record ids.

This is necessary for the demo site in order to have correct links among records.

jmartinm:
By the way, how are we going to do so that the demo records contain all the references and citations in the JSON so that it is ready to be displayed?

kaplun:
> By the way, how are we going to do so that the demo records contain all the references and citations in the JSON so that it is ready to be displayed?

At the moment, these are available in the MARCXML as part of $0 field of each reference. Of course all those references that currently don't have an explicit link in the MARCXML will not have a valid link in JSON.

kaplun:
@jalavik, @jmartinm, @Panos512 the new demo records are now passing, but there are some failures in the unittests due to the fact that record IDs are no longer the first few small records identifiers.

E.g.
```python
=================================== FAILURES ===================================
________________ CitationsTests.test_citations_references_empty ________________
self = <inspire.modules.citations.testsuite.test_citations.CitationsTests testMethod=test_citations_references_empty>
    def test_citations_references_empty(self):
        """Test if record is working corectly when all citations are deleted"""
        from invenio_records.api import get_record
        rec8 = get_record(8)
>       self.assertEqual(rec8.get("references_id"), [])
E       AttributeError: 'NoneType' object has no attribute 'get'
inspire/modules/citations/testsuite/test_citations.py:97: AttributeError
_____________ CitationsTests.test_citations_references_nonexistant _____________
self = <inspire.modules.citations.testsuite.test_citations.CitationsTests testMethod=test_citations_references_nonexistant>
    def test_citations_references_nonexistant(self):
        """Test if record is working corectly without any citations"""
        from invenio_records.api import get_record
        rec3 = get_record(3)
>       self.assertEqual(rec3.get("references_id"), None)
E       AttributeError: 'NoneType' object has no attribute 'get'
inspire/modules/citations/testsuite/test_citations.py:91: AttributeError
_______________ CitationsTests.test_citations_references_updated _______________
self = <inspire.modules.citations.testsuite.test_citations.CitationsTests testMethod=test_citations_references_updated>
    def test_citations_references_updated(self):
        """Test if record is populated corectly with new citations"""
        from invenio_records.api import get_record
        rec1 = get_record(1)
>       self.assertEqual(rec1.get("references_id"), [2, 3])
E       AttributeError: 'NoneType' object has no attribute 'get'
inspire/modules/citations/testsuite/test_citations.py:85: AttributeError
```

I am thinking whether in the `setUp()` phase we should allocate 10 empty records with IDs from 1 to 10 , and delete those upon `tearDown()` but this seems a bit clumsy. Any better idea?

Panos512:
Yeah this could be done. Or we could add them to the end (from the last id and on) so that they will not collide with useful data.

jalavik:
Another way could also be to change the mock data to work with the recids of the demo records. These are loaded before tests are run, so it could work. Otherwise setup indeed needs to allocate the appropriate records.

kaplun:
This depends on https://github.com/inspirehep/inspire-next/issues/426

kaplun:
@jalavik this is still not mergable because of tests failing because of invenio-records thingy.

jmartinm:
Did a small modification here and the tests are passing https://github.com/inspirehep/inspire-next/pull/460

PR #358, opened 2015-09-16
dojson: profile and claimed fields

jacenkow:
* Changed hep schema to handle new fields: author's profile and claimed
  status.

* Added method to create hep authors' profiles links based on profile_id.

* Hepnames' start_year and end_year handle ValueErroe error exception.

* Removed controlfields from sample XML file.

### DIFF
--- a/inspire/dojson/hep/fields/bd1xx.py
+++ b/inspire/dojson/hep/fields/bd1xx.py
@@ -24,6 +24,7 @@
 
 from dojson import utils
 
+from inspire.dojson import utils as inspire_dojson_utils
 from ..model import hep, hep2marc
 
 
@@ -44,6 +45,9 @@ def authors(self, key, value):
                 'affiliation': utils.force_list(
                     value.get('u')
                 ),
+                'profile': inspire_dojson_utils.create_profile_url(
+                    value.get('x')),
+                'claimed': value.get('y')
             }
 
     authors = self.get('authors', [])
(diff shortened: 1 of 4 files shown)


kaplun:
Hi @jacenkow, please do not include the giant MARCXML with this commit. I would prefer to including it with my separate PR when we have first solved the recid 001 issue.

kaplun:
Can you amend commit log to speak with present tense:

E.g.:

```
* Changes HEP schema...
* Adds method ...
* Handles HEPNames' ...
```

inspire/dojson/hep/fields/bd1xx.py
```diff
@@ -24,6 +24,7 @@
 
 from dojson import utils
 
+from inspire.dojson import utils as utils_
```
kaplun: Maybe you can rename this to `inspire_dojson_utils` so that it's clear what they are about.

inspire/dojson/utils.py
```diff
@@ -74,3 +74,13 @@ def encode_for_marcxml(value):
                 export += ['\t</datafield>\n'.expandtabs(tabsize)]
     export += ['</record>\n']
     return "".join(export)
+
+
+def create_profile_url(profile_id):
```
kaplun: Can you add RestructeredText docstring to describe this function?

kaplun:
Beside that, very good!

kaplun:
Ah! Would be great if you can add unit tests to test your code. See: https://docs.python.org/2/library/unittest.html and https://github.com/inspirehep/inspire-next/tree/master/inspire/testsuite

PR #367, opened 2015-09-21
dojson: earliest_date field

jacenkow:
* Adds earliest_date field to HEP schema.

* New method for generating earliest_date on before_record_insert and
  before_record_update signals.

* New methods for formating dates in utils.py

### DIFF
--- a/inspire/dojson/hep/receivers.py
+++ b/inspire/dojson/hep/receivers.py
@@ -1,0 +1,61 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2015 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+from inspire.dojson import utils as inspire_dojson_utils
+
+from invenio_records.signals import (
+    before_record_insert,
+    before_record_update,
+)
+
+
+@before_record_insert.connect
+@before_record_update.connect
+def earliest_date(sender, *args, **kwargs):
+    """ Find and assign the earliest date to a HEP paper. """
+    dates = []
+
+    if 'preprint_info' in sender:
+        for preprint_info_key in sender['preprint_info']:
+            if 'date' in preprint_info_key:
+                dates.append(preprint_info_key['date'])
+
+    if 'defense_date' in sender:
+        for defense_date_key in sender['defense_date']:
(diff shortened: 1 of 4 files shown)


jalavik:
@jacenkow What's the status here? I see you have wip in the last commit. Shall we wait with merge or?

jacenkow:
@jalavik I screwed up and instead of two pull requests I made one, however it is ready to merge. The WIP thingy is the beginning of machine learning phonetic notation, which I will push after solving issues with Travis.

jalavik:
@jacenkow Alright, can you update commit log message to not be WIP, rebase with master, and I will merge.

jacenkow:
@jalavik Done

inspire/dojson/hep/receivers.py
```diff
@@ -0,0 +1,79 @@
...
+                        thesis_key['date']))
+
+    if 'publication_info' in sender:
+        for publication_info_key in sender['publication_info']:
+            if 'year' in publication_info_key:
+                dates.append(
+                    inspire_dojson_utils.full_date_format(
+                        publication_info_key['year']))
+
+    if 'references' in sender:
```
jalavik: I have now tested this branch locally, and I think the logic of getting the year from references is flawed. Simply because it is the date of other papers (and not this one!). E.g. if a paper references an old paper from 1970, then with the current logic the earliest_date is `1970-01-01`. Which clearly is wrong.

inspire/dojson/utils.py
```diff
@@ -85,3 +85,23 @@ def create_profile_url(profile_id):
         return base_url + str(profile_id)
     except (TypeError, ValueError):
         return ''
+
+
+def full_date_format(date):
```
jalavik: I would recommend a different way of doing the date parsing by using existing libraries. E.g. `datetime` and `dateutil`.

```python
from dateutil.parser import parse
import datetime

# Need default to get proper default values if parts are missing
parse("2015-01-22", default=datetime.datetime(year=1900, day=1, month=1))
# Out[10]: datetime.datetime(2015, 1, 22, 0, 0)

parse("2015-12", default=datetime.datetime(year=1900, day=1, month=1))
# Out[11]: datetime.datetime(2015, 12, 1, 0, 0)

parse("2015", default=datetime.datetime(year=1900, day=1, month=1))
# Out[12]: datetime.datetime(2015, 1, 1, 0, 0)

# Then use strftime to print the correct format
parse("2015", default=datetime.datetime(year=1900, day=1, month=1)).strftime("%Y-%m-%d")
# Out[13]: '2015-01-01'
```

jacenkow: :+1:

jacenkow:
@jalavik Updated and tested. Cheers for the hint.

jalavik:
@jacenkow Awesome. Thanks for the fixes and patience.

I will have to ask you for a bit more patience as I have some more comments:

1. Can you squash the commit into one (since they are now both having almost the same description) and add the lost bullet point of `* Adds ml_block and uuid to authors in HEP schema`

2. Now the dates makes more sense, but there is a case where `earliest_date` could be better. E.g. when a less specific date is chosen of a more specific one. E.g. in one record the result looks like this:

```json
{
   "_index": "hep",
   "_type": "record",
   "_id": "16",
   "_score": 1,
   "fields": {
      "preprint_info.date": [
         "2010-09-03"
      ],
      "publication_info.year": [
         "2010"
      ],
      "earliest_date": [
         "2010-01-01"
      ]
   }
}
```

As you can see. Since `publication_info.year` is a less specific field it sort of nullifies the month and day information from the preprint_info. I know this a tricky one, because the publication date might still be different than the preprint date. So we can ignore this one off for now in case it requires further discussion @kaplun

PR #368, opened 2015-09-22
Catchup master from qa/production

jalavik:
Bundle of commits missing in `master` from `qa` and  `production`.


jalavik:
Merged.

PR #369, opened 2015-09-22
authors: BAI generation function

glouppe:
This PR adds a reimplementation of the BAI generation function. 

Note: it does not handle the numbered suffix to be added at the end of the BAI. I am not sure how to do that properly in fact, as this operation needs to have the exact number of profiles with the same BAI currently the db (i.e., you cannot rely on ES since it can be out of sync). 

CC: @kaplun @jacenkow

### DIFF
--- a/inspire/modules/authors/testsuite/test_utils.py
+++ b/inspire/modules/authors/testsuite/test_utils.py
@@ -1,0 +1,61 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014, 2015 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Tests for utils."""
+
+from ..utils import bai
+
+from invenio.testsuite import make_test_suite, run_test_suite, InvenioTestCase
+
+
+class BaiTests(InvenioTestCase):
+
+    """Test the BAI generation function."""
+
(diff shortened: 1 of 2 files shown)


jalavik:
I think you'll need to add `beard` to requirements here: https://github.com/inspirehep/inspire-next/blob/master/setup.py#L32

kaplun:
> I am not sure how to do that properly in fact, as this operation needs to have the exact number of profiles with the same BAI currently the db (i.e., you cannot rely on ES since it can be out of sync).

@glouppe the proper way is to register the created ID with the pidstore. See e.g. how invenio-records does in https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/providers/recid.py

glouppe:
@jalavik Hmm, then why does this exist? https://github.com/inspirehep/inspire-next/blob/master/inspire/modules/predicter/arxiv.py#L42

glouppe:
> @glouppe the proper way is to register the created ID with the pidstore. See e.g. how invenio-records does in https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/providers/recid.py

I dont get it. You need to count how many profile have the same prefix. Pidstore is not meant for that.

jalavik:
@glouppe Yeah, it was only used in particular harvesting workflows in the beginning and due to the heavy dependencies of `numpy` and `sklearn` it was put as optional in `requirements-harvesting.txt`. Nowadays however, we should just add it.

jalavik:
@glouppe So @jacenkow was also looking at adding beard as a dependency but ran into trouble with Travis. So we might just go the Docker way as presented this Monday.

kaplun:
> I dont get it. You need to count how many profile have the same bai prefix. Pidstore is not meant for that.

Well `pidstore` should not be *out of sync*, as it depends on a SQL table where operations should be atomic. Also you can delegate to an ad-hoc `pidstore` provider the responsibility compute what is the next valid `id` (including the number).

kaplun:
Another appropriate tool is `inveniosoftware/invenio-sequencegenerator` which is currently mostly based on the Invenio legacy code.

glouppe:
For adding the numbered suffix using whatever solution, let's do that in another PR, with its proper integration upon author record creation.

kaplun:
:+1:

jalavik:
@glouppe beard dependency is now sorted so if you rebase, hopefully things will go better.

glouppe:
Thanks Jan! I rebased the PR.

glouppe:
Travis is happy!

(Btw I just noticed that test files are included in the coverage. These should be excluded in my opinion. Also, we may want to put all tests inside `inspire/testsuite` dont we?)

jalavik:
Great. I will merge then.

Regarding the test files. Yes, ideally they should live outside the package and not count as coverage. We'll attack this next. In addition, coveralls is not working as well as we are missing sphinx docs.

jacenkow:
@jalavik :clap:

PR #370, opened 2015-09-23
travis: add beard to dependencies

jacenkow:
* Add beard to requirements of setup.py file.

* Add travis-install.sh file, taken from beard git repository.

* Move beard git repository from harvesting to requirements.txt

Signed-off-by: Grzegorz Jacenków <grzegorz.jacenkow@cern.ch>

PR #371, opened 2015-09-23
travis: add beard to dependencies

jalavik:
Continuation from #370

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,16 +30,17 @@ sudo: false
 
 language: python
 
+matrix:
+  include:
+    - python: "2.7"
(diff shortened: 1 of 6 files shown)


glouppe:
I am not sure you want to install Anaconda for testing Inspire. Couldnt you rather have a normal build inside a Docker so that Numpy and Scipy compilations (along with all other depencies) are done once and for all? CC: @eamonnmag

jalavik:
@glouppe Yeah, might be the easiest in the end.

eamonnmag:
Yes, I think this is the way things should be done in future.

jalavik:
Actually it's passing nicely now. So I will merge it for now. Then we work on Docker integration.

glouppe:
Note that conda environments usually dont play that well with virtualenvs... I think the build is mixing both now, which may not be a good idea.

jalavik:
Superseded by #372

PR #376, opened 2015-09-28
authors: phonetic name notation

jacenkow:
* Add new method for assigning phonetic notation of HEP authors'
  names on before_record_insert and before_record_update signals.

* Remove before_record_update signal for assign_uuid method.

### DIFF
--- a/inspire/modules/authors/receivers.py
+++ b/inspire/modules/authors/receivers.py
@@ -3,34 +3,53 @@
 # This file is part of INSPIRE.
 # Copyright (C) 2014, 2015 CERN.
 #
-# INSPIRE is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
 #
-# INSPIRE is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
 #
 # You should have received a copy of the GNU General Public License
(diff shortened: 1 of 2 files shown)


inspire/modules/authors/receivers.py
```diff
@@ -3,34 +3,56 @@
...
+
+                ml_block = block_phonetic(
+                    np_array([name], dtype=np_object).reshape(-1, 1),
+                    threshold=0,
+                    phonetic_algorithm='nysiis'
+                )
+
+                author['ml_block'] = ml_block[0]
+
+        sender['authors'] = authors
```
kaplun: There is no need for this step: `authors` was initialized as a reference to `sender['authors']`  therefore they are already the same object in Python.

inspire/modules/authors/receivers.py
```diff
@@ -3,34 +3,56 @@
...
+            if 'full_name' in author:
+                name = {'author_name': author['full_name']}
+
+                ml_block = block_phonetic(
+                    np_array([name], dtype=np_object).reshape(-1, 1),
+                    threshold=0,
+                    phonetic_algorithm='nysiis'
+                )
+
+                author['ml_block'] = ml_block[0]
```
glouppe: `ml_block` is not a good name in my opinion. I would rather name to something `signature_block`

inspire/modules/authors/receivers.py
```diff
@@ -3,34 +3,56 @@
...
 
+from beard.clustering import block_phonetic
+
 from invenio_records.signals import (
     before_record_insert,
     before_record_update,
 )
 
+from numpy import array as np_array
+from numpy import object as np_object
```
glouppe: Common usage of numpy is to import it using `import numpy as np`. No need to import all functions in the global namespace.

kaplun:
@jacenkow can you implement @glouppe recommendations? Then this PR is IMHO good to be merged.

jacenkow:
:+1:

jacenkow:
Fixed. :santa:

inspire/modules/authors/receivers.py
```diff
@@ -3,34 +3,53 @@
...
 from invenio_records.signals import (
     before_record_insert,
     before_record_update,
 )
 
+import numpy as np
+
 
 @before_record_insert.connect
 @before_record_update.connect
```
kaplun: Actually would be great to move this from being a `before_record*` receiver to become a record enhancer, as soon as the functionality suggested into inveniosoftware/invenio-records#48 is available.

kaplun: See what @Panos512 did for https://github.com/inspirehep/inspire-next/pull/529/files#diff-e0e9a1f8aa763c7b4c65e1b934b67459R30

jacenkow:
Continued a2d64378691dbb8aee0e95d95ce378b47134a699

@kaplun, how do you do the proper reference formatting?

PR #379, opened 2015-09-29
records: WIP xml insertion hack

Panos512:
* Adds a hack to import data from the new .xml file containing 
   ``<controlfield tag="001">``.

### DIFF
--- a/inspire/modules/records/receivers.py
+++ b/inspire/modules/records/receivers.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2015 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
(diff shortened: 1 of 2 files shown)


inspire/modules/records/receivers.py
```diff
@@ -0,0 +1,42 @@
...
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+from invenio.ext.sqlalchemy import db
+
+from invenio_records.models import Record
+
+from invenio_records.signals import before_record_insert
+
+
+@before_record_insert.connect
```
kaplun: Hi @Panos512. Discussing IRL with @jalavik, we came to the conclusion the best fit for hooking up this signal is indeed upon loading of INSPIRE demo records, while disable it otherwise.

For this reason a generic `inveniomanage inspire` (or better name) is going to be created such that we can have e.g.:

`$ inveniomanage inspire demorecords foo.xml` 

In the `demorecords` command, and only there, we can then register the signal.

For this reason, we can pause for a second this PR, to define better the above command, so that you can then hook in the proper place your hack :smile: .

jalavik: When #380 is tweaked and merged, you can hook this up to a `populate` command there.

kaplun:
Hi @Panos512, can you rebase it on top of latest master branch and remove WIP?

inspire/modules/records/receivers.py
kaplun: @Panos512 actually #380 was merged so you can now amend your hook to have it triggered as @jalavik  suggests.

kaplun:
Ah shit :smile:  I was confused. This branch is actually superseeded by #395

Panos512:
Yeah indeed :stuck_out_tongue:

PR #382, opened 2015-09-30
dojson: fix earliest date parsing

jalavik:
Had to make some fixes for `earliest_date` field. @jacenkow

### DIFF
--- a/inspire/dojson/hep/receivers.py
+++ b/inspire/dojson/hep/receivers.py
@@ -17,6 +17,8 @@
 # along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 
+"""Contains signal receivers for HEP dojson processing."""
+
 from inspire.dojson import utils as inspire_dojson_utils
 
 from invenio_records.signals import (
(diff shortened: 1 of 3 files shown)


jmartinm:
Looks :+1: 

Were these changes needed due to the `dojson` updates just merged?

jalavik:
It has been bugged from the start I think. I discovered it when doing a submission.

jmartinm:
Ah ok, strange that it was not crashing while loading demo records then.

jalavik:
No wait. I think it was fine, perhaps submission was not following the new structure actually. I will amend this commit a bit and check how submissions are treating these fields.

kaplun:
BTW, `earliest_date` is one of those derived fields that are needed to be computed for elasticsearch but we could avoid to store in MySQL. So in principle dojson should not be responsible for allocating it. Just pointing this out in case you are going to do further amendments.

jalavik:
@jmartinm Yeah actually the schema reports these as not being lists while the dojson processing puts all these fields in a list. So in fact the original code was doing the right thing.

jacenkow:
Yeah!

jalavik:
PR adjusted

PR #392, opened 2015-10-01
experiments: fixes problems in brief

ioannistsanaktsidis:


ioannistsanaktsidis:
@jmartinm  also this one is ready to go after the latest update!

inspire/config.py
```diff
@@ -225,7 +225,7 @@
     },
     "abstract": ["abstract.summary"],
     "collection": ["_collections"],
-    "collaboration": ["collaboration.collaboration"],
+    "collaboration": ["collaboration"],
```
jmartinm: Same comment. Do we need this config when it is pointing to a first level of the JSON?

PR #396, opened 2015-10-06
template: jobs brief institution search

ioannistsanaktsidis:
* Adds link to find the institution the job belongs.

* Adds link to find the experiment the job belongs.

### DIFF
--- a/inspire/base/templates/format/record/Job_HTML_brief.tpl
+++ b/inspire/base/templates/format/record/Job_HTML_brief.tpl
@@ -34,9 +34,11 @@
           </b>
           {% if record['institution'][0]['name']%}
               {% if record['continent'] %}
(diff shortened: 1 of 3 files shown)


inspire/config.py
```diff
@@ -230,6 +230,7 @@
     "reportnumber": ["report_number.value"],
     "experiment": ["accelerator_experiment.experiment"],
     "title": ["title.title"],
+    "department_acronym": ["department_acronym"],
```
jmartinm: Did you try if this is needed? Usually you need to configure it in cases where the value is in a second level like `accelerator_experiment.experiment` but for first level I thought it worked by default.

ioannistsanaktsidis: Hmmmmm you are maybe right but I let me check one second and I will inform
you.

2015-10-06 17:22 GMT+02:00 Javier Martin Montull <notifications@github.com>:

PR #400, opened 2015-10-06
global: data model update

jalavik:
* Synchronizes the HEP JSON Schema with doJSON conversion.

* Reduces code duplication in some formats.

* Adds a new upgrade script to update existing workflow objects
  to new data model.

Signed-off-by: Jan Aage Lavik <jan.age.lavik@cern.ch>

PR #409, opened 2015-10-09
dojson: fixes doi mapping

ioannistsanaktsidis:

### DIFF
--- a/inspire/dojson/hep/fields/bd01x09x.py
+++ b/inspire/dojson/hep/fields/bd01x09x.py
@@ -54,9 +54,24 @@ def isbn2marc(self, key, value):
 @utils.filter_values
 def doi(self, key, value):
     """Other Standard Identifier."""
-    return {
-        'doi': value.get('a')
-    }
+    if value.get("2").lower() == "doi":
+        return {
+            'doi': value.get('a'),
+            'source': value.get('9')
+        }
+
+
+@hep.over('persistent_identifiers', '^024[1032478_][10_]')
+@utils.for_each_value
+@utils.filter_values
+def persistent_identifiers(self, key, value):
+    """Persistent identifiers."""
+    if value.get("2").lower() != "doi":
+        return {
+            'value': value.get('a'),
+            'source': value.get('9'),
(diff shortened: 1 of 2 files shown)


jmartinm:
We are missing some things here:

* For DOIs, the source needs to be added (subfield $$9)
* For other types of identifiers ([see here](https://github.com/inspirehep/inspire-next/blob/master/inspire/dojson/current_marcxml_usage/hep.txt#L488)) we would be dropping them. After talking with Annette and Alexander I think we should add a new field to the schema:

```json
"persistent_identifiers": {
            "uniqueItems": true,
            "items": {
                "type": "object",
                "properties": {
                    "source": {
                        "type": "string",
                        "description": "Provenance of the persistent identifier",
                    },
                    "type": {
                        "type": "string",
                        "title": "Type of persistent identifier"
                    },
                    "value": {
                        "type": "string",
                        "title": "Value of the persistent identifier"
                    }
                },
                "title": "Persistent Identifier"
            },
            "type": "array",
            "title": "Persistent Identifiers"
```

jalavik:
#400 is touching the doi field as well, but I can sort out the conflict if this is merged first.

ioannistsanaktsidis:
for the persistent_identifiers will we map it like:
@hep.over('persistent_identifiers', '^024[1032478_][10_]')
@utils.for_each_value
@utils.filter_values
def persistent_identifiers(self, key, value):
    """Other Standard Identifier."""
    if value.get("2").lower() != "doi":
        return {
            'value': value.get('a'),
            'source': value.get('9'),
            'type': value.get('2')
        }

jmartinm:
@ioannistsanaktsidis 

Yes, that looks right.

ioannistsanaktsidis:
@jmartinm  I have just amended the commit.

jmartinm:
Ok, thanks, you can sync with @jalavik and the PR https://github.com/inspirehep/inspire-next/pull/400

ioannistsanaktsidis:
Cool!

jalavik:
These fixes are now incorporated in #400

PR #413, opened 2015-10-13
format: brief removal of duplicate check

nikpap:
* Removes unneccesary check of duplicate values.


inspire/base/templates/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -30,8 +30,8 @@
       <div class="col-md-9"  id="left-column">
         <h4 class="custom-h">
           <b>
-            {% if record.titles %}
-              {% for title in record['titles'] %}
+            {% if record.title %}
```
jalavik: This should remain `titles` as it has been recently changed in the schema.

nikpap: Thanks for noting this, removed said commit from the PR.

kaplun:
I confirm this fixes the trouble as I tested in my PR #357. So for me :shipit:

PR #415, opened 2015-10-14
hepnames: fixes bug in field_categories

ioannistsanaktsidis:

### DIFF
--- a/inspire/dojson/hepnames/fields/bd1xx.py
+++ b/inspire/dojson/hepnames/fields/bd1xx.py
@@ -304,13 +304,31 @@ def positions2marc(self, key, value):
 
 
 @hepnames.over('field_categories', '^65017')
-@utils.for_each_value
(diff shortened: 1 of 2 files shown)


inspire/dojson/hepnames/fields/bd1xx.py
```diff
@@ -304,14 +304,31 @@ def positions2marc(self, key, value):
...
-@utils.filter_values
 def field_categories(self, key, value):
-    return {
-        'name': value.get('a').lower(),
-        'type': value.get('2'),
-    }
 
+    def get_value(value):
+        if isinstance(value, dict):
+            name = utils.force_list(value.get('a').lower())
```
kaplun: Better `value.get('a', '')`. Otherwise `.lower()` will crash when there is no `a`.

ioannistsanaktsidis:
Updated!

PR #421, opened 2015-10-14
WIP docs: GROBID documentation

jacquerie:
This probably needs to be expanded a little more.

CC: @jalavik

### DIFF
--- a/docs/grobid.rst
+++ b/docs/grobid.rst
@@ -1,0 +1,74 @@
+GROBID
+======
+
+1. About
+--------
+
+This document specifies how to train and use GROBID.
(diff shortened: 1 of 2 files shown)


jalavik:
Thanks. You mention that it is available under port 8080. Actually I added a commit on our fork to use port 80. However I am not sure if it was a good idea after all (might conflict with web servers etc.). Perhaps we can revert that and try to rebase our fork on latest grobid code as well.

jacquerie:
Oh, right! Yes, I think that both reverting that change and rebasing our fork are good ideas.

PR #427, opened 2015-10-19
search results: implements export-as format functionality

ioannistsanaktsidis:

### DIFF
--- a/inspire/base/static/js/search/search_results.js
+++ b/inspire/base/static/js/search/search_results.js
@@ -44,7 +44,214 @@ require(['jquery', 'bootstrap'], function($) {
         $("#download" + data.recid).attr("href", "/formatter/download-latex/latex_us/" + data.recid)
       });
     });
+
   })
+
+  $("#export-as-bibtex").on('click', function() {
+    if (sList != 0) {
+      create_url('/formatter/export-as/bibtex/', sList);
+    } else {
+      $("#download").removeAttr('href');
+    }
+    $("#dropdown-export").html('BibTex <span class="caret"></span>');
+  });
+
+  $("#export-as-latex-eu").on('click', function() {
+    if (sList != 0) {
+      create_url('/formatter/export-as/latex_eu/', sList);
+    } else {
+      $("#download").removeAttr('href');
+    }
+    $("#dropdown-export").html('LaTex(EU) <span class="caret"></span>');
+  });
+
+  $("#export-as-latex-us").on('click', function() {
+    if (sList != 0) {
+      create_url('/formatter/export-as/latex_us/', sList);
+    } else {
+      $("#download").removeAttr('href');
+    }
+    $("#dropdown-export").html('LaTex(US) <span class="caret"></span>');
+  });
+
+  $("#export-as-cv-latex").on('click', function() {
+    if (sList != 0) {
+      create_url('/formatter/export-as/cv_latex/', sList);
+    } else {
+      $("#download").removeAttr('href');
+    }
+    $("#dropdown-export").html('CV format (LaTex) <span class="caret"></span>');
+  });
+
+  $("#export-as-cv-html").on('click', function() {
+    if (sList != 0) {
+      create_url('/formatter/export-as/cv_latex_html/', sList);
+    } else {
+      $("#download").removeAttr('href');
+    }
+    $("#dropdown-export").html('CV format (html) <span class="caret"></span>');
+  });
+
+  $("#export-as-cv-text").on('click', function() {
+    if (sList != 0) {
+      create_url('/formatter/export-as/cv_latex_text/', sList);
+    } else {
+      $("#download").removeAttr('href');
+    }
+    $("#dropdown-export").html('CV format (text) <span class="caret"></span>');
+  });
+
+  var sList = [];
+  $('#checkbox-select-all').click(function(event) {
+    if (this.checked) {
+      $('.checkbox-results').each(function() {
+        this.checked = true;
+        sList.push($(this).attr("id"));
+      });
+      $('#results-panel').after('<div class="panel panel-default" id="info-message">' +
+        '<div class="panel-body" >You have selected ' + sList.length + ' records of this page. <a class="pointer" id="select-all-records">' +
+        'Select all ' + $('#total-results').text() + ' results.</a></div></div>');
+      $('#select-all-records').on('click', function() {
+        $('#info-message').remove();
+        $('#results-panel').after('<div class="alert alert-warning" id="alert-selection" role="alert">' +
+          $('#total-results').text() + ' have been selected.<a class="pointer" id="undo-selection"> Undo selection.</a></div>');
+        $.get("/search?of=id&rg=" + $('#total-results').text(), function(data, status) {
+          if (status == 'success') {
+            sList = data;
+            create_url(getType(), sList);
+          } else {
+            sList = [];
+            $('#alert-selection').remove();
+            $("#download").removeAttr('href');
+            $('#results-panel').after('<div class="alert alert-danger" id="danger-selection" role="alert">' +
+              'Too many records to download.</div>')
+          }
+        });
+      });
+    } else {
+      $('.checkbox-results').each(function() {
+        this.checked = false;
+      });
+      sList = [];
+      $("#download").removeAttr('href');
+      $('#info-message').remove();
+      $('#alert-selection').remove();
+    }
+  });
+
+  $(document).on('click', '#undo-selection', function(e) {
+    sList = [];
+    $('#alert-selection').remove();
+    $("#download").removeAttr('href');
+    $('.checkbox-results').each(function() {
+      this.checked = false;
+    });
+    $('#checkbox-select-all').prop('checked', false);
+  });
+
+  function querystring(key) {
+    var re = new RegExp('(?:\\?|&)' + key + '=(.*?)(?=&|$)', 'gi');
+    var r = [],
+      m;
+    while ((m = re.exec(document.location.search)) != null) r.push(m[1]);
+    return r;
+  }
+
+  $('input[type=checkbox]').change(function() {
+    var $this = $(this);
+    if (this.checked) {
+      if ($this.attr("id") != 'checkbox-select-all') {
+        sList.push($this.attr("id"));
+      }
+    } else {
+      var index = sList.indexOf($this.attr("id"))
+      if (index > -1) {
+        sList.splice(index, 1);
+      }
+    }
+    if ($('#dropdown-export').text().trim() == 'BibTex') {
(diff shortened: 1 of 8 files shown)


ioannistsanaktsidis:
@jmartinm this is the PR for the export as functionality! Cheers!

jmartinm:
Check https://travis-ci.org/inspirehep/inspire-next/builds/86163084#L632

jmartinm:
In the commit log summary, we keep it with the subset of module names that we have, instead of changing every time (for easy `grep` afterwards)

If this PR touches:

```
 inspire/base/static/js/search/search_results.js
+4 −0   inspire/base/static/less/format/brief-results.less
+2 −1   inspire/base/templates/format/record/Inspire_Default_HTML_brief.tpl
+50 −2   inspire/base/templates/search/results.html
+37 −0   inspire/modules/formatter/views.py
+8 −7   inspire/utils/bibtex.py
+2 −2   inspire/utils/cv_latex.py
+2 −2   inspire/utils/latex.py
```

The main module you are changing is `inspire/modules/formatter/views.py`

So change to:

```
formatter: implements export-as functionality
```

inspire/base/static/js/search/search_results.js
```diff
@@ -44,8 +44,111 @@ require(['jquery', 'bootstrap'], function($) {
...
+    if (this.checked) {
+      $('.checkbox-results').each(function() {
+        this.checked = true;
+        sList.push($(this).attr("id"));
+      });
+    } else {
+      $('.checkbox-results').each(function() {
+        this.checked = false;
+      });
+      sList.length = 0;
```
jmartinm: if `sList` is already a list it should have the property `length` and it is strange to modify it directly. Is `sList = []` what you want to do here?

inspire/base/static/js/search/search_results.js
```diff
@@ -44,8 +44,111 @@ require(['jquery', 'bootstrap'], function($) {
...
+    }
+  });
+
+  $('input[type=checkbox]').change(function() {
```
jmartinm: Is this selecting all checkboxes in the page? Make sure only the ones you want get selected (search results) as there might be other checkboxes in the page that should be ignored (like the facets)

inspire/base/static/js/search/search_results.js
```diff
@@ -44,8 +44,111 @@ require(['jquery', 'bootstrap'], function($) {
...
+      if ($this.attr("id") != 'checkbox-select-all') {
+        sList.push($this.attr("id"));
+      }
+    } else {
+      var index = sList.indexOf($this.attr("id"))
+      if (index > -1) {
+        sList.splice(index, 1);
+      }
+    }
+    if (sList != 0) {
```
jmartinm: is `sList` a list? Then this comparison would not make much sense.

inspire/base/templates/search/results.html
```diff
@@ -28,6 +28,53 @@
...
+      <div class="panel panel-default" >
+        <div class="panel-body" >
+          <div class="form-group" style="display:block;">
+            <p class="help-block hidden-xs" style="font-size: 20px;">
+              {%- set r_from = (pagination.page-1)*pagination.per_page+1 -%}
+              {%- set r_to = pagination.page*pagination.per_page -%}
+              {%- set r_of = pagination.total_count -%}
+              {%- set r_to = r_to if r_to < r_of else r_of -%}
+              {# _('Showing records %(x_from)d to %(x_to)d out of %(x_total)d results.', x_from=r_from, x_to=r_to, x_total=r_of) #}
+              {{ _('<b style="color:#e57e25;">%(x_total)d</b> results found.', x_total=r_of) }}
```
jmartinm: Avoid hardcoding this value and use the `less` variable with the color. Maybe it is not yet available until we merge Nick's code

inspire/base/templates/search/results.html
```diff
@@ -28,6 +28,53 @@
...
+            </p>
+          </div>
+          <hr>
+          {% if  request.args.get('cc') not in ["HepNames", "Institutions", "Conferences", "Jobs", "Journals", "Experiments"] %}
```
jmartinm: It is better to just check ```{% if  request.args.get('cc') == "HEP" %}```

As that is the only collection that should have the Export.

Otherwise if we add a new collection the dropdown will appear unexpectedly.

inspire/modules/formatter/views.py
```diff
@@ -84,3 +88,36 @@ def get_latex_file(recid, latex_format):
             )
         }
     )
+
+
+@blueprint.route("/export-as/<path:export_format>/<path:ids>",
+                 methods=['GET', 'POST'])
+def download_all_latex(ids, export_format):
```
jmartinm: Add here a docstring explaining what are the `ids` and `export_format` parameters and what is the result value.

We use `Sphinx` format. See https://github.com/inspirehep/invenio-workflows/blob/master/invenio_workflows/api.py#L71

inspire/modules/formatter/views.py
```diff
@@ -84,3 +88,36 @@ def get_latex_file(recid, latex_format):
...
+        out.append(generator)
+    output = itertools.chain()
+    for gen in out:
+        output = itertools.chain(output, gen)
+
+    return Response(
+        output,
+        mimetype="text/plain",
+        headers={
+            "Content-Disposition": "attachment;filename=allLatex_{0}.tex".format(
```
jmartinm: You can adapt the name of the file depending on the output format.

inspire/modules/formatter/views.py
```diff
@@ -84,3 +88,36 @@ def get_latex_file(recid, latex_format):
             )
         }
     )
+
+
+@blueprint.route("/export-as/<path:export_format>/<path:ids>",
```
jmartinm: This function receives a list of `ids` specified by the user and returns all the given formats.

We should be defensive with the user input and:

* Set a limit to the `ids` list. Or we have to explore if using Celery we can scale and be able to export as many `ids` as requested.
* What about user permissions? Do we have hidden records for certain users? Should we do the intersection of the requested ids with the ids the user is allowed to see? Is there an API to get all records allowed by a user?

What do you think @kaplun @jalavik  ? Something else?

inspire/base/static/js/search/search_results.js
ioannistsanaktsidis: I wanted to set the list to empty and I found out that you can also do it by setting the length to 0. In the new update I changed it like you have it.

inspire/base/static/js/search/search_results.js
ioannistsanaktsidis: As I see it selects only the ones we want.

inspire/base/static/js/search/search_results.js
ioannistsanaktsidis: Yes I removed it from here, true it is useless.

inspire/base/templates/search/results.html
ioannistsanaktsidis: True but that way when we have sth like http://localhost:4000/search it will not appear too. But will we have a url like this one? Or it will always be with the collection as a parameter?

inspire/modules/formatter/views.py
ioannistsanaktsidis: if the output format is bibtex we have the_file.bibtex, if latex the_file.tex. If it is one of the others should I save it as txt? any ideas?

inspire/modules/formatter/views.py
kaplun: So I think up to 5000 ids could be served (as there are e.g. certain papers having ~3000 references). After that we can truncate the export and add an appropriate comment line (e.g. in BibTeX for the BibTeX format) describing the truncation being implemented and to contact us if there are further needs.

Regarding permission, luckily there are no real secret records so far in INSPIRE. Rather some hidden fields and in case some protected fulltext files, but these are unrelated with this very specific PR.

ioannistsanaktsidis:
@jmartinm I have amended the commit and added the select all(google style) functionality. We can check today how to implement the celery task and we are ready to go!

ioannistsanaktsidis:
I have updated the PR. Feel free to test and send me some feedback :) Cheers!

inspire/modules/formatter/views.py
```diff
@@ -84,3 +89,57 @@ def get_latex_file(recid, latex_format):
...
+
+
+@blueprint.route("/get-results/<path:params>",
+                 methods=['GET', ])
+@blueprint.route("/get-results/", defaults={'params': ''},
+                 methods=['GET', ])
+def get_results(params):
+    results = Query('%s' % (params,)).search().recids
+    if len(results.tolist()) > 5000:
+        return jsonify(result=[])
```
jmartinm: Here you should return 5000 instead of empty.

inspire/modules/formatter/views.py
```diff
@@ -84,3 +89,57 @@ def get_latex_file(recid, latex_format):
...
+                export_format, suffix
+            )}
+    )
+
+
+@blueprint.route("/get-results/<path:params>",
+                 methods=['GET', ])
+@blueprint.route("/get-results/", defaults={'params': ''},
+                 methods=['GET', ])
+def get_results(params):
```
jmartinm: Maybe we don't even need to specify our own endpoint. This looks very much like the search API.

You can try to call directly from JavaScript:

`/search?p="blah"&of=id`

This returns already ids.

ioannistsanaktsidis: I found the url that returns all the ids of the results, it is like this
/search?of=id&rg=numberOfResults, the problem is that it is returning a list with the recids and not the control_number. Proposals?

jmartinm: Are the `recids` and the `control_number` different?? If that is the case we should check why is that.

ioannistsanaktsidis: Yes, they are different. If we find the problem then we can use this search
function.

2015-10-23 13:43 GMT+02:00 Javier Martin Montull <notifications@github.com>:

kaplun: According to @egabancho their are aliased in the default configuration

egabancho: @kaplun is right, `recid` is just an alias for `control_number`, see https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/config.py#L44

ioannistsanaktsidis: Strange, when I tried it I remember they were different. Maybe my fault
then. We can use the search?of=id url then :)

2015-10-23 14:42 GMT+02:00 Esteban J. G. Gabancho <notifications@github.com>
:

ioannistsanaktsidis:
@jmartinm the commit is amended with the latest version! You can check it and then we are ready to go. In the end we used the search api style to get all the ids. Much easier :dancer:

Issue #381, opened 2015-09-30
Missing stuff in brief results


ioannistsanaktsidis:
- [x] journals: Missing doc_string from url and Publisher. 
- [x] jobs: Creation date missing.
- [x] experiments: Missing HEP articles and collaboration link. 
- [x] institutions: Missing affiliation line (e.g 860 Papers from KIPAC, Menlo Park)
- [x] conferences: Contributions link missing.
- [ ] authors: Author profile link missing.

ioannistsanaktsidis:
@kaplun Is creation date of the job now added to the metadata?

jmartinm:
We will tackle the author part once we start using the new profiles.

PR #440, opened 2015-10-26
brief: fixes mistakes after introducing new model

ioannistsanaktsidis:
* (addresses #381)

### DIFF
--- a/inspire/ext/formatter_jinja_filters/record.py
+++ b/inspire/ext/formatter_jinja_filters/record.py
@@ -168,6 +168,7 @@ def search_for_experiments(value):
         '<a href="/search?p=experiment_name:%s&cc=Experiments">%s</a>'
         % (i, i)
         for i in value])
+    return result
 
 
 def experiment_date(record):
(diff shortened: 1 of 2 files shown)


ioannistsanaktsidis:
@jmartinm we are almost done completely from brief results. We only miss the link to author profile, which I guess it is not a problem right now, and finally we just need the creation date in the Jobs. For further information you can check #381 . Cheers :dancer:

inspire/ext/formatter_jinja_filters/record.py
```diff
@@ -262,6 +263,11 @@ def link_to_hep_affiliation(record):
                 str(record['department_acronym'])
 
 
+def join_nested_lists(l):
+    new_list = [item for sublist in l for item in sublist]
+    return ', '.join(element for element in new_list)
```
jmartinm: so `new_list` is just a list of elements? 

```python
', '.join(new_list)
```

does not work?

ioannistsanaktsidis: Yeah, that also works! I amended the commit. ;)

2015-10-27 8:39 GMT+01:00 Javier Martin Montull <notifications@github.com>:

Issue #434, opened 2015-10-21
Page range/ Article ID - data model consistency


jmartinm:
I have found some inconsistencies between our dojson import and the HEP Schema that should be addressed:

* [x] `page_artid` is in dojson but not on schema. It is called `page_range` [in the schema](https://github.com/inspirehep/inspire-next/blob/master/inspire/dojson/hep/schemas/hep-0.0.1.json#L383) which I am not sure covers the full meaning of the field.
* [x] cnum in dojson is conference_paper_info in schema
* [x] collaboration is not array in dojson but it is in the schema.
* [x] url should be renamed to urls in schema and dojson

Tasks missing https://github.com/inspirehep/inspire-next/issues/434#issuecomment-152526331

annetteholtkamp:
It still hasn't been decided whether we want to split article nr and page range. Would be useful in case we have both - e.g. for JACoW conferences.

- Annette


On Oct 21, 2015, at 11:21 AM, Javier Martin Montull <notifications@github.com<mailto:notifications@github.com>> wrote:


I have found some inconsistencies between our dojson import and the HEP Schema that should be addressed:

  *    page_artid is in dojson but not on schema. It is called page_range in the schema<https://github.com/inspirehep/inspire-next/blob/master/inspire/dojson/hep/schemas/hep-0.0.1.json#L383> which I am not sure covers the full meaning of the field.
  *    cnum in dojson is conference_paper_info in schema
  *    collaboration is not array in dojson but it is in the schema.
  *    url should be renamed to urls in schema and dojson

jalavik:
Regarding, `page_range` I fixed it as part of #422 https://github.com/inspirehep/inspire-next/pull/422/files#diff-3db794b5dfd367ebcb5f09729f2490f5 

But yeah, I guess it make sense to split it rather?

kaplun:
> ￼ collaboration is not array in dojson but it is in the schema.

Should indeed be an array. (e.g. paper from ATLAS and CMS).

ioannistsanaktsidis:
@jmartinm @kaplun the collaboration problem I think is fixed. I have just checked and it is array in both now.

kaplun:
IMHO this belongs to the `Enabling search` milestone

jacquerie:
[...]

PR #441, opened 2015-10-26
dojson: changes url to urls

ioannistsanaktsidis:
* (addresses #434)

### DIFF
--- a/inspire/base/templates/format/record/Holding_Pen_HTML_detailed.tpl
+++ b/inspire/base/templates/format/record/Holding_Pen_HTML_detailed.tpl
@@ -100,7 +100,7 @@
             <!-- URLs -->
             <span id='editable-urls'><small><strong>URL(s): </strong></small></span>
             <span id="url-links">
-            {% for url_obj in record['url'] %}
+            {% for url_obj in record['urls'] %}
               <a href="{{ url_obj['url'] }}">{{ url_obj['url'] }}</a>
               {{ add_delimiter(loop, '|') }}
             {% endfor %}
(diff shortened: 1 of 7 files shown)


ioannistsanaktsidis:
@jmartinm  this PR is for the url problem in #434 that you mentioned.

jmartinm:
We have to amend this test too https://github.com/inspirehep/inspire-next/blob/2294c1f848d2cd102cdfcb505f739c42944899a6/inspire/modules/deposit/testsuite/test_hep_submission.py#L195 and the code here https://github.com/inspirehep/inspire-next/blob/master/inspire/modules/deposit/dojson/fields/literature.py#L304

ioannistsanaktsidis:
The first one I tried to change it and run the tests locally and it was
failing. Thats why I left it like this. We can check it tomorrow and see.

2015-10-26 17:18 GMT+01:00 Javier Martin Montull <notifications@github.com>:

jmartinm:
Ok, they should be amended because otherwise all the submissions from users would not contains the URLs. Try it and if it does not work I can help out.

ioannistsanaktsidis:
I am changing to urls the test_hep_submission.py, running the test and it
is failing saying  KeyError: 'urls'. My change is
this self.record['urls'][0]['url']. I also recreated the database so as to
apply my changes in the dojson. Ideas?

2015-10-27 8:37 GMT+01:00 Javier Martin Montull <notifications@github.com>:

jmartinm:
You will have to update https://github.com/inspirehep/inspire-next/blob/master/inspire/modules/deposit/dojson/fields/literature.py#L298 and https://github.com/inspirehep/inspire-next/blob/master/inspire/modules/deposit/dojson/fields/literature.py#L310 so that the key is now `urls` instead of `url`

ioannistsanaktsidis:
I have just amended the commit. The tests are good now in my local.

2015-10-27 10:45 GMT+01:00 Javier Martin Montull <notifications@github.com>:

PR #443, opened 2015-10-27
formatter: implements export-as functionality

ioannistsanaktsidis:
* Implements select all results functionality.

### DIFF
--- a/inspire/base/static/js/search/search_results.js
+++ b/inspire/base/static/js/search/search_results.js
@@ -44,7 +44,214 @@ require(['jquery', 'bootstrap'], function($) {
         $("#download" + data.recid).attr("href", "/formatter/download-latex/latex_us/" + data.recid)
       });
     });
+
   })
+
+  $("#export-as-bibtex").on('click', function() {
+    if (sList != 0) {
+      create_url('/formatter/export-as/bibtex/', sList);
+    } else {
+      $("#download").removeAttr('href');
+    }
+    $("#dropdown-export").html('BibTex <span class="caret"></span>');
+  });
+
+  $("#export-as-latex-eu").on('click', function() {
+    if (sList != 0) {
+      create_url('/formatter/export-as/latex_eu/', sList);
+    } else {
+      $("#download").removeAttr('href');
+    }
+    $("#dropdown-export").html('LaTex(EU) <span class="caret"></span>');
+  });
+
+  $("#export-as-latex-us").on('click', function() {
+    if (sList != 0) {
+      create_url('/formatter/export-as/latex_us/', sList);
+    } else {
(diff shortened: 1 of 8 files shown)


inspire/base/static/js/search/search_results.js
```diff
@@ -44,8 +44,215 @@ require(['jquery', 'bootstrap'], function($) {
...
+    sList = [];
+    $('#alert-selection').remove();
+    $("#download").removeAttr('href');
+    $('.checkbox-results').each(function() {
+      this.checked = false;
+    });
+    $('#checkbox-select-all').prop('checked', false);
+  });
+
+  function querystring(key) {
```
jmartinm: this will be removed.

inspire/base/static/js/search/search_results.js
```diff
@@ -44,8 +44,215 @@ require(['jquery', 'bootstrap'], function($) {
...
+    var re = new RegExp('(?:\\?|&)' + key + '=(.*?)(?=&|$)', 'gi');
+    var r = [],
+      m;
+    while ((m = re.exec(document.location.search)) != null) r.push(m[1]);
+    return r;
+  }
+
+  $('input[type=checkbox]').change(function() {
```
jmartinm: Better to put an id to select from

inspire/base/static/js/search/search_results.js
```diff
@@ -44,8 +44,215 @@ require(['jquery', 'bootstrap'], function($) {
...
+    var $this = $(this);
+    if (this.checked) {
+      if ($this.attr("id") != 'checkbox-select-all') {
```
jmartinm: This will not be needed if the selector only matches the checkboxes close to each result.

ioannistsanaktsidis:
The reason that we have the problem is that search results(search/results_base.html) are wrapped to this form:
```
 <form class="form-inline" name="" action="{{ url_for('search.dispatch') }}" method="post">
```
so every time that I fire my ajax call: 
```
javascript
$("#download-format").on('click', function() {
      console.log(getType(url));
     $.ajax({
        type: "POST",
        url: getType(url),
        data: {ids: uniqueIds},
        success: function(data) {
          console.log(data);
          var response_data = "text/plain;charset=utf-8," + encodeURIComponent(data);
          $("body").append('<a id="data-download" href="data:' + response_data + '" download="references.bib">download</a>');
          var trigger_element = document.getElementById('data-download');
          trigger_element.click();
          $("#data-download").remove();
        }
      });
     });
```
in the success function the data from the form are also added and it is becoming a mess.
Any ideas how to avoid this problem? Is there any specific reason that results are wrapped to a form?

PR #444, opened 2015-10-27
header: mega dropdown menu

ioannistsanaktsidis:

### DIFF
--- a/inspire/base/templates/header.html
+++ b/inspire/base/templates/header.html
@@ -33,53 +33,360 @@
   ('https://labs.inspirehep.net/help', {}, '', 'Help'),
 ] -%}
 
+    <section>
+      <nav id="topnav" class="navbar navbar-fixed-top navbar-inverse navbar-custom" role="navigation">
+        <div class="container">
+          <div class="navbar-header">
+            <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target=".navbar-ex1-collapse">
+              <span class="sr-only">{{ _("Toggle navigation") }}</span>
+              <span class="icon-bar"></span>
+              <span class="icon-bar"></span>
+              <span class="icon-bar"></span>
+            </button>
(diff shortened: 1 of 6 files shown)


ioannistsanaktsidis:
@jmartinm just squashed @nikpap commit and mine. Looks good!

inspire/base/templates/header.html
```diff
@@ -28,58 +28,364 @@
...
+        <div class="container">
+          <div class="navbar-header">
+            <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target=".navbar-ex1-collapse">
+              <span class="sr-only">{{ _("Toggle navigation") }}</span>
+              <span class="icon-bar"></span>
+              <span class="icon-bar"></span>
+              <span class="icon-bar"></span>
+            </button>
+            <a class="navbar-brand" href="{{ url_for('search.index') }}">
+              <img src="{{ url_for('static', filename='img/labs_white.svg') }}" alt="{{ config.CFG_SITE_NAME_INTL['en'] }}">
```
jmartinm: This is the wrong logo

jmartinm: And the link on the logo goes to `http://localhost:4000/search/index` which gives 404.

Please have a look as it seems something went wrong in the rebase and the logo part was overwritten.

jmartinm:
As an improvement that we can do after I would add an arrow down symbol that indicates the dropdown is going to open, the dropdown will fill the whole width of the screen (instead of partially like now) and there would be a dialog indicator coming from the `Resources` link.

Like we do in the login

screen shot 2015-11-06 at 1 45 00 pm

ioannistsanaktsidis:
@jmartinm problems solved!

jmartinm:
Merged 89a37474c2a28ecfef2a94b3a730fbea774f31bf

Issue #403, opened 2015-10-08
Fix formats according to data model changes


jmartinm:
After https://github.com/inspirehep/inspire-next/pull/400 gets merged all formats `bibtex`, `CV` and friends need to be amended to fit the new model.

jalavik:
While updating the schema, I have tried to change all usages everywhere (including bibtex, latex formats, submissions, workflows etc.). However, I might have missed some parts, so this should still be verified.

jalavik:
So @ioannistsanaktsidis is updating these formats now for BibTeX output and something came up when running it over the demo records.

Since across our collection we may have a different data model, e.g. HEP vs Experiments, we would have to amend the BibTeX/LateX formats to export differently per collection. E.g. titles.title vs. title. 

Does it make sense to even offer BibTeX/LaTeX export for non-HEP collections? Authors perhaps, but others? @kaplun @annetteholtkamp @jmartinm

kaplun:
Actually Authors nope. IMHO only records in HEP (i.e. of document type) should have a BibTeX/LaTeX export. In the future also records in the Data collection might have such export, though.

jmartinm:
Fixed https://github.com/inspirehep/inspire-next/pull/445

PR #445, opened 2015-10-27
formats: fixes errors caused by new model

ioannistsanaktsidis:
* Creates test for bibtex article.

* Creates test for latex(EU, US).

* Creates test for Cv Latex(html,text).

* Creates test for Cv Latex.

* Creates test for export.

* (addresses #403)

### DIFF
--- a/inspire/testsuite/test_bibtex.py
+++ b/inspire/testsuite/test_bibtex.py
@@ -1,0 +1,153 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2015 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+import pkg_resources
+import os
+
+from dojson.contrib.marc21.utils import create_record
+
+from invenio.testsuite import InvenioTestCase, make_test_suite, run_test_suite
+from inspire.dojson.hep import hep
+from invenio.base.wrappers import lazy_import
+Bibtex = lazy_import('inspire.utils.bibtex.Bibtex')
+
+
+class BibtexTests(InvenioTestCase):
+
+    def setUp(self):
+        self.marcxml = pkg_resources.resource_string('inspire.testsuite',
+                                                     os.path.join(
+                                                         'fixtures',
+                                                         'test_hep_formats.xml')
+                                                     )
+        record = create_record(self.marcxml)
+
+        self.hep_record = hep.do(record)
+
+        self.bibtex = Bibtex(self.hep_record)
+
+        self.sample_bibtex = {
+            'collections': ('article', 'article'),
+            'citation_key': 'Aad:2015wqa',
+            'author': 'Aad, Georges',
+            'title': '\nSearch for supersymmetry in events containing a same-flavour opposite-sign dilepton pair, jets, and large missing transverse momentum in $\sqrt{s}=8$ TeV $pp$ collisions with the ATLAS detector\n',
+            'collaboration': 'ATLAS',
+            'journal': 'Eur. Phys. J.',
+            'volume': 'C75',
+            'year': 2015,
+            'number': '7',
+            'pages': '318',
+            'doi': '10.1140/epjc/s10052-015-3661-9, 10.1140/epjc/s10052-015-3518-2',
+            'note': '[Erratum: Eur. Phys. J.C75,no.10,463(2015)]',
+            'eprint': '1503.03290',
+            'archivePrefix': 'arXiv',
+            'primaryClass': 'hep-ex',
+            'reportNumber': 'CERN-PH-EP-2015-038',
+            'SLACcitation': '%%CITATION = ARXIV:1503.03290;%%',
+
+        }
+
+    def test_entry_type(self):
+        """Test if entry type is created correctly"""
+        self.assertEqual(self.sample_bibtex['collections'],
+                         self.bibtex._get_entry_type())
+
+    def test_citation_key(self):
+        """Test if citation key is created correctly"""
+        self.assertEqual(self.sample_bibtex['citation_key'],
+                         self.bibtex._get_citation_key())
+
+    def test_author(self):
+        """Test if author is created correctly"""
+        self.assertEqual(self.sample_bibtex['author'],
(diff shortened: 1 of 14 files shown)


ioannistsanaktsidis:
@jalavik  this PR fails in the workflows test. I remember that we had fixed that problem right? Btw now that we have the arxiv in arxiv_eprints, this function 
```
def get_arxiv_id_from_record(record):
    """Return the arXiv identifier from given record.

    This function works with Deposition and Payload data models.
    """
    arxiv_id = record.get("arxiv_id")
    if not arxiv_id:
        report_numbers = record.get('report_numbers', [])
        for number in report_numbers:
            if number.get("source", "").lower() == "arxiv":
                arxiv_id = number.get("value")

    if not arxiv_id:
        arxiv_eprints = record.get('arxiv_eprints', [])
        for element in arxiv_eprints:
            if element.get("value", ""):
                arxiv_id = element.get("value", "")

    if arxiv_id and not arxiv_id.lower().startswith("oai:arxiv") and not \
       arxiv_id.lower().startswith("arxiv") and \
       "/" not in arxiv_id:
        arxiv_id = "arXiv:{0}".format(arxiv_id)
    return arxiv_id
```
should be refactored like this:
```
def get_arxiv_id_from_record(record):
    """Return the arXiv identifier from given record.

    This function works with Deposition and Payload data models.
    """
    arxiv_id = record.get("arxiv_id")
    if not arxiv_id:
        arxiv_eprints = record.get('arxiv_eprints', [])
        for element in arxiv_eprints:
            if element.get("value", ""):
                arxiv_id = element.get("value", "")

    if arxiv_id and not arxiv_id.lower().startswith("oai:arxiv") and not \
       arxiv_id.lower().startswith("arxiv") and \
       "/" not in arxiv_id:
        arxiv_id = "arXiv:{0}".format(arxiv_id)
    return arxiv_id
```
Right?

jalavik:
Yeah, sounds good @ioannistsanaktsidis

ioannistsanaktsidis:
Nice! I am amending the commit!

2015-10-27 14:50 GMT+01:00 Jan Åge Lavik <notifications@github.com>:

inspire/dojson/hep/fields/bd76x78x.py
```diff
@@ -28,31 +28,42 @@
...
-        'confpaper_info': value.get('t'),
-        'journal_volume': value.get('v'),
-        'cnum': value.get('w'),
-        'pubinfo_freetext': value.get('x'),
-        'year': year,
-        'isbn': value.get('z'),
-        'note': value.get('m'),
-    }
+
+    def get_value(value):
```
jmartinm: A test will be added to show the problem before this code was amended

inspire/testsuite/test_bibtex.py
```diff
@@ -0,0 +1,151 @@
...
+                                                     os.path.join(
+                                                         'fixtures',
+                                                         'test_hep_formats.xml')
+                                                     )
+        record = create_record(self.marcxml)
+
+        self.hep_record = hep.do(record)
+
+        self.sample_bibtex = {
+            'collections': ('article', 'article'),
```
jmartinm: why are the collections repeated?

ioannistsanaktsidis: Because _get_entry_type returns the 2 things, the entry_type and the
original_entry. So in my record the original and the entry are the same.

2015-10-29 10:37 GMT+01:00 Javier Martin Montull <notifications@github.com>:

inspire/testsuite/test_export.py
```diff
@@ -0,0 +1,85 @@
...
+            'arxiv_field': {u'categories': [u'hep-ex'], u'value': u'arXiv:1503.03290'},
+            'arxiv': 'arXiv:1503.03290 [hep-ex]',
+            'reportNumber': 'CERN-PH-EP-2015-038',
+            'SLACcitation': '%%CITATION = ARXIV:1503.03290;%%',
+        }
+
+    def test_citation_key(self):
+        """Test if citation key is created correctly"""
+        self.assertEqual(self.sample_export_good['citation_key'],
+                         Export(self.hep_record)._get_citation_key())
```
jmartinm: You can do the `Export(self.hep_record)` only once in the `setUp()`.

inspire/testsuite/test_export.py
```diff
@@ -0,0 +1,85 @@
...
+                                                         'fixtures',
+                                                         'test_hep_formats.xml')
+                                                     )
+        record = create_record(self.marcxml)
+
+        self.hep_record = hep.do(record)
+
+        self.sample_export_good = {
+            'citation_key': 'Aad:2015wqa',
+            'doi': '10.1140/epjc/s10052-015-3661-9, 10.1140/epjc/s10052-015-3518-2',
```
jmartinm: I am not 100% sure where this is used, but is it not more convenient for this to be a list with the dois? Do you have to later split by comma again or you use it straight like this?

inspire/testsuite/test_latex.py
```diff
@@ -0,0 +1,127 @@
...
+            'arxiv': 'arXiv:1503.03290 [hep-ex]',
+            'report_number': '',
+            'SLACcitation': '%%CITATION = ARXIV:1503.03290;%%',
+
+        }
+
+    def test_citation_key(self):
+        """Test if citation_key is created correctly"""
+        self.assertEqual(self.sample_latex_eu['citation_key'],
+                         Latex(self.hep_record, 'latex_eu').
```
jmartinm: Same as before, you can do in the `setUp`

```python
Latex(self.hep_record, 'latex_eu')
Latex(self.hep_record, 'latex_us')
```

only once and then use it in the tests.

jmartinm:
We can :shipit: when the comments are addressed.

inspire/testsuite/test_bibtex.py
jmartinm: I see. OK then.

inspire/testsuite/test_export.py
ioannistsanaktsidis: I am using it like this, the function get_doi does all the job and returns me either one single, either multiple separated by comma.

ioannistsanaktsidis:
@jmartinm  I have just amended the commit. The test case for publication info will be added to a different PR. Cheers!

inspire/dojson/hep/fields/bd76x78x.py
ioannistsanaktsidis: I am checking again the previous code and it is working good (!). It gives back the correct results. I am going to remove the file from the commit as long as there is no problem. I don't get why I had this problem before :/

ioannistsanaktsidis:
I can't reproduce the problem with the publication info that I had. I am testing the previous code and it works as charm. I removed the file from the commit as long as there is no need to change anything. The PR is ready to go.

inspire/testsuite/test_bibtex.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def test_title(self):
+        """Test if title is created correctly"""
+        self.assertEqual(self.sample_bibtex['title'],
+                         Bibtex(self.hep_record)._get_title())
+
+    def test_collaboration(self):
+        """Test if collaboration is created correctly"""
+        self.assertEqual(self.sample_bibtex['collaboration'],
+                         Bibtex(self.hep_record)._get_collaboration())
```
jmartinm: In these tests and the next ones it is better if you do like I mentioned before, just compute once `Bibtex(self.hep_record)` in `setUp`

PR #449, opened 2015-10-29
format: bug fixes on HEP detailed record

nikpap:
* Displays PDF's correctly

* Removes popover affiliations from modal window

* Displays external links correctly

### DIFF
--- a/inspire/base/templates/format/record/Inspire_Default_HTML_general_macros.tpl
+++ b/inspire/base/templates/format/record/Inspire_Default_HTML_general_macros.tpl
@@ -17,8 +17,8 @@
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 #}
 
-{% macro render_author_names(author) %}
-  <a{% if author.affiliations|length > 0 %}
+{% macro render_author_names(author, show_affiliation) %}
+  <a{% if author.affiliations|length > 0  and show_affiliation %}
       data-toggle="tooltip"
       data-placement="bottom"
       title="{{ author.get('affiliations')[0]['value'] }}"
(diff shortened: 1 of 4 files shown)


inspire/base/templates/format/record/Inspire_HTML_detailed_macros.tpl
```diff
@@ -51,16 +51,13 @@
...
-          {% do exists.append(1) %}  
-            <a class="btn custom-btn btn-warning pdf-btn" href="{{ url.get('url') }}" target="_blank" role="button" title="View PDF"><i class="fa fa-eye"></i> View PDF</a>
-        {% endif %}
-      {% endif %}
-    {% endfor %}
+  {% if record.get('arxiv_eprints') %}
+    {% if record.get('arxiv_eprints') | is_list() %}
+      {% set filtered_arxiv = record.get('arxiv_eprints') %}
+      {% for i in filtered_arxiv %}
+        <a class="btn custom-btn btn-warning pdf-btn" href="http://arxiv.org/pdf/{{ i.get('value') }}" role="button" title="View PDF"><i class="fa fa-eye"></i> View PDF</a>
```
jmartinm: Looking at the code that we use nowadays to link to arXiv records, I assume this link will not always work.

You can create a filter that does the same logic as today:

https://github.com/inspirehep/inspire/blob/master/bibformat/format_elements/bfe_INSPIRE_arxiv_link.py#L32-L47

inspire/base/templates/format/record/Inspire_Default_HTML_general_macros.tpl
```diff
@@ -17,8 +17,8 @@
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 #}
 
-{% macro render_author_names(author) %}
-  <a{% if author.affiliations|length > 0 %}
+{% macro render_author_names(author, show_affiliation) %}
```
jmartinm: So when do we show the affiliation and when we don't show it?

nikpap: This refers specifically to "hover" affiliations. These should not be shown in the modal popup, since the affiliations will be already displayed in parentheses next to the author. Therefore, there is no need to display them on hover, in the modal window.

jmartinm: That makes sense, thanks.

PR #451, opened 2015-10-29
Rejection filter for current labs production

jalavik:
Adds an automatic reject filter for arXiv workflow on current Labs production.

### DIFF
--- a/inspire/modules/workflows/tasks/actions.py
+++ b/inspire/modules/workflows/tasks/actions.py
@@ -74,6 +74,19 @@ def reject_record(message):
     """Reject record with message."""
     @wraps(reject_record)
     def _reject_record(obj, eng):
+        from inspire.modules.audit.api import log_prediction_action
+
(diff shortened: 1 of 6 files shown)


inspire/modules/workflows/tasks/actions.py
```diff
@@ -74,7 +74,45 @@ def reject_record(message):
...
+    results = obj.get_tasks_results()
+    prediction_results = results.get("arxiv_guessing", {})
+    classification_results = get_classification_from_task_results(obj)
+
+    if prediction_results and classification_results:
+        prediction_results = prediction_results[0].get("result")
+        score = prediction_results.get("max_score")
+        decision = prediction_results.get("decision")
+        core_keywords = classification_results.get("Core keywords")
+        if decision.lower() == "rejected" and score > 0 and \
```
jalavik: @ksachs @fschwenn FYI, here is the current filter code.

PR #453, opened 2015-10-30
search: add collection pages for each collection

nikpap:
* Displays all collection titles on second top bar.

* Adds a collection landing page for each of the 8 collections.

* Adds twitter feed on HEP collection page.

* Removes search bar from from top bar.

* Makes collections header stick to top.


inspire/base/templates/search/collection.html
```diff
@@ -0,0 +1,34 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#}
+
+{%- extends "search/collection_base.html" -%}
+
+{% macro search_bar(collection) %}
+  <form action="/search">
```
jmartinm: give the form `name=search` as in some places that is implied (like in the facets https://github.com/inveniosoftware/invenio-search/blob/master/invenio_search/static/js/search/search.js#L186)

inspire/base/templates/search/collection.html
```diff
@@ -0,0 +1,34 @@
...
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#}
+
+{%- extends "search/collection_base.html" -%}
```
jmartinm: I don't think we need to extend from Invenio's default template as we are not using any of it. Please double check.

Probably enough extending from `page.html`

inspire/base/templates/search/collection_hep.html
```diff
@@ -0,0 +1,84 @@
...
+      <div class="subtitle">
+        Example queries:
+        <a href="/search?p=arXiv%3A1007.5048&cc=HEP">arXiv:1007.5048</a>
+        | 
+        <a href="/search?p=find+a+ellis+and+refersto+a+witten&cc=HEP">find a ellis and refersto a witten</a>
+      </div>
+    </div>
+  {% endblock %}
+
+  {% block collection_how_to_search %}
```
jmartinm: I guess you don't need the blocks here if there is no child template that will overwrite them.

inspire/base/templates/search/collection.html
nikpap: Correct, resolved this issue

inspire/base/templates/header.html
```diff
@@ -128,14 +104,37 @@
   </nav>
 </section>
 
-{% if request.path  == "/search" %}
+{% if request.path  == "/search" or request.path | is_collection_path %}
```
jmartinm: instead of hardcoding the URLs, endpoints should be used.

For example:

```python
{%- if request.endpoint == 'collections.index' %}
```

You can see all the endpoints defined in the application with:

```bash
$ inveniomanage show-urls
```

And the `flask-toolbar` also displays some info about the current view being used

screen shot 2015-10-30 at 2 58 43 pm

inspire/base/templates/header.html
```diff
@@ -128,14 +104,37 @@
...
+
+  {% if request.path  == "/search" %}
```
jmartinm: same, use endpoints

inspire/base/templates/header.html
```diff
@@ -128,14 +104,37 @@
...
+    {% set collection_name = request.args.get('cc') %}
+  {% else %}
+    {% set collection_name = request.path | check_collection %}
```
jmartinm: I don't think you need a filter here, the request object should have enough information (http://flask.pocoo.org/docs/0.10/api/#flask.Request.url_root)

inspire/base/templates/header.html
```diff
@@ -128,14 +104,37 @@
...
+  {% endif %}
+
   <div id="collections">
     <div class="container">
-      {% set collection_name = request.args.get('cc') %}
-        <span id="{{collection_name | collection_select_current("Literature") }}">Literature</span>
-        <span id="{{collection_name | collection_select_current("Authors") }}">Authors</span>
-        <span id="{{collection_name | collection_select_current("Conferences") }}">Conferences</span>
-        <span id="{{collection_name | collection_select_current("Jobs") }}">Jobs</span>
+        <span id="{{collection_name | collection_select_current("Literature") }}">
```
jmartinm: Not sure I understand why that filter is needed. To add to the id a suffix `-collection` ?

nikpap: Yes, the filter `collection_select_current` checks the collection name against the current collection and adds the `-collection` suffix only in the current collection, so as to highlight it.

jmartinm: I would say `ids` are supposed to be `ids` and not used for that. You can just add a class='active' to the current collection

inspire/base/templates/header.html
nikpap: `request.path` will return `collection/hep` for http://localhost:4000/collection/hep , however we only need the name of the collection: the `hep` part

inspire/base/templates/header.html
nikpap: Each `span` needs to have a unique `id` because each collection has a different color. If the class `active` is added to the active collection, I'm not sure how to have a different color for each one.

inspire/base/templates/header.html
```diff
@@ -128,14 +104,41 @@
   </nav>
 </section>
 
-{% if request.path  == "/search" %}
-  <div id="collections">
+{% if request.endpoint == 'collections.collection'
```
jmartinm: You could simplify with

```python
if request.endpoint in ('collections.collection', 'search.search')
```
and so on.

inspire/base/templates/header.html
```diff
@@ -128,14 +104,41 @@
...
+
+  <div id="collections" class="navbar-fixed-top">
     <div class="container">
-      {% set collection_name = request.args.get('cc') %}
-        <span id="{{collection_name | collection_select_current("Literature") }}">Literature</span>
-        <span id="{{collection_name | collection_select_current("Authors") }}">Authors</span>
-        <span id="{{collection_name | collection_select_current("Conferences") }}">Conferences</span>
-        <span id="{{collection_name | collection_select_current("Jobs") }}">Jobs</span>
+      <span id="hep-collection" class="{{collection_name | collection_select_current("Literature") }}">
+        <a href="/collection/literature">Literature</a>
```
jmartinm: change `/collection/literature` to `/literature`

inspire/base/templates/header.html
```diff
@@ -128,14 +104,41 @@
...
+      </span>
+      <span id="hepnames-collection" class="{{collection_name | collection_select_current("Authors") }}">
```
jmartinm: replace `hepnames` to `authors`

inspire/base/templates/header.html
```diff
@@ -434,19 +421,36 @@ <h5 class="text-left"><ins><a class="cursor-pointer popover-link">External resou
...
-        <span id="{{collection_name | collection_select_current("Conferences") }}">Conferences</span>
-        <span id="{{collection_name | collection_select_current("Jobs") }}">Jobs</span>
-    </div>
-  </div>
-  <div class="collection-header" id="{{collection_name | lower }}-collection">
+
+{#
+## If not home path, show collections bar
+#}
+{% if request.path != '/' %}
```
nikpap: @jmartinm Collections bar is not going to be shown in homepage, right?

inspire/base/templates/search/search_base.html
```diff
@@ -0,0 +1,182 @@
...
+              <strong>{{ _('Your search did not match any records. Please try again.') }}</strong>
+            </p>
+            {{ create_nearest_terms_box()|safe }}
+          </div>
+        </div>
+      {%- endblock index_noresult -%}
+    {%- else -%}
+      {%- block index_results -%}
+        <div class="row">
+          {%- block search_bar %}
```
nikpap: @jmartinm I actually only need to add this block to this template. The rest of the template is taken directly from Invenio. Is there a way to minimize repetition and only add the specific block at exactly this point?

inspire/base/templates/header.html
```diff
@@ -32,355 +32,342 @@
...
+      <div class="navbar-header">
+        <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target=".navbar-ex1-collapse">
+          <span class="sr-only">{{ _("Toggle navigation") }}</span>
+          <span class="icon-bar"></span>
+          <span class="icon-bar"></span>
+          <span class="icon-bar"></span>
+        </button>
+        <a class="navbar-brand" href="{{ url_for('collections.index') }}">
+          <span id="{{ collection_name | sanitize_collection_name }}-collection">
+            {% include 'labs-logo.svg' %}
```
nikpap: @jmartinm I've moved the logo file to `/templates` because it seems like I can't include it from any other folder over here. Is that valid?

The logo needs to be output this way in order to have access through CSS to the elements of the SVG file, in order to change the colors of the logo.

jmartinm: Was going to comment on it as it looks weird but if that is needed I guess it should be ok.

inspire/base/templates/header.html
jmartinm: You can leave it hidden for the moment, I think it will be better to have the Literature collection as our front page.

inspire/base/templates/search/search_base.html
jmartinm: If the rest of the template is exactly the same then just having in your template `extends` and:

```
{%- block search_bar %}
            <div id="collection-search-bar" class="col-md-12">
              <div id="search-bar">
                <form name="search" action="/search">
                  <div class="input-group">
                    <input type="text" name="p" placeholder="Search" autofocus>
                    <input type="hidden" name="cc" value="{{ collection.name | lower }}">
                      <span class="input-group-btn" id="{{ collection.name | lower }}-search">
                      <button class="btn btn-primary btn-block">
                        <i class="fa fa-search fa-2x"></i>
                      </span>
                      </button>
                  </div>
                </form>
              </div>
            </div>
          {%- endblock search_bar %}
```

should work.

inspire/base/templates/search/results.html
```diff
@@ -162,7 +163,7 @@
             </div>
           </div>
           <hr>
-          {% if not request.args.get('cc') or request.args.get('cc') == "HEP" %}
+          {% if not request.args.get('cc') or request.args.get('cc') == "hep" %}
```
jmartinm: Add ` request.args.get('cc').lower()`

inspire/base/templates/search/collection_basic.html
```diff
@@ -0,0 +1,44 @@
...
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#}
+
+{%- extends "page.html" -%}
+
+{% bundles "collection.css" %}
+{% bundles "searchbar.css" %}
+
+{%- block header %}
+  {% if collection is defined %}
```
jmartinm: this can be removed.

inspire/base/templates/records/base.html
```diff
@@ -27,11 +27,6 @@
 {%- block header %}
   {%- include "base/scripts.html" %}
   {{- super() }}
```
jmartinm: Check if this template is still needed.

inspire/base/templates/records/Inspire_Default_HTML_detailed_macros.tpl
```diff
@@ -38,34 +27,14 @@
...
-    {% for collection in collections %}
-      {% set collection_name = collection.get('primary') %}
-      {{ collection_name | collection_get_title | safe }}
-    {% endfor %}
-  {% endif %}  
+  {{ collection.name | get_collection_title | safe }}
 {% endmacro %}
 
 {% macro search_bar() %}
   <div class="search-box pull-right" style="display: inline-block;">
```
jmartinm: Instead of this we can reuse the same main search bar.

inspire/base/templates/format/record/Institution_HTML_brief.tpl
```diff
@@ -19,6 +19,8 @@
 
 {% extends "format/record/Default_HTML_brief_base.tpl" %}
```
jmartinm: Create a base parent that will have the common elements like:

```
  {%- block bundles %}
{{super()}}
{% bundles "brief-results.css" %}
  {%- endblock %}
```

inspire/base/static/less/base/header.less
```diff
@@ -160,6 +185,47 @@ body {
...
+
+    #institutions-collection & {
+        fill: @institutions-collection !important;
+    }
+
+    #data-collection & {
+        fill: @data-collection !important;
+    }
+}
+
```
jmartinm: Remove extra space.

inspire/base/templates/records/base.html
nikpap: Can confirm it's still needed.

nikpap:
@jmartinm This is what collection landing pages currently look like:

- Literature landing page
capture d ecran 2015-11-19 a 14 36 25

- Detailed record
capture d ecran 2015-11-19 a 14 31 18

- Search results page
capture d ecran 2015-11-19 a 14 36 56

jmartinm:
Merged

e53d525 format: adds search bar on search results
aef3e04 search: Add landing pages for all collections

PR #454, opened 2015-11-04
docs: mock certain modules when building

jalavik:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -30,6 +30,15 @@ except ImportError:
           file=sys.stderr)
 
 
+from mock import Mock as MagicMock
+
+class Mock(MagicMock):
+    @classmethod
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -30,6 +30,15 @@
           file=sys.stderr)
 
 
+from mock import Mock as MagicMock
```
jmartinm: Does not work with:

```python
In [1]: from mock import MagicMock
```
?

jalavik: Actually I just followed http://docs.readthedocs.org/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules so I am not sure why they are not just doing that.

jalavik: I'll merge for now I think and then see if it helped the building at all.

Issue #459, opened 2015-11-04
Export functionality improvements


jmartinm:
* [ ] Loading spinner is always displayed close to the Download button. The element can be hidden from the beginning with css `display:none` and then shown when needed.
* [ ] All the files exported are called `export_as`. Would be nice to have the name based on the export format. Like `latexus.tex`
* [ ] CV Format Latex and CV Format Text have `<br>` tags in them. Not sure that is intended when downloading?
* [ ] The dropdown is not well aligned

![screen shot 2015-11-04 at 5 06 32 pm](https://cloud.githubusercontent.com/assets/580727/10943809/614cedde-8316-11e5-86a1-bef0e2bd9aad.png)

* [ ] We might still find the case when people prefer to just see the output format as part of the results ( like [currently on INSPIRE](http://inspirehep.net/search?ln=en&ln=en&p=author%3A%27Ellis%27&of=hx&action_search=Search&sf=earliestdate&so=d&rm=&rg=25&sc=0) ). We should ask certain users in the next round of user testing what they think about this.

ioannistsanaktsidis:
@jmartinm the PR for the improvements that you mentioned https://github.com/inspirehep/inspire-next/pull/461

jmartinm:
DOne

PR #461, opened 2015-11-05
formatter: export functionality improvements

ioannistsanaktsidis:
* Addresses #459.

### DIFF
--- a/inspire/base/static/js/search/search_results.js
+++ b/inspire/base/static/js/search/search_results.js
@@ -53,6 +53,18 @@ require(['jquery', 'bootstrap'], function($) {
     });
   });
 
+  var checkboxes = document.querySelectorAll('input.checkbox-results'),
+    checkall = document.getElementById('checkbox-select-all');
+
+  for (var i = 0; i < checkboxes.length; i++) {
+    checkboxes[i].onclick = function() {
+      var checkedCount = document.querySelectorAll('input.checkbox-results:checked').length;
+
+      checkall.checked = checkedCount > 0;
+      checkall.indeterminate = checkedCount > 0 && checkedCount < checkboxes.length;
+    }
+  }
+
   var sList = [];
   var EXPORT_LIMIT = 5000;
   $('#checkbox-select-all').click(function(event) {
@@ -100,6 +112,7 @@ require(['jquery', 'bootstrap'], function($) {
       this.checked = false;
     });
     $('#checkbox-select-all').prop('checked', false);
+    $('#checkbox-select-all').prop('indeterminate', false);
   });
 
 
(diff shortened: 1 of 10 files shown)


jmartinm:
Some small things I see:

*  [ ] Accessing the page with http://localhost:4000/search?p=&cc=HEP&of=tlcv breaks the page formatting
screen shot 2015-11-05 at 2 35 20 pm

* [ ] It would be nice if the formats received the separator `\n` or `<br/>` as a parameter in `__init__()` so that you could use the same code both in HTML and in text (on Download)

* [ ] When selecting a few search results it would be good if the top `select-all` checkbox became 'half selected' like here:

screen shot 2015-11-05 at 2 40 18 pm

To easily deselect results.

* [ ] Instead of separating `obj['format']` and `obj['prefix']` isn't it simpler to just have `obj['filename']` ?

ioannistsanaktsidis:
@jmartinm  for the comment ```  Accessing the page with http://localhost:4000/search?p=&cc=HEP&of=tlcv breaks the page formatting ```
I fixed for now not to break, but we have to decide if we allow users to see the results with of=param then I guess we will have to apply the checkboxes functionality to that templates too. Right?

ioannistsanaktsidis:
With the latest update:
I have just left the second bullet and want to check some staff on the 1st. The last 2 bullets are ok now. Cheers.

ioannistsanaktsidis:
@jmartinm all bullets have been updated with the latest amend. Cheers :dancer:

jmartinm:
Please rebase with latest `master`.

inspire/base/static/js/search/search_results.js
```diff
@@ -133,52 +134,70 @@ require(['jquery', 'bootstrap'], function($) {
...
       }).done(function() {
-        $loading.hide();
+        $('#spinner-download').hide();
         $('#download-format').text('Download');
         $('#download-format').removeClass('disabled');
       });
     }
   });
 
+  var checkboxes = document.querySelectorAll('input.checkbox-results'),
```
jmartinm: you moved away from jQuery and to vanilla JavaScript?

inspire/base/static/js/search/search_results.js
```diff
@@ -118,13 +119,13 @@ require(['jquery', 'bootstrap'], function($) {
...
 
-  var $loading = $('#spinner-download').hide();
   $("#download-format").on('click', function() {
     var obj = get_download_information($('#dropdown-export').text().trim());
     if (sList.length != 0) {
-      $loading.show();
+      $('#spinner-download').show();
       $('#download-format').text('Downloading...');
       $('#download-format').addClass('disabled');
+      console.log(sList);
```
jmartinm: leftover

ioannistsanaktsidis:
@jmartinm the commit is rebased now. Concerning the indeterminate state in the select all check-box there is a problem when you select all the results one by one. The checkbox becomes checked but there are some staff missing. It will take me some time to deal with it, so should I take care or continue in the facets?

jmartinm:
Merged 835286c94658cd151bbdd2ce21dad85955239ddb

PR #490, opened 2015-11-10
format: detailed HEP record MathJax rendering

### LINKED ISSUE #489: MathJax is not enabled on detailed records

nikpap:
* Loads MathJax from CDN in detailed HEP records. (closes #489)

### DIFF
--- a/inspire/base/templates/search/search.html
+++ b/inspire/base/templates/search/search.html
@@ -19,23 +19,9 @@
 
 {% extends "search/search_base.html" %}
 
+{% from "format/record/Inspire_Default_HTML_general_macros.tpl" import mathjax %}
+
 {%- block javascript %}
 {{ super() }}
-<script src="//cdnjs.cloudflare.com/ajax/libs/mathjax/2.5.3/MathJax.js?config=TeX-AMS-MML_HTMLorMML"></script>
-<script type="text/javascript">
-    require([
-      "jquery",
-      ], function ($) {
-          $(document).ready(function () {
-            MathJax.Hub.Config({
-                tex2jax: {inlineMath: [['$', '$'], ['\\(', '\\)']],
-                processEscapes: true},
-                showProcessingMessages: false,
-                messageStyle: "none"
-            });
-            MathJax.Hub.Queue(["Typeset", MathJax.Hub]);
-          })
-      }
-    );
-</script>
+{{ mathjax() | safe }}
 {% endblock %}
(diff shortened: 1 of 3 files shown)


jalavik:
Don't we have MathJax loading from CDN already somewhere? @kaplun

nikpap:
@jalavik MathJax for search results is loaded here: https://github.com/inspirehep/inspire-next/blob/master/inspire/base/templates/search/search.html

inspire/base/templates/format/record/Inspire_HTML_detailed_macros.tpl
```diff
@@ -328,4 +328,24 @@
       <a class="btn btn-default pull-right" href="/search?p=refersto%3Arecid%3A+{{record['control_number']}}&cc=HEP" data-toggle="modal">View in Search Results</a>
     </div>
   </div>
+{% endmacro %}
+
+{% macro mathjax() %}
+<script src="//cdnjs.cloudflare.com/ajax/libs/mathjax/2.5.3/MathJax.js?config=TeX-AMS-MML_HTMLorMML"></script>
```
kaplun: It's better to move out the 2.5.3 version, so that we can easily upgrade MathJax.

kaplun: The whole MathJax macro should be put in some generic utils so that it can be reused in all contexts.

jalavik: Indeed. Then also update the search results to use this new generic macro.

nikpap:
@kaplun I've moved the version to an outside variable and @jalavik the macro has been moved to a general file and reused in search results.

inspire/base/templates/search/search.html
```diff
@@ -19,23 +19,9 @@
...
-                processEscapes: true},
-                showProcessingMessages: false,
-                messageStyle: "none"
-            });
-            MathJax.Hub.Queue(["Typeset", MathJax.Hub]);
-          })
-      }
-    );
-</script>
+{{ mathjax() | safe }}
```
jalavik: Is it intended that `safe` was used here and not above?

jalavik:
@nikpap Thanks!

inspire/base/templates/search/search.html
nikpap: Yes, otherwise the output of the macro was not rendering correctly. I've updated the macro call on the detailed template to mark it safe as well.

Am I right to assume that we can automatic escaping  enabled by default in Jinja2? http://jinja.pocoo.org/docs/dev/templates/#working-with-automatic-escaping

PR #492, opened 2015-11-10
WIP citations: Track changes

Panos512:
* Tracks changes on references.

* In the future it will alert the cache to delete modifieded
  citations.

### DIFF
--- a/inspire/modules/citations/__init__.py
+++ b/inspire/modules/citations/__init__.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2015 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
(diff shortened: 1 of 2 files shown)


inspire/modules/citations/tasks.py
```diff
@@ -54,6 +55,41 @@ def update_records_citations(new_citations):
...
+        references = sender['references']
+        for reference in references:
+            recid = reference.get('recid')
+            if recid is not None:
+                a.append(recid)
+    b = []
+    if 'control_number' in sender:
+        from invenio_ext.es import es
+        recid = sender['control_number']
+        rec = es.get(index='hep', id=recid)
```
jmartinm: This might throw an exception when a non-hep record is modified as they are stored in a different index

Panos512: Indeed this could be a problem. Is there any chance that a 'hep' record and a record from another collection can never have the same id? If this is the case we can catch the 'NotFoundError' and return without doing anything. If not we have to think of something else.

tomaszgy:
continued in #621

jmartinm:
Superseded by https://github.com/inspirehep/inspire-next/pull/492

PR #497, opened 2015-11-11
testsuite: new ViewsTests

kaplun:
* Introduces new testsuite for testing views.

* Adds new tests to access all existing records while checking none
  returns an error.

### DIFF
--- a/inspire/dojson/experiments/fields/bd1xx.py
+++ b/inspire/dojson/experiments/fields/bd1xx.py
@@ -121,31 +121,25 @@ def related_experiments(self, key, value):
     }
 
 
+@utils.filter_values
 @experiments.over('date_started', '^046..')
 def date_started(self, key, value):
     """Date created."""
-    def get_value(value):
-        return value.get('s')
-
-    date_created = self.get('date_created', [])
-
-    for element in value:
-        if 's' in element:
-            date_created.append(get_value(element))
-    if date_created:
-        return date_created[0]
+    if isinstance(value, list):
+        for elem in value:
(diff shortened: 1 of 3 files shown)


kaplun:
cc @jmartinm

jmartinm:
Very nice! So we have 3 broken records at the moment.

```
 [1110515, 1115448, 1210630]
```

@nikpap Can you look into it?

kaplun:
Let me first double check because I have a set of fixes (based on new demo records) that might already solve this issue.

I'll add them as commit on top.

kaplun:
Done. Let's see if there are still failing records. In that case I have no further fixes :smile:

nikpap:
Hey @jmartinm , @kaplun the issue with the 2 later records might be that no collections are assigned to them.

When not logged in, this message appears:

```The record you are trying to access has just been submitted to the system and needs to be assigned to the proper collections. It is currently restricted for security reasons until the assignment will be fully completed. Please come back later to properly access this record.```

When logged in, the pages are rendered correctly but appear empty.

The first record seems to have another issue that will require further investigating ;)

PS. Let me know of the outcome

kaplun:
Ha! Interesting that there are records without collections. My fix actually is for the Experiment dojson, related to dates, so these 3 records will likely still fail.

kaplun:
OK Indeed Travis confirms the same errors. So you can go ahead with the investigation :+1:

jmartinm:
Merged https://github.com/inspirehep/inspire-next/pull/505

PR #466, opened 2015-11-05
elasticsearch: ICU plugin

Panos512:
* Removes unused analyzers from elasticsearcg configuration

* Uses ICU plugin to deal with non-ascii characters.
  NOTE: The plugin must be installed in the elasticsearch root directory
  by running "bin/plugin -install elasticsearch/elasticsearch-
  analysis-icu/2.7.0"

Signed-off-by: Panos Paparrigopoulos <panos.paparrigopoulos@cern.ch>

PR #501, opened 2015-11-12
Icu

kaplun:


Panos512:
@kaplun I added a test at #466 to check if ICU is installed. I think we should include it here too.

kaplun:
Ouch! I missed it :)

Issue #500, opened 2015-11-11
Problems after applying facets


ioannistsanaktsidis:
- [x] After searching something e.g higgs and you check and uncheck a facet the url after all that clicks becomes like ``` http://localhost:4000/search?ln=en&post_filter=author%3A%22Aaij%2C%20Roel%22&p=higgs&p=higgs&p=higgs&p=higgs&p=higgs&p=higgs ``` . The page is not crashing but the url needs to be refactored. -@ioannistsanaktsidis

- [x] Filter by year is not working correctly. E.g it shows that year 2015 has 33 results but only 9 appear. -@jmartinm

- [x] If you first click a facet and then make a search the facets should clear again(?). E.g you click Adinolfi, Marco and you get 5 results. If you make a search afterwards it should reset the filter I guess. -@ioannistsanaktsidis

- [x] Selecting record to export and then clicking a facet it keeps the previous event in the Download button. So if you select 10 records and then you click the facet that gives you 5 results, you are still able to download the previous 10. We have to unbind the event every time a facet is clicked. -@ioannistsanaktsidis

ioannistsanaktsidis:
@jmartinm here are some of the problems that I found. I will keep searching.

ioannistsanaktsidis:
For the first bullet I found what causes the problem, the problem is in updateURL function in search.js. After you merge the commit for invenio-search in master I can apply my PR on top.

PR #503, opened 2015-11-12
facets: fixes search bug in facets

ioannistsanaktsidis:
* Addresses #500.

### DIFF
--- a/inspire/base/templates/search/form/form.html
+++ b/inspire/base/templates/search/form/form.html
@@ -1,0 +1,32 @@
+{#
+# This file is part of INSPIRE.
+# Copyright (C) 2015 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


ioannistsanaktsidis:
@jmartinm this one is for this bullet: 
```
If you first click a facet and then make a search the facets should clear again(?). E.g you click Adinolfi, Marco and you get 5 results. If you make a search afterwards it should reset the filter I guess. 
```

nikpap:
@ioannistsanaktsidis please not that in the copyright, `Invenio` needs to be changed to `INSPIRE`.

What's the difference with https://github.com/inspirehep/invenio-search/blob/master/invenio_search/templates/search/form/form_base.html ?

ioannistsanaktsidis:
I removed this line from the template  <input type="hidden" name="
post_filter" value="{{request.args.get('post_filter', '')}}"> which we dont
need in the overlay.

jmartinm:
Did you make sure that removing the input field does not have any side effect or create some error in the facets JS code? https://github.com/inveniosoftware/invenio-search/blob/master/invenio_search/static/js/search/search.js#L97

ioannistsanaktsidis:
I checked multiple combinations and I can't reproduce any problem. Also the
console is clear.

2015-11-12 14:07 GMT+01:00 Javier Martin Montull <notifications@github.com>:

ioannistsanaktsidis:
I will close this PR as we are going to take care the problem in the form.js file.

PR #506, opened 2015-11-12
deposit: literature form update thesis date input values

tomaszgy:
Works here fine with datetimepicker from our repo


jmartinm:
Merged 
ffe43ef base: import bootstrap-datetimepicker from inspirehep's repo
2bd4cc9 deposit: literature form update thesis date input values

Issue #507, opened 2015-11-12
HepNames record errors


jmartinm:
With the new demo records proposed, the following errors appear:

https://travis-ci.org/inspirehep/inspire-next/builds/90724932#L521

They seem to be caused by HepNames template inheriting from the INSPIRE default detailed template which should not be the case.

jmartinm:
fixed.

PR #509, opened 2015-11-12
formatter: render the appropriate authors detailed template

nikpap:
* Resolves an issue where the default template was incorrectly rendered, instead of the authors template. (adresses #507)


jmartinm:
Good catch :)

PR #517, opened 2015-11-16
format: show more abstract in HEP collection records

nikpap:
* Implements the 'show more' and 'show less' abstract in pure CSS.

* Resolves an issue with HEP detailed record abstract not displaying correctly.


jmartinm:
This PR has not been updated since we spoke IRL about the changes needed (like inline `Show more`) right?

nikpap:
@jmartinm I've updated this PR to reduce the size of text and removed the arrows. Placing it on the same line would require a completely different approach with JS. Ss far as I researched, that can't be done simply with CSS. Here's what it currently looks like:

capture d ecran 2015-11-19 a 10 50 54

capture d ecran 2015-11-19 a 10 51 01

jmartinm:
Did you give a try to this? http://stackoverflow.com/questions/26002974/css-text-ellipsis-including-more-link

nikpap:
I tried the solutions given in the thread, but none seem to currently work. It seems like the only way to have the "show more" text inline would be to use javascript.

jmartinm:
Alright, I will merge this one for the moment.

An alternative in JS is to use the `readmore` library (which we already install anyway)

Issue #411, opened 2015-10-12
Faceting of records in Literature


jmartinm:
Facets (now called aggregations in ElasticSearch) need to be implemented and the following facets need to be present in the Literature collection:

*  [x] Subject (field 65017 with subfield 2 == INSPIRE)
* [x] Document type (subset of 980__a): It also needs to be made more human readable, e.g `Conference Paper` instead of `ConferencePaper`
* [x] Author
* [x] Year
* [x] Experiment (field 693__e)

After merging https://github.com/inspirehep/inspire-next/issues/426 basic faceting is available. Things to address:

* [x] Order facets. Currently the order the aggregations are defined in `search.py` does not reflect the UI. In the template the aggregations are iterated blindly and displayed. One option is in our INSPIRE template to not iterate over them but for each Facet just access directly its content.
* [x] BUG: Fix infinite &p addition to URL while changing facets
* [ ] BUG: In the Author facet. The first result shows 12 matches but when clicked 11 results are displayed.
* [x] BUG?: In the demo site with 200 records, clicking on a facet has a slow response. Time seems to be spend in the JavaScript code as the server does not show `GET /search?post_filter` while waiting.
* [x] We need postprocessing of the aggregations for certain cases like `date-by-year` aggregation. Results are returned in `epoch` format and need to converted to human readable year. Also for document types we can introduce a mapping between collection names and human readable document types. We can either do this in JavaScript or do it in the backend like [hepdata](https://github.com/HEPData/hepdata/blob/master/hepdata/ext/elasticsearch/aggregations.py#L29)
* [x] Selecting a given facet should refresh the whole page and facets should be regenerated so that counts reflect the current user query AND the selected facets.
* [x] All unnecessary JavaScript should be removed from the page (like legacy facets JavaScript)
* [x] Design can be refined and the title of each facet should be `Filter by Author`, `Filter by Document type`
* [ ] As an extra we should add the widget to show the currently selected filters and a Clear Filters button.
![screen shot 2015-10-30 at 10 25 24 am](https://cloud.githubusercontent.com/assets/580727/10842243/9b93d888-7ef0-11e5-97c6-1f0882ba287e.png)

NOTE: There is the following Invenio related ticket https://github.com/inveniosoftware/invenio/issues/3415 that does not seem active at the moment.

kaplun:
How these facets have been chosen?

jmartinm:
Depends on https://github.com/inspirehep/inspire-next/issues/426

[...]

PR #521, opened 2015-11-17
facets: adds filter by subject

ioannistsanaktsidis:
* Addresses #411.

### DIFF
--- a/inspire/config.py
+++ b/inspire/config.py
@@ -292,6 +292,7 @@ SEARCH_ELASTIC_KEYWORD_MAPPING = {
         'e': ['exactauthor.raw'],
     },
     "abstract": ["abstracts.value"],
+    "subject": ["subject_terms.scheme"],
     "collection": ["_collections"],
     "doi": ["dois.value"],
     "affiliation": ["authors.affiliations.value"],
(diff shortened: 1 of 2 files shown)


inspire/config.py
```diff
@@ -313,6 +314,11 @@
 
 SEARCH_ELASTIC_AGGREGATIONS = {
     "hep": {
+        "subject": {
+            "terms": {
+                "field": "subject_terms.scheme"
```
jmartinm: I don't think this will be correct. In the `subject_terms.scheme` field we store where the subject term comes from (like `INSPIRE`) but what we want to show in the values is the actual values (`subject_terms.term`)

The tricky part is that we don't want to show all of the `subject_terms.term` (only the ones that have INSPIRE as a `scheme`) so here we will need to add a new field to the json document only with the values we want. I will look for an example on how to extend the json before indexing and you can use it.

jmartinm:
I have created a PR https://github.com/inspirehep/inspire-next/pull/525 that continues this one. Seems to work fine, please have a look.

ioannistsanaktsidis:
Hey Javier, I am testing your PR, do  you remember if it was returning any
results in the buckets?

2015-11-17 18:17 GMT+01:00 Javier Martin Montull <notifications@github.com>:

jmartinm:
Yes, when I tested it yesterday it was working fine (including the new Subjects)

ioannistsanaktsidis:
Strange, for me it was returning 0 results in the buckets for subjects.
Although the subject was in the bucket. I will keep you up to date tomorrow
after solving the database recreate problem. Cheers.

2015-11-18 18:57 GMT+01:00 Javier Martin Montull <notifications@github.com>:

jmartinm:
Merged https://github.com/inspirehep/inspire-next/pull/525

PR #523, opened 2015-11-17
format: authors modal window

nikpap:
* Removes the authors modal window from the HEP brief results.

### DIFF
--- a/inspire/base/templates/format/record/Inspire_Default_HTML_brief.tpl
+++ b/inspire/base/templates/format/record/Inspire_Default_HTML_brief.tpl
@@ -47,7 +47,7 @@
             {% endif %}
           </b>
       </h4>
-      {{ render_record_authors(is_brief=true) }}
+      {{ render_record_authors(is_brief=true, show_affiliations=false) }}
       <div class="row"><div class="col-md-12"><p></p></div></div>
       <div class="row">
       {% if record.get('publication_info') %}
--- a/inspire/base/templates/format/record/Inspire_Default_HTML_detailed.tpl
+++ b/inspire/base/templates/format/record/Inspire_Default_HTML_detailed.tpl
@@ -33,7 +33,7 @@
     {{ record_title() }}
   </div>
   <div id="record-authors">
-    {{ render_record_authors(is_brief=false, number_of_displayed_authors=20) }}
+    {{ render_record_authors(is_brief=false, number_of_displayed_authors=25) }}
   </div>
   <div class="journal">
     {{ record_publication_info() }}
--- a/inspire/base/templates/format/record/Inspire_Default_HTML_general_macros.tpl
+++ b/inspire/base/templates/format/record/Inspire_Default_HTML_general_macros.tpl
@@ -28,7 +28,7 @@
   </a>
 {% endmacro %}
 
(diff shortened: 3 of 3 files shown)


jmartinm:
Just tried it and it works :)

For the case when there are more authors, in brief results we can just write `et al.` (like currently in INSPIRE)

screen shot 2015-11-17 at 6 47 14 pm

And also remove the latest semicolon before the `et al.` (like in the screenshot)

salmele:
It would be good to check that

1) we have the same number of authors than INSPIRE has in the function to export references (it was a delicate balance of how many we show)

2) et al. would look nicer in italic (as it is an abbreviation from latin)

3) do we need another discussion about Doe, John vs. John Doe in the way we display names? so far INSPIRE had done John Doe AND NOT Doe, John as you are doing here and maybe is appropriate not to introduce this change

jmartinm:
Thanks @salmele , good points.

> 3) do we need another discussion about Doe, John vs. John Doe in the way we display names? so far INSPIRE had done John Doe AND NOT Doe, John as you are doing here and maybe is appropriate not to introduce this change

I did not realise about this, indeed in our metadata all authors are stored as `Doe, John` but then are shown as `John Doe`.

@nikpap You can see the current code that renders authors in INSPIRE [here](https://github.com/inspirehep/inspire/blob/master/bibformat/format_elements/bfe_INSPIRE_authors.py#L27).

And the usage in [brief format](https://github.com/inspirehep/inspire/blob/e606ebdf268eb9273626615923a51670407e52c7/bibformat/format_templates/Default_HTML_brief.bft#L9) and [detailed format](https://github.com/inspirehep/inspire/blob/e606ebdf268eb9273626615923a51670407e52c7/bibformat/format_templates/Default_HTML_detailed.bft#L14) where you can see the limits of authors that we currently show. Please keep the same ones.

What does not seem very efficient to me is that for each search result we run 300 lines of code just to render the authors, @kaplun @jalavik do you think that for these cases it would be appropriate to create a new field upon indexing (like [here](https://github.com/inspirehep/inspire-next/pull/525/files#diff-43d4373c7917cf189c2f8999397658a3R54)) so that the field contains the authors in the form we will display them?

Or it is better to cache them (as part of the brief HTML snippet) in which case we need to know when to delete the cache (`before_record_index` for example)

kaplun:
IMHO, the whole list of authors of a collaboration is rarely useful, hence it's not needed to be included into the brief results. If one wants to really check she can access the detailed record :-)

salmele:
+1

And I guess some minor speed improvement could also come

Salvatore

--

nikpap:
I've updated this PR with the new author limits for brief and detailed record - 10 and 25 respectively.

Regarding the way we display authors, I'm wondering why in the current production website, authors are rendered as "John Doe" in HEP records, but as "Doe, John" in HEPNAMES records. Wouldn't it be better to display authors the same way in both cases?

salmele:
I believe the author pages will have to look like John DOE as well. We can stay with John DOE on the search results and when we launch the author pages we check again.

jmartinm:
Merged 7524b5fdb92fed24a35b639ed08aedd18ecfac81

The rest of the issues related to author display will be continued as a separate issue.

PR #504, opened 2015-11-12
elasticsearch: ICU plugin test

kaplun:
* Adds a test that checks if ICU is installed.

Signed-off-by: Panos Paparrigopoulos <panos.paparrigopoulos@cern.ch>

PR #528, opened 2015-11-19
elasticsearch: icu plugin test

Panos512:
* Enchances #504 with an error message

### DIFF
--- a/inspire/testsuite/test_icu.py
+++ b/inspire/testsuite/test_icu.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2015 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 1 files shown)


inspire/testsuite/test_icu.py
```diff
@@ -0,0 +1,31 @@
...
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+from invenio.testsuite import InvenioTestCase
+
+
+class IcuPluginTests(InvenioTestCase):
+
+    def test_icu(self):
+        from invenio_ext.es import es
```
jacquerie: Maybe you can try in the outer scope:
```python
from invenio_base.wrappers import lazy_import

ES = lazy_import('invenio_ext.es')
```
and then use it inside of `test_icu`.

jacquerie: Scratch that, in order for it be visible inside the test you'd have to define `ES` inside of a `setup` method, but I don't see the point of the added complexity. I think this is fine as it is.

PR #529, opened 2015-11-19
authors: authors names tokenizer

Panos512:
* Ports BibIndexAuthorTokenizer from old invenio.

* Enchances json record before indexing it to elasticsearch
with all the possible variations of the authors name.

### DIFF
--- a/inspire/modules/authors/receivers.py
+++ b/inspire/modules/authors/receivers.py
@@ -1,32 +1,46 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of INSPIRE.
-# Copyright (C) 2014, 2015 CERN.
+# Copyright (C) 2015 CERN.
 #
-# INSPIRE is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
 #
-# INSPIRE is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
 #
 # You should have received a copy of the GNU General Public License
-# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
-#
-# In applying this licence, CERN does not waive the privileges and immunities
-# granted to it by virtue of its status as an Intergovernmental Organization
-# or submit itself to any jurisdiction.
-
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 
 import uuid
 
 from invenio_records.signals import (
+    before_record_index,
     before_record_insert,
     before_record_update,
 )
+
+
+@before_record_index.connect
+def generate_name_variatons(recid, json):
+    """Populates a json record before indexing it to elastic.
+    Adds a field for all the possible variations of an authors name
+
+    :param recid: The id of the record that is going to be indexed.
+    :param json: The json representation of the record that is going to be indexed.
+    """
+    from inspire.modules.authors.utils import author_tokenize
+    authors = json.get("authors")
+    if authors:
+        for author in authors:
+            name = author.get("full_name")
+            if name:
+                author.update({"name_variations": author_tokenize(name)})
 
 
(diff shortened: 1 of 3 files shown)


inspire/utils/author_tokenizer.py
```diff
@@ -0,0 +1,235 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2014, 2015 CERN.
```
nikpap: Might this also need to be changed to INSPIRE?

Panos512: Missed that. Thanks :)

inspire/utils/author_tokenizer.py
```diff
@@ -0,0 +1,233 @@
+# -*- coding: utf-8 -*-
```
kaplun: Almost perfect: can you move the content of this file under `inspire/modules/authors/utils.py`. IMHO it should belongs there. Sorry if that was not clear in my previous comments.

inspire/utils/author_tokenizer.py
```diff
@@ -0,0 +1,233 @@
...
+            if not __check_parts(parts):
+                continue
+            val.extend([''.join(parts)])
+
+        return val
+
+    return _add_squashed(_expand_contract(namelist))
+
+
+def tokenize(phrase):
```
kaplun: This should be called `author_tokenize()`

inspire/utils/author_tokenizer.py
```diff
@@ -0,0 +1,233 @@
...
+                continue
+            # XXX: remember to document that titles can only be applied to
+            # complete last names
+            expanded.extend(_fully_expanded_last_name(
+                exp, [' '.join(last_parts)], title))
+
+    return sorted(list(set(expanded)))
+
+
+def __expand_nonlastnames(namelist):
```
kaplun: I think the `__` prefix can be removed.

inspire/utils/author_tokenizer.py
```diff
@@ -0,0 +1,233 @@
...
+            retval[tag] = retval[tag].strip()
+            retval[tag] = re.split(split_on_re, retval[tag])
+            # filter empty strings
+            retval[tag] = [x for x in retval[tag] if x != '']
+        retval['titles'] = [x.strip() for x in retval['titles'] if x != '']
+
+    return retval
+
+
+def parse_scanned_for_phrases(scanned):
```
kaplun: This should be called `parse_scanned_author_for_phrases`.

inspire/utils/author_tokenizer.py
```diff
@@ -0,0 +1,233 @@
...
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+import re
+
+
+def scan_string_for_phrases(s):
```
kaplun: This should be called `scan_author_string_for_phrases`.

inspire/utils/author_tokenizer.py
```diff
@@ -0,0 +1,233 @@
...
+    """Scan a name string and output an object representing its structure.
+    @param s: the input to be lexically tagged
```
kaplun: Please, amend the docstring in this module to respect the modern Python convention based on `RestructuredText`.

inspire/utils/author_tokenizer.py
```diff
@@ -0,0 +1,233 @@
...
+        {
+            'TOKEN_TAG_LIST' : ['lastnames', 'nonlastnames', 'titles', 'raw'],
+            'lastnames'      : ['Jingleheimer', 'Schmitt'],
+            'nonlastnames'   : ['John', 'Jacob'],
+            'titles'         : ['XVI.'],
+            'raw'            : 'Jingleheimer Schmitt, John Jacob, XVI.'
+        }
+    @rtype: dict
+    """
+    split_on_re = re.compile('[\.\s-]')
```
kaplun: We should compile regular expression in the global scope, otherwise the compilation is not adding any benefit.

Panos512:
@kaplun I fixed the problems (hopefully all of them :stuck_out_tongue: ). Thanks for the notes!

kaplun:
Looks good! Do you have any tests to add?

inspire/modules/authors/receivers.py
```diff
@@ -1,34 +1,48 @@
...
+def generate_name_variatons(recid, json):
+    """Populates a json record before indexing it to elastic.
+    Adds a field for all the possible variations of an authors name
+
+    :param recid: The id of the record that is going to be indexed.
+    :param json: The json representatio of the record that is going to be indexed.
+    """
+    from inspire.modules.authors.utils import author_tokenize
+    authors = json.get("authors")
+    if authors is not None:
```
kaplun: Actually you might want to just write:
`if authors:`

since `authors` could be an empty list.

inspire/modules/authors/receivers.py
```diff
@@ -1,34 +1,48 @@
...
+        for author in authors:
+            name = author.get("full_name")
+            if name is not None:
```
kaplun: Ditto: just write `if name:`, since `name` could be an empty string (who knows...)

Panos512:
@kaplun Indeed there was a test lost somewhere in the migration from invenio_records to inspire. I think it's ok now

PR #530, opened 2015-11-19
format: plots in HEP detailed record

nikpap:
* Displays plots.

### DIFF
--- a/inspirehep/base/static/js/plots.js
+++ b/inspirehep/base/static/js/plots.js
@@ -1,0 +1,53 @@
+ /*
+  * This file is part of INSPIRE.
+  * Copyright (C) 2015 CERN.
+  *
+  * INSPIRE is free software: you can redistribute it and/or modify
+  * it under the terms of the GNU General Public License as published by
+  * the Free Software Foundation, either version 3 of the License, or
+  * (at your option) any later version.
+  *
+  * INSPIRE is distributed in the hope that it will be useful,
+  * but WITHOUT ANY WARRANTY; without even the implied warranty of
+  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  * GNU General Public License for more details.
+  *
+  * You should have received a copy of the GNU General Public License
+  * along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+  *
+  * In applying this licence, CERN does not waive the privileges and immunities
+  * granted to it by virtue of its status as an Intergovernmental Organization
+  * or submit itself to any jurisdiction.
+  */
+
+ define(
+  [
+    'jquery',
+    'flight/lib/component',
+    'bootstrap'
+  ],
+  function($, defineComponent) {
+      'use strict';
+
+      return defineComponent(Plots);
+
+      function Plots() {
+
+        $('#carousel-text').html($('#slide-content-0').html());
+
+        //Handles the carousel thumbnails
+        $('[id^=carousel-selector-]').click( function(){
+          var id = this.id.substr(this.id.lastIndexOf("-") + 1);
+          var id = parseInt(id);
+          $('#plotsCarousel').carousel(id);
+        });
+        
+        // When the carousel slides, auto update the text
+        $('#plotsCarousel').on('slid.bs.carousel', function (e) {
+         var id = $('.item.active').data('slide-number');
+         $('#carousel-text').html($('#slide-content-'+id).html());
+       });
+
+      }
+      
+    });
--- a/inspirehep/base/static/less/format/detailed-record.less
+++ b/inspirehep/base/static/less/format/detailed-record.less
@@ -508,6 +508,57 @@ body {
     text-align: right;
 }
 
+// Plots
+
+#record-plots {
+    &:extend(#record-abstract-keywords);
+
+    #record-plots-title {
+    &:extend(#record-abstract-title);   
+    }
+}
+
+.hide-bullets {
+  list-style:none;
+  margin-left: -40px;
+  margin-top:20px;
+}
+
+#plotsCarousel{
+    min-height: 300px;
+}
+
+.img-responsive,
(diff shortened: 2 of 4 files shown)


nikpap:
This is a first draft of how plots could look like:

capture d ecran 2015-11-19 a 11 59 06

Current issues:

1) Some records have too many plots, these appear like this:

capture d ecran 2015-11-19 a 14 18 43

2) Some of the plots descriptions have a very large text that "overflows" the right part of the screen. We should decide where to cut off.

inspire/base/static/js/inspire_base_init.js
```diff
@@ -25,7 +25,8 @@ require([
   "js/feedback_inspire",
   "js/header/header",
   "js/notification",
-  "js/dropdown-menu/dropdown-menu"
+  "js/dropdown-menu/dropdown-menu",
+  "js/plots"
```
nikpap: @jmartinm We only need to load plots.js in the detailed record, I have a feeling there's another way to do it, other than here.

inspire/base/static/js/plots.js
```diff
@@ -0,0 +1,46 @@
...
+  * granted to it by virtue of its status as an Intergovernmental Organization
+  * or submit itself to any jurisdiction.
+  */
+
+  define([
+    'jquery',
+    ],function($) {
+     
+      // $('#plotsCarousel').carousel({
+      //   interval: 500
```
nikpap: @jmartinm We can define a carousel internal, after which the next image is automatically loaded. Do we want such a functionality?

inspire/base/templates/format/record/Inspire_HTML_detailed_macros.tpl
```diff
@@ -328,4 +328,80 @@
...
+    <!-- Slider -->
+    <div class="row">
+      <div class="col-xs-12" id="slider">
+        <!-- Top part of the slider -->
+        <div class="row">
+          <div class="col-sm-8" id="carousel-bounding-box">
+            <div class="carousel slide" id="plotsCarousel">
+              <!-- Carousel items -->
+              <div class="carousel-inner">
+                {% for url in record.get('urls') %}
```
nikpap: @jmartinm I'm trying to figure a way to skip the second for loop, do you have any suggestions?

The first loop (just as the macro starts) is used to detect if the record has plots, then this loop is used to display them.

inspire/base/templates/format/record/Inspire_HTML_detailed_macros.tpl
```diff
@@ -328,4 +328,80 @@
...
+                  {% if url.get('url').endswith(".png") or url.get('url').endswith(".jpg") %}
```
nikpap: @jmartinm I'm assuming plots are .jpg or .png files. Is that actually always the case?

salmele:
>Current issues:
>1) Some records have too many plots, these appear like this:

Should we just have a number of plots in the lower line as large as the window, and then have scroll arrows also on that line? That's what e.g. AirBnB does on the accommodation pics

>2) Some of the plots descriptions have a very large text that "overflows" the right part of the screen. We should decide where to cut off.

Maybe we just cut when we've filled the vertical space ?

nikpap:
Updates on the current issues:

1) In the current implementation, I've opted to initially display the first 5-6 thumbnails. The user can then click on the "Show more plots" button to display additional plots. Because some of our records have a lot of plots, I have the feeling that it'd be best not to implement a vertical/horizontal scroll in the thumbnails. Scrolling through 75 thumbnails to find the one you need might not be that easy for users of our system. It might be based to gather user feedback on this.

2) I've made the text area on the right side have a fixed height. Any text that exceeds this height will "overflow" and the users will be able to scroll through it.

capture d ecran 2015-11-25 a 16 53 00

jmartinm:
> 1) In the current implementation, I've opted to initially display the first 5-6 thumbnails. The user can then click on the "Show more plots" button to display additional plots. Because some of our records have a lot of plots, I have the feeling that it'd be best not to implement a vertical/horizontal scroll in the thumbnails. Scrolling through 75 thumbnails to find the one you need might not be that easy for users of our system. It might be based to gather user feedback on this.

So when you click `Show more`, where are the plots displayed?

> 2) I've made the text area on the right side have a fixed height. Any text that exceeds this height will "overflow" and the users will be able to scroll through it.

This looks like a good solution to me. Depending on how critical this description is it could also be truncated ending with `...` when it does not fit in the available space. You can leave it like this and we gather feedback next week.

nikpap:
>So when you click Show more, where are the plots displayed?

They are currently displayed like before (see second image in this page). Another option would be to add a scrolling bar to the plots thumbnails, like this: 
capture d ecran 2015-11-25 a 17 08 43

jmartinm:
> They are currently displayed like before (see second image in this page). Another option would be to add a scrolling bar to the plots thumbnails, like this: 

Yes, even from the beginning I would display 2 rows and a vertical scrollbar. I think scrolling horizontally is a bit awkward but vertically should be fine.

salmele:
Vertical scrolling seems neat

inspire/base/static/js/plots.js
```diff
@@ -0,0 +1,59 @@
...
+ define(
+  [
+    'jquery',
+    'flight/lib/component',
+    'bootstrap'
+  ],
+  function($, defineComponent) {
+      'use strict';
+
+      var sList = [];
```
jmartinm: leftover it seems.

inspire/base/static/js/plots.js
```diff
@@ -0,0 +1,59 @@
...
+
+      return defineComponent(Plots);
+
+      function Plots() {
+
+        // $('#plotsCarousel').carousel({
```
jmartinm: leftover?

jmartinm:
Merged 498aa5de98d3af4c161304a969129ac44e7bbbeb

PR #533, opened 2015-11-19
facets: fixes year problem

ioannistsanaktsidis:


ioannistsanaktsidis:
@jmartinm  this one is for the year problem we have in facets. It fixes the problem from the 'inspire' side. We have to  fix the search so as to  accept this kind(e.g http://localhost:4000/search?post_filter=earliest_date%3A%"2007-01-01->2007-12-31") of queries in the post_filter.

jmartinm:
Merged 73c2bf588f4cad744b74cc57b6d3e20be0b8e8e4

PR #536, opened 2015-11-20
migrator: populate via celery

jmartinm:
* populate command now creates records in Celery tasks. The
  signal to force record ids is not needed as this is done in
  the Celery task.


inspire/modules/migrator/tasks.py
```diff
@@ -22,10 +22,66 @@
...
 
 
 @celery.task
-def migrate(url, **kwargs):
+def migrate(source, **kwargs):
     """Main migration function."""
-    pass
+    from dojson.contrib.marc21.utils import split_stream
+    with db.session.begin_nested():
+        for record in split_stream(open(source)):
```
kaplun: There is no Python native way of chunking but you can borrow something from: http://stackoverflow.com/questions/434287/what-is-the-most-pythonic-way-to-iterate-over-a-list-in-chunks

and then have:
```python
for chunk in chunks(split_stream(open(source)), 1000):
    with.db.session.begin_nested():
        for record in chunk:
            ...
    db.session.commit()
    bulk_enhance_records(...)
    bulk_elastic_index_records(...)
```

PR #537, opened 2015-11-20
ES 2.0 compatibility

jalavik:
global: ElasticSearch 2.0 compatibility
    
    * Updates some index mappings to be ES 2.0 compliant.
    
    * Adds temporary dev dependency on `elasticsearch-2.2.0.dev0` due to
      fixes for ES 2.0. Can be removed when 2.2.0 is released.


base.requirements.txt
```diff
@@ -3,6 +3,9 @@
 git+https://github.com/mloesch/sickle@master#egg=Sickle
 git+https://github.com/inveniosoftware/beard@master#egg=beard
 
+# Until 2.2 is out
+git+https://github.com/elastic/elasticsearch-py@master#egg=elasticsearch
```
jmartinm: shall we add the version in case the master branch gets suddenly updated?

setup.py
```diff
@@ -30,6 +30,7 @@
 
 
 requirements = [
+    'elasticsearch>=2.2.0.dev0,<3.0.0',
```
jmartinm: Do we need both here and in `base.requirements.txt` ?

base.requirements.txt
jalavik: Indeed. Will add it.

setup.py
jalavik: So I realized its convenient as it checks that the correct version is installed when running the site. E.g. someone pulls master and then `inveniomanage` will break until they run `pip install -r requirements.txt`. Without it, things may "seem" to work, but people may be running outdated stuff.

I can add a note about this in the commit message.

kaplun:
@jalavik the travis build fails due to the sniffing issue. I think I might have a workaround in #546.

jalavik:
Success! The sniff thing happens also when elasticsearch is down. It was down now due to not being started properly as one has to run it as non-root user. Honestly, our docker image is in need of some love (more decomposed etc.) but it should do for now.

PR #545, opened 2015-11-23
Enhancement/jinja2filter cleanup

jmartinm:


jmartinm:
@ioannistsanaktsidis @nikpap This is a way to reduce a bit of code when defining filters.

Just need to do:

```python
@app.template_filter()
    def is_list(value):
        """Checks if an object is a list"""
        if isinstance(value, list):
            return True
```

And if you want the filter to be named differently than the function name

```python
@app.template_filter('myfiltername')
    def is_list(value):
        """Checks if an object is a list"""
        if isinstance(value, list):
            return True
```

PR #546, opened 2015-11-23
honcho: delayed start

kaplun:
* Starts first elasticsearch, sleep 10 seconds and then starts all
  the other depending processes, thus working around sniffing
  timeouts.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,5 +1,5 @@
-web: inveniomanage runserver
-cache: redis-server
-worker: celery worker -E -A invenio_celery.celery --loglevel=INFO --workdir=$VIRTUAL_ENV
-workermon: flower --broker=amqp://guest:guest@localhost:5672//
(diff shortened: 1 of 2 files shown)


jmartinm:
I was having problems regarding the `sniffing` running `honcho start` although now I tried with latest `master` of https://github.com/inspirehep/invenio-ext and it worked fine. So I am not really sure if this is needed? Somebody else saw problems?

kaplun:
I still have problems... but not tryed with ES 2

jalavik:
As mentioned IRL I have now removed `sniff_on_start` to hopefully avoid this problem.

jalavik:
Just merged #537. Try installing ES 2.0 and see https://github.com/inspirehep/inspire-next/blob/master/README.rst 

Hopefully this will fix your problem.

kaplun:
Indeed!

PR #548, opened 2015-11-23
authors: phonetic name notation

jacenkow:
* Adds assign_signature_block in receivers.py resposible for
  assigning phonetic names of authors (signatures) in HEP
  collection. The blocks are assigned on before_record_index
  signal.

* Renames HEP schema ml_block to signature_block.

* Cleans up assign_uuid in receivers.py. Removes
  before_record_update signal.

* Extends HEP schema with regex pattern for uuid.

### DIFF
--- a/inspire/base/searchext/mappings/hep.json
+++ b/inspire/base/searchext/mappings/hep.json
@@ -121,7 +121,8 @@
                     "type": "string"
                 },
                 "authors": {
-                    "type": "object",
+                    "type": "nested",
+                    "include_in_parent": true,
                     "properties": {
                         "full_name": {
                             "type": "string",
@@ -134,6 +135,9 @@
                           "type": "string",
                           "index": "not_analyzed",
                           "copy_to": "global_fulltext"
+                        },
+                        "signature_block": {
+                            "type": "string"
                         }
                     }
                 },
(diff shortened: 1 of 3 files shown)


kaplun:
Can you rebase on latest master? Then it's good for merging.

jacenkow:
@kaplun updated. However I had to fix all the conflicts, so it would be good to double-check me in ```receivers.py``` and ```mappings/hep.json```. :smile:

inspire/base/searchext/mappings/hep.json
```diff
@@ -121,7 +121,8 @@
                     "type": "string"
                 },
                 "authors": {
-                    "type": "object",
+                    "type": "nested",
+                    "include_in_parent": true,
```
kaplun: Uh cool. Sounds something we need in general? @Panos512 @jmartinm ?

jacenkow: So I need ```nested``` to use my queries and ```"include_in_parent": true``` allows nested type to behave like an object at the same time (during testing, something is accessing authors like it is an object, so this trick necessary to keep everything working).

Panos512: Yeah, as far as I understood `"include_in_parent": true` keeps the intended behavior for what we have done until now. Nested type is good if we need to query array items separately we haven't needed it yet but it could be useful in the future.

PR #550, opened 2015-11-23
Chuncked migrator

kaplun:


kaplun:
@jalavik, @jmartinm this is currently failing to me in the CLI parsing:
```
inveniomanage migrator populate --broken-ouput /tmp/broken.xml --input prodsync20151117173222.xml.gz --dry-run

usage: inveniomanage [-?]
                     {upgrader,detect-system-details,check-for-software-updates,knowledge,show-urls,version,migrator,formatter,config,shell,records,bower,workflows,apache,predicter,assets,database,harvester,oaiharvester,classifier,collect,clean,runserver}
                     ...
inveniomanage: error: too many arguments
```

Can you see what I've done wrong in the `manage` decorators? I don't see anything blatantly wrong...

jmartinm:
Did not see anything obvious. Maybe you need to add a short version of the command?

```
@manager.option('--broken-output', '-b', dest='broken_output', default=None,
                help='Where to write back records that were not possible to migrate')
```

All the ones that only have one, seem to be booleans. When they accept a parameter they are always defined with short and long name...

kaplun:
Yep indeed adding a short version works... Good guess! Not obvious though...

jmartinm:
So just checked the code and looks good. I guess as a step 2 we add the bulk enhancement and the bulk indexing per chunk.

jmartinm:
BTW, there is a small typo in the commit log

```
Improves migragtor task to work on 1000 records at a time.
```

And the branch is currently not rebased with latest `master`

kaplun:
Yep. It's not yet working :( There is still a memory leak in lxml :(

inspire/modules/migrator/manage.py
```diff
@@ -63,11 +55,18 @@
...
 @manager.option('--force', action='store_true', dest='force_import', default=None,
                 help="Force records that are not registered to import on the system")
-def populate(records, collections, file_input=None, input_type=None, force_import=None):
+@manager.option('--broken-output', '-b', dest='broken_output', default=None,
+                help='Where to write back records that were not possible to migrate')
+@manager.option('--dry-run', '-d', action='store_true', dest='dry_run', default=False,
+                help='Whether records should really be imported or not')
+def populate(records, collections, file_input=None, input_type=None,
+             force_import=None, broken_output=None, dry_run=False):
     """Train a set of records from the command line.
```
jalavik: This docstring should perhaps be amended.

inspire/modules/migrator/tasks.py
```diff
@@ -22,10 +22,100 @@
...
+    if buf:
+        yield buf
 
 
 @celery.task
-def migrate(url, **kwargs):
+def migrate(source, **kwargs):
     """Main migration function."""
-    pass
+    broken_output = kwargs.get('broken_output')
```
jalavik: Instead of getting these from the kwargs, it would be better to just add them to the function signature. Say:

```def migrate(source, broken_output=None, dry_run=False, **kwargs):```

PS: Looks also like `kwargs` will be unused then or?

kaplun: Yep!

PR #551, opened 2015-11-23
format: detailed HEP record fixes

nikpap:
- [x] Rearranges the way external links are shown.

- [x] Restores cite record functionality.

- [x] Removes space before keywords.

- [x] Resolves error with `Uncaught Error: undefined missing js/citation_modal` when `ASSETS_DEBUG=False`

### DIFF
--- a/inspirehep/base/static/js/citation_modal.js
+++ b/inspirehep/base/static/js/citation_modal.js
@@ -1,0 +1,68 @@
+/*
+ ** This file is part of INSPIRE.
+ ** Copyright (C) 2015 CERN.
+ **
+ ** INSPIRE is free software: you can redistribute it and/or modify
+ ** it under the terms of the GNU General Public License as published by
+ ** the Free Software Foundation, either version 3 of the License, or
+ ** (at your option) any later version.
+ **
+ ** INSPIRE is distributed in the hope that it will be useful,
+ ** but WITHOUT ANY WARRANTY; without even the implied warranty of
+ ** MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ ** GNU General Public License for more details.
+ **
+ ** You should have received a copy of the GNU General Public License
+ ** along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+ **
+ ** In applying this licence, CERN does not waive the privileges and immunities
+ ** granted to it by virtue of its status as an Intergovernmental Organization
+ ** or submit itself to any jurisdiction.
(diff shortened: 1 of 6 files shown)


jmartinm:
Does the Cite record function work? I get the following error in the console:

```javascript
TypeError: this.after is not a function(…)
```

nikpap:
Nope, it's still a WIP. I've made it work now, but some further changes are required. I have a feeling this might need to be fixed in the brief results as well - the issue you describe is present there too.

inspire/base/templates/format/record/Inspire_Default_HTML_detailed.tpl
```diff
@@ -73,4 +72,14 @@
       {{ record_citations() }}
     </div>
   </div>
+  <script type="text/javascript">
```
nikpap: @jmartinm placing this in the `javascript` or `bundles` block doesn't seem to work. This template inherits from https://github.com/inveniosoftware/invenio-formatter/blob/master/invenio_formatter/templates/format/record/Default_HTML_detailed.tpl , which in turn inherits from https://github.com/inveniosoftware/invenio-formatter/blob/master/invenio_formatter/templates/format/record/Default_HTML_detailed_base.tpl. Neither of these have a `javascript` or `bundles` block.

jmartinm:
I am getting an error in the console:

```Uncaught Error: Script error for "js/citation_modal"```

And `inveniomanage collect` does not seem to copy the file to the static/ folder. Can you remove the file from your folder `var/invenio.base-instance/static/js` and make sure that it is being correctly collected?

nikpap:
Issue resolved, commit amended and rebased.

jmartinm:
There is a problem with this PR when running in production mode. `ASSETS_DEBUG=False` in `invenio.cfg`

And error appears in the Javascript console.

nikpap:
I'm trying to resolve the issue with `ASSETS_DEBUG=False`.

Does turning it to `True` resolve the issue in production mode?

jmartinm:
Yes, `ASSETS_DEBUG=True` works. I can help you debug the problem.

jmartinm:
The error will be solved when this is merged https://github.com/inspirehep/inspire-next/pull/607

Just need to add the citations file here https://github.com/inspirehep/inspire-next/pull/607/files#diff-3fc0e9ddbcfc37b1d8180a8bda46fc6dR24

jmartinm:
Merged cf0633680f182229804678a391360fb5497a26a3

PR #552, opened 2015-11-23
export: fixes select all records problem

### LINKED ISSUE #538: Select-all export problem
When using facets to filter results and then selecting all search results there seems to still be a problem @ioannistsanaktsidis 

screen shot 2015-11-20 at 6 07 48 pm

ioannistsanaktsidis:
* Closes #538.

### DIFF
--- a/inspire/base/static/js/search/search_results.js
+++ b/inspire/base/static/js/search/search_results.js
@@ -12,7 +12,6 @@ define(
     return defineComponent(SearchResults);
 
     function SearchResults() {
-      var minimum_page_limit = $("#results-per-page option:selected").text();
       this.attributes({
         EXPORT_LIMIT: 5000
       });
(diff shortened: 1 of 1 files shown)


inspire/base/static/js/search/search_results.js
```diff
@@ -95,6 +94,7 @@ define(
       }
 
       this.onExportSelectAll = function(ev) {
+        var minimum_page_limit = $("#select-numpages option:selected").text();
```
jmartinm: As it is a dropdown, doesn't it work with:

```
var minimum_page_limit = $("#select-numpages").val();
```
?

ioannistsanaktsidis: No, but it works like this var minimum_page_limit = $("#select-numpages
option:selected").val(); . Should we keep .text() or .val() ?


2015-11-24 11:58 GMT+01:00 Javier Martin Montull <notifications@github.com>:

PR #554, opened 2015-11-24
facets: experiments validation

Panos512:
* Trying to fix misspelled experiment names.

* Adds a new field 'facet_experiment' that should be used
  to generate facets.


kaplun:
cc @jmartinm

inspire/modules/records/receivers.py
```diff
@@ -1,30 +1,57 @@
...
+from invenio_records.signals import before_record_index, before_record_insert
+
+
+@before_record_index.connect
+def match_valid_experiments(recid, json):
+    """Matches misspelled experiment names with valid experiments.
+       Tries to match with valid experiments by matching lowercased and
+       whitespace-free versions of known experiments.
+    """
+    experiments_list_original = [
```
kaplun: Can you add a FIXME comment here to remember ourselves this is just a quick hack and we should use a better method based on checking against the whole experiment list? (You could actually add a github issue on inspire-next related to that and reference it in the FIXME).

inspire/modules/records/receivers.py
```diff
@@ -35,3 +32,33 @@ def populate_inspire_subjects(recid, json):
...
+    """Matches misspelled experiment names with valid experiments.
+       Tries to match with valid experiments by matching lowercased and
+       whitespace-free versions of known experiments.
+    """
+    # FIXME: This list is temporary. We should have a list that is generated
+    # from the current state of our data.
+    from .experiment_list import EXPERIMENTS_LIST
+    experiments_list_original = EXPERIMENTS_LIST
+    experiments_list = map(str.lower, experiments_list_original)
+    experiments_list = [exp.replace(' ', '') for exp in experiments_list]
```
kaplun: Move this into `.experiment_list`.

inspire/modules/records/receivers.py
```diff
@@ -35,3 +32,33 @@ def populate_inspire_subjects(recid, json):
...
+
+    experiments = json.get("accelerator_experiments")
+    if experiments:
```
kaplun: You can actually move the import here, so that there's no computation in case of no experiment.

kaplun:
@Panos512 ouch! Can you rebase once more this branch? Then I think it's good for :shipit:

Panos512:
@kaplun ready!

PR #555, opened 2015-11-25
migrator: index creation changes

jmartinm:
* New migrator create and delete indices command.

* Removes Holding Pen ElasticSearch configuration as it now uses
  the record mappings extended with WORKFLOWS_HOLDING_PEN_ES_PROPERTIES.


jmartinm:
Related branches:

https://github.com/inspirehep/invenio-ext/pull/4
https://github.com/inspirehep/invenio-search/pull/6
https://github.com/inspirehep/invenio-workflows/pull/4

inspire/modules/migrator/__init__.py
```diff
@@ -0,0 +1 @@
+from .receivers import *
```
jacquerie: Uh? Sounds suspicious.

inspire/modules/migrator/manage.py
```diff
@@ -157,25 +152,47 @@ def remove_legacy_tables():
...
+        es.indices.put_alias(index=index + "_v1", name=index)
+        # Create Holding Pen index
+        if mapping:
+            mapping['mappings']['record']['properties'].update(
+                cfg['WORKFLOWS_HOLDING_PEN_ES_PROPERTIES']
+            )
+        name = cfg['WORKFLOWS_HOLDING_PEN_ES_PREFIX'] + index
+        es.indices.delete(index=name, ignore=404)
+        es.indices.create(index=name + "_v1", body=mapping)
+        es.indices.put_alias(index=name + "_v1", name=name)
```
jacquerie: Maybe these three lines deserve a helper function?

inspire/modules/migrator/__init__.py
jmartinm: It's kind of a pattern we have been using in some other modules in order to connect the receivers to the signals.

inspire/modules/migrator/manage.py
jmartinm: Yep good point

jmartinm:
The tests will fail as they rely on the related PRs being merged.

inspire/modules/migrator/manage.py
jmartinm: Done.

inspire/modules/migrator/__init__.py
jacquerie: Uh, ok, It looked like some kind of leftover.

jacquerie:
Other than the addressed comments, LGTM!

jmartinm:
Merged all the related branches in external packages. Waiting for Travis now.

jmartinm:
So now only the (in)famous citation tests are failing. They fail also locally in my computer and I don't see what changes in this branch can affect them.

Looking at it, the records seem to be created during the tests but the field `references_id` is not added to them. In [this line](https://github.com/inspirehep/inspire-next/blob/master/inspire/modules/citations/tasks.py#L50) there is the assumption that calling `update` on the record would update it (even in the database) which seems quite weird.

I am not sure how they were working before.

@kaplun @Panos512 @jalavik any ideas?

inspire/modules/migrator/__init__.py
kaplun: Isn't it enough: `import .receivers`? Maybe not though...

jacquerie: This won't add them to the top-level scope, though.

kaplun: Nope, but, if if it will execute the global part of the module, in order to compile it, it will register the signals etc.

kaplun:
@jmartinm, given how the synchronization between prod and Labs managed to evolve (i.e. by using the enhanced MARCXML which contains all information embedded), there is indeed no longer need for the special hack for citations, which served well meanwhile to help everybody better understanding all the testability, and usage of invenio-records API.

So indeed :+1: for the simplification.

jmartinm:
Ok thanks :+1:

PR #558, opened 2015-11-25
WIP: search: disable search bar in authors landing page

nikpap:

### DIFF
--- a/inspirehep/base/static/less/search/collection.less
+++ b/inspirehep/base/static/less/search/collection.less
@@ -162,6 +162,22 @@
     }
 }
 
+#collection-authors-not-available {
+    &:extend(#collection-how-to-search);
+    padding-bottom: 25px;
+
+    .col-md-12 {
+        font-size: larger;
+        text-align: center;
+        margin-bottom: 20px;
+
+        a.btn {
+            background-color: @authors-collection;
+            color: white;
+        }
+    }
+}
+
 #twitter-feed {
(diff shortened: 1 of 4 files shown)


inspire/base/templates/search/collection_authors.html
```diff
@@ -27,6 +27,13 @@
       </div>      
       {% block search_bar %}
         {{ super() }}
+        <script type="text/javascript">
```
nikpap: Since this is only temporary until the author pages go live, I've opted to use inline JS in this case.

nikpap:
Here is an updated version of how this currently looks like:

capture d ecran 2015-11-26 a 11 13 27

salmele:
Should the copy say "The authors collection is not YET available on INSPIRE labs"

nikpap:
Indeed, that choice of words might be better. I've updated this PR to include the new copy.

jmartinm:
> Should the copy say "The authors collection is not YET available on INSPIRE labs"

Shall we also take the opportunity to get people to know the blog and Twitter ?

------
The authors collection is not yet available on INSPIRE labs. Follow us on [Twitter](http://twitter.com/inspirehep) and through [our blog](http://blog.inspirehep.net/) to get updated when the newly designed collection is ready.

jmartinm:
Merged 3236951fc09d2c2ed2fd1456e6b2cfbc4e98007b

PR #559, opened 2015-11-25
WIP Integrate Invenio-Matcher

jacquerie:
Currently there's some preliminary work to better test the current matching. Actual integration coming soon.


kaplun:
What's the plan for this integration?

jacquerie:
The idea is to integrate it in `inspire.modules.workflows.tasks.matching`. The proposal is to create alternative versions of each task using `invenio-matcher`, probably using the helper functions that this PR currently added tests for.

jacquerie:
So, the first commit contains a few tests for `inspirehep/modules/workflows/tasks/matching.py`. In particular it contains a trick to test tasks in isolation which might prove useful in the future.

The second commit contains an example of integration of `Invenio-Matcher`. You probably don't want to use it as it is, since it's not unwrapping the `MatchResult` objects returned by `Invenio-Matcher`, but it's a trivial modification to make, and the API for `MatchResult` is still very much in flux.

jalavik:
Continued in #678

PR #562, opened 2015-11-26
format: HEP brief and detailed enhancements

nikpap:
- [x] Increases text size in search page search bar.
- [x] Removes full DOI text and displays it on hover.
- [x] Displays arXiv category next to e-Print .
- [x] Displays 'published in' next to the journal.
- [x] Makes buttons gray, as in detailed record references and citations, changing color on hover.
- [x] Moves DOI next to journal in brief results and removes the semicolon.
- [x] Removes external link icons.
- [x] Adds extra space on left side of brief title.
- [x] Moves e-Print, DOI and journal into same line.
- [x] Reduces right column width in brief.
- [x] Makes detailed record search bar text white.
- [x] Removes DOI text, links journal to DOI (under discussion).
    * If journal exists, link to DOI
    * Otherwise, output DOI as usual
- [x] Moves `citations-references` to the bottom of the right column.
- [x] Shorten width of facets, increase results width.
- [x] Adress #591
- [x] Remove background color from between white record blocks.
- [x] Remove duplicate `Published in` text
    * Fixed: 
capture d ecran 2015-12-04 a 12 59 14

- [x] Hide `Show more` text in abstract if all text is already shown.
- [x] Remove duplicate DOI text from detailed record:
capture d ecran 2015-12-04 a 13 18 01
- [x] Display multiple DOI's and journal links in separate places:
capture d ecran 2015-12-04 a 13 18 36

- [x] Check brief results minimum height.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,5 +1,5 @@
 web: inveniomanage runserver
 cache: redis-server
-worker: celery worker -E -A invenio_celery.celery --loglevel=DEBUG --workdir=$VIRTUAL_ENV
+worker: celery worker -E -A invenio_celery.celery --loglevel=INFO --workdir=$VIRTUAL_ENV
 workermon: flower --broker=amqp://guest:guest@localhost:5672//
 indexer: elasticsearch -Dcluster.name="inspire" -Ddiscovery.zen.ping.multicast.enabled=false -Dpath.data="$VIRTUAL_ENV/var/data/elasticsearch"  -Dpath.logs="$VIRTUAL_ENV/var/log/elasticsearch"


nikpap:
Here's what an updated brief result record looks like (hovering over DOI to show the tooltip effect):

capture d ecran 2015-11-26 a 14 29 14

salmele:
What is the use of the DOI word in this context @annetteholtkamp @jmartinm? Not to cut and paste it, as having just the string in the tooltip is not usable for that, right?

It is to allow users, at one click, to go to the journal, right? But then, we can achieve the same effect exactly as we do for arXiv, which is to make the "Eur.Phys..." string an hyperlink.

Then, those who need to know the DOI would go cut paste it from the detailed record.

annetteholtkamp:
We don't always have an unambiguous association of DOI and pubnote. We have quite a few records with multiple DOIs and/or multiple pub notes.

- Annette

On Nov 30, 2015, at 2:20 PM, Salvatore Mele <notifications@github.com<mailto:notifications@github.com>> wrote:


What is the use of the DOI word in this context @annetteholtkamp<https://github.com/annetteholtkamp> @jmartinm<https://github.com/jmartinm>? Not to cut and paste it, as having just the string in the tooltip is not usable for that, right?

It is to allow users, at one click, to go to the journal, right? But then, we can achieve the same effect exactly as we do for arXiv, which is to make the "Eur.Phys..." string an hyperlink.

Then, those who need to know the DOI would go cut paste it from the detailed record.

salmele:
> We don't always have an unambiguous association of DOI and pubnote. We have quite a few records with multiple DOIs and/or multiple pub notes.

We could resolve these, expose all, in the detailed record, and have an heuristic for a single clickable pub note (through DOI) at the brief record level.

nikpap:
Updated version of how the search results look like:

capture d ecran 2015-12-02 a 11 59 25

Hovering on the buttons would display them like this: 
capture d ecran 2015-12-02 a 12 00 59

jmartinm:
> We could resolve these, expose all, in the detailed record, and have an heuristic for a single clickable pub note (through DOI) at the brief record level.

That would be ideal. So far the users we have talked to use the DOI link just to click on it and go to the source. If they want the actual DOI, they would click on the Cite button to get the full citation (including DOI)

Doing the cleanup for ambiguous DOIs/Pubnotes seems like it would be costly and we cannot count on it immediately.

So some open questions regarding the display of the Pubnote/DOI:

* What to display when there is more than one DOI?
* What to display when there is a DOI but no publication note?
* If there is no publication Info, does the e-Print stay on the right or we move it to the left side?
* If there is one DOI and one Publication information, shall we make the publication information a link to the DOI source?

kaplun:
I am not sure to write `DOI` without anything afterwards is of any use to the user. The DOI is a label do say that what follows *is* a DOI. The word per-se is not something that the user would search in order to reach the final publication. She would rather click on something called *final publication* in that case.

So if we want to hide the `10.xxx/xxx` part, then IMHO better not put *DOI* at all, and simply put the links behind something else.

For the pubnote Vs. DOI association: this is an issue for the old MARCXML schema, but we can solve it for the future JSON. And, as Salvatore says, have an heuristic to try to guess the correct association.

salmele:
I believe I was not clear in my suggestion. I suggested *not* show the word DOI, as you do in the screenshot now. I suggested to hyperlink the Phys.Rev.Foo.Bar string. Open issues

1) If there is more than a DOI we can chose one according to whichever heuristics in the brief format
2) We show all the DOIs in the full record. We also show the extended 10.xxx/.... form there

nikpap:
> What to display when there is a DOI but no publication note?

That's indeed a good question. I think that's not common, but I've been able to find at least one record where this occurs: https://qa.inspirehep.net/record/1395503

I will amend this PR to implement the rest of the changes as suggested here.

kaplun:
I think indeed, in case of ambiguity (or when there is pubnote but not a DOI) we can indeed just display explcitly the DOI and avoid any confusion. I don't think this is going to harm the user.

jmartinm:
> 1) If there is more than a DOI we can chose one according to whichever heuristics in the brief format

Speaking to @annetteholtkamp it seems like such an heuristic would not be trivial in cases with multiple DOI/Pubnotes in order to match a DOI with a correct publication note (that will have a hyperlink).

As suggested by @kaplun , for non-ambiguous cases with one publication note and one DOI we could directly put the link in the publication note and for the ambiguous ones we can display the explicit DOI

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -25,92 +25,93 @@
...
+              {% endif %}
+            {% endif %}
+            <span class="dropdown">
+              <button class="btn btn-default dropdown-toggle dropdown-cite" type="button" id="dropdownMenu{{record['control_number']}}" data-recid="{{record['control_number']}}"  data-toggle="modal" data-target="#citeModal{{record['control_number']}}">
+                <i class="fa fa-quote-right"></i> Cite
+              </button>
+            </span>
+            {{ record_cite_modal() }}
+            <div class="citations-references">
+              {% if record.get('date_updated') %}
```
jmartinm: Here we should use the `earliest_date` field which is present in all records.

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -25,92 +25,93 @@
...
+                <i class="glyphicon glyphicon-calendar"></i> {{ record.get('date_updated').split('-')[0] }}<br/>
+              {% endif %}
+              {% if  record.get('_cited_by_count') > 0  %}
```
jmartinm: The field will be called `citation_count`, you can already updated.

annetteholtkamp:
Currently 4600 records:
https://inspirehep.net/search?wl=0&ln=en&p=0247%3Adoi+-773__c%3A**&of=hb&action_search=Search&sf=earliestdate&so=a&rm=&rg=250&sc=0

- Annette

salmele:
>>> What to display when there is a DOI but no publication note?

>>That's indeed a good question. I think that's not common, but I've been able to find at least one record where this occurs: https://qa.inspirehep.net/record/1395503

> Currently 4600 records:
https://inspirehep.net/search?wl=0&ln=en&p=0247%3Adoi+-773__c%3A**&of=hb&action_search=Search&sf=earliestdate&so=a&rm=&rg=250&sc=0

There seem to be Journals Names and Page Numbers there, so I am confused...

jmartinm:
I see that in some cases the ePrint information is indented to the right while in most of them it is indented to the left. Can you have a look?

screen shot 2015-12-08 at 2 26 11 pm

inspirehep/base/templates/format/record/Inspire_Default_HTML_general_macros.tpl
```diff
@@ -69,13 +69,21 @@
         {% endif %}
       </small>
     {% else %}
-      <small>
         {% if not is_brief and show_affiliations %}
-          <a id="authors-show-more" class="text-muted" data-toggle="modal" href="" data-target="#authors_{{ record['control_number'] }}">
-            Show affiliations
-          </a>
+          {% set affilitions_exist = [] %}
```
jmartinm: can you correct the typo? :)

inspirehep/base/static/less/base/variables.less
```diff
@@ -33,6 +33,11 @@
 @institutions-collection: #79B873;
 @journals-collection: #98A3A4;
 
+// Buttons colors
```
jmartinm: Can you add an indication on how to use them? what color is used for what part of the button?

inspirehep/base/static/less/format/detailed-record.less
```diff
@@ -496,7 +490,7 @@ body {
 #record_content .record-detailed .record-details {
     padding-top: 30px;
     text-align: left;
-    background-color: @teal;
+    background-color: #f5f5f5;
```
jmartinm: should this be a variable?

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -19,98 +19,112 @@
...
-          {{ render_record_authors(is_brief=true, show_affiliations=false) }}
-        </div>
-      {% endif %}
-      {% if record.get('publication_info') or record.get('dois')%}
+      <div class="panel-body" >
         <div class="row">
-          {% if record.get('publication_info') %}
-            <div class="col-md-6 ">
-              {{ record_journal_info() }}
+          <div class="col-md-9" id="left-column">
```
jmartinm: This is a problem as many elements in the page have the same id `left-column`. If we need an id then you can add dynamically a suffix with the record id. Otherwise a class can be used?

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -19,98 +19,112 @@
...
+            <div class="row">
+              <div class="col-md-1">
+                <span id='checkbox-parent'>
+                  <input type="checkbox" class="checkbox-results" id="{{ record['control_number'] }}">
```
jmartinm: the `id` here being only a number is not very nice and error prone. Would be good if it something like `checkbox-results-12345` where `12345` is the record id.

jmartinm: That would imply doing some changes in the JavaScript selectors but it is better to do early on.

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -19,98 +19,112 @@
...
+                </span>
+              </div>
+              <div class="col-md-11">
+                <h4 class="custom-h">
+                  <b>
+                    <a class="title" href="{{ url_for('record.metadata', recid=record['control_number']) }}">
+                      {{ record['titles[0].title']|capitalize }}
```
jmartinm: Here there is capitalization everywhere, that should be fixed here https://github.com/inspirehep/inspire-next/pull/601 Once we merge this PR the other one needs to be rebased and make sure this is changed.

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -19,98 +19,112 @@
...
+                    {{ record_abstract(is_brief=true) }}
+                {% endif %}
+              </div>
             </div>
-          {% endif %}
-          {% if record.get('dois') %}
-            <div class="col-md-6 ">
-              DOI: {{ render_doi() }}
+          </div>
+          <div class="col-md-3" id="right-column" >
```
jmartinm: Same as before, all elements in the page will have the same id `right-column`

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -19,98 +19,112 @@
...
+            {% if record.get('arxiv_eprints') %}
+              {% if record.get('arxiv_eprints') | is_list() %}
+                {% set filtered_arxiv = record.get('arxiv_eprints') %}
```
jmartinm: why is it filtered?

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief_macros.tpl
```diff
@@ -32,12 +36,57 @@
...
+          Published in
+        {% endif %}
+        {% if not loop.first %}
+          and
+        {% endif %}
+        <i>{{ pub_info.get('journal_title') }}</i> {{ pub_info.get('journal_volume') }} ({{pub_info.get('year')}}) {{ pub_info.get('page_artid') }}
+      {% endif %}
+    {% endfor %}
+  {% else %}
+    {% if record.get('publication_info').get('journal_title') and record.get('publication_info').get('journal_volume') and  record.get('publication_info').get('year') and record.get('publication_info').get('page_artid') %}
```
jmartinm: Here you are doing some different if `record.get('publication_info')` is a list or not.

In the first case you show any part of the publication info available but in the second only if all the parts are there you show the publication info. This should be consistent in both places.

And the information shown should reflect the same way we do it nowadays in INSPIRE - see https://github.com/inspirehep/inspire/blob/master/bibformat/format_elements/bfe_INSPIRE_publi_info.py#L28

Can you create a separate issue for this? We will merge this PR first and then you address it.

nikpap: Created issue #637 regarding this.

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
nikpap: Created task in issue #589 to address this in a future PR, along with other changes to the exporting functionality.

PR #565, opened 2015-11-26
format: HEP detailed record keywords

nikpap:
- [x] Adds free keywords to the HEP detailed record.
- [x] Styles keywords as labels.
- [x] Differentiates between INSPIRE and author-supplied keywords.
- [x] Fixes alignment.
- [x] Reduces keywords shown to 10.

### DIFF
--- a/inspirehep/base/static/less/format/detailed-record.less
+++ b/inspirehep/base/static/less/format/detailed-record.less
@@ -296,9 +296,6 @@ body {
     .cite-modal {
         width: 900px;
     }
-    #record-keywords {
-        margin-top: 20px;
-    }
 }
 
 .cite-modal {
@@ -332,6 +329,15 @@ body {
     padding: 20px;
     font-size: 15px;
     color: black;
+}
+
+#record-keywords .label-keyword {
+    background-color: #D3D3D3;
+
+    a {
+        color: black;
+        font-weight: 500;
+    }
 }
 
 .record-abstract {
--- a/inspirehep/base/templates/format/record/Inspire_HTML_detailed_macros.tpl
+++ b/inspirehep/base/templates/format/record/Inspire_HTML_detailed_macros.tpl
@@ -123,36 +123,41 @@
     </div>
     {% set sep = joiner("; ") %}
 
-    {% if record.get('thesaurus_terms') %}
+    {% if record.get('thesaurus_terms') or record.get('free_keywords') %}
       {% set showMore = [] %}
       {% set keywords_number = [] %}
-        {% for keywords in record.get('thesaurus_terms') %}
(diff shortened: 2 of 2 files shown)


nikpap:
Here's how free keywords could look in the HEP detailed record. @jmartinm @annetteholtkamp are all "free_keywords" author supplied or are there any other categories as well?

capture d ecran 2015-11-26 a 15 43 38

annetteholtkamp:
I find that less readable than the current display. It’s harder to see which keywords form a pair. 

- Annette

jmartinm:
> I find that less readable than the current display. It’s harder to see which keywords form a pair.

I agree it is not very easy to distinguish them. We could use something like `chips` - http://materializecss.com/chips.html to separate them clearly.

By the way @annetteholtkamp , what about the questions asked by Nick regarding the headings of each of the keywords?

annetteholtkamp:
On 30 Nov 2015, at 10:56, Javier Martin Montull <notifications@github.com> wrote:

> I find that less readable than the current display. It’s harder to see which keywords form a pair.
> 
> I agree it is not very easy to distinguish them. We could use something like chips - http://materializecss.com/chips.html to separate them clearly.
> 
> 
Yes, that could be nice.
> By the way @annetteholtkamp , what about the questions asked by Nick regarding the headings of each of the keywords?
> 
> 
Still on my list to ask DESY at standup today. 

- Annette

nikpap:
@annetteholtkamp have there been any updates on this?

annetteholtkamp:
I mentioned to Javier already that we’ll stay with just Inspire and author keywords - no further categories for the time being.

- Annette

nikpap:
An updated version of it looks like, with keywords styled as labels:

capture d ecran 2015-12-02 a 15 10 56

jmartinm:
@nikpap I guess the keywords are clickable?

And what about making them less dark? Something more like the one I mentioned the other day http://materializecss.com/chips.html

nikpap:
Changed color to match the chips you've mentioned.

I think we also need to either reduce their number to <10 or increase the abstract length, because there's currently lots of white space under the abstract.

capture d ecran 2015-12-02 a 16 22 16

jmartinm:
Indeed, the alignment seems off. The titles `Abstract` and `Keywords` should be at the same height and both the abstract text and the keywords should start at the same height.

And as you mention, would be good to make both the same total height. As we are in the detailed record I would show more of the abstract.

nikpap:
Updated: 
capture d ecran 2015-12-03 a 16 31 27

jmartinm:
Merged 9c28183898ac8d7e8fdcb04b8a5ee7c7eba75021

PR #568, opened 2015-11-27
format: HEP search results compacting

nikpap:
* Makes search results more compact.

* Truncates title up to one line.

### DIFF
--- a/inspirehep/base/static/less/base/header.less
+++ b/inspirehep/base/static/less/base/header.less
@@ -137,8 +137,8 @@ body {
 
 .collection-header {
     font-weight: 700;
-    height: 200px;
-    padding: 80px 20px 0 20px;
+    height: 180px;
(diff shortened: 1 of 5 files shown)


nikpap:
This is a before and after screenshot. Note that in the first picture, the first search result is only partially shown. In the second picture, the text up to authors of the second result is visible.

capture d ecran 2015-11-27 a 13 22 22

capture d ecran 2015-11-27 a 15 11 16

jmartinm:
I like the better use of space on the top part of the page.

The titles are one of the most representative parts so it is better to leave them in full without cutting them. 

Also to have more space when scrolling down I would remove the fixed navbar for the Collections, we can see later on if we can make a compact header when scrolling down that includes the important parts.

jmartinm:
Merged 5545b9072ca5f29444b5028fc1710bc7efd50991

PR #569, opened 2015-11-27
global: remove search API recids usage

jmartinm:
* NOTE: Invenio search API recids property should not be used with
  ElasticSearch 2.0 as size cannot be bigger than 10000.


jacquerie:
Related: https://github.com/elastic/elasticsearch/pull/13188

kaplun:
:shipit: BTW, what about `invenio-search`? Is it using `.recids` in the UI part somewhere?

PR #570, opened 2015-11-27
migrator: log any exception during migration

jalavik:
During `migrate_chunk` any exception is hidden. This patch at least shows any problems to the log so it is visible.

@kaplun

### DIFF
--- a/inspire/modules/migrator/tasks.py
+++ b/inspire/modules/migrator/tasks.py
@@ -133,7 +133,7 @@ def migrate_chunk(chunk, broken_output=None, dry_run=False):
         logger.info("Committing chunk")
         db.session.commit()
         logger.info("Sending chunk to elasticsearch")
-        es_bulk(es, records_to_index)
+        es_bulk(es, records_to_index, timeout=60)
     finally:
         models_committed.connect(record_modification)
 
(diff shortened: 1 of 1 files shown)


inspire/modules/migrator/tasks.py
```diff
@@ -124,10 +125,12 @@ def migrate_chunk(chunk, broken_output=None, dry_run=False):
                 before_record_index.send(recid, json=json)
                 json.update({'_index': index, '_type': 'record', '_id': recid})
                 records_to_index.append(json)
-            except Exception:
+            except Exception as err:
+                current_app.logger.exception(err)
```
kaplun: Down below in `create_record` I was using: 
```python
logger.exception("Error in elaborating record ID {}".format(recid))
```

this should use the native Celery logging. What would be best (it passes automatically the exception)

jalavik: Indeed that covers `create_record` but not if some exception happens on lines `124-127`. The problem we saw happened there.

jalavik: Ah, I see now what you meant :dancer:  SO I guess we can just the logger then to not mix things.

kaplun:
I am including this into my PR

PR #593, opened 2015-12-01
static: resources dropdown-menu fixes

jalavik:
Updates the selectors used in the Resources dropdown menu so that the code does not affect other dropdown menus across the entire site. E.g. this was causing dropdowns in Holding Pen to malfunction.

PS: Be careful when using very generic selectors of bootstrap components that may be used elsewhere. Better always use id's to select with.

@ioannistsanaktsidis @nikpap


ioannistsanaktsidis:
Nice @jalavik. Cheers! :dancer:

Issue #465, opened 2015-11-05
do not allow anonymous feedback


jmartinm:
At the moment the feedback form

<img width="627" alt="screen shot 2015-11-05 at 2 59 38 pm" src="https://cloud.githubusercontent.com/assets/580727/10970257/c7cc3382-83cd-11e5-80a4-b6182265c8ea.png">

Uses the email address from the user if he/she is logged in.

If the user is not logged in, a ticket is created in the RT system but no email appears.

Solution
======

If the user is not logged in and the feedback form is presented there should be an extra email field mandatory for feedback to be submitted.

salmele:
It is very valuable to keep the threshold low for users to help with feedback, at least while we are in labs+prod phase. I agree that forcing (ORCID) login for this would be too much. I agree with the solution of asking extra e-mail field.

Concerning the verbiage, we should be careful about what we are asking for: it is ONLY for labs, and/or for bizarre things we see in that page. Not citations missing/other suggestions which should go another way.

jacquerie:
I can take this. Proposal for copy: "INSPIRE Labs welcomes problem reports, feature ideas and general comments. For missing citations in the current INSPIRE please refer to [this page](https://inspirehep.net/info/hep/corrections_reference)".

annetteholtkamp:
Add “and other corrections/suggestions” 

- Annette

PR #599, opened 2015-12-02
feedback: disallow anonymous feedback

jacquerie:
As requested in #465, disallow anonymous feedback. Instead of implementing the solution proposed by @jmartinm I decided for another flow:

If the user is logged in, she can submit her feedback exactly like today. If the user is not logged in, when she tries to submit her feedback, a new field is added to the popup, requesting the email.

### DIFF
--- a/inspirehep/base/static/js/feedback/feedback.js
+++ b/inspirehep/base/static/js/feedback/feedback.js
@@ -1,6 +1,6 @@
  /*
   * This file is part of INSPIRE.
-  * Copyright (C) 2015 CERN.
+  * Copyright (C) 2015, 2016 CERN.
   *
   * INSPIRE is free software: you can redistribute it and/or modify
   * it under the terms of the GNU General Public License as published by
@@ -23,9 +23,10 @@
  define([
    'jquery',
    'hgn!js/feedback/templates/button',
+   'hgn!js/feedback/templates/email',
    'hgn!js/feedback/templates/error',
    'hgn!js/feedback/templates/modal',
- ], function($, buttonTemplate, errorTemplate, modalTemplate) {
+ ], function($, buttonTemplate, emailTemplate, errorTemplate, modalTemplate) {
    return {
      'Feedback': function(opts) {
        var options = {};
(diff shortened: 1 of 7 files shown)


jacquerie:
BTW, I'm not really satisfied with how the modal looks. Maybe people more versed with our CSS can make it look good!

jmartinm:
@jacquerie Can you add a screenshot of the modal?

jacquerie:
The modal with just the feedback request:
screen shot 2016-02-03 at 10 14 41

The modal with the email request as well:
screen shot 2016-02-03 at 10 14 55

jmartinm:
One thing that could improve it is to have text with less font-weight so that the title of the modal and the content inside have a been more contrast (there is too much bold at the moment)

The send button could also have an icon representing the 'Send' action close to the text.

jmartinm:
And maybe a bit more margin-bottom between the first input (for the email) and the second one.

jacquerie:
New version pushed should address all comments.

The modal with just the feedback request:
screen shot 2016-02-03 at 12 51 50

The modal with the email request as well:
screen shot 2016-02-03 at 12 51 58

annetteholtkamp:
Why is there only a link for missing citations? Shouldn’t we link to the paper submission form as well?

- Annette

jacquerie:
> Why is there only a link for missing citations? Shouldn’t we link to the paper submission form as well?

Mh, that label is already pretty long. Maybe I should move this information in a `class="notice"` above? BTW, I noticed I forgot to include [your suggestion](https://github.com/inspirehep/inspire-next/issues/465#issuecomment-154422529). Do you have a proposal for the full copy of that sentence, anyway?

jmartinm:
@jacquerie you can leave the main text as now and add the clarification at the bottom:

screen shot 2016-02-03 at 1 52 45 pm

This is the HTML snippet for the bottom text:

```html
<p class="text-muted" style="
    font-size: 12px;
    margin-top: 4px;
">For missing citations and other corrections/suggestions in the current INSPIRE please refer to <a href="#">this page</a></p>
```

jacquerie:
Nifty! New screenshots:
screen shot 2016-02-03 at 14 13 00
screen shot 2016-02-03 at 14 13 07

jmartinm:
@jacquerie :+1: for me.

inspirehep/base/static/js/feedback/feedback.js
```diff
@@ -33,11 +34,20 @@
...
+         options['emailLabel'] = opts['emailLabel'] ||
+           'Please enter your email:';
+         options['feedbackError'] = opts['feedbackError'] ||
            '<strong>Oh snap!</strong> It seems like you forgot to input your feedback?';
-         options['label'] = opts['label'] ||
-           'We welcome problem reports, feature ideas and general comments:';
+         options['feedbackLabel'] = opts['feedbackLabel'] ||
+           'INSPIRE Labs welcomes problem reports, feature ideas and general \
+           comments. For missing citations in the current INSPIRE please refer \
+           to <a href="https://inspirehep.net/info/hep/corrections_reference"> \
```
jmartinm: Can you add `target=_blank`? that way we avoid people leaving Labs :)

jacquerie: :+1: 

Good catch!

jacquerie:
Let's wait for a copy review by @annetteholtkamp or @salmele and :shipit:?

jacquerie:
Rebased. Minor consistency issue: I made the header blue, while currently other popups have no style. Shall I revert this, or extend my modification to the others?

jalavik:
@jacquerie I like the blue header style. Should perhaps be the default styling of modals? E.g. not doing it specifically here but applied to general theme? @jmartinm @eamonnmag

jmartinm:
The blue looks fine to me.

You could separate into two different commits, one where you add your functionality and the other one changing the general style of modals.

jmartinm:
By the way, talking to @annetteholtkamp, as the current link where we point to production does not have all information regarding suggestions/corrections might be better to remove the link and just clarify that the feedback is just for INSPIRE Labs (without mentioning production)

`Please, do not use this form for inspirehep.net comments/corrections.`

jacquerie:
> You could separate into two different commits, one where you add your functionality and the other one changing the general style of modals.

Ok, will split it into another PR.

> `Please, do not use this form for inspirehep.net comments/corrections.`

:+1:

I also fixed the tiny spacing gap we discussed IRL. :mouse:

PR #603, opened 2015-12-02
ext: redis client

tomaszgy:
* added extension for redis client

Signed-off-by: Tomasz <tomaszgy@gmail.com>

### DIFF
--- a/inspirehep/ext/redis/__init__.py
+++ b/inspirehep/ext/redis/__init__.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015 CERN.
+#
+# Invenio is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+from .proxy import redis
+
+__all__ = (
+    'redis'
+)
--- a/inspirehep/ext/redis/client.py
+++ b/inspirehep/ext/redis/client.py
@@ -1,0 +1,66 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2015 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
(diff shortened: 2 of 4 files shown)


jmartinm:
This looks :+1: to me. @kaplun @jalavik wdyt?

inspirehep/ext/redis/__init__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
nikpap: I believe this can be removed nowadays.

jalavik: Unless something has changed, it might still be fine actually. We have it everywhere nowadays.

jalavik: Ah you mean the whole thing, yeah, right that can be removed. I thought you meant the encoding :dancer:

jalavik:
Yeah :shipit:

inspirehep/ext/redis/__init__.py
tomaszgy: So the __init__.py file is unnecessary?

kaplun:
What is exactly this? Is it a way to be able to just use redis? Shouldn't this be on invenio_ext instead?

Why `__init__.py` is optional?

inspirehep/ext/redis/__init__.py
jalavik: Sorry, it was not clear with our comments. We were discussing the license header. The file can just be empty.

inspirehep/ext/redis/client.py
```diff
@@ -0,0 +1,34 @@
...
+# or submit itself to any jurisdiction.
+
+
+redis = None
+
+
+def setup_app(app):
+
+    from werkzeug.contrib.cache import RedisCache
+    global redis
```
kaplun: Is this correct the correct pattern? Shouldn't this based context locals? http://werkzeug.pocoo.org/docs/0.11/local/

jalavik:
@kaplun To connect to a Redis instance without the imposed timeout given by `invenio_ext.cache`. Who knows what will happen with `invenio_ext` in Invenio 3, so we figured to add it here for now and then later use the Invenio 3 way (if it exposes it's own thing).

kaplun:
You mean the fact that on Invenio cache things expires by default? I see.

inspirehep/ext/redis/__init__.py
tomaszgy: OK, I see.

jalavik:
@kaplun yeah, exactly.

inspirehep/ext/redis/client.py
tomaszgy: Well, I am not sure though I followed the pattern from https://github.com/inspirehep/invenio-ext/blob/master/invenio_ext/es.py#L81

kaplun: OK, just had the proof from https://gitter.im/inveniosoftware/invenio that that pattern was not correct. (it has been removed in 3).

I think we should really use context locals, here (and in invenio-ext as well). `globals` are evil (e.g. if use multithreading).

jmartinm:
@tomaszgy Here you have another example in the new `invenio-search` of how to use `LocalProxy`

https://github.com/inveniosoftware/invenio-search/blob/master/invenio_search/proxy.py#L27-L37

You can use it as an example to avoid the `global` use.

jmartinm:
I like the fact that you can import Redis with:

```python
from inspirehep.ext.redis import redis
```

I am wondering if it is too much boilerplate, for example, registering another extension https://github.com/inspirehep/inspire-next/pull/603/files#diff-56eeba39c106acfdd01c4ce0609975b6R39

Do you think it can be simplified just with `setup_app` and having the `LocalProxy` ?

tomaszgy:
Well I agree it's a bit boilerplate although you can see it in many invenio_exts. I guess it is kind of a pattern. I tried to find a simpler way of doing that but I haven't really found anything that'd look better.

jmartinm:
Alright, if there are no more comments I think we can :shipit:

jalavik:
:+1:

PR #610, opened 2015-12-03
citations: citation_count fix

tomaszgy:
* add citation_count to all records, not only those that have references

Signed-off-by: Tomasz <tomaszgy@gmail.com>

### DIFF
--- a/inspirehep/modules/migrator/tasks.py
+++ b/inspirehep/modules/migrator/tasks.py
@@ -29,8 +29,9 @@ import re
 
 from celery.utils.log import get_task_logger
 
-from flask import current_app
 from dojson.contrib.marc21.utils import create_record as marc_create_record
+
+from flask import current_app
 
 from inspirehep.dojson.conferences import conferences
 from inspirehep.dojson.experiments import experiments
@@ -155,7 +156,7 @@ def migrate_chunk(chunk, broken_output=None, dry_run=False):
                 index = get_record_index(json) or \
                     cfg['SEARCH_ELASTIC_DEFAULT_INDEX']
                 before_record_index.send(recid, json=json)
-                json.update({'_index': index, '_type': 'record', '_id': recid})
+                json.update({'_index': index, '_type': 'record', '_id': recid, 'citation_count': 0})
                 records_to_index.append(json)
             except Exception as err:
                 logger.exception(err)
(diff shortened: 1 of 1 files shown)


inspirehep/modules/migrator/tasks.py
```diff
@@ -187,11 +187,6 @@ def add_citation_counts():
             es,
             query={
                 "_source": "references.recid",
-                "filter": {
-                    "exists": {
-                        "field": "references.recid"
-                    }
-                },
```
kaplun: I don't think this is necessary to solve the bug. Actually I realize the best way would really be to exploit the `citations_lookup`.

You should rather create the `records_to_update` list via a generator independent from the input.:
```python
def get_records_to_update_generator(citation_lookup):
    for recid, citation_count in citation_lookup.iteritems():
        yield {'_op_type': 'update',
                     '_index': 'hep',
                     '_type': 'record',
                     '_id': recid,
                     'doc': {'citation_count': citation_count}
                 }

for chunk in chunker(get_records_to_update_generator(citations_lookup), CHUNK_SIZE):
    es_bulk(es, chunk)
```

With the current implementation you were risking of filling up the memory with a `records_to_update` list with 1.5M records :smile: .

tomaszgy:
Thanks for tips. Added a generator, please check.

tomaszgy:
Also seems to me like removal of `filter` is necessary. We need recids of all records, not only those that contain references.

kaplun:
You have the recids of all records that you need to re-index: these recids are those in the `references.recid` :wink:

inspirehep/modules/migrator/tasks.py
```diff
@@ -187,17 +194,15 @@ def add_citation_counts():
...
                 "size": CHUNK_SIZE
             },
             scroll=u'1m',
             index="hep",
             doc_type="record"):
 
+        # initialize the key for record in citations_lookup dict in case it doesn't exist there yet
+        citations_lookup[record['_id']] = citations_lookup.get(record['_id'], 0)
```
kaplun: Here's the misunderstanding. it should not be `citations_lookup[record['_id']]`, rather:
```python 
for recid in record['references.recid']:
    citations_lookup[recid]
```

tomaszgy:
Voila!

PR #616, opened 2015-12-04
base: typeahead searchbar fix

jmartinm:


jmartinm:
This adds back the typeahead autocompletion in the search bar. It supports Spires and Invenio syntax. Would like to try out with some users to see if it is helpful.

typeahead

PR #618, opened 2015-12-07
WIP: search: creates easy search

### LINKED ISSUE #582: Implement Easy Search
An equivalent to the current Easy Search:

screen shot 2015-11-30 at 1 45 35 pm

can be implemented as part of the search bar. An Easy Search link will make the fields appear under the search bar and the user can fill only the required fields.

Extra 1
=====
As every collection has different requirements of what fields should appear in the Easy Search form it would be great if this could be easily configured in a central place.

Extra 2
=====
Each of the fields can have the appropriate autocompletion.

ioannistsanaktsidis:
* Closes #582.

### DIFF
--- a/inspirehep/base/templates/search/easy_search.html
+++ b/inspirehep/base/templates/search/easy_search.html
@@ -1,0 +1,222 @@
+{#
+# This file is part of INSPIRE.
+# Copyright (C) 2015 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#}
+ <div id="easy-search" class="collapse">
+  <form class="form-horizontal">
+  <div class="form-group">
+    <label for="inputAuthor" class="col-sm-2 control-label">Author</label>
+    <div class="col-sm-10">
+      <input style="width: 97%;" class="form-control" id="inputAuthor" placeholder="Author">
+    </div>
+  </div>
+  <div class="form-group">
+    <label for="inputTitle" class="col-sm-2 control-label">Title</label>
+    <div class="col-sm-10">
+      <input style="width: 97%;" class="form-control" id="inputTitle" placeholder="Title">
+    </div>
+  </div>
+  <div class="form-group">
+    <label for="inputReportNumber" class="col-sm-2 control-label">Report Number</label>
+    <div class="col-sm-10">
+      <input style="width: 97%;" class="form-control" id="inputReportNumber" placeholder="Report Number">
+    </div>
+  </div>
+  <div class="form-group">
+    <label for="inputAffiliation" class="col-sm-2 control-label">Affiliation</label>
+    <div class="col-sm-10">
+      <input style="width: 97%;" class="form-control" id="inputAffiliation" placeholder="Affiliation">
+    </div>
+  </div>
+  <div class="form-group">
+    <label for="inputCollaboration" class="col-sm-2 control-label">Collaboration</label>
+    <div class="col-sm-10">
+      <input style="width: 97%;" class="form-control" id="inputCollaboration" placeholder="Collaboration">
+    </div>
+  </div>
+  <div class="form-group">
+    <label for="inputKeywords" class="col-sm-2 control-label">Keywords</label>
+    <div class="col-sm-10">
+      <input  style="width: 97%;" class="form-control" id="inputKeywords" placeholder="Keywords">
+    </div>
+  </div>
+  <div class="form-group">
+    <label for="inputEprint" class="col-sm-2 control-label">Eprint</label>
+    <div class="col-sm-4 dropdown">
+      <button style="width:100%" class="btn btn-default dropdown-toggle" type="button" id="eprintDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
+        Any Type
+        <span class="caret"></span>
+      </button>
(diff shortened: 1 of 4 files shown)


jmartinm:
@ioannistsanaktsidis I see you are adding the changes to `collection_literature.html`. That is good for the landing page but I think you will have to also edit `results.html` so that easy search is available in the search results page.

And also check the JS available in some functions here https://github.com/inspirehep/inspire-next/blob/master/inspirehep/base/static/js/search/form.js#L126 as it might be useful as a starting point to generate the query out of the form.

ioannistsanaktsidis:
Yes you are right I also have to edit results.html. I will create a
template that actually contains only the easy search implementation and
include it in the parts that is needed. As for the form.js I saw the
functionality. Is it an old implementation of easy search? I need to
recreate it I guess right? Also from the fields that the old one has, do
you think we will need them all for the new one ?

2015-12-08 10:09 GMT+01:00 Javier Martin Montull <notifications@github.com>:

inspirehep/base/templates/search/easy_search.html
```diff
@@ -0,0 +1,222 @@
...
+  </div>
+  <div class="form-group">
+    <label for="inputJournal" class="col-sm-2 control-label">Journal</label>
+    <div class="col-sm-4 dropdown">
+    <button style="width:100%;" class="btn btn-default dropdown-toggle" type="button" id="journalDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
+        Any Journal
+        <span class="caret"></span>
+      </button>
+      <ul class="dropdown-menu" id="journalDropdownmenu" aria-labelledby="topciteDropdown">
+        <li><a>Any Journal</a></li>
```
kaplun: How this list was generated?
In principle there are 3475 journal known to INSPIRE...

inspirehep/base/templates/search/easy_search.html
```diff
@@ -0,0 +1,222 @@
...
+  </div>
+  <div class="form-group">
+    <label for="inputTopCite" class="col-sm-2 control-label">Topcite</label>
+    <div class="col-sm-4 dropdown">
+    <button style="width:100%;" class="btn btn-default dropdown-toggle" type="button" id="topciteDropdown" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
+        Don't care
+        <span class="caret"></span>
+      </button>
+      <ul class="dropdown-menu" id="topciteDropdownMenu" aria-labelledby="topciteDropdown">
+        <li><a>Don't care</a></li>
```
kaplun: shall we change this to the more polite: `any`?

kaplun:
For reference, here's the current snapshot of the new Easy search implementation.

sso1

I have several UI/UX questions:
* [ ] Should the "easy search" link be more promiment and maybe be transformed into something similar to a switch? Given it switches the current searchbox into an easy search form.
* [ ] Is the regular search box still used in the easy  search form, or can it disappear? Or better shall it still stay to let the user enter any additional terms?
* [ ] The reportnumber is not a concept that is necessarily well understood by our user. We might want to add examples... Also could be way after the other fields.
* [ ] Affiliation, and Collaboration should be autocompleted.
* [ ] Eprints should be more: `Category`
* [ ] Number should be more: `arXiv eprint number` (with example of new and old pre-print numbers)
* [ ] Not sure about the proposed list of journals. How is this curated? (side note to @inspirehep/inspire-dir, do we need a tag CORE also for journals and derive this list from those records?)
* [ ] `vol` and `pg` should read `volume` and `page(s)`
* [ ] It is not clear what type of dates can the user enter... (also there is asymmetry between `Date` and `Until`

kaplun:
(in view of the Enabling search milestone, I am removing here the milestone, since this functionality is not a blocker for "Enabling search").

annetteholtkamp:
Eprint + nr seems to be based on the old arxiv notation and does not fit the newer papers. In principle we can use the report number field for arxiv papers as well.
Maybe give an example like "arXiv:1501.0001 or hep-ph/9509416 or CERN-PH-EP-2015-001"
I wonder whether anybody would use the easy search to search for an arxiv paper since putting the number into the search box usually gives a unique result.

Why does the field name get repeated in volatile form?

How does one enter several authors?

Maybe we should add a subject field?

The example queries don't seem to be appropriate.

Should we add the doc types?

- Annette




On 09 Dec 2015, at 21:55, Samuele Kaplun <notifications@github.com<mailto:notifications@github.com>> wrote:


For reference, here's the current snapshot of the new Easy search implementation.

[sso1]<https://cloud.githubusercontent.com/assets/119349/11698319/45b90738-9ebe-11e5-8349-abc1b15cf67d.png>

I have several UI/UX questions:

  *    Should the "easy search" link be more promiment and maybe be transformed into something similar to a switch? Given it switches the current searchbox into an easy search form.
  *    Is the regular search box still used in the easy search form, or can it disappear? Or better shall it still stay to let the user enter any additional terms?
  *    The reportnumber is not a concept that is necessarily well understood by our user. We might want to add examples... Also could be way after the other fields.
  *    Affiliation, and Collaboration should be autocompleted.
  *    Eprints should be more: Category
  *    Number should be more: arXiv eprint number (with example of new and old pre-print numbers)
  *    Not sure about the proposed list of journals. How is this curated? (side note to @inspirehep/inspire-dir<https://github.com/orgs/inspirehep/teams/inspire-dir>, do we need a tag CORE also for journals and derive this list from those records?)
  *    vol and pg should read volume and page(s)
  *    It is not clear what type of dates can the user enter... (also there is asymmetry between Date and Until

inspirehep/base/templates/search/easy_search.html
```diff
@@ -0,0 +1,222 @@
...
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#}
+ <div id="easy-search" class="collapse">
+  <form class="form-horizontal">
+  <div class="form-group">
+    <label for="inputAuthor" class="col-sm-2 control-label">Author</label>
```
jalavik: For consistency and avoiding overriding potentially similar selectors, shall we add a prefix to all the classes and ids used in the easy search? Say `easy-search-author`, `easy-search-title` etc.

jmartinm: +1

ioannistsanaktsidis:
I guess I can close this PR as it is outdated for a while. We can reopen it if we decide to implement easy search in labs. cc @jalavik ?

jalavik:
@ioannistsanaktsidis :+1:

Issue #576, opened 2015-11-30
HTML selector collision in Author forms


jmartinm:
An unwanted style is being applied in author forms due to some `id` collision. 

See `http://localhost:4000/author/new`

![screen shot 2015-11-30 at 11 02 03 am](https://cloud.githubusercontent.com/assets/580727/11468691/cc295a94-9751-11e5-931c-cb4fd1286960.png)

PR #622, opened 2015-12-08
Fix/author form fixes

jmartinm:


nikpap:
Does this also fix #576 ?

jmartinm:
> Does this also fix #576 ?

Yes, it does.

PR #629, opened 2015-12-09
LESS: external links styling

### LINKED ISSUE #585: Create a Jinja2 helper for external links
For consistency across the site, would be good if all external links (going away from INSPIRE) were displayed with an external icon, like:

screen shot 2015-11-30 at 1 56 26 pm

Ideally the developer would not have to worry about it and if a link points outside of the `inspirehep.net` domain the icon would be automatically added (with a possibility to disable that).

This could be done on the JavaScript side or on the server (Jinja2). I think the server side would be better as it would avoid elements appearing in the page after page load (once the JavaScript to do this trick runs).

So a good mechanism needs to be found to do this on Jinja2 (filter, macro or other)

nikpap:
- [x] Add a relevant icon to all external links. (closes #585)
- [x] Add a `no-external-link` class to disable the icon in specific cases.
- [x] Apply aforementioned rule in authors, PDF
- [x] Add `https://`in the rule
- [x] Make icon smaller and more inline

### DIFF
--- a/inspirehep/base/static/less/base/core.less
+++ b/inspirehep/base/static/less/base/core.less
@@ -62,3 +62,23 @@ body {
     top: 128px !important;
     right: 30px !important;
 }
+
+/* External links styling
+================================================== */
+
+a[href^="http://"]:after, a[href^="https://"]:after {
+  content: "\f08e";
+  font-size: 80%;
+  font-family: FontAwesome;
+  font-weight: normal;
+  font-style: normal;
+  display: inline-block;
+  text-decoration: none;
+  padding-left: 3px;
+}
+
+a.no-external-icon:after {
+  content: "";
+  padding-left: 0;
+  display: inline;
+}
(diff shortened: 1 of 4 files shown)


nikpap:
Note that most of our links are externals at the moment.

capture d ecran 2015-12-09 a 16 37 26
capture d ecran 2015-12-09 a 16 37 16

nikpap:
@jmartinm Would we consider these external links? The ones on the left are currently pointing to `inspirehep` production website. Would `blog` be considered external?

capture d ecran 2015-12-10 a 11 26 31

kaplun:
IMHO I guess the first two in the future should be considered internal links. The blog is infact under blog.inspirehep.net, hence you get out of the service.

nikpap:
If we can ensure that relative links will be used for in-site linking, then this piece of code will always work correctly. The only links that will be external will therefore be absolute and will display the `external icon` styling.

See also: http://stackoverflow.com/questions/2005079/absolute-vs-relative-urls

jmartinm:
Merged 5f22564761c7fba10fe4ad10c01df1d799327550

PR #632, opened 2015-12-09
travis: forcing new invenio-query-parser

kaplun:
This contains the new code from @Panos512 in inspirehep/invenio-query-parser#1 to work around when querying for things that look like nested keywords.


jalavik:
You might wanna also remove/update the version requirement in `setup.py` for `'invenio-query-parser==0.3.0'`.

kaplun:
Well spot. Thanks

PR #633, opened 2015-12-09
deposit: extra submission data fix

jalavik:
* Fixes the literature submission extra data handling such as
  user comments, references and url to PDF.


jmartinm:
:+1: 

Wonder when `add_submission_extra_data ` got moved. It is correctly placed in production https://github.com/inspirehep/inspire-next/blob/production/inspire/modules/deposit/workflows/literature.py#L108

jalavik:
@jmartinm Hehe, I moved it the other day to fix one issue, but it had some unwanted side-effects, so I found another way. ;-)

PR #634, opened 2015-12-10
dojson: valid schema

Panos512:
* Removes two unintended commas from the json schema to make it valid.


jalavik:
Thanks!

Issue #609, opened 2015-12-03
Searching 101


kaplun:
* On current legacy Invenio a query of the form of `foo bar baz`, was searching for **all and only** the documents having *all* the 3 words anywhere.
* On current legacy Invenio a query of the form of `key:value key2:value2` is matching **all and only** the documents having `value` in `key` and `value2` in `key2`.
*  On Google and default elasticsearch a query for the form `foo bar baz` is matching **all** the documents that have **at least** one of the words (hence an Invenio `foo or bar or baz`), but is ranking them by how much the document matches the query(!), hence documents having all the 3 words will be ranked higher and so on.

We keep on talking about INSPIRE offering Google like search capability, but what we have by default a `strict boolean and`, which is perfect for filtering, but not for information retrieval. 

Proposal
------------
Shall we change our default to be the elasticsearch one, and fall back to the Invenio `default boolean and `as soon as one uses a *keyworded* search? Baisc boolean filtering is then offered anyway by facets.

Usage of quotes
----------------------
We could still use the Invenio convention: something quoted should not be *massaged*. So no stemming, no removal of plurals, no normalization of spaces (or maybe yes?).

jmartinm:
+1 for Elasticsearch way when not using any specific syntax `foo bar baz`

The best practice is to have as many results as possible and ElasticSearch should do a good job in the ranking.

annetteholtkamp:
If we go for this option we should make it transparent to the user whether all words have been found or not. Google shows you the unmatched words by striking through them. 
Or we should add a tick box: Match all terms (use ‘and' instead of ‘or')

- Annette

aw-bib:
> The best practice is to have as many results as possible 

I may buy this for google. I'm pretty sure I do not (at least not in this generality) if I'm searching scientific literature.

Usually, the problem there is that you get to many results that you need to narrow down, and unlike google where basically any document that is _good enough_ serves in the scientific case you're searching _the_ relevant paper. And relevance in itself here gets very difficult. More keys that match (the google case) is usually way to simple.

In your language I usually had the worst of user experiences as soon as bibliographic databases just threw whatever sounds only half plausible by some algorithm at me.

In any case an option to have boolean search is mandatory.

ksachs:
It has to be transparent what this thing is doing. I.e. keep it simple or at least tell the user what was done (and why) so they have a chance of understanding the query. I guess I have to see it before I understand your proposal.

kaplun:
[...]

PR #641, opened 2015-12-10
searchext: add alias to mappings

jalavik:
We'll have to do something like this to get away with no `SEARCH_ELASTIC_KEYWORD_MAPPING`, like after recent Holding Pen updates.

Still WIP as more fields need to be added.


jalavik:
@kaplun @jmartinm Alongside new mappings (e.g. pure arXiv:1111.2222 now works), I also added new `inspirehep` manager for reindexing tasks etc. including click progress bars and colors (yay!).

kaplun:
That is beautiful. I wonder if we can contribute all this upstream. I think all the other services will benefit.

inspirehep/manage.py
```diff
@@ -0,0 +1,188 @@
...
+    """Delete all the indices for records and holdingpen."""
+    from invenio.base.globals import cfg
+
+    indices = set(cfg["SEARCH_ELASTIC_COLLECTION_INDEX_MAPPING"].values())
+    indices.add(cfg['SEARCH_ELASTIC_DEFAULT_INDEX'])
+    holdingpen_indices = [
+        cfg['WORKFLOWS_HOLDING_PEN_ES_PREFIX'] + name for name in list(indices)
+    ]
+    indices.update(holdingpen_indices)
+    click.echo(click.style('Going to delete ALL indicies...', fg='red'), nl=False)
```
jmartinm: Typo :)

kaplun:
BTW. Not to forget about this, but in connection with #609, @Panos512 was having in his pipeline to intercept the `invenio-query-parser` and, if a given query resulted in something not involving any special keyword, to simply pass it over to elasticsearch as a flat query againsta a combination of all the important fields:

```json
{
    "multi_match" : {
        "message" : {
            "query" : "{}",
            "operator" : "or",
            "zero_terms_query": "all",
            "fields": ["title^3", "title.raw^10", "abstract^2", "abstrac.raw^4", "author^10", "author.raw^15", "reportnumber^10", "eprint^10", "doi^10", ...]
        }
    }
}

**Comment by jmartinm** [comment]
:+1: This looks good to me.

**Comment by jalavik** [comment]
@kaplun I see. Anything I have to do in regards to that here? Btw, shall we add those boost values to the configurations as well?

**Comment by kaplun** [comment]
Nope. Not at the moment. I just think we need to make a sprint and really go trhough all the fields, and decide for which we have a need for `raw` subfields, and decide on one global query. I am not sure when it is the right time for that.

There are so many options to play with in building the perfect generic query :-)

So indeed it was a bit OT WRT your PR.

For boosting, Elasticsearch doc recommend to do it only at query time, not index time, so I think we are good for that.

File: inspirehep/base/searchext/mappings/hep.json [hunk 2]
```diff
@@ -57,15 +57,32 @@
                         }
                     }
                 },
+                "abstracts": {
+                    "type": "object",
+                    "properties": {
+                        "value": {
+                            "type": "string",
+                            "copy_to": ["abstract", "global_fulltext"]
+                        }
```
kaplun: I think for `abstract` and `title` we should check how to better tokenize also considering LaTeX formulas. I am not sure yet how, though.

kaplun: Maybe via something like in https://www.elastic.co/guide/en/elasticsearch/reference/2.1/analysis-custom-analyzer.html It's probably worth a dedicated github issue.

inspirehep/base/searchext/mappings/hep.json
```diff
@@ -57,15 +57,32 @@
...
+                "abstract": {
+                    "type": "string"
+                },
                 "accelerator_experiments": {
                     "type": "object",
                     "properties": {
                         "experiment": {
                             "type": "string",
-                            "index": "not_analyzed"
+                            "index": "not_analyzed",
```
kaplun: Shouldn't we split the experiment as well? E.g. would be great to match CERN-LHC-LHCb even when the user has typed `LHCb`.

I think this could be the occasion to have both a `.raw` subfield not analyzed and an analyzed one.

inspirehep/base/searchext/mappings/hep.json
```diff
@@ -78,28 +95,79 @@
                         "properties": {
                             "value": {
                                 "type": "string",
-                                "index": "not_analyzed"
+                                "index": "not_analyzed",
+                                "copy_to": ["doi"]
```
kaplun: We have to make sure, at record creation/modification, that the DOI is normalized e.g. to `10.xxx/xxx`, without any superflous prefix, and that the search term is receiving the same treatment. Alternatively we need, for every DOI to index it also with `doi:` and `http://dx.doi.org/` prefixes. This seems something that could be made easily with a Groovy script.

https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-scripting.html

inspirehep/base/searchext/mappings/hep.json
```diff
@@ -78,28 +95,79 @@
...
+                    "index": "not_analyzed"
+                },
                 "arxiv_eprints": {
                     "type": "object",
                     "properties": {
                         "value": {
                             "type": "string",
-                            "index": "not_analyzed"
+                            "index": "not_analyzed",
+                            "copy_to": ["arXiv"]
```
kaplun: Same thing as for DOIs.

inspirehep/base/searchext/mappings/hep.json
```diff
@@ -78,28 +95,79 @@
...
+                },
+                "arXiv": {
+                    "type": "string"
+                },
+                "isbns": {
+                    "type": "object",
+                    "properties": {
+                        "value": {
+                            "type": "string",
+                            "copy_to": ["isbn"]
```
kaplun: Same thing. Would be great to index both the original string, then with and without dashes.

inspirehep/manage.py
```diff
@@ -0,0 +1,188 @@
...
+    click.echo(
+        click.style(
+            'Going to reindex {0} records...'.format(len(query)), fg='green'
+        )
+    )
+    click.echo()
+
+    with click.progressbar(query) as record_objects:
+        for obj in record_objects:
+            index_record(obj.id, obj.json)
```
kaplun: This is going to be extremely slow because it doesn't use bulk update. However it's going to be surely correct.

jalavik:
Cool. Thanks. :shipit:

kaplun:
but but... I was still reviewing it! :stuck_out_tongue_closed_eyes:

kaplun:
OK I will transform all my comments into github issues...

jalavik:
:+1:

jalavik:
Github did not update the page for me :boom:

PR #666, opened 2015-12-15
dojson: custom do function for hep records

jmartinm:
* Allows for record post-processing in order to modify the record
  after it has been generated by dojson.

### DIFF
--- a/inspirehep/dojson/hep/model.py
+++ b/inspirehep/dojson/hep/model.py
@@ -22,8 +22,17 @@
 
 """HEP model definition."""
 
+import types
 from dojson import Overdo
-from dojson import utils
+
+
+def custom_do(self, blob):
+    """Custom do function that allows extra post-processing."""
+    record = self._do(blob)
+    return record
 
 hep = Overdo()
+hep._do = hep.do
+hep.do = types.MethodType(custom_do, hep)
+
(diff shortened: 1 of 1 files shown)


inspirehep/dojson/hep/model.py
```diff
@@ -22,8 +22,16 @@
...
 
+import types
 from dojson import Overdo
-from dojson import utils
+
+
+def custom_do(self, blob):
+    """Custom do function that allows extra post-processing."""
+    record = self.do(blob)
+    return record
```
kaplun: Ha! We could move my `strip_empty_values()` call here. Cool!

inspirehep/dojson/hep/model.py
```diff
@@ -22,8 +22,16 @@
...
 
 hep = Overdo()
+hep.custom_do = types.MethodType(custom_do, hep)
```
kaplun: I think you should be able to do similar to a decorator:
```python
hep._do = hep.do
hep.do = types.MethodType(custom_do, hep)
```

and in `custom_do()` you use `self._do`.

jalavik: This is better, but what about good ol' inheritance?

PR #668, opened 2015-12-15
templates: implements 0 results page

### LINKED ISSUE #452: Improve 0 results page
This needs to be modified in the template `search.html`

When there are no results in a query, currently the following message is displayed:

```
Your search did not match any records. Please try again.

Try to modify the query.
```

In this case, we can:

1) On the server side we try to modify the query in a smart way to get matches and then communicate to the user the change performed.

2) We offer some search helpful tips and point to the search help.

Obviously 2) is easier so we should start by that.

ioannistsanaktsidis:
* Closes #452.


jmartinm:
Merged

PR #670, opened 2015-12-16
migrator: migrate out of Celery

kaplun:
* No longer schedules the parent migration task in Celery. Instead,
  suspend Celery, so that the chunks are quickly loaded into
  RabbitMQ and then processed as fast as possible.


kaplun:
cc @jalavik @jmartinm 

In this way I hope the process can no longer be truncated.

PR #684, opened 2016-01-06
search: implements cite all functionality

ioannistsanaktsidis:
* Changes download functionality to modal box.

### DIFF
--- a/inspirehep/base/static/js/citation_modal.js
+++ b/inspirehep/base/static/js/citation_modal.js
@@ -65,10 +65,16 @@ define(
         })
       }
 
+      this.onSelectContent = function(ev) {
+        document.execCommand('selectAll', false, null);
+        $(this).off(ev);
+      }
+
       this.after('initialize', function() {
-        $('.dropdown-cite, .bibtex').on('click', this.onCiteClick)
-        $('.latex_eu').on('click', this.onLatexEUClick)
-        $('.latex_us').on('click', this.onLatexUSClick)
+        this.on('.dropdown-cite, .bibtex', 'click', this.onCiteClick);
+        this.on('.latex_eu', 'click', this.onLatexEUClick);
+        this.on('.latex_us', 'click', this.onLatexUSClick);
+        this.on('.editable', 'click', this.onSelectContent);
       });
     }
   });
--- a/inspirehep/base/static/js/detailed_record_init.js
+++ b/inspirehep/base/static/js/detailed_record_init.js
@@ -23,6 +23,9 @@
 require([
   "js/plots",
   "js/citation_modal"
-], function() {
+], function(Plots, CitationModal) {
+	CitationModal.teardownAll();
+    CitationModal.attachTo(document);
+	console.log('js/detailed_record_init is loaded')
   // Needed to have the JavaScript available when ASSETS_DEBUG=False.
 });
--- a/inspirehep/base/static/js/search/search_results.js
+++ b/inspirehep/base/static/js/search/search_results.js
@@ -90,7 +90,7 @@ define(
         var that = this;
         sList = [];
         if ($(ev.target).prop('checked')) {
-          $('#download-format').removeClass('disabled');
+          $('#cite-all-format').removeClass('disabled');
           $('.checkbox-results').each(function() {
             $(this).prop('checked', true);
             sList.push($(this).attr("id"));
@@ -112,7 +112,7 @@ define(
           }
           this.on('#select-all-records', "click", this.onSelectAllRecords);
         } else {
-          $('#download-format').addClass('disabled');
+          $('#cite-all-format').addClass('disabled');
           $('.checkbox-results').each(function() {
             $(this).prop('checked', false);
           });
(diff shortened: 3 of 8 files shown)


ioannistsanaktsidis:
Some minor changes left to check and then we can merge it. I will address them tomorrow morning.

ioannistsanaktsidis:
@jmartinm I think we are ready. ;)

ioannistsanaktsidis:
@jmartinm I am improving some design mistakes that I saw now, I will comment here when evth is ready so as to test it.

jmartinm:
Ok, I was testing it now and also had some comments :) We can check later.

jmartinm:
These were the comments I had (to see if they are the same):

* Cite Article -> Cite Articles

* If Select all 400 results, while the loading icon shows 'Processing' button is not disabled so I can click and it opens the empty modal.

* Height proportional to screen size to have more space. In a big screen the modal ends in the middle of the screen - it could go almost all the way down so that there is more space to show multiple BibTex.

screen shot 2016-01-07 at 12 54 28 pm


* Move down the text (as it is hidden when you scroll) and the Download button.

screen shot 2016-01-07 at 12 50 26 pm

ioannistsanaktsidis:
Most of them are the same. I have add some spinners while loading the modal because in some cases it takes time and confuses the user. I will amend the commit in a bit. ;)

jmartinm:
Remember to amend the small PEP8 issue :) https://travis-ci.org/inspirehep/inspire-next/builds/100794726#L869

ioannistsanaktsidis:
@jmartinm I amended all the comments, you can test and inform  me if we need to change anything else. I think it looks good now. Cheers!

ioannistsanaktsidis:
screen shot 2016-01-07 at 15 04 28

screen shot 2016-01-07 at 15 05 30

jmartinm:
* It would be good to use percentages for the height, so that when there are not many records (like in the screenshot) the modal is not that tall.

screen shot 2016-01-07 at 4 56 18 pm

* I am not sure how the Loading is working. When I select all 429 records the spinner next to the `Cite` button takes a while, and then when I click Cite then there is the loading inside the modal that also takes a while. I guess the computation is being done 2 times?

When clicking on `Select all 429 records` it should be almost immediate I would say and then when you click on Cite then the modal will have the spinner and ideally some text under it that says `Loading citations...`. There it is when the BibTex computation is being done.

* When doing a second export I think it would be better instead of showing the previous values in the background, to just have the modal empty with the loading message in the center. Also the Download button should be disabled.

inspire labs-s iner

jmartinm:
To have the same styling, please modify as well the individual Cite button, so that the modal has the same layout

screen shot 2016-01-07 at 5 16 42 pm

jmartinm:
And actually for consistency, as for an individual item you click `Cite` and then inside select the format we should do the same for multiple items.

So the Select all would look like (being disabled when nothing is selected):

screen shot 2016-01-07 at 5 31 44 pm

And once you click on it it would have the format selector close to the Download button.

jmartinm:
One last thing that can be annoying is that when you have multiple citations:

screen shot 2016-01-07 at 5 36 49 pm

It is not possible to select a single one (or very difficult). We could on first click do the selection but then disable that functionality in case somebody wants to only get one.

ioannistsanaktsidis:
```
I am not sure how the Loading is working. When I select all 429 records the spinner next to the Cite button takes a while, and then when I click Cite then there is the loading inside the modal that also takes a while. I guess the computation is being done 2 times?
```
The spinner next to the Cite button when you click ' select all 429 records' is added because it takes a while the $.get request to be completed and there is no message during that period. The computation of the format is done only when you click the cite button.

jmartinm:
> The spinner next to the Cite button when you click ' select all 429 records' is added because it takes a while the $.get request to be completed and there is no message during that period. The computation of the format is done only when you click the cite button.
 Show all checks


I see. There must be something not very efficient going on that should be profiled as just doing a search like `http://localhost:4000/search?of=id&rg=500` should not take that long.

ioannistsanaktsidis:
I am making now the changes that you mentioned, after I finish them, I can also make some profiling to check what causes the problem.

ioannistsanaktsidis:
@jmartinm what do you think of the spinner and the text like this?
screen shot 2016-01-08 at 15 16 26

jmartinm:
> @jmartinm what do you think of the spinner and the text like this?

:+1: Looks good.

ioannistsanaktsidis:
I have made all the changes we discussed. I would like to test some more things and then I will comment here when is completely ready.

ioannistsanaktsidis:
screen shot 2016-01-08 at 17 16 38
screen shot 2016-01-08 at 17 16 02

@jmartinm I think looks good. What do you say?

ioannistsanaktsidis:
@jmartinm  I will let you know here when you can test it cause I am fixing some bugs that come up with  the changes. I will comment here when it is complete.

ioannistsanaktsidis:
@jmartinm we are ready with  this. For this one https://github.com/inspirehep/inspire-next/issues/688 I am going to handle it in a different PR, as for now I need to check the elastic configuration task. Generally evth we discussed it is implemented. You can also check it and inform if sth goes wrong that I didn't see. Cheers.

jmartinm:
2 things I have seen:

1) When downloading a single bibtex, the name of the file is `Bibtex1256.bibtex` - should be `Bibtex1256.bib` as that is the extension understood by LaTeX editors.

2) In detailed records, the `Cite` button is broken:

screen shot 2016-01-12 at 10 51 35 am

Both the brief and the detailed button should be using the same JavaScript so that when we alter one the other one also gets changed to avoid this kind of problem.

The main functionality works very nicely.

ioannistsanaktsidis:
Cool, I will take care and push now. For the detailed view, I didn't touch the functionality at all, I forgot that we cite from there too. I will fix the problem and amend the commit. Cheers :dancer:

ioannistsanaktsidis:
@jmartinm  updated!

PR #690, opened 2016-01-07
configuration: config additions

Panos512:
* Adds the paths of Inspire installation and json schema file
  to inspire's configuration file.

* Adds some keywords to elastic mapping.

* Changes the way parsers are referenced in the config file.

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of INSPIRE.
-# Copyright (C) 2014, 2015 CERN.
+# Copyright (C) 2014, 2015, 2016 CERN.
 #
 # INSPIRE is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -33,8 +33,8 @@ Happy hacking!
 """
 
 from invenio_query_parser.contrib.spires.config import SPIRES_KEYWORDS
-from invenio_query_parser.contrib.spires.walkers.pypeg_to_ast import PypegConverter
-from invenio_query_parser.contrib.spires.walkers.spires_to_invenio import SpiresToInvenio
+
+import pkg_resources
 
 
 EXTENSIONS = [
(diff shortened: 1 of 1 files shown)


inspirehep/modules/migrator/tasks.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of INSPIRE.
-# Copyright (C) 2015 CERN.
+# Copyright (C) 2014, 2015, 2016 CERN.
```
kaplun: You don't need to put this change into your commit since you don't touch this file.

inspirehep/modules/migrator/manage.py
```diff
@@ -1,24 +1,21 @@
...
+# General Public License for more details.
 #
 # You should have received a copy of the GNU General Public License
-# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
-#
-# In applying this license, CERN does not waive the privileges and immunities
-# granted to it by virtue of its status as an Intergovernmental Organization
-# or submit itself to any jurisdiction.
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
```
kaplun: Same as below.

inspirehep/config.py
```diff
@@ -443,6 +455,10 @@
 
 SEARCH_ELASTIC_DEFAULT_INDEX = 'hep'
 
+INSPIRE_PATH = pkg_resources.resource_filename("inspirehep", "")
```
kaplun: Uh. I thought you wanted to use `INSPIRE_PATH` afterwards into `JSON_SCHEMA_PATHS`. In which other place do you plan to use it?

Panos512: No I just thought that maybe someone will need this one day so I left it as extra information. You think we should remove it?

kaplun: I see. OK. I guess it doesn't harm :-)

inspirehep/config.py
```diff
@@ -299,6 +308,7 @@
 # invenio keywords to elasticsearch fields
 SEARCH_ELASTIC_KEYWORD_MAPPING = {
     "author": ["authors.full_name"],
+    "a": ["authors.full_name"],
```
jmartinm: I don't think we need these mappings? Those keywords are only used in Spires syntax (like `find a Ellis`) so they will first go through [this config file](https://github.com/inspirehep/invenio-query-parser/blob/master/invenio_query_parser/contrib/spires/config.py#L79) and be converted from `a` to `author`.

So `author` is the keyword that needs configuration.

Same is true for the other two.

Panos512: ahh! I thought someone could search for `a: Ellis` .

PR #699, opened 2016-01-11
format: Search and collections bar styling

nikpap:
* New top bar styling.


inspirehep/base/templates/header.html
```diff
@@ -43,245 +42,172 @@
...
+      </a>
+      {% if request.endpoint in ['search.search', 'record.metadata'] %}
+        <div id="top-search-bar-small" class="col-sm-5 {{collection.name}}">
+          {% include "search/form/form.html" %}
+        </div>
+      {% endif %}
+    </div>
+    {% if request.endpoint in ['search.search', 'record.metadata'] %}
+      <div id="top-search-bar-full" class="col-md-4 col-lg-5 {{collection.name}}">
+        {% include "search/form/form.html" %}
```
jmartinm: Why is the search bar included twice?

inspirehep/base/static/less/base/header.less
```diff
@@ -629,3 +596,173 @@ a.external-link::after {
...
+        }
+
+        &.data {
+            .collection-background-color(@data-collection);
+        }
+    }
+}
+
+// TO-DO: fix these..
+@media (min-width: 767px) {
```
jmartinm: would be good to use the standard break points that we use overall (with their variable name)

inspirehep/base/templates/search/searchbar_frame_base.html
```diff
@@ -0,0 +1,34 @@
...
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#}
+{% extends "page.html" %}
```
jmartinm: When is this template used?

inspirehep/base/templates/search/searchbar_frame_base.html
```diff
@@ -0,0 +1,34 @@
+{#
```
jmartinm: When is this template used?

inspirehep/base/templates/header.html
nikpap: The `top-search-bar-small` is included in the `navbar-header` div, whereas the `top-search-bar-full` is outside it, I found this necessary to accommodate for the mobile view. Open to suggestions for improvement though!

inspirehep/base/templates/search/searchbar_frame_base.html
nikpap: I'm overwriting https://github.com/inspirehep/invenio-search/blob/master/invenio_search/templates/search/searchbar_frame_base.html#L29 and removing that line to no longer show the search bar above the search results

jmartinm: I think it should work if the file is named `searchbar_frame.html` and then inside you extend `searchbar_frame_base.html` and just have as content:

```html
{% block body %}
   {% block inner_content %}{% endblock %}
{% endblock %}
```

Can you try?

PR #703, opened 2016-01-12
authors: tokenizer additions

Panos512:
* Name tokenizer now includes the last name in the name variations.

* Name tokenizer now produces only lowercased variations.

* Addresses inspirehep/invenio-search#14

### DIFF
--- a/inspirehep/modules/authors/utils.py
+++ b/inspirehep/modules/authors/utils.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of INSPIRE.
-# Copyright (C) 2015 CERN.
+# Copyright (C) 2015, 2016 CERN.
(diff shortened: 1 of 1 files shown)


inspirehep/modules/authors/utils.py
```diff
@@ -313,5 +312,6 @@ def __check_parts(parts):
 
 def author_tokenize(phrase):
     """Return all possible variatons of a name"""
-    return parse_scanned_author_for_phrases(
+    name_variations = parse_scanned_author_for_phrases(
         scan_author_string_for_phrases(phrase))
+    return [x.lower() for x in name_variations]
```
kaplun: Mmh. I thought we decided we let ES to lower everything, directly in the mapping configuration. Or I am wrong? @jmartinm

PR #705, opened 2016-01-12
dojson: fixes document type problem

### LINKED ISSUE #694: Document type multiple preprint
Records get multiple times the preprint document type. See for example https://qa.inspirehep.net/record/146539/export/recjson (field `facet_inspire_doc_type`)

This is due to the following code https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/receivers.py#L66 where for each collection as a fallback the type `preprint` is being added.

ioannistsanaktsidis:
* Closes #694.

### DIFF
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -42,17 +42,17 @@ def populate_inspire_document_type(recid, json):
     if 'collections' in json:
         for element in json.get('collections', []):
             if 'primary' in element and element.get('primary', ''):
-                if element['primary'].lower() == 'thesis':
+                if element['primary'].lower() == 'published':
                     inspire_doc_type.append(element['primary'].lower())
                     break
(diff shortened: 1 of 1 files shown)


inspirehep/modules/records/receivers.py
```diff
@@ -63,12 +63,14 @@ def populate_inspire_document_type(recid, json):
...
                     break
-                elif json.get('publication_info', []):
-                    for field in json.get('publication_info', []):
-                        if 'page_artid' not in field:
-                            inspire_doc_type.append('preprint')
-                            break
-
+        complete_pub_info = []
+        if not inspire_doc_type:
+            if json.get('publication_info', []):
```
jmartinm: I don't think this line is needed.

ioannistsanaktsidis: Because if publication_info doesn't exist it will return [] due to this json.get('publication_info', []) right?

PR #708, opened 2016-01-13
settings: use proper buckets JS path

jalavik:
Fix for submission form.


jmartinm:
@kaplun is having a look at why the build has the ElasticSearch problem.

jmartinm:
Apart from that, :shipit:

kaplun:
Yep, let's see if #710 fixes it.

kaplun:
@jalavik if you can rebase it should now work.

PR #723, opened 2016-01-14
workflows: safe arXiv file handling

jalavik:
* Fixes an issue where old arXiv IDs generated bad filenames for
  downloaded tarball and PDF.

* Adds top_words statistics to arXiv predictions.


kaplun:
:shipit:

PR #726, opened 2016-01-14
cite: selects automatically bibtex

### LINKED ISSUE #688: Select automatically bibtex when clicking on Cite
To speed up, the use case of getting, e.g the BibTex of a paper, it would be good if when clicking on `Cite` on a brief or detailed view the content would be automatically selected.

screen shot 2016-01-07 at 10 26 42 am

See, for example, how the cite button works in https://www.semanticscholar.org/paper/a30bb4ac51d18af3c4e72b5944742c3ade56eb10

* [ ] Add `Copy to clipboard` button using https://zenorocha.github.io/clipboard.js/

ioannistsanaktsidis:
* Closes #688.

### DIFF
--- a/inspirehep/base/static/js/citation_modal.js
+++ b/inspirehep/base/static/js/citation_modal.js
@@ -24,57 +24,92 @@ define(
   [
     'jquery',
     'flight/lib/component',
-    'bootstrap'
+    'bootstrap',
+    'clipboard'
   ],
-  function($, defineComponent) {
+  function($, defineComponent, Bootstrap, Clipboard) {
     'use strict';
 
     return defineComponent(CitationsModal);
 
     function CitationsModal() {
+      this.onSelectContent = function(record_id) {
+        setTimeout(function(){
+            var id = 'singleRecord' + record_id
+            var range = document.createRange();
+            range.selectNodeContents(document.getElementById(id));
+            var sel = window.getSelection();
+            sel.removeAllRanges();
+            sel.addRange(range); 
+          }, 300);
+      }
 
       this.onCiteClick = function(ev) {
+        var that = this;
+        $('.copy-clp').addClass('disabled');
+        $('.copy-clp' + $(ev.target).data("recid")).tooltip();
         $.getJSON('/formatter/bibtex', {
           recid: $(ev.target).data("recid")
         }, function(data) {
+          $('.copy-clp').removeClass('disabled');
+          $('.copy-clp').tooltip('destroy');
           $("#text" + data.recid).text(data.result);
           $("#format" + data.recid).text('BibTex')
           $("#download" + data.recid).attr("href", "/formatter/download-bibtex/" + data.recid)
+          that.onSelectContent($(ev.target).data("recid"));
         })
       }
 
       this.onLatexEUClick = function(ev) {
+        var that = this;
+        $('.copy-clp').addClass('disabled');
+        $('.copy-clp').tooltip();
         $.getJSON('/formatter/latex', {
           recid: $(ev.target).data("recid"),
           latex_format: 'latex_eu'
         }, function(data) {
+          $('.copy-clp').removeClass('disabled');
+          $('.copy-clp').tooltip('destroy');
           $("#text" + data.recid).text(data.result);
           $("#format" + data.recid).text('LaTex(EU)')
           $("#download" + data.recid).attr("href", "/formatter/download-latex/latex_eu/" + data.recid)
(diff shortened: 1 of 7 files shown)


inspirehep/base/static/js/citation_modal.js
```diff
@@ -40,6 +40,14 @@ define(
           $("#text" + data.recid).text(data.result);
           $("#format" + data.recid).text('BibTex')
           $("#download" + data.recid).attr("href", "/formatter/download-bibtex/" + data.recid)
+          setTimeout(function(){
```
jmartinm: Wasn't it possible to do `$('.editable;).click()` ?

ioannistsanaktsidis: Tried it, wasn't working for me :/ Also tried $('.editable').trigger('click'). No luck!

jmartinm: ok then :+1:

jmartinm:
Just tried it, works nicely :)

Now the text would not be applicable anymore:

`* Click inside the box to select the whole content.`

What can be nice is to add instead of that text a button `Copy to clipboard` making use of https://zenorocha.github.io/clipboard.js/

ioannistsanaktsidis:
Yes we can do that, you can add a bullet to the issue https://github.com/inspirehep/inspire-next/issues/688 and I will amend the commit.

jmartinm:
Done. You can use this icon http://fortawesome.github.io/Font-Awesome/icon/clipboard/ for the button

ioannistsanaktsidis:
:+1:

ioannistsanaktsidis:
@jmartinm  I amended the commit. You can test it  :)

jmartinm:
Nice, for the copy button, I think it would be better to make it clear with a button (instead of only the icon). Like:

screen shot 2016-01-18 at 9 39 22 am

ioannistsanaktsidis:
Cool, I will amend the commit rightaway.

ioannistsanaktsidis:
@jmartinm  we are ready.

jmartinm:
It seems like on hover there is always the message `Copied!` It should be only when pressing the button

screen shot 2016-01-18 at 2 57 02 pm

ioannistsanaktsidis:
@jmartinm I hope now it is fine! ;)

ioannistsanaktsidis:
@jmartinm it works nice for me now after the latest update. You can try it and inform me if anything goes wrong. Cheers!

jmartinm:
You got it in the end! But in the detailed view it does not seem to work?

ioannistsanaktsidis:
Grrrrrrr I forgot the detailed one. I will amend it now ;)

jmartinm:
> Grrrrrrr I forgot the detailed one

Shouldn't the Cite part be a Flight component that works equally in both places so that the changes you make are reflected automatically?, it does not seem to be the case.

ioannistsanaktsidis:
Actually you are right. It should work automatically. Let me check and I will comment here again.

ioannistsanaktsidis:
I found the problem. The search_results.js was not loaded in the detailed view. I will commit the changes now.

ioannistsanaktsidis:
Evth should be ok now. I hope :P

jmartinm:
> I found the problem. The search_results.js was not loaded in the detailed view. I will commit the changes now.

Hmm, I don't think we should add that file to the detailed view as there is quite some unrelated functionality.

The detailed view already includes the file `citation_modal.js`. I think all the functionality related to the citation modal including this functionality of copying should be in that file and removed from `search_results.js`, what do you think?

ioannistsanaktsidis:
You are right, better to keep  these things in the Citation_modal, makes more sense. I will transfer the functionality there ;)

ioannistsanaktsidis:
Updated ;)

inspirehep/base/static/js/search/search_results.js
```diff
@@ -157,7 +157,9 @@ define(
         var format_id = '#' + $(ev.target).attr("id");
         var obj = this.get_download_information($(format_id).text().trim());
         $("#response-data").html('');
-        $("#dropdown-export").html('BibTex <span class="caret"></span>')
+        $("#dropdown-export").html('BibTex <span class="caret"></span>');
+        $("#copy-to-clipboard").addClass('disabled');
```
jmartinm: Are all these things related to the copy clipboard needed on the search results but not on the detailed page?

PR #728, opened 2016-01-15
deposit: use bundlejs from bower

jalavik:
Co-authored-by: Samuele Kaplun <samuele.kaplun@cern.ch>

Tested with `ASSETS_DEBUG=False` and `True`

cc @kaplun @jmartinm


kaplun:
I like your implementation...

PR #743, opened 2016-01-19
utils: fixes citation number in formats

### LINKED ISSUE #737: Fix the citation number latex
Currently in the output formats (latex, cv latex) we are missing the citation number. 
e.g %245 citations counted in INSPIRE as of 21 Aug 2015

ioannistsanaktsidis:
* Closes #737.

### DIFF
--- a/inspirehep/utils/cv_latex.py
+++ b/inspirehep/utils/cv_latex.py
@@ -30,10 +30,7 @@ from .export import MissingRequiredFieldError, Export
 
 class Cv_latex(Export):
 
-    """Class used to output CV LaTex format.
-    TODO Fix the citation number latex
-    e.g %245 citations counted in INSPIRE as of 21 Aug 2015
-    """
+    """Class used to output CV LaTex format."""
 
     def __init__(self, record):
         super(Cv_latex, self).__init__(record)
@@ -47,7 +44,7 @@ class Cv_latex(Export):
 
     def _format_record(self):
         required_fields = ['title', 'author', 'arxiv']
-        optional_fields = ['doi', 'publi_info', 'url']
+        optional_fields = ['doi', 'publi_info', 'url', 'citation_count']
         try:
             return self._format_entry(required_fields, optional_fields)
         except MissingRequiredFieldError as e:
@@ -70,6 +67,7 @@ class Cv_latex(Export):
             'doi': self._get_doi,
             'publi_info': self._get_publi_info,
             'url': self._get_url,
+            'citation_count': self._get_citation_number
         }
         out = ''
         for field in req_fields:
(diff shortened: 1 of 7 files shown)


ioannistsanaktsidis:
@jmartinm  when you have time could you pls check my function so as to make sure it is the correct way I am fetching them? Thanks :)

inspirehep/utils/export.py
```diff
@@ -122,10 +122,34 @@ def _get_slac_citation(self):
...
                 cite_element + ';%%'
         return cite_line
+
+    def _get_citation_number(self):
+        """Returns how many times record was cited. If 0, returns nothing"""
+        from invenio_ext.es import es
+        from inspirehep.utils.citations import Citation
+        import time
+        today = time.strftime("%d %b %Y")
+        record = Citation(es.get(index='hep',
```
jmartinm: Here what you need is to fetch only the record you are currently formatting and accessing the field `citation_count `

No need to use `from inspirehep.utils.citations import Citation` as that will recompute the citations.

ioannistsanaktsidis:
Updated!

inspirehep/utils/cv_latex.py
```diff
@@ -128,6 +126,8 @@ def _format_output_row(self, field, value):
                 field, value)
         elif field == 'url':
             out += u' %\href{{{1}}}{{HEP entry}}.\n'.format(field, value)
+        elif field == 'citation_count':
+            out += u' %{1}'.format(field, value)
```
jmartinm: What is the use of having `field` in the `format` function? Just curious.

inspirehep/utils/export.py
```diff
@@ -122,10 +122,31 @@ def _get_slac_citation(self):
...
+        record = es.get(index='hep', id=self.record['control_number'])
+        citations = ''
+        try:
+            times_cited = record['_source']['citation_count']
+            if times_cited != 0:
+                if times_cited > 1:
+                    citations = '%d citations counted in INSPIRE as of %s' \
+                                % (times_cited, today)
+                else:
+                    citations = '%d citation counted in INSPIRE'\
```
jmartinm: here you don't have the date like above.

inspirehep/utils/export.py
```diff
@@ -122,10 +122,31 @@ def _get_slac_citation(self):
...
+                                % (times_cited, today)
+        except:
```
jmartinm: It's better to specify the exceptions to avoid hiding issues. Would it be here `except KeyError` ?

inspirehep/utils/cv_latex.py
ioannistsanaktsidis: Well in that case there is no need actually as we can have u' %{0}'.format(value)
It was mainly used for Bibtex where we had sth like       pages          = "079",
where pages where field and 079 value. I think we can remove it actually.

inspirehep/utils/export.py
ioannistsanaktsidis: Yes KeyError will be the case.

ioannistsanaktsidis:
I see some staff now in the format_output_row that can be done better. I will remove the parts of code that are not needed and I will push again.

jmartinm:
> I see some staff now in the format_output_row that can be done better. I will remove the parts of code that are not needed and I will push again.

@ioannistsanaktsidis Let me know when the code is final and can have a look.

ioannistsanaktsidis:
@jmartinm I just rebased and pushed the latest master. If it gets green we are good to go.

ioannistsanaktsidis:
Ohh wait, I can also add a test for the new functionality. What do you think?

ioannistsanaktsidis:
The test is added. The PR is ready.

PR #745, opened 2016-01-19
WIP Es improvements

kaplun:
I am trying here to clean up hep.json in order to make explicit all tokenizers and many more features... this is completely WIP

### DIFF
--- a/inspirehep/base/searchext/mappings/hep.json
+++ b/inspirehep/base/searchext/mappings/hep.json
@@ -10,28 +10,43 @@
           "default": {
             "mapping": {
               "analyzer": "basic_analyzer",
-              "copy_to": "global_fulltext",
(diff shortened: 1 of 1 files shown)


jmartinm:
For all the fields that we define as `not_analyzed`, shall we instead have a custom one that lowercases?

kaplun:
Yep. Although not e.g. URLs and DOIs which are case sensitive.

jmartinm:
> Yep. Although not e.g. URLs and DOIs which are case sensitive.

Even though technically speaking that might the case, at least I would hope that a url matches even if I write it in upper case. Isn't it the case noways in production?

kaplun:
eheh yes

kaplun:
Obsolete

PR #753, opened 2016-01-20
templates: fixes errors caused in brief and detailed records

### LINKED ISSUE #738: Errors in detailed records for qa.inspirehep.net
After running a test against all records in qa.inspirehep.net to see which detailed records would cause an error, here is [a list with 27142 of them](https://www.dropbox.com/s/mkva4fu3a4jmk2j/failed-recids.txt?dl=0).

I have extracted the XML from production so that the records can get loaded locally and fixed.

https://www.dropbox.com/s/x5gn76k4myasrco/bad-xml.xml?dl=0

To load locally:

```bash
inveniomanage migrator populate -f ~/Desktop/bad-xml.xml
```

Probably there are just few errors being repeated multiple times.

### LINKED ISSUE #747: Errors in brief records for qa.inspirehep.net
After doing a full test on all the brief formats on qa.inspirehep.net the following errors appear:

Recid | Error
------------ | -------------
874680 | unhashable type: 'list'
1230870 | expected string or buffer
1238277 | unhashable type: 'list'
1317440 | expected string or buffer

The xml can be downloaded, for example, https://qa.inspirehep.net/record/874680/export/xm

Loaded locally:

```bash
$ inveniomanage migrator populate -f ~/Desktop/xm.xml
```

And then test the brief format:

```bash
$ inveniomanage shell
```

```python
from invenio_ext.es import es
from invenio_records.api import Record as RecordApi
from invenio_formatter.engine import format_record
format_record(RecordApi(es.get(index='hep', id=recid)['_source']), 'hb')```
```

ioannistsanaktsidis:
* Closes #747.

* Closes #738.

### DIFF
--- a/inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
+++ b/inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
@@ -36,7 +36,6 @@
               </div>
               <div class="col-md-11">
                 <h4 class="custom-h">
-                  <b>
                     {% if record['titles[0].title']|is_list() %}
                       {% set title = record['titles[0].title[0]'] %}
                     {% else %}
--- a/inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
+++ b/inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
@@ -36,16 +36,25 @@
               </div>
               <div class="col-md-11">
                 <h4 class="custom-h">
+                  <b>
+                    {% if record['titles[0].title']|is_list() %}
+                      {% set title = record['titles[0].title[0]'] %}
+                    {% else %}
+                      {% set title = record['titles[0].title'] %}
+                    {% endif %}
+                    {% if not title %}
+                      {% set title = record['titles.title[0]'] %}
+                    {% endif %}
                     <a class="title" href="{{ url_for('record.metadata', recid=record['control_number']) }}">
-                      {{ record['titles[0].title']|capitalize }}
+                      {{ title|capitalize }}
                     </a>
-                    {% if record['titles[0].title']|count_words() > 5 %}
+                    {% if title|count_words() > 5 %}
                       <a class="mobile-title" href="{{ url_for('record.metadata', recid=record['control_number']) }}">
-                        {{ record['titles[0].title']|capitalize | words(5) + "..."}}
+                        {{ title|capitalize | words(5) + "..."}}
                       </a>
                     {% else %}
                       <a class="mobile-title" href="{{ url_for('record.metadata', recid=record['control_number']) }}">
-                        {{ record['titles[0].title']|capitalize }}
+                        {{ title|capitalize }}
                       </a>
                     {% endif %}
                 </h4>
--- a/inspirehep/base/templates/format/record/Inspire_Default_HTML_brief_macros.tpl
+++ b/inspirehep/base/templates/format/record/Inspire_Default_HTML_brief_macros.tpl
@@ -18,7 +18,7 @@
 #}
 
 {% macro render_doi() %}
-    {% set filtered_doi = record.get('dois.value')|tranform_nested_list()|remove_duplicates() %}
+    {% set filtered_doi = record.get('dois.value')|remove_duplicates() %}
     DOI: 
     {% for doi in filtered_doi %}
       {% if not loop.first %}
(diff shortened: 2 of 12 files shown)


ioannistsanaktsidis:
@jmartinm  I have finished this one, the citation test fails. Is it sth that needs to be done about it or is it another issue we have?

ioannistsanaktsidis:
@jmartinm after rebasing the PR passes all the tests. It is also ready.

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -36,16 +36,25 @@
               </div>
               <div class="col-md-11">
                 <h4 class="custom-h">
+                  <b>
```
jmartinm: There is no closing of this tag. It is not needed actually since the change to Helvetica font.

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -36,16 +36,25 @@
...
+                    {% if record['titles[0].title']|is_list() %}
+                      {% set title = record['titles[0].title[0]'] %}
+                    {% else %}
+                      {% set title = record['titles[0].title'] %}
+                    {% endif %}
+                    {% if not title %}
+                      {% set title = record['titles.title[0]'] %}
```
jmartinm: what would be this case?

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief_macros.tpl
```diff
@@ -18,8 +18,7 @@
 #}
 
 {% macro render_doi() %}
-  {% if record.get('dois')| is_list() %}
-    {% set filtered_doi = record.get('dois.value')| remove_duplicates() %}
+    {% set filtered_doi = record.get('dois.value')|tranform_nested_list()|remove_duplicates() %}
```
jmartinm: so what you are trying to do here is to flatten a nested list (http://stackoverflow.com/questions/952914/making-a-flat-list-out-of-list-of-lists-in-python) - you could rename the filter.

But the question is, why do you get a list of lists? I don't see in the `dojson` definition when that would be the case - https://github.com/inspirehep/inspire-next/blob/master/inspirehep/dojson/hep/fields/bd01x09x.py#L53

It might be better to solve the source of the problem. You can run the problematic record though `dojson` and see what is going on.

inspirehep/base/templates/format/record/Inspire_HTML_detailed_macros.tpl
```diff
@@ -222,30 +220,59 @@
...
+        {% if url.get('url')[0].startswith('http://projecteuclid.org') %}
+          <a href='{{ url.get('url')[0] }}'>Project Euclid</a>
+        {% endif %}
+        {% if url.get('url')[0].startswith('http://www-lib.kek.jp') %}
+          <a href='{{ url.get('url')[0] }}'>KEK scanned document</a>
+        {% endif %}
+        {% if url.get('url')[0].startswith('http://cds.cern.ch/record/') %}
+          <a href='{{ url.get('url')[0] }}'>CDS</a>
+        {% endif %}
+     {% else%}
```
jmartinm: Space after `else`

inspirehep/base/templates/format/record/Inspire_HTML_detailed_macros.tpl
```diff
@@ -222,30 +220,59 @@
...
-        <a href='{{ url.get('url') }}'>Project Euclid</a>
-      {% endif %}
-      {% if url.get('url').startswith('http://www-lib.kek.jp') %}
-        <a href='{{ url.get('url') }}'>KEK scanned document</a>
-      {% endif %}
-      {% if url.get('url').startswith('http://cds.cern.ch/record/') %}
-        <a href='{{ url.get('url') }}'>CDS</a>
-      {% endif %}
+    {% if url.get('url') != None %}
+     {% if url.get('url')|is_list() %}
```
jmartinm: As with the `dois`, can you have a look at why that field would ever not be a list?

In `dojson` we have:

```
@utils.for_each_value
```

So it should always be a list.

Solving the source of the problem would avoid so much duplication in the jinja templates.

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
ioannistsanaktsidis: As I remember in one record a title was not appearing. I checked and saw that titles.title was a list, so I added this is the case for this.

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief_macros.tpl
ioannistsanaktsidis: You are right in the dojson we do not have this case. The issue comes when we do this think:
{{ record.get('dois.value') }}
Actually this transforms 
screen shot 2016-01-27 at 15 23 28
to this : 
screen shot 2016-01-27 at 15 24 14
Do you think that it would be better if we avoid this style of getting the value?

inspirehep/base/templates/header.html
```diff
@@ -55,13 +55,13 @@
         </span>
       </a>
       {% if request.endpoint in ['search.search', 'record.metadata'] %}
-        <div id="top-search-bar-small" class="col-sm-5 {{collection_name}}">
+        <div id="top-search-bar-small" class="col-sm-5 {% if collection %}{{collection.name}} {% endif %}">
```
jmartinm: I think this fie might not be to be touched with the latest changes to `header.html`. The problems with `collection_name` should be solved now.

inspirehep/base/templates/search/form/controls.html
```diff
@@ -34,7 +34,7 @@
        type="text"
        tabindex="1"
        value="{{ request.args.get('p', '') }}"
-       placeholder="Search {{ collection_name }}"
+       placeholder="Search {% if collection %} {{collection.name | sanitize_collection_name }} {% endif %}"
```
jmartinm: same on this file. The problem should already be fixed in `master`

inspirehep/base/templates/header.html
ioannistsanaktsidis: Cool then, I will rebase, check and remove the if statement.

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief_macros.tpl
jmartinm: This would be better solved in the `dojson` rule for dois:

```python
@hep.over('dois', '^024[1032478_][10_]')
def dois(self, key, value):
    """Other Standard Identifier."""
    value = utils.force_list(value)
    out = []
    for val in value:
        if val and val.get("2", '').lower() == "doi":
            out.append({
                'value': val.get('a'),
                'source': val.get('9')
            })
    return out
```

So that we never put a list inside of `value`, instead separate it in two different fields.

ioannistsanaktsidis: Ok cool! What about the filter for flattening a list? As we will not have this problem due to dojson changes, should I remove it, or should we keep it in case we want one day?

jmartinm: Let's remove it if it is not needed anymore.

ioannistsanaktsidis: :+1:

PR #755, opened 2016-01-20
records: citations update receiver update

jalavik:
To avoid copying the record upon init of Record class when it is not always needed.

See https://github.com/inspirehep/invenio-records/commit/b685e69fb14cb35bb1416fb30ee967dc17039d70


kaplun:
Moved to #756

PR #757, opened 2016-01-21
header: fixes

nikpap:

### DIFF
--- a/inspirehep/base/templates/search/form/controls.html
+++ b/inspirehep/base/templates/search/form/controls.html
@@ -20,7 +20,7 @@
 {% extends "search/form/controls_base.html" %}
 
 {% if collection is defined %}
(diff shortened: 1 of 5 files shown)


inspirehep/base/templates/search/form/controls.html
```diff
@@ -19,11 +19,7 @@
 
 {% extends "search/form/controls_base.html" %}
 
-{% if collection is defined %}
- {% set collection_name = collection.name | sanitize_collection_name %}
-{% else %}
-  {% set collection_name = '' %}
-{% endif %}
+{%- set collection_name = collection.name | sanitize_collection_name -%}
```
jmartinm: This should be as before. In some pages like the search when an error occurs, collection is not defined, so the code will crash if left like that.

PR #758, opened 2016-01-21
global: Data collection addition

jmartinm:
* Adds data collection to collection bar.

* Creates basic landing page.


kaplun:
Can you add a snapshot so that @suenjedt and @pherterich can already check?

kaplun:
@jmartinm I think if you rebase it should be green.

PR #760, opened 2016-01-21
requirements: update invenio-ext version

jalavik:


kaplun:
@jalavik just rebase and it will be :green_heart:

PR #762, opened 2016-01-21
ext: error pages handler

jalavik:
* Adds a preliminary 500 error page.

* Fixes issues with collection name in templates when collection
  object is not available.

We should improve the text/look, but this acts as a preliminary page at least instead of good ol white Internal Server Error page.

Screenshot


kaplun:
This is indeed a critical functionality for our image. Can you add also a regression test with a mocked bad page to prove this page is presented?

kaplun:
You know what? This is too cool to wait for a regression test :)

PR #767, opened 2016-01-22
base: landing page route change

jmartinm:
* Shadows / view to directly render the Literature search page as
  landing page.


kaplun:
Isn't this creating any conflict?

PR #769, opened 2016-01-22
requirements: bump invenio-search/ext

kaplun:


kaplun:
This forces latest fix from @Panos512, therefore includes revert of @jalavik workaround.

PR #772, opened 2016-01-22
Workflows details

jalavik:


kaplun:
@jalavik Uh sorry can you just rebase it? I have merged the bumped version from invenio-ext/search

PR #775, opened 2016-01-25
filters: number of records per collection fix

jmartinm:
* Modifies the query to report the correct total number of records
  on each collection landing page.


kaplun:
And what collection:foo would have returned?

jmartinm:
It should return the same thing in principle but could be that as `collection` field looks into `_collections` and that field is populated with a signal before record insert, the number is not correct while the records are being added.

Instead with this change the search is the same one as empty query so the number is correct (tested locally)

PR #776, opened 2016-01-25
templates: displays conference info

### LINKED ISSUE #541: Detailed/Brief: Display information about conferences
For Literature records that have conference information, the following info can be found in the JSON:

```json
"publication_info":[  
   {  
      "cnum":"C14-02-23.1",
      "conference_recid":1276884,
      "journal_issue":"6",
      "journal_recid":1214555,
      "journal_title":"Nuovo Cim.",
      "journal_volume":"C037",
      "page_artid":"101-109",
      "parent_recid":1357781,
      "year":2014
   }
]
```

Showing info in conference papers
==========================

Example 1
--------------

001292801 773__ $$c101-109$$n6$$pNuovo Cim.$$vC037$$wC14-02-23.1$$y2014

screen shot 2016-01-15 at 3 15 55 pm

As there is publication info and conference info, the text would be:

```Published in Nuovo Cim. C037 (2014) 6, 101-109 (Proceedings of 28th Rencontres de Physique de la Vallée d'Aoste)```

Where `Proceedings` will be a link pointing to record id `1357781` and `28th Rencontres de Physique de la Vallée d'Aoste` a link pointing to record id `1276884`

Example 2
--------------

001292801 773__ $$c101-109$$n6$$wC14-02-23.1

Same as before but no journal information. The text would be:

```Published in proceedings of 28th Rencontres de Physique de la Vallée d'Aoste, pages 101-109```

Example 3
--------------

001405098 773__ $$wC15-07-14  (as long as there is no Proceedings record)

Here there only available field will be `conference_recid`. In that case we just will have the text:

```Contribution to 4th European Particle Accelerator Conference (EPAC 94)```

Example 4
--------------

000395649 773__ $$c1-10$$wC94-06-27$$xIn *London 1994, Proceedings, EPAC 94, vol. 1* 165-169, and Juelich KFA - Juel-2965 (94/09,rec.Dec.) 1-10

If a record does not have information in `parent_recid` but contains the field `pubinfo_freetext` (old $$x in MARC) then we display the content of that field.

### PR DESCRIPTION
**Pull request by ioannistsanaktsidis**
* Adds conference information according to cases.

* Closes #541.

### DIFF
--- a/inspirehep/base/templates/format/record/Conference_info_macros.tpl
+++ b/inspirehep/base/templates/format/record/Conference_info_macros.tpl
@@ -1,0 +1,47 @@
+{#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#}
+
+
+{%- macro conf_with_pub_info(parent_recid, conference_recid, conference_title) -%}
+  (<a href="/record/{{ parent_recid }}">Proceedings</a> of <a href="/record/{{ conference_recid }}">
+  {{ conference_title }}</a>)
+{%- endmacro -%}
+
+{%- macro conf_without_pub_info(parent_recid, conference_recid, conference_title, page_artid='') -%}
+    Published in <a href="/record/{{ parent_recid }}">proceedings</a> of <a href="/record/{{ conference_recid }}">{{ conference_title }}</a>
+  {%- if page_artid -%}
+    , pages  
+  {{ page_artid }}
+  {%- endif -%}
+{%- endmacro -%}
+
+{%- macro conference_only(conference_recid, conference_title, pub_info = false) -%}
+  {%- if pub_info -%}
+    <br/>
+  {%- endif -%}
+  Contribution to <a href="/record/{{ conference_recid }}">{{ conference_title }}</a>
+{%- endmacro -%}
+
+{%- macro proceedings_only(parent_recid, parent_title, pub_info = false) -%}
+  {%- if pub_info -%}
+    (<a href="/record/{{ parent_recid }}">Proceedings</a> of {{ parent_title }})   
+  {%- else -%}
+    Published in <a href="/record/{{ parent_recid }}">proceedings</a> of {{ parent_title }}
+  {%- endif -%}
+{%- endmacro -%}
--- a/inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
+++ b/inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
@@ -25,6 +25,7 @@
 
 {% block record_header %}
 {% call cache_if_not_debug('record_brief', record['control_number']) %}
+{% set pub_info = record|publication_info %}
 <div class="row">
   <div class="col-md-12">
     <div id="panel-default-brief" class="panel panel-default" >
@@ -53,9 +54,9 @@
                   {% endif %}
                 </div>
                 <div class="row">
-                  {% if record.get('publication_info') and record.get('dois') %}
-                    {% if record.get('publication_info') | length == 1 
-                    and record.get('dois') | length == 1 %}
+                  {% if pub_info['pub_info'] and record.get('dois') %}
+                    {% if pub_info['pub_info'] | length == 1 
+                    and record.get('dois.value')|remove_duplicates() | length == 1 %}
                       <div class="col-md-12 ">
                         {{ record_journal_info_and_doi() }}
                       </div>
@@ -66,7 +67,7 @@
                         {{ render_doi() }}
                       </div>
                     {% endif %}
-                  {% elif record.get('publication_info') %}
+                  {% elif pub_info %}
                     <div class="col-md-12 ">
                       {{ record_journal_info() }}
                     </div>
--- a/inspirehep/ext/jinja_filters/record.py
+++ b/inspirehep/ext/jinja_filters/record.py
@@ -30,6 +30,9 @@ from inspirehep.utils.cv_latex import Cv_latex
 from inspirehep.utils.cv_latex_html_text import Cv_latex_html_text
 from inspirehep.utils.latex import Latex
 
+from invenio_ext.es import es
+from invenio_ext.template.utils import render_macro_from_template
+
 from invenio_base.globals import cfg
 
 from invenio_search.api import Query
@@ -408,39 +411,6 @@ def setup_app(app):
                    search(collection=collection_name))
 
     @app.template_filter()
-    def publication_info(record):
-        result = []
-        if 'publication_info' in record:
-            journal_title = journal_volume = year = journal_issue = pages = ''
-            for field in record['publication_info']:
-                out = 'Published in '
-                if 'journal_title' in field:
-                    if not ('journal_volume' in field or 'journal_issue' in
-                            field or 'page_artid' in field or
-                            'dois' in record):
-                        journal_title = 'Submitted to:' +\
-                            field['journal_title']
-                    else:
-                        journal_title = field['journal_title']
-
-                    if 'journal_volume' in field:
-                        journal_volume = ' ' + field['journal_volume']
-
-                    if 'year' in field:
-                        year = ' (' + str(field['year']) + ')'
-
-                    if 'journal_issue' in field:
-                        journal_issue = ' ' + field['journal_issue'] + ', '
-
-                    if 'page_artid' in field:
-                        pages = ' ' + field['page_artid']
-
-                    out += '<i>' + journal_title + '</i>&nbsp;' +\
-                        journal_volume + year + journal_issue + pages
-                    result.append(out)
-        return result
-
-    @app.template_filter()
     def is_upper(s):
         return s.isupper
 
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

**Comment by annetteholtkamp** [comment]
Could we see a few screen shots for the different cases?

- Annette

**Comment by ioannistsanaktsidis** [comment]
@annetteholtkamp  Yes of course, I will add some here when I have a stable version. Cheers.

**Comment by ioannistsanaktsidis** [comment]
Only Conference Id available
screen shot 2016-01-25 at 17 09 59

**Comment by ioannistsanaktsidis** [comment]
Conference and Parent Id available
screen shot 2016-01-25 at 17 13 26

**Comment by ioannistsanaktsidis** [comment]
Conference and Parent Id available, but no journal_information
screen shot 2016-01-25 at 17 17 08

**Comment by annetteholtkamp** [comment]
The two columns look a bit strange. Better to have the whole line for the conference and the arxiv nr underneath I think

- Annette

**Comment by annetteholtkamp** [comment]
For the detailed format it would be good to have a bit more detailed info about the conference, e.g. place and date. The titles are sometimes not very informative, e.g. “ Is SUSY alive and well? "

Annette

**Comment by ioannistsanaktsidis** [comment]
@jmartinm I pushed the latest changes for this PR. You can check and if evth ok we are good to go.

File: inspirehep/config.py [hunk 1]
```diff
@@ -339,6 +339,7 @@
     "980__a": ["collections.primary"],
     "980__b": ["collections.secondary"],
     "542__l": ["information_relating_to_copyright_status.copyright_status"],
+    "recid": ["control_number"]
```
jmartinm: do you use this in this PR?

inspirehep/ext/jinja_filters/record.py
```diff
@@ -390,3 +390,99 @@ def citation_phrase(count):
             return 'Cited 1 time'
         else:
             return 'Cited ' + str(count) + ' times'
+
+    @app.template_filter()
+    def conference_info(record):
+        out = ''
```
jmartinm: In Python it is more efficient creating an empty list and appending text to it.

```python
out = []
out.append("hello")
out.append("world")
return "".join(out)
```

inspirehep/ext/jinja_filters/record.py
```diff
@@ -390,3 +390,99 @@ def citation_phrase(count):
...
+                if 'journal_title' in element:
+                    title_exists = True
+                if 'conference_recid' in element \
+                        and 'parent_recid' in element:
+                    conference_rec = es.get(
+                        index='conferences', id=element['conference_recid'])
+                    if 'journal_title' not in element:
+                        if 'page_artid' in element:
+                            if title_exists:
+                                out += '(<a href="/record/' + str(
```
jmartinm: Instead of having all the HTML here it can be nice adding all the snippets as macros in a Jinja template and then using this utility (https://github.com/inspirehep/invenio-ext/blob/master/invenio_ext/template/utils.py#L25) to render them.

See usage here https://github.com/inspirehep/invenio-deposit/blob/master/invenio_deposit/field_widgets.py#L554

( I like it because I wrote that helper :) )

inspirehep/ext/jinja_filters/record.py
```diff
@@ -390,3 +390,99 @@ def citation_phrase(count):
...
+        out = ''
+        title_exists = False
+        if 'publication_info' in record:
+            from invenio_ext.es import es
+            for element in record['publication_info']:
+                if 'journal_title' in element:
+                    title_exists = True
+                if 'conference_recid' in element \
+                        and 'parent_recid' in element:
+                    conference_rec = es.get(
```
jmartinm: Can you use [get_source](https://elasticsearch-py.readthedocs.org/en/master/api.html#elasticsearch.Elasticsearch.get_source) instead ? It will avoid having to access the `_source` key and you can specify which fields to get.

inspirehep/ext/jinja_filters/record.py
```diff
@@ -390,3 +390,99 @@ def citation_phrase(count):
...
+                            if title_exists:
+                                out += '(<a href="/record/' + str(
+                                    element['parent_recid']) + \
+                                    '">Proceedings' + '</a>' + \
+                                    ' of <a href="/record/' + \
+                                    str(element['conference_recid']) + \
+                                    '">' + conference_rec['_source']['title'] + '</a>' + \
+                                    ', pages ' + element['page_artid'] + ')'
+                                break
+                            else:
```
jmartinm: I am not sure I see why depending if there is a journal title or not the display of the conference information is different?

inspirehep/ext/jinja_filters/record.py
```diff
@@ -390,3 +390,99 @@ def citation_phrase(count):
...
+        if 'publication_info' in record:
+            from invenio_ext.es import es
+            for element in record['publication_info']:
+                if 'journal_title' in element:
+                    title_exists = True
+                if 'conference_recid' in element \
+                        and 'parent_recid' in element:
+                    conference_rec = es.get(
+                        index='conferences', id=element['conference_recid'])
+                    if 'journal_title' not in element:
```
jmartinm: Here you can use your previous `title_exists` variable.

inspirehep/ext/jinja_filters/record.py
```diff
@@ -390,3 +390,99 @@ def citation_phrase(count):
...
+                                out += 'Published in <a href="/record/' + str(
+                                    element['parent_recid']) + \
+                                    '">proceedings' + '</a>' + \
+                                    ' of <a href="/record/' + \
+                                    str(element['conference_recid']) + \
+                                    '">' + conference_rec['_source']['title'] + '</a>' + \
+                                    ', pages ' + element['page_artid']
+                                break
+                        else:
+                            if title_exists:
```
jmartinm: You have in a [previous if](https://github.com/inspirehep/inspire-next/pull/776/files#diff-50fbc1d5feafee144203f26406a1267cR407) checked that title does not exist so this case is not possible.

inspirehep/ext/jinja_filters/record.py
```diff
@@ -390,3 +390,99 @@ def citation_phrase(count):
             return 'Cited 1 time'
         else:
             return 'Cited ' + str(count) + ' times'
+
+    @app.template_filter()
+    def conference_info(record):
```
jmartinm: In general I think this should be merged with the logic of displaying the publication information. (while merging, the [publication information legacy code](https://github.com/inspirehep/inspire/blob/master/bibformat/format_elements/bfe_INSPIRE_publi_info.py#L28) should be double checked as I am not sure the output of the jinja macro is equal to that code)

They can both be converted into a common filter to display the publication info line.

As there are quite a few cases and the logic is not simple we need a good amount of tests for this functionality. You can have different snippets of json with all the combinations of publication information, conference info, etc and the desired output.

ioannistsanaktsidis:
@jmartinm this one is also ready for testing. After it becomes green you can test it and merge it :) Cheers.

inspirehep/base/templates/format/record/templates_from_macros.tpl
```diff
@@ -0,0 +1,48 @@
+{#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
```
jmartinm: The name of this file would be better something like `conference_info_macros.tpl`

jmartinm:
What is the difference between these 2 cases?

screen shot 2016-02-09 at 1 42 58 pm

screen shot 2016-02-09 at 1 45 08 pm

Shouldn't the first one also say, `Published in proceedings...`

ioannistsanaktsidis:
The first one does not contain the conference_recid thats why it is not linked.
```
[{u'parent_recid': 720114, u'cnum': u'C05-08-14.1', u'confpaper_info': u'Invited talk at'}]
```
The second one is exactly the second example of https://github.com/inspirehep/inspire-next/issues/541
Should I make the first one to be like 
```
Published in proceedings... 
```
without the link?

jmartinm:
Ok, it is strange because it has a `cnum` so it should have the record id of the conference as well. I guess it is a problem of the demo records not having the full information.

Yes, the first one should be changed to `Published in proceedings... ` , otherwise it seems like the sentence is missing something at the beginning.

ioannistsanaktsidis:
Updated :)

inspirehep/base/templates/format/record/Conference_info_macros.tpl
```diff
@@ -0,0 +1,48 @@
...
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#}
+
+
+{% macro conf_with_title(parent_recid, conference_recid, conference_title) %}
+  (<a href="/record/{{ parent_recid }}">Proceedings</a> of <a href="/record/{{ conference_recid }}">
+  {{ conference_title }}</a>)
+{% endmacro %}
+
+{% macro conf_without_title(parent_recid, conference_recid, conference_title, page_artid='') %}
```
jmartinm: why is it `conf_without_title` if it uses inside `conference_title `? ;)

inspirehep/base/templates/format/record/Conference_info_macros.tpl
```diff
@@ -0,0 +1,48 @@
...
+  {% if page_artid %}
+    Published in <a href="/record/{{ parent_recid }}">proceedings</a> of <a href="/record/{{ conference_recid }}">{{ conference_title }}</a>, pages  
+    {{ page_artid }}
+  {% else %}
+    Published in <a href="/record/{{ parent_recid }}">proceedings</a> of <a href="/record/{{ conference_recid }}">{{ conference_title }}</a>
```
jmartinm: You can avoid duplicating this line, just adding the pages at the end if `{{ page_artid }}` is present. (without the whole sentence)

inspirehep/base/templates/format/record/Conference_info_macros.tpl
```diff
@@ -0,0 +1,48 @@
...
+  {% endif %}
+{% endmacro %}
+
+{% macro conference_only(conference_recid, conference_title, pub_info = '') %}
```
jmartinm: should `pub_info` be a boolean?

inspirehep/base/templates/format/record/Conference_info_macros.tpl
ioannistsanaktsidis: Missleading name, it should be conf_without_pub_info and conf_with_pub_info.

inspirehep/base/templates/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -63,7 +64,7 @@
                         {{ render_doi() }}
                       </div>
                     {% endif %}
-                  {% elif record.get('publication_info') %}
+                  {% elif record|publication_info %}
```
jmartinm: You have already computed this before in the variable `pub_info`

inspirehep/ext/jinja_filters/record.py
```diff
@@ -458,3 +427,105 @@ def is_upper(s):
...
+                    }
+                    if result:
+                        result['conf_info'] = render_macro_from_template(
+                            name="conf_with_title",
+                            template="format/record/Conference_info_macros.tpl",
+                            ctx=ctx)
+                        break
+                    else:
+                        if 'page_artid' in element:
+                            ctx.update(
```
jmartinm: if `page_artid` is there you can update the `ctx`. Then you can remove the `else` statement and the duplicated code:

```python
result['conf_info'] = render_macro_from_template(
 +                                name="conf_without_title",
 +                                template="format/record/Conference_info_macros.tpl",
 +                                ctx=ctx)
 +                            break
```

tests/test_publication_info.py
```diff
@@ -0,0 +1,102 @@
...
+
+        self.pub_info_unique = {"pub_info": [u'<i>JCAP</i> 1503 (2015) 044']}
+        self.pub_info_multiple = {
+            'pub_info': [u'<i>JHEP</i> 1401 (2014) 163', u'<i>JHEP</i> 1501 (2015) 014']}
+        self.pub_info_freetext = {'pub_info': [
+            u'Current Science Vol. 109 (N0. 12, 25 December 2015), 2220-2229\n (2015)']}
+        self.conf_info_conf_and_par = {
+            'conf_info': u'(<a href="/record/706120">Proceedings</a> of <a href="/record/976391">  2005 International Linear Collider Workshop (LCWS 2005)</a>)'}
+        self.conf_info_only_par = {
+            'conf_info': u'\n \n <a href="/record/720114">Proceedings</a> of 2005 International Linear Collider Physics and Detector Workshop and 2nd ILC Accelerator Workshop (Snowmass 2005)\n \n'}
```
jmartinm: Is it intended to have '\n' ? Is this then supposed to be directly used in a Jinja template?

ioannistsanaktsidis: Yes exactly it is supposed to be used directly in template.

tests/test_publication_info.py
```diff
@@ -0,0 +1,102 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+from invenio.testsuite import InvenioTestCase, make_test_suite, run_test_suite
+from flask import current_app
+
+
+class PubInfoTests(InvenioTestCase):
```
jmartinm: Is there a test case for the cases when there is publication information and conference information?

Where the output should be:

`Published in Nuevo Cim, 23, 99 (Proceedings of Conference name)`

Looking at the code it seemed that in most cases first the publication information will be shown and in the next line the information about the conference. Not sure this case is taken into account.

tests/test_publication_info.py
jmartinm: And does jinja render correctly this line when there are \n ? Does it introduce a `<br />` or just ignores these characters ?

ioannistsanaktsidis: It renders it correctly. E.g 
screen shot 2016-02-09 at 15 34 34

jmartinm: That is an example without proceedings right? Looks fine to me.

What about when proceedings are present? As far as I remember the idea was to present it like:

`Published in J.Phys. XX (Proceedings of 18th International Conference)`

ioannistsanaktsidis: When Proceedings are present we present it like this:
screen shot 2016-02-09 at 16 13 01

jmartinm: :+1:

jmartinm: And just to clarify, what is the goal of '\n \n' Do you add yourself for some reason ?

inspirehep/ext/jinja_filters/record.py
```diff
@@ -458,3 +427,105 @@ def is_upper(s):
     def split_author_name(name):
         new_name = name.split(',')
         return ' '.join(reversed(new_name))
+
+    @app.template_filter()
+    def publication_info(record):
+        """Displays inline publication and conference information"""
+        result = {}
+        out = []
+        """Publication info line"""
```
jmartinm: Can you convert this into a comment instead of dosctring?

ioannistsanaktsidis: :+1:

inspirehep/ext/jinja_filters/record.py
```diff
@@ -458,3 +427,105 @@ def is_upper(s):
...
+                    if 'year' in pub_info:
+                        year = ' (' + str(pub_info['year']) + ')'
+                    if 'journal_issue' in pub_info:
+                        journal_issue = ' ' + pub_info['journal_issue'] + ', '
+                    if 'page_artid' in pub_info:
+                        pages = ' ' + pub_info['page_artid']
+                    out.append(journal_title + journal_volume +
+                               year + journal_issue + pages)
+            if out:
+                result['pub_info'] = out
```
jmartinm: Here you are returning a list. Might be better if you keep appending the parts of the string like:

```python
out.append(journal_title)
out.append(year)
```

And at the end:

```python
result['pub_info'] = "".join(out)
```

or

 ```python
result['pub_info'] = " ".join(out)
```

if you need each element separated by space.

File: inspirehep/ext/jinja_filters/record.py [hunk 19]
```diff
@@ -458,3 +427,105 @@ def is_upper(s):
...
+            if not result:
+                for field in record['publication_info']:
+                    if 'pubinfo_freetext' in field:
+                        out.append(field['pubinfo_freetext'])
+                        result['pub_info'] = out
+                        break
+            """Conference info line"""
+            for element in record['publication_info']:
```
jmartinm: Instead of `element` you could call it like above `pub_info` for consistency.

tests/test_publication_info.py
ioannistsanaktsidis: It is added because of the macros file that we are using. This happens because I use {% macro conf_with_pub_info(parent_recid, conference_recid, conference_title) %} and not sth like 
{%- macro conf_with_pub_info(parent_recid, conference_recid, conference_title) -%}. I can remove it actually in most of them except the case that I am using this 
```
{% macro conference_only(conference_recid, conference_title, pub_info = false) %}
  {% if pub_info %}
    <br/>
  {% endif %}
  Contribution to <a href="/record/{{ conference_recid }}">{{ conference_title }}</a>
{% endmacro %}
```
because I need to break the line in this one.

inspirehep/ext/jinja_filters/record.py
jmartinm: Or you need to keep it as a list, in case there are several publication_info and in the Jinja template you separate them by `and`. It might be fine then.

ioannistsanaktsidis:
Updated with the latest changes. I removed the \n as they are only misleading and make the code uglier.

PR #778, opened 2016-01-25
inspirehep: no refreshing upon reindexing

kaplun:


kaplun:
@jmartinm this is implementing what you pointed out. Let's see how fast we can even be :-D

PR #782, opened 2016-01-26
authors: skip uuid if already present

jacenkow:
* Skips assignation of uuid if a given signature has already one.


jacenkow:
CC: @kaplun

PR #783, opened 2016-01-26
configuration: elastic mapping path

Panos512:
* Adds the path to elastic search mapping file in the config file.

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -460,8 +460,28 @@ SEARCH_ELASTIC_COLLECTION_INDEX_MAPPING = {
 SEARCH_ELASTIC_DEFAULT_INDEX = 'hep'
 
 INSPIRE_PATH = pkg_resources.resource_filename("inspirehep", "")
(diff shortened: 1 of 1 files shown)


jmartinm:
We might want to add the other files as well for conferences, institutions and so on, in order to avoid https://github.com/inspirehep/inspire-next/pull/786/files#diff-3f822b17443888990c9f70f9f89d66f6R342

The only worry is, do you need to open the files upon every search or this is done only once per process?

Opening the files might affect performance when searching.

Panos512:
That was indeed a problem. With [this]( https://github.com/inspirehep/invenio-query-parser/blob/a869ddca55a39e5bbf91ba7f69988018141d697c/invenio_query_parser/utils.py#L72 ) fix that Sam proposed we save the list of keywords after we generate it the first time.

jmartinm:
> That was indeed a problem. With this fix that Sam proposed we save the list of keywords after we generate it the first time.

Nice, then can you amend the commit to also include the other mappings? I use them for autocompletion of certain fields in the submission form.

PR #786, opened 2016-01-26
facets: fix 0 results problem in all collections

ioannistsanaktsidis:

### DIFF
--- a/inspirehep/base/searchext/mappings/institutions.json
+++ b/inspirehep/base/searchext/mappings/institutions.json
@@ -102,6 +102,19 @@
                 "ICN": {
                     "type": "string",
                     "copy_to": ["affautocomplete"]
+                },
+                "address": {
+                    "properties": {
+                        "country": {
+                            "type": "string",
+                            "fields": {
+                                "raw": {
+                                    "type": "string",
+                                    "index": "not_analyzed"
+                                }
+                            }
+                        }
+                    }
                 }
             }
         }
--- a/inspirehep/base/static/less/format/brief-results.less
+++ b/inspirehep/base/static/less/format/brief-results.less
@@ -405,3 +405,28 @@ i.fa-chevron-right.facet-slider {
 .export-all-div {
     padding-left: 9px;
 }
+
+.hep#clear-facets-message {
+    color: @literature-collection;
+}
+
+.conferences#clear-facets-message {
+    color: @conferences-collection;
+}
+
+.jobs#clear-facets-message {
+    color: @jobs-collection;
+}
+
+.institutions#clear-facets-message {
+    color: @institutions-collection;
+}
+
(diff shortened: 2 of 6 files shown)


inspirehep/base/searchext/mappings/institutions.json
```diff
@@ -102,6 +102,14 @@
                 "ICN": {
                     "type": "string",
                     "copy_to": ["affautocomplete"]
+                },
+                "address": {
```
jmartinm: Can you check if it is possible to use `fields` in order to have the `address` field analyzed and a `raw` version not analyzed? And for the facets we will use `address.raw`

ioannistsanaktsidis:
I tried to exactaddress and seems to working fine. We can keep it like this I guess.

inspirehep/base/searchext/mappings/institutions.json
```diff
@@ -102,6 +102,23 @@
...
+                        }
+                    }
+                },
+                "address": {
+                    "properties": {
+                        "country": {
+                            "type": "string",
+                            "copy_to": ["exactaddress"]
+                        }
+                    }
```
jmartinm:
```
"address": {
    "properties": {
        "country": {
             "type": "string",
             "fields": {
                 "raw": { 
                     "type":  "string",
                    "index": "not_analyzed"
                }
            }
     }
}
```

And then you would have `address.country` and `address.country.raw`

ioannistsanaktsidis:
Updated!

jmartinm:
After merging https://github.com/inspirehep/inspire-next/commit/dd847d83c46e646248e21438fc1cf61b2ba6d2c5 the keywords that are in the ElasticSearch mapping should be discovered automatically so we might not need to add most of keywords in `SEARCH_ELASTIC_KEYWORD_MAPPING`

Can you rebase and which ones we can remove? Once it is working with latest `master` I can merge this.

ioannistsanaktsidis:
@jmartinm I updated the PR. We can only remove the series keyword. I tried to remove the others too, but the facets  not working properly without them.

jmartinm:
> @jmartinm I updated the PR. We can only remove the series keyword. I tried to remove the others too, but the facets not working properly without them.

@Panos512 isn't it strange? I would assume now that mappings are taken automatically doing something like `"continent": ["continent"]` should not be needed. Can you take install this PR and see if you get the same problem?

ioannistsanaktsidis:
@jmartinm  Updated!

ioannistsanaktsidis:
@jmartinm  we solved the problem with @Panos512 . The problem was that in my local the query-parser
and the invenio-search  were not updated. Cheers!

ioannistsanaktsidis:
@jmartinm  this one is also ready.

PR #791, opened 2016-01-28
caching: add caching and cache management (invalidation)

tomaszgy:
* Adds caching and memoization to the following views:
  single-reference, single-citation, references, citations
  and brief search.

* Introduces InspireCache class which extends Flask Cache
  (adding conditional caching and memoization:
   e.g. cache/memoize when not in debug mode).

* Adds CacheManager extension which decides
  which cached views should be invalidated on record insert/update.

* Adds 'invalidate_cache' command to inspirehep in order to
  allow manual invalidation of certain views (references, citations etc.).

### DIFF
--- a/inspirehep/base/templates/format/record/Cache_HTML_macros.tpl
+++ b/inspirehep/base/templates/format/record/Cache_HTML_macros.tpl
@@ -1,7 +1,6 @@
-# -*- coding: utf-8 -*-
-#
+{#
 # This file is part of INSPIRE.
-# Copyright (C) 2015, 2016 CERN.
+# Copyright (C) 2016 CERN.
 #
 # INSPIRE is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -16,5 +15,15 @@
 # You should have received a copy of the GNU General Public License
 # along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#}
 
-from .receivers import *
+{% macro cache_if_not_debug(template_name, recid) %}
+  {% set content=caller() %}
+    {%  if not config.DEBUG %}
+      {% cache timeout | long_timeout, template_name, recid %}
+        {{  content }}
+      {% endcache %}
+    {% else %}
+      {{ content }}
+    {%  endif %}
+{% endmacro %}
--- a/inspirehep/base/views.py
+++ b/inspirehep/base/views.py
@@ -1,44 +1,43 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of INSPIRE.
-# Copyright (C) 2014, 2015 CERN.
+# Copyright (C) 2014, 2015, 2016 CERN.
 #
-# INSPIRE is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
 #
-# INSPIRE is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
 #
 # You should have received a copy of the GNU General Public License
-# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
-#
-# In applying this licence, CERN does not waive the privileges and immunities
-# granted to it by virtue of its status as an Intergovernmental Organization
-# or submit itself to any jurisdiction.
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 
 """Inspire views."""
 
 import json
 
 from flask import (
-    redirect,
-    request,
     Blueprint,
-    render_template,
     current_app,
+    redirect,
+    render_template,
+    request,
     url_for
 )
-from flask.ext.menu import register_menu, current_menu
-from flask.ext.login import current_user
 
-from invenio_base.i18n import _
-from invenio_ext.email import send_email
+from flask.ext.login import current_user
+from flask.ext.menu import current_menu, register_menu
+
 from invenio_base.decorators import wash_arguments
 from invenio_base.globals import cfg
+from invenio_base.i18n import _
+
+from invenio_ext.email import send_email
 
 
 blueprint = Blueprint('inspire', __name__, url_prefix="",
@@ -163,15 +162,29 @@ def data():
                  'collection': (unicode, "")})
 def ajax_references(recid, collection):
     """Handler for datatables references view"""
+
+    from elasticsearch import TransportError
     from flask import jsonify
-    from inspirehep.utils.references import Reference
-    from invenio.base.globals import cfg
+    from inspirehep.utils.references import render_references
     from invenio_ext.es import es
+    from invenio_base.globals import cfg
 
     index = cfg['SEARCH_ELASTIC_COLLECTION_INDEX_MAPPING'].get(collection, 'hep')
+
+    try:
+        record = es.get(index=index, id=recid)['_source']
+    except TransportError:
+        current_app.logger.exception('Cannot render references for record no. {0}. Could not find record in ES.'
+                                     .format(str(recid)))
+        return jsonify(
+            {
+                "data": []
+            }
+        )
+
     return jsonify(
         {
-            "data": Reference(es.get(index=index, id=recid)['_source']).references()
+            "data": render_references(record)
         }
     )
 
@@ -181,15 +194,12 @@ def ajax_references(recid, collection):
                  'collection': (unicode, "")})
 def ajax_citations(recid, collection):
     """Handler for datatables citations view"""
-    from flask import jsonify
-    from inspirehep.utils.citations import Citation
-    from invenio.base.globals import cfg
-    from invenio_ext.es import es
 
-    index = cfg['SEARCH_ELASTIC_COLLECTION_INDEX_MAPPING'].get(collection, 'hep')
+    from flask import jsonify
+    from inspirehep.utils.citations import render_citations
     return jsonify(
         {
-            "data": Citation(es.get(index=index, id=recid)['_source']).citations()
+            "data": render_citations(recid)
         }
     )
 
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -72,6 +72,8 @@ EXTENSIONS = [
     'inspirehep.ext.jinja_filters.general',
     'inspirehep.ext.jinja_filters.record',
     'inspirehep.ext.redis.client',
+    'inspirehep.ext.cache_manager.cache',
+    'inspirehep.ext.cache_manager.manager',
     'inspirehep.ext.deprecation_warnings:disable_deprecation_warnings',
     'inspirehep.ext.error_pages',
 ]
@@ -596,3 +598,16 @@ WORKFLOWS_HOLDING_PEN_ES_PROPERTIES = {
 }
 """Updates default properties that should be added to the Holding Pen index
 mappings."""
+
+CACHED_VIEWS = {
+    'record_oneline': {
+        'type': 'jinja_cache',
+        'template_name': 'record_oneline',
+        'kwargs': ['recid']
+    },
+    'record_brief': {
+        'type': 'jinja_cache',
+        'template_name': 'record_brief',
+        'kwargs': ['recid']
+    }
+}
--- a/inspirehep/ext/cache_manager/cache.py
+++ b/inspirehep/ext/cache_manager/cache.py
@@ -1,0 +1,140 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+import functools
+
+from flask import current_app, request
+
+from flask_cache import Cache
+
+
+class InspireCache(Cache):
+    """Extends Cache from Flask Cache and adds mode-checking (debug or not) for memoize and cached methods.
+
+    The parameters in both inspire_memoize and inspire_cached are the same
+    as the parameters of the original methods in Flask Cache class
+
+    """
+
+    def inspire_memoize(self, timeout=None, make_name=None, unless=None):
+        def memoize(f):
+            @functools.wraps(f)
+            def decorated_function(*args, **kwargs):
+                if not current_app.debug:
+                    #: bypass cache
+                    if callable(unless) and unless() is True:
+                        return f(*args, **kwargs)
+
+                    try:
+                        cache_key = decorated_function.make_cache_key(f, *args, **kwargs)
+                        rv = self.cache.get(cache_key)
+                    except Exception:
+                        if current_app.debug:
+                            raise
+                        current_app.logger.exception("Exception possibly due to cache backend.")
+                        return f(*args, **kwargs)
+
+                    if rv is None:
+                        rv = f(*args, **kwargs)
+                        try:
+                            self.cache.set(cache_key, rv, timeout=decorated_function.cache_timeout)
+                        except Exception:
+                            if current_app.debug:
+                                raise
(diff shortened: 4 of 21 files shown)


inspirehep/ext/jinja_filters/record.py
```diff
@@ -390,3 +390,11 @@ def citation_phrase(count):
             return 'Cited 1 time'
         else:
             return 'Cited ' + str(count) + ' times'
+
+    @app.template_filter()
+    def long_timeout(timeout):
+        return 5184000  # 60 days
+
+    @app.template_filter()
+    def in_debug_mode():
```
jmartinm: This is unused it seems? Might be enough to just do in templates `if config.DEBUG` without a filter.

inspirehep/manage.py
```diff
@@ -210,3 +209,19 @@ def recreate_index(name, mapping, rebuild=False, delete_old=True):
...
+def invalidate_cache(cache_name):
+    """Invalidate cache of certain type e.g. brief, citation etc."""
+    from inspirehep.ext.cache_manager import cache_manager
+    template_name_to_function = {
+        'all': cache_manager.invalidator.invalidate_all_cached_views,
+        'citation': cache_manager.invalidator.invalidate_all_citation_views,
+        'brief': cache_manager.invalidator.invalidate_all_brief_search_views,
+        'references': cache_manager.invalidator.invalidate_all_references_views,
+        'citations': cache_manager.invalidator.invalidate_all_citations_views,
+        'reference': cache_manager.invalidator.invalidate_all_reference_views
```
jmartinm: So what is the difference between the singular `reference` cache and the plural `references` ? What is each caching exactly?

inspirehep/ext/jinja_filters/record.py
tomaszgy: You're right. Filter is not needed.

inspirehep/base/templates/format/record/Cache_HTML_macros.tpl
```diff
@@ -0,0 +1,29 @@
...
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#}
+
+{% macro cache_if_not_debug(template_name, recid) %}
+    {% set content=caller() %}
+    {%  if not config.DEBUG %}
+        {% cache timeout | long_timeout, template_name, recid %}
+            {{  content }}
```
kaplun: Unbalanced number of spaces :space_invader:

inspirehep/base/views.py
```diff
@@ -158,17 +156,9 @@ def data():
...
-    from invenio_ext.es import es
 
-    index = cfg['SEARCH_ELASTIC_COLLECTION_INDEX_MAPPING'].get(collection, 'hep')
-    return jsonify(
-        {
-            "data": Reference(es.get(index=index, id=recid)['_source']).references()
-        }
-    )
+    from inspirehep.utils.references import AjaxReferences
+    return AjaxReferences.get_references(unicode(recid), unicode(collection.lower()))
```
kaplun: Do you reallly need to transform into `unicode` an `int`?

kaplun: This is passed directly to ES. It should be able to cast on its own an int to a unicode.

inspirehep/base/views.py
```diff
@@ -176,17 +166,9 @@ def ajax_references(recid, collection):
...
-
-    index = cfg['SEARCH_ELASTIC_COLLECTION_INDEX_MAPPING'].get(collection, 'hep')
-    return jsonify(
-        {
-            "data": Citation(es.get(index=index, id=recid)['_source']).citations()
-        }
-    )
+
+    from inspirehep.utils.citations import AjaxCitations
+    return AjaxCitations.get_citations(unicode(recid))
```
kaplun: Ditto

inspirehep/ext/cache_manager/cache.py
```diff
@@ -0,0 +1,141 @@
...
+
+            decorated_function.uncached = f
+            decorated_function.cache_timeout = timeout
+            decorated_function.make_cache_key = self._memoize_make_cache_key(make_name, decorated_function)
+            decorated_function.delete_memoized = lambda: self.delete_memoized(f)
+
+            return decorated_function
+        return memoize
+
+    def inspire__cached(self, timeout=None, key_prefix='view/%s', unless=None):
```
kaplun: Should be `inspire_cached`.

jmartinm:
Looks good @tomaszgy ! Looking forward to seeing it in action :)

 You know the reason why the test fails ? https://travis-ci.org/inspirehep/inspire-next/builds/105627720#L2222

inspirehep/ext/cache_manager/cache.py
kaplun: Can you add docstring? What's `unless`?

inspirehep/ext/cache_manager/cache.py
```diff
@@ -0,0 +1,141 @@
...
+
+from invenio_base.globals import cfg
+
+logger = logging.getLogger(__name__)
+
+
+class InspireCache(Cache):
+    """Extends Cache from flask_cache and adds mode-checking (debug or not) for memoize and cached methods."""
+
+    def inspire_memoize(self, timeout=None, make_name=None, unless=None):
```
kaplun: Why are you calling it `inspire_memoize`, rather than overriding the parent class?

inspirehep/manage.py
tomaszgy: Singular `reference ` is a single entry in the plural `references` field. Just realized though that it can be reimplemented so we could use singular `citation` for that and get rid of singular `reference`.

inspirehep/ext/cache_manager/utils.py
```diff
@@ -0,0 +1,204 @@
...
+        old_hash = self.cache_client._memoize_version(function_reference)
+
+        # change hash
+        self.cache_client.delete_memoized(function_reference)
+
+        # delete all keys connected with memoized function left in Redis
+        lua_script = "local keys = redis.call('KEYS', '*{0}') if #keys > 0 then redis.call('DEL', unpack(keys)) end"\
+            .format(old_hash[1])
+        lua_cleaner = self.cache_client.cache._client.register_script(lua_script)
+        lua_cleaner()
```
kaplun: What's the performance of this?

inspirehep/ext/cache_manager/manager.py
```diff
@@ -0,0 +1,176 @@
...
+        single_citations_to_invalidate.append(recid)
+        single_references_to_invalidate.append(recid)
+
+        # invalidate citations views of all record's references as they contain outdated version of the record
+        for ref_recid in references:
+            citations_to_invalidate.append(ref_recid)
+
+        # get all records that refer to changed record and invalidate their references views
+        # as they contain outdated version of the record
+        es_query = Query('refersto:' + str(recid)).search()
```
kaplun: We should use search/scroll here, since it's we have papers with more than 10000 citations/references.

kaplun: Just use plain ES APIs here (since we should substitute this with Neo4J once we attack it). In this way you can optimize a lot the query.

inspirehep/ext/cache_manager/manager.py
```diff
@@ -0,0 +1,176 @@
...
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+
+from invenio_search.api import Query
+
+from .utils import InspireCacheInvalidator
+
+
+class InspireCacheManager(object):
```
kaplun: All these class to me seems the only specific part to INSPIRE and could be reimplemented by using simple receivers listening to `after_record_create/update` signals.

In this way the rest of your code could be really made generic and contributed.

inspirehep/utils/citations.py
```diff
@@ -42,8 +44,7 @@ def citations(self):
 
         out = []
         row = []
-        recid = self.record['control_number']
-        es_query = Query('refersto:' + str(recid)).search()
+        es_query = Query('refersto:' + str(self.recid)).search()
```
kaplun: See identical comment above.

inspirehep/utils/references.py
```diff
@@ -75,3 +77,24 @@ def references(self):
...
+class AjaxReferences(object):
+
+    def __init__(self, recid, index):
+        self.recid = recid
+        self.index = index
+
+    @staticmethod
+    @inspire_cache.inspire_memoize(timeout=5184000)  # 60 days
+    def get_references(recid, index):
+        from invenio.base.globals import cfg
```
kaplun: `invenio_base`

inspirehep/manage.py
kaplun: But then what do you cache in `references`? The IDs of references? Or the whole HTML? In the latter case it's probably using too much memory and you will need to delete it too often for being worth.

inspirehep/base/views.py
tomaszgy: `get_references` is a memoize function and because of that there's a difference between passing `1196797` and `u'1196797`. They're treated as different arguments and consequently two different keys with exactly the same content are being saved to Redis. That's why casting to `unicode` is needed :)

kaplun: Sad! :)

jmartinm:
When using the cli, would be good to have more human friendly messages:

```
$ inveniomanage inspirehep
```

```
invalidate_cache    Invalidate cache of certain type e.g. brief, citation
                        etc.
```

Like this it is not possible to know what options I can provide.

```bash
$ inveniomanage inspirehep invalidate_cache
usage: Manage INSPIRE instance. invalidate_cache [-?] cache_name
Manage INSPIRE instance. invalidate_cache: error: too few arguments
```

You could do it with the `click` library to have some user input.

`Select cache to invalidate (all, brief, citations, references) [default all]`

jalavik:
Indeed you can use `@manager.option` instead of `@manager.command` to add help docs to your arguments.  See e.g. how `create_indices` command look in the same file.

inspirehep/ext/cache_manager/cache.py
tomaszgy: You're right, there's no need for that.

inspirehep/utils/citations.py
kaplun: @jmartinm I notice here we are deliberately limiting the number of citation to the first 9999 ones. Since anyway we are truncating the results, maybe we can simply gave up and provide the first 200 (or 500)? In this way the page is not unnecessarily bloated and one can always open a search page.

jmartinm: I think that makes sense. Nowadays we show 5 https://inspirehep.net/record/724189/citations and I don't think there have been requests to show more.

I think the code does not need to be modified much, we can limit the query to 10 results, each row will be outputted and on the client side we will display them (without datatables).

We also need to get the `len()` of the `refersto:` query to show the total number of citations.

I can do the modifications on the client side after we merge this PR.

@tomaszgy do you think making these changes will alter the way we do the caching?

jmartinm: screen shot 2016-02-05 at 4 29 41 pm

tomaszgy: @jmartinm, no it won't affect the way we do the caching. It is quite likely that less single-entry views will be cached after changes :smile:  Are we also going to display max 10 references of the record? Finally, would you prefer to do `len()` on the client side or rather have it provided within the AJAX response?

jmartinm: For the references you can leave it with all of them.

Can you check how it is done now so that we get the number when the page loads?

screen shot 2016-02-05 at 5 01 42 pm

So either we leave it as it is now, or yes, receiving it on the AJAX request sounds good.

tomaszgy:
The command line looks now like that:

screenshot from 2016-02-08 10 54 12

You can also do ```inveniomanage inspirehep invalidate_cache --cache=brief``` to get the same results.

inspirehep/ext/cache_manager/manager.py
```diff
@@ -0,0 +1,147 @@
...
+        self.memoized_invalidator = MemoizedInvalidator(cache)
+        self.cached_invalidator = CachedInvalidator(cache)
+        self.cache_client = cache
+        self.app = app
+        self.views = None
+        if app is not None:
+            self.init_app(app)
+
+    @staticmethod
+    def import_function_by_string(func_string):
```
jacquerie: This is already supported by Werkzeug's `import_string`: just use `:` as the last delimiter.

inspirehep/utils/citations.py
```diff
@@ -22,39 +22,44 @@
...
-
-        return out
+
+    return out
+
+
+def get_ajax_citations(recid):
+    """Jsonified citation export for single record in datatables format."""
+
+    return jsonify(
```
jalavik: Perhaps this should ideally be placed in the `views: ajax_citations` instead, so that it is clear in the views what is returned. That would make this function go away and instead the views will be:

```python
from inspirehep.utils.citations import render_citations
return jsonify({
    "data": render_citations(recid)
})
```
Same with `ajax_references`.

inspirehep/ext/cache_manager/cache.py
```diff
@@ -0,0 +1,146 @@
...
+                    if callable(unless) and unless() is True:
+                        return f(*args, **kwargs)
+
+                    try:
+                        cache_key = decorated_function.make_cache_key(f, *args, **kwargs)
+                        rv = self.cache.get(cache_key)
+                    except Exception:
+                        if current_app.debug:
+                            raise
+                        logger.exception("Exception possibly due to cache backend.")
```
jalavik: You should use the `logger` already available under  `current_app`. E.g. `current_app.logger.exception`. This is then logged to files and Sentry etc.

You can then kill the ``logger = logging.getLogger(__name__)`` at the top.

Also applies in the rest of the PR where this pattern is used.

inspirehep/ext/cache_manager/cache.py
```diff
@@ -0,0 +1,146 @@
...
+    The parameters in both inspire_memoize and inspire_cached are the same
+    as the parameters of the original methods in Flask Cache class
+
+    """
+
+    def inspire_memoize(self, timeout=None, make_name=None, unless=None):
+        def memoize(f):
+            @functools.wraps(f)
+            def decorated_function(*args, **kwargs):
+                if not cfg['DEBUG']:
```
jalavik: Minor thing, but there is an inconsistency in the checking of `DEBUG` mode. Here it is `cfg["DEBUG"]` and below `current_app.debug` is used. Ideally it should be one way only for consistency sake.

inspirehep/ext/cache_manager/manager.py
```diff
@@ -0,0 +1,145 @@
...
+        self.cache_client = cache
+        self.app = app
+        self.views = None
+        if app is not None:
+            self.init_app(app)
+
+    def init_app(self, app, cache):
+
+        if app:
+            if not hasattr(app, 'extensions'):
```
jalavik: Will this ever be the case with Flask that `app.extensions` is missing?

jalavik:
Thanks @tomaszgy! :+1:  Just gonna give it a test locally and see.

inspirehep/ext/cache_manager/cache.py
jacquerie: Is there a better option of the two? I've always had this doubt.

inspirehep/utils/citations.py
tomaszgy: Sure, I'll do it also for references.

inspirehep/ext/cache_manager/cache.py
tomaszgy: Right, thanks for pointing it out.

inspirehep/ext/cache_manager/cache.py
tomaszgy: We checked with @jacquerie that `current_app.debug` has better performance so I'll replace `cfg["DEBUG"]` with it.

jalavik: @tomaszgy Cool! Thanks for checking it out.

jacquerie: Here's an IPython session to prove this:
```python
In [1]: from invenio_base.globals import cfg

In [2]: from flask import current_app

In [3]: def from_cfg():
   ...:     return cfg['DEBUG']
   ...:

In [4]: def from_current_app():
   ...:     return current_app.config['DEBUG']
   ...:

In [5]: def from_flask():
   ...:     return current_app.debug
   ...:

In [6]: %timeit from_cfg()
The slowest run took 4.79 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 4.58 µs per loop

In [7]: %timeit from_current_app()
The slowest run took 5.86 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 3.25 µs per loop

In [8]: %timeit from_flask()
The slowest run took 5.13 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 3.72 µs per loop
```
(yay microbenchmarks!)

inspirehep/ext/cache_manager/manager.py
tomaszgy: Not sure but I've seen this pattern in few Flask extensions so I would guess it's possible. E.g. https://github.com/thadeusb/flask-cache/blob/master/flask_cache/__init__.py#L182

jalavik: Ah OK. Cool :)

inspirehep/config.py
```diff
@@ -596,3 +598,16 @@
 }
 """Updates default properties that should be added to the Holding Pen index
 mappings."""
+
+CACHED_VIEWS = {
+    'single_entry': {
```
jalavik: What is `single_entry`? The detailed record with references etc? Shall we call it `detailed`?

tomaszgy: No, it's a single entry in the `references` or `citations` box in detailed view of the record.

jalavik:
@tomaszgy BTW, tested and it runs fine for me :+1:

inspirehep/config.py
jalavik: Ah, I see. Shall we rename it? :dancer:  Say, `record_reference`  (reference/citation is the same?) or `record_oneline`. I am thinking it could be good to also rename `brief` to `record_brief` so that the full cache name becomes more clear `next::_template_fragment_cache_brief_1227953` vs `next::_template_fragment_cache_record_brief_1227953

tomaszgy:
@jalavik Thanks, good to hear :smile:

inspirehep/config.py
tomaszgy: Sure, why not :smile: So I'll change it to `record_oneline` and `record_brief`.

PR #801, opened 2016-02-01
requirements: forcing dojson 0.4.0

kaplun:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ requirements = [
     'invenio-webhooks==0.1.0',
     'invenio_classifier==0.1.0',
     'librabbitmq>=1.6.1',
+    'dojson==0.4.0',
 ]
 
 test_requirements = [


base.requirements.txt
```diff
@@ -1,4 +1,5 @@
 ## Base requirements
+dojson==0.4.0
```
jmartinm: shall we add here https://github.com/inspirehep/inspire-next/blob/master/setup.py#L32 instead?

kaplun: Is that line ever run? We don't do `pip install .` or python setup.py build`, at least on Travis

jmartinm: Good question. @jalavik ?

jalavik: It is picked up when `-e .` is run as part of `requirements.txt`, so I agree with @jmartinm  that it should be pinned in setup.py.

jalavik: requirements file entries should only be for git repo requirements.

PR #802, opened 2016-02-01
WIP search: suggestion messages

Panos512:
* Adds suggestion messages on search.

### DIFF
--- a/inspirehep/base/templates/search/results.html
+++ b/inspirehep/base/templates/search/results.html
@@ -70,13 +70,13 @@
   {% elif request.args.get('cc').lower() == "jobs" %}
     {% include 'search/facets/jobs.html' %}
   {% elif request.args.get('cc').lower() == "institutions" %}
-    {% include 'search/facets/institutions.html' %}  
+    {% include 'search/facets/institutions.html' %}
   {% elif request.args.get('cc').lower() == "journals" %}
     {% include 'search/facets/journals.html' %}
   {% elif request.args.get('cc').lower() == "experiments" %}
-    {% include 'search/facets/experiments.html' %}  
+    {% include 'search/facets/experiments.html' %}
   {% elif request.args.get('cc').lower() == "conferences" %}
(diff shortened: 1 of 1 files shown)


Panos512:
WIP preview:

screen shot 2016-02-02 at 10 29 33

kaplun:
Great! It's currently quite far from the query. Wonder if this can be displayed e.g. as a baloon pop up next to the search bar. 

cc: @jmartinm

inspirehep/base/templates/search/results.html
```diff
@@ -198,6 +220,28 @@
...
+    </ul>
+  </div>
+    {% endif %}
+    {% endwith %}
+    {% with messages = get_flashed_messages(category_filter=["query_suggestion_with_redirect"]) %}
+      {% if messages %}
+      <div class="alert alert-warning">
+      <ul class=flashes>
+        {% for message in messages %}
+        <li class="fa fa-exclamation-triangle">   {{ message }} Try: <a target="_blank" href='http://inspirehep.net/search?{{request.query_string}}'>inspirehep.net</a></li>
```
kaplun: Does `requests.query_string` need to be URL encoded? I guess Jinja will take care of escaping it for HTML but I am not sure it's correctly urlencoding it (e.g. spaces into %20 etc.)

jmartinm:
@kaplun Yes, that would be nice.

@Panos512 we already use http://codeseven.github.io/toastr/demo.html so you could place it just below the search bar with a closable message:

screen shot 2016-02-04 at 11 25 31 am

And to make it even nicer indeed it would be good if it has an arrow pointing to the search bar. You can try altering the css of `toastr` with something like http://www.cssarrowplease.com/ :)

Panos512:
Great! That looks much better! I'll get to it when I fix the `topcited` keyword.

jacquerie:
Superseded by #991, no? And it's targeted to `master`, so I guess this can be closed.

PR #811, opened 2016-02-02
WIP: tests: completes test views

### LINKED ISSUE #804: Complete test views
The file https://github.com/inspirehep/inspire-next/blob/master/tests/test_views.py currently contains tests for the `/detailed` pages. 

Tests should be added for other main views such as:

* [ ] Landing pages (/literature, /jobs, ...)
* [ ] /admin/holdingpen/list (shall we create some objects and make sure they appear in the list ?)
* [ ] /author/new
* [ ] /author/update?recid=997876 ( so that imports data from that recid, otherwise it redirects to /new )
* [ ] /doi/search
* [ ] /arxiv/search
* [ ] /search with some different searches, sorting options, ...
* [ ] /submit/literature/create

Any others?

ioannistsanaktsidis:
* Closes #804.

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -20,19 +20,112 @@
 """Tests for views"""
 
 from invenio.testsuite import InvenioTestCase
+from flask import url_for
 
 
 class ViewsTests(InvenioTestCase):
     """Tests views."""
+    def setUp(self):
+        from invenio.config import CFG_SITE_SECURE_URL
+        self.login = self.client.post(
+            url_for('webaccount.login', scheme='https'),
+            base_url=CFG_SITE_SECURE_URL,
+            data=dict(nickname='admin',
+                      password='admin'),
+            follow_redirects=True)
 
-    def test_detailed_records(self):
-        """Test visiting each detailed record returns 200"""
-        from invenio_records.models import Record
-        failing_recids = []
-        for recid in Record.allids():
-            try:
-                if self.client.get("/record/{recid}".format(recid=recid)).status_code != 200:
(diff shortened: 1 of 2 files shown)


tests/test_views.py
```diff
@@ -36,3 +36,117 @@ def test_detailed_records(self):
...
+
+    def test_literature_landing_page(self):
+        """Test visiting each landing page returns 200"""
+        failing_landing_page = []
+        try:
+            if self.client.get("/literature").status_code != 200:
+                failing_landing_page.append('/literature')
+        except Exception:
+            failing_landing_page.append('/literature')
+        self.failIf(failing_landing_page)
```
jalavik: Just a curiosity here, if an exception happens, will it be printed? (since it is caught before `failIf`)

ioannistsanaktsidis: It will print sth like this:
['/literature'] as the above test prints a list with all the failing Ids. Makes sense?

ioannistsanaktsidis:
@jmartinm this one fails due to the tests I added :P It is because of the flask toolbar the status code is 302 and then it redirects. Do we fix it here or in another PR?

jmartinm:
I think Travis runs the tests with DEBUG=False so no debug toolbar should be present. Maybe it is something else? For example, if the page requires login you will be redirected to the login page.

ioannistsanaktsidis:
True some of them require Login.

jacquerie:
> True some of them require Login.

Maybe then adding `TESTING = True` will work for you, since it bypasses the `@login_required` decorator.

Issue #817, opened 2016-02-03
Proofchecking queries


kaplun:
- [ ] `journal:` field (to actually use refextract to build corresponding query)
- [ ] `refersto:` nested support
- [ ] `citedby:` basic support (this need to bypass ES, and simply load all the records from the corresponding recid)
- [ ] `citedby:` nested support
- [ ] `eprint:` should understand both `arxiv:1007.5048` and without the `arxiv` prefix
- [ ] `fulltext:` should work
- [x] `author:"b storaci"`, `author:"b. storaci"` should be the same (see also https://inspirehep.net/info/hep/search-tips?ln=en#author)
- [ ] `author:B.Storaci.1` should work (assuming BAIs are fetched from HepNames)
- [ ] `topcite:200+` should return all the record with more than 200 citations (@panos512)
- [ ] `find r atlas-conf-*`: search by reportnumber (@panos512)
- [ ] `find cc italy`: search by country (or country code)
- [ ] `find caption "Diagram for the fermion flow violating process"` caption 
- [ ] `find tc book`: type codes (@panos512) (#970)
- [ ] `find fc a`: field codes (@panos512) (#971)

For reference see: https://inspirehep.net/info/hep/search-tips

kaplun:
cc: @jalavik, @jmartinm the above are documented searches on Legacy that do not work on QA. Many of them can not be addressed in the current milestone.

Panos512:
'refersto' and 'citedby' can popup a message that they are unsupported (with the flash technique), until we fix them.

I will check if we can quickly fix some of the others and come back for more.

StellaCh:
outdated

PR #821, opened 2016-02-03
WIP elasticsearch: new filter

Panos512:
* Adds a new filter to elasticsearch that is used to ignore special
  characters on author searching.

* Ticks the author bullet (no.7) of inspirehep/inspire-next#817.

### DIFF
--- a/inspirehep/base/searchext/mappings/hep.json
+++ b/inspirehep/base/searchext/mappings/hep.json
@@ -6,6 +6,16 @@
                     "type": "icu_folding",
                     "preserve_original": true
                 },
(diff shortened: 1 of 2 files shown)


kaplun:
Can you add a regression test to prove that indeed "b storaci" and "b. storaci" (or corresponding similar use case) are working as intended? In this way we are sure this behavior is not reverted (since we are going to clean up in the future the mapping configuration etc.

Panos512:
I added the test and the apostrophes filter. We can use this until we figure a better way for author searching

Issue #796, opened 2016-01-29
Support for Hidden/restricted collections


kaplun:
- [ ] Restrict hidden documents D0 CDF etc.
- [ ] Move hidden jobs (i.e. what to do when a record exist and should be moved to a different collection)

kaplun:
See also #919

jmartinm:
And see also https://github.com/inspirehep/inspire-next/pull/1116

StellaCh:
Is this resolved @jmartinm ?

PR #826, opened 2016-02-03
config: updated collection enhancer

jalavik:
* Enables the updated collection enhancer which filters records
  from the search interface by collection.

* Removes special Jobs Hidden mappings as it is shared with Jobs
  mappings. All JOB records now stay in the Jobs index as they
  share the mapping. The updated filtering takes care to filter
  based on `_collections` value in the records.

* Adds a sample mapping and restricted collection CDF Internal Notes.

Depends on https://github.com/inspirehep/invenio-search/pull/26

Addresses #796


kaplun:
:clap:

kaplun:
I guess this will fail on Travis since we have not yet merged inspirehep/invenio-search#26

kaplun:
Can you add: https://inspirehep.net/record/1403304/export/xm to the demo records and check that it does not searchable in HEP, but is searchable in `CDF Internal Notes`?

jalavik:
@kaplun I am working on tests now that does a similar thing. (changing an existing record)

But I can add that record too.

tests/test_indexing.py
```diff
@@ -19,13 +19,58 @@
...
+    def test_recids_restricted_collection(self):
+        """Test that recids API works"""
+        from invenio_ext.sqlalchemy import db
+        from invenio_search.api import Query
+        from invenio_records.api import Record
+
+        self.assertEqual(len(Query("").search(collection='CDF Internal Notes').recids), 0)
+
+        test_recid = 1396160
+        original_record = dict(Record.get_record(test_recid))
```
kaplun: You should do a deepcopy otherwise, since `rec['collection']` is a `list` it will be passed by reference. And at line 62 you would modify the original one.

tests/test_indexing.py
```diff
@@ -19,13 +19,58 @@
...
+
+        self.assertEqual(len(Query("").search(collection='CDF Internal Notes').recids), 0)
+        self.assertEqual(len(Query("").search(
+            collection='CDF Internal Notes',
+            user_info={'precached_permitted_restricted_collections': ['CDF Internal Notes']}
+        ).recids), 1)
+
+        rec = Record.get_record(test_recid)
+        rec.model.json = original_record
+        db.session.commit()
```
kaplun: Better put this last part in the`finally` of a  `try..finally`, so that we are sure it's always executed also in case of errors. Otherwise the test would live traces behind.

tests/test_records.py
```diff
@@ -26,13 +26,26 @@ class RecordsTest(InvenioTestCase):
 
     """Tests for the Records."""
 
+    def setUp(self):
+        from invenio_records.api import Record
+        self.test_recid = 1402176
+        self.original_record = dict(Record.get_record(self.test_recid))
```
kaplun: `deepcopy`

tests/test_indexing.py
```diff
@@ -19,13 +19,58 @@
...
+        rec['collections'].append({'primary': 'CDF-INTERNAL-NOTE'})
+        rec.commit()
+
+        # Wait for indexing
+        time.sleep(10)
+
+        self.assertEqual(len(Query("").search(collection='CDF Internal Notes').recids), 0)
+        self.assertEqual(len(Query("").search(
+            collection='CDF Internal Notes',
+            user_info={'precached_permitted_restricted_collections': ['CDF Internal Notes']}
```
kaplun: Woah! It still works? I guess we should add a #NOTE: to be amended for Invenio 3 :)

jalavik: Yeah, most of this with regards to collections needs probably to be amended. But yeah.

tests/test_indexing.py
jalavik: Hmm, it worked for me locally.

tests/test_indexing.py
jalavik: Will do!

tests/test_indexing.py
kaplun: I am never sure with SQLAlchemy, whether two calls to get_record for the same record aren't returning the same actual object reference.

jalavik: Yeha, I ran the tests twice to be sure and all is green :)  Pushing now.

PR #827, opened 2016-02-04
dojson: updates json schema

### LINKED ISSUE #741: Update hep-0.0.1.json schema.
After all the related model tasks are merged we should update the hep-0.0.1.json so as to keep the schema up to date.

ioannistsanaktsidis:
* Closes #741.

### DIFF
--- a/inspirehep/dojson/hep/schemas/hep-0.0.1.json
+++ b/inspirehep/dojson/hep/schemas/hep-0.0.1.json
@@ -245,6 +245,18 @@
             "type": "array",
             "description": "FIXME: Old 247, shall it still have its own field?"
         },
+        "edition": {
+            "uniqueItems": true,
+            "items": {
+                "type": "object",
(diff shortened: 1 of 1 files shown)


inspirehep/dojson/hep/schemas/hep-0.0.1.json
```diff
@@ -1007,9 +1035,12 @@
             }
         },
         "page_nr": {
-            "type": "string",
-            "description": "FIXME: shall it be integer perhaps?",
-            "title": "Number of pages"
+            "type": "array",
```
jalavik: Since this is an array now, I think the literature submission has a potential problem @jmartinm 

```shell
inspirehep/modules/deposit/workflows/literature.py
328:        if 'page_nr' not in metadata:
333:                        metadata['page_nr'] = int(pages[1]) - int(pages[0]) + 1
```

PR #828, opened 2016-02-04
search: topcite keyword

Panos512:
* Adds `topcite` and `cited` keyword support.

* Should be merged with inspirehep/invenio-query-parser#11 .

* Closes box no.9 of #817.

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -390,7 +390,9 @@ SEARCH_ELASTIC_KEYWORD_MAPPING = {
     "address": ["corporate_author"],
     "datecreated": ["creation_modification_date.creation_date"],
     "datemodified": ["creation_modification_date.modification_date"],
-    "recid": ["control_number"]
+    "recid": ["control_number"],
+    "cited": ["citation_count"],
+    "topcite": ["citation_count"]
 }
 
 FACETS_SIZE_LIMIT = 10


kaplun:
Can you add a test to prove that topcite works as it should? (also amending the version.py) Thanks!

jacquerie:
> Can you add a test to prove that topcite works as it should? (also amending the version.py) Thanks!

:+1:

Here's an integration test that does this:

```python
from invenio_testing import InvenioTestCase

import json


class SearchTests(InvenioTestCase):

    def _search(self, p):
        url = '/search?cc=hep&sf=earliest_date&of=id&p={p}'.format(p=p)
        data = self.client.get(url).data
        result = map(int, json.loads(data))

        return result

    def test_topcite(self):
        self.assertTrue(self._search('find topcite 3'))
```

jacquerie:
In fact, the same idea could be used to recast all queries in #817 (and in general, all queries we want to keep track for regressions) as failing tests.

kaplun:
@Panos512 I think this code hasn't been integrated into Invenio3. Can you open a new PR against Invenio3 branch, and then we close this one?

Panos512:
@kaplun I think it's already merged ( https://github.com/inspirehep/inspire-next/pull/946 )

PR #829, opened 2016-02-04
tests: completes test views

### LINKED ISSUE #804: Complete test views

ioannistsanaktsidis:
* Closes #804.

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -34,3 +34,80 @@ class ViewsTests(InvenioTestCase):
             response = self.client.get("/record/{recid}".format(recid=recid))
             if response.status_code != 200:
                 raise Exception(response, recid)
+
+    def test_login(self):
+        """Test visiting login page returns 200"""
+        self.assertEqual(self.login('admin', 'admin').status_code, 200)
+
+    def test_literature_landing_page(self):
+        """Test visiting literature landing page returns 200"""
+        self.assertEqual(self.client.get("/literature").status_code, 200)
+
+    def test_conferences_landing_page(self):
+        """Test visiting conferences landing page returns 200"""
+        self.assertEqual(self.client.get("/conferences").status_code, 200)
+
+    def test_jobs_landing_page(self):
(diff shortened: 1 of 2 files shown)


tests/test_views.py
```diff
@@ -24,13 +24,88 @@
 
 class ViewsTests(InvenioTestCase):
     """Tests views."""
+    def setUp(self):
+        self.login('admin', 'admin')
```
jalavik: I suspect the login/logout part may slow down the tests a bit, so if you do not need it for every test then we can just call it in the few tests that do.

ioannistsanaktsidis:
cc @jalavik

jmartinm:
It seemed to work with the changes to `install.sh` right? cc @jalavik

jalavik:
Yeah, it does. Great. @ioannistsanaktsidis So you can remove the commented out parts completely and is the plan to keep the detailed record check uncommented?

jmartinm:
> Yeah, it does. Great. @ioannistsanaktsidis So you can remove the commented out parts completely and is the plan to keep the detailed record check uncommented?

Yes, I guess `test_detailed_records` was commented out to speed up this test? I think it should be left in the tests (although it is a bit slow)

ioannistsanaktsidis:
Exactly the test_detailed was commented for speed reasons. I will amend the commit and we are good to go.

PR #832, opened 2016-02-05
utils: new formula utils

kaplun:
* Introduces new LaTeX/MathML formula utilities, based on a MathJax
  server running on NodeJS.

### DIFF
--- a/inspirehep/utils/formulas.py
+++ b/inspirehep/utils/formulas.py
@@ -1,0 +1,189 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""
+Utilities to handling LaTeX/MathML formulas.
+"""
+
+
+import re
+
+import requests
+
+from flask import current_app
+
+from HTMLParser import HTMLParser
+
+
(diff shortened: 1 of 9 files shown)


inspirehep/utils/formulas.py
```diff
@@ -0,0 +1,180 @@
...
+            count -= 1
+        if count < 0:
+            return False
+    return count == 0
+
+
+def tex2mml(formula):
+    """Converts a LaTeX formula to MathML."""
+    try:
+        mathjax_node_server = current_app.config['MATHJAX_NODE_SERVER']
```
jalavik: Could be simplified with
```python
current_app.config.get('MATHJAX_NODE_SERVER', MATHJAX_NODE_SERVER)
```

inspirehep/utils/tex2mml.js
```diff
@@ -0,0 +1,63 @@
+#! /usr/bin/env node
+
+// This file is part of INSPIRE.
+// Copyright (C) 2015 CERN.
```
jalavik: 2016

inspirehep/utils/tex2mml.js
```diff
@@ -0,0 +1,63 @@
...
+            format: ("inline-TeX"),
+            mml:true,
+        }, function (data) {
+            if (!data.errors) {
+                response.write(data.mml);
+            }
+        });
+    response.end();
+  });
+}).listen(48999);
```
jacquerie: Maybe make this configurable?

inspirehep/utils/formulas.py
kaplun: The issue was to protect against working outside of the context.
`current_app.config.get('MATHJAX_NODE_SERVER', MATHJAX_NODE_SERVER)` would still raise `RuntimeError` outside of the context.

inspirehep/utils/tex2mml.js
kaplun: Sure. I just need to learn a bit of JS and NodeJS first :smile:

tests/test_formulas.py
```diff
@@ -0,0 +1,72 @@
...
+        tokens = {
+            u'Bc+',
+            u'Bc+→J/ψDs+',
+            u'ψDs∗+',
+            u'J/ψDs+',
+            u'J',
+            u'Bc+→J/ψDs∗+',
+            u'J/ψDs∗+',
+            u'ψDs+'}
+        self.assertEquals(formulas.get_all_unicode_formula_tokens_from_text(text), tokens)
```
jacquerie: Deprecated, use `self.assertEqual` instead (https://docs.python.org/2/library/unittest.html#deprecated-aliases).

inspirehep/utils/formulas.py
jalavik: I see. Is it normal that it would run outside of the context? Do we want it to?

kaplun: Probably not :-) (since it depends on an available, configurable server)

kaplun:
Mmh. Our Docker image uses an old NodeJS. I will update it with the latest version 5.

PR #835, opened 2016-02-05
filters: number of records optimization

jmartinm:
* Uses information about the query stored in session instead of performing
  a query every time a facet is selected.

### DIFF
--- a/inspirehep/base/templates/search/results.html
+++ b/inspirehep/base/templates/search/results.html
@@ -148,7 +148,7 @@
     <div class="panel-body" >
       <div class="row">
         <div class="col-md-9">
-          <span class="text-left {{ request.args.get('cc')|lower() }}" id="clear-facets-message">{{ pagination.total_count }} of {{ request.args.get('p')|number_of_search_results(request.args.get('cc')) }} results match the filters you selected.</span><br/>
+          <span class="text-left {{ request.args.get('cc')|lower() }}" id="clear-facets-message">{{ pagination.total_count }} of {{ request.args.get('p', '')|number_of_search_results(collection.name) }} results match the filters you selected.</span><br/>
           <small id="clear-facets-info">Clear filters to see all results.</small>
         </div>
         <div class="col-md-3 text-right">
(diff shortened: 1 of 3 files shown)


jmartinm:
To be merged with https://github.com/inspirehep/invenio-search/pull/29

jmartinm:
@jalavik if you have a moment, can you have a look?

As when a user is logged in the session is permanent and there is no possibility to add a timeout to a `session.get` I have added a timestamp to the session value and only use the session when it is newer than 5 minutes.

inspirehep/base/templates/search/results.html
```diff
@@ -148,7 +148,7 @@
     <div class="panel-body" >
       <div class="row">
         <div class="col-md-9">
-          <span class="text-left {{ request.args.get('cc')|lower() }}" id="clear-facets-message">{{ pagination.total_count }} of {{ request.args.get('p')|number_of_search_results(request.args.get('cc')) }} results match the filters you selected.</span><br/>
+          <span class="text-left {{ request.args.get('cc')|lower() }}" id="clear-facets-message">{{ pagination.total_count }} of {{ request.args.get('p', '')|number_of_search_results(collection.name) }} results match the filters you selected.</span><br/>
```
jalavik: Just in case, we are always sure that the `collection` object is available when the template renders right? Referring to the past issues we've had with the 500 page etc.

inspirehep/ext/jinja_filters/record.py
```diff
@@ -396,5 +396,24 @@ def citation_phrase(count):
 
     @app.template_filter()
     def number_of_search_results(query, collection_name):
+        """
+        Filter used to show total number of results out of filtered ones.
+        """
+        from flask import session
+        import datetime
```
jalavik: Minor thing, these import should be safe to keep at module level to avoid hitting these lines every time the filter is used (even if they will quickly return as it is already imported, but yeah).

jalavik:
Looks good overall :+1:

inspirehep/base/templates/search/results.html
jmartinm: Yes, this page is rendered by the view in `search.py` which adds `collection` to the context so it should always be there.

jmartinm:
Thanks, comments addressed.

PR #838, opened 2016-02-05
dojson: tests upgrade and deleted fields

jalavik:
Adds missing handling + tests of 970__d and 981__a in dojson.

Significantly improves testing speed of `test_hep` from 23s to 2.3s by using pytest and fixtures. @kaplun @jacquerie @jmartinm


jacquerie:
:+1:

Sorry I missed this! This is exactly the direction I want to take with the tests: using `InvenioTestCase` adds unnecessary overhead. The convenience of `setup_class` can easily be obtained with fixtures. `InvenioTestCase` should probably only be used when we want to unit test Invenio mechanisms like registries and configuration, and when we write integration tests (which are allowed to be slow anyway).

But unit tests should be fast!

PR #842, opened 2016-02-08
utils: date utils addition

jalavik:
* Improves the date validation functions that is used e.g. in
  earliest_date receiver. (Taken from INSPIRE PROD.)

* Moves the date functions from DoJSON utilities to a general
  utility.

Fixes date issues upon ES indexing for multiple records.

### DIFF
--- a/inspirehep/utils/date.py
+++ b/inspirehep/utils/date.py
@@ -1,0 +1,89 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Helpers for handling dates."""
+
+import six
+
+from invenio_utils.date import strftime, strptime   # pre-1900 friendly parser
+
+DATE_FORMATS_YEAR = ["%Y", "%y"]
+DATE_FORMATS_MONTH = [
+    "%Y-%m", "%Y %b", "%b %Y", "%Y %B", "%B %Y",
+    "%y-%m", "%y %b", "%b %y", "%y %B", "%B %y",
+]
+DATE_FORMATS_FULL = [
+    "%Y-%m-%d", "%d %m %Y", "%x", "%d %b %Y",
+    "%d %B %Y", "%d %b %y", "%d %B %y",
+]
+
+
+def create_earliest_date(dates):
+    """Return the earliest valid date from a list of date strings."""
+    if not dates:
+        return None
+    if not isinstance(dates, (list, set)):
+        dates = [dates]
+    valid_dates = []
+    for date in dates:
+        # Add 99 to the end of partial date to make sure we get the REAL
+        # earliest date
+        valid_date = create_valid_date(date,
+                                       date_format_month="%Y-%m-99",
+                                       date_format_year="%Y-99-99")
+        if valid_date:
+            valid_dates.append(valid_date)
+    if valid_dates:
+        date = min(valid_dates)
+        return date.replace('-99', '')
+
+
+def create_valid_date(date, date_format_full="%Y-%m-%d",
+                      date_format_month="%Y-%m", date_format_year="%Y"):
+    """Iterate over possible formats and return a valid date if found."""
+    valid_date = None
+    date = six.text_type(date)
+    for format in DATE_FORMATS_FULL:
+        try:
+            valid_date = strftime(date_format_full, (strptime(date, format)))
+            break
+        except ValueError:
+            pass
+    else:
+        for format in DATE_FORMATS_MONTH:
+            try:
+                if date.count('-') > 1:
+                    date = "-".join(date.split('-')[:2])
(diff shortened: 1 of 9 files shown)


inspirehep/utils/date.py
```diff
@@ -0,0 +1,90 @@
...
+    "%y-%m", "%y %b", "%b %y", "%y %B", "%B %y",
+]
+DATE_FORMATS_FULL = [
+    "%Y-%m-%d", "%d %m %Y", "%x", "%d %b %Y",
+    "%d %B %Y", "%d %b %y", "%d %B %y",
+]
+
+
+def create_valid_dates(dates):
+    """Add '01' as a value of missing month and/or day values."""
```
kaplun: The issue with adding `01` suffixes is that it introduces fake information that it's later trusted.

E.g. take this:
Say a document has dates: 2016 and 2016-05. What is its earliest date? With the proposed implementation is 2016-01-01, because of the fake `-01-01` being introduced. However it would very likely be more probable that the real earliest date is `2016-05-01`.  (2016 being just an unfortunate truncation due to how meta data is dealt with). I believe it's important to consider this info, otherwise it will look like that all papers receive citations only at the first of January of every year.

In order to achieve this, on legacy, we were not touching the actual metadata, but for when it was really necessary: 2016 was stored as 2016, and only at runtime, when the question "earliest_date" was asked, it was converted temporary to:

`2016-99-99` so that then it could have been compared with e.g. `2016-05-99`. Then the earliest date is clearly `2016-05-99`, which contains this canary value `-01` that can be converted at the end to `-01` giving the desired `2016-05-01`.

kaplun: Basically I'd propose that ES and formats are made compatible with partial dates such as `2016` and `2016-05`, so that we don't introduce fake information, that might impact further computations and presentations.

jalavik: :+1:

tests/test_date_utils.py
```diff
@@ -17,16 +17,25 @@
...
+    assert create_valid_date('1977') == '1977'
+    assert create_valid_date('1977-06-02') == '1977-06-02'
+    assert create_valid_date('1977-06-022') == '1977-06'
+    assert create_valid_date('1977-06-220') == '1977-06'
+
+
+def test_create_earliest_date():
+    """Test the date validator that excepts list."""
+    assert create_earliest_date([1877, '2002-01-05']) == '1877'
+    assert create_earliest_date(['1877-02-03', '1877']) == '1877-02-03'
```
kaplun: :+1: 

What about `create_earliest_date(['1877-02', '1877'])` ? I believe your code is returning: `1877-02`, without any specification for the day. Are we happy for this?

What ES is doing when sorting two records with `earliest_date` respectively of `1877-02` and `1877-02-03`? I guess either case, it's not dramatic :smile: .

inspirehep/utils/date.py
```diff
@@ -0,0 +1,89 @@
...
+    "%y-%m", "%y %b", "%b %y", "%y %B", "%B %y",
+]
+DATE_FORMATS_FULL = [
+    "%Y-%m-%d", "%d %m %Y", "%x", "%d %b %Y",
+    "%d %B %Y", "%d %b %y", "%d %B %y",
+]
+
+
+def create_earliest_date(dates):
+    """Add '01' as a value of missing month and/or day values."""
```
kaplun: The docstring is no longer up-to-date with the return statement.

inspirehep/utils/date.py
```diff
@@ -0,0 +1,89 @@
...
+        try:
+            valid_date = strftime(date_format_full, (strptime(date, format)))
+            break
+        except ValueError:
+            pass
+    else:
+        for format in DATE_FORMATS_MONTH:
+            try:
+                if date.count('-') > 1:
+                    date = "-".join(date.split('-')[:2])
```
kaplun: Are you trying here to workaround malformed dates? Aren't we risking to correct automatically a totally malformed date that would really need to be curated by a human?

kaplun:
Beside my few question, this code is great!

inspirehep/utils/date.py
jalavik: Yeah, so it depends on the strategy we apply. I was discussing with @jmartinm in that we should really start validating the records with the schema and report any inconsistencies even before any of this "correcting" code is run. So I was thinking of adding a "non-breaking" schema validation to the migrator so that bad metadata will be logged but the record is still migrated and finally indexed. E.g. by adding some new columns to the prod migration model: `valid=True/False` and text with errors.

This code is now executed upon indexing with the goal of creating an indexable document. So one could in many cases save some part of a date that is malformed (say `2012-03-022`) would be returned as `2012-03`, until the date is fixed by a human after receiving the report from the schema validation.

PS: There may perhaps even exist some "correcting" code in the dojson that ideally should be reporting the bad metadata as well.

tests/test_date_utils.py
jalavik: In the current ES config, dates are described as:

```
"date": {
    "type": "date",
    "format": "yyyy||yyyyMM||yyyyMMdd||yyyyMMddHHmmss||yyyyMMddHHmmss.S||dd MM yyyy||dd MMM yyyy||MMM yyyy||MMM yyyy?||yyyy ('repr'.1964.)"
},
```
which I plan to change everywhere to:
```
"date": {
    "type": "date",
    "format": "yyyy||yyyy-MM||yyyy-MM-dd"
},
```
except for the special `date_and_time_of_latest_transaction`

jalavik: To answer you question, so yeah, we could return `01`, but as you said that it would perhaps be a bad assumption to make, perhaps it is better to not add it?

inspirehep/utils/date.py
jalavik: Thanks, will fix.

kaplun:
Actually looks like it is of paramount importance to have this branch and fix the default ranking "newest first", (which behind should actually use the earliest date field). This is because currently, due to certain records having broken year (such as 2093, or similar), we are producing embarrassing results:

https://qa.inspirehep.net/search?cc=HEP&p=&sf=earliest_date&so=desc&rg=25
sso1

jalavik:
@kaplun Sure things. I'll have this branch ready tomorrow. In addition I have a branch almost ready with proper schema validation as warnings. PR incoming tomorrow.

kaplun:
:+1: (a bit late)

jalavik:
@kaplun No probs. It should fix the indexing issues and the future dates.

PR #843, opened 2016-02-08
global: remove upgrade scripts

jacquerie:
As discussed IRL, past upgrade scripts are not run during the build, so they can be safely removed.

Skipping CI to avoid clogging the queue for more useful stuff. As soon as this is merged, coverage is expected to go up by ~5%.


jalavik:
@jacquerie My bad for not being fully clear earlier, but unfortunately we cannot remove many of these **yet**. As I realize that most still needs to be run on production labs :/. I will give you a list of those we can remove.

jacquerie:
Oh! I see. This is not high prority anyway, since it's just a coverage quick win, so for me this can be closed and the discussion postponed to when we run those scripts.

jalavik:
You can safely remove:
```
inspirehep/modules/workflows/upgrades/inspire_2014_11_13_update_actions.py
inspirehep/modules/workflows/upgrades/inspire_2015_08_13_update_record_instances.py
inspirehep/modules/workflows/upgrades/workflows_2015_05_19_innodb_engine_and_keys.py 
```
On the other hand, we can also think about if it is better to just exclude upgrades from coverage reports instead of deleting them.

jacquerie:
Arguments for removing: code that is not run should be deleted, since deleted code can be recovered from git history if needed. Arguments for keeping but excluding from coverage: documentation, and to save time if code must be retrieved frequently from git history.

I'm leaning towards deletion, maybe @kaplun and @jmartinm want to weigh in?

kaplun:
I am ok with deleting because anyway, when we will move to Invenio 3, we are also very likely go for PostgreSQL. At that time no update will save us.

jalavik:
@jacquerie @kaplun Fair points. I am okay with deleting.

jmartinm:
Ok for me.

jacquerie:
Not really relevant anymore.

PR #844, opened 2016-02-08
inspirehep: improve reindexing performance

kaplun:

### DIFF
--- a/inspirehep/manage.py
+++ b/inspirehep/manage.py
@@ -24,14 +24,21 @@
 
 from __future__ import print_function
 
-import click
 import json
 import sys
-
 from time import sleep
-
+import click
+from elasticsearch.helpers import streaming_bulk
+from invenio_base.globals import cfg
 from invenio_ext.es import es
 from invenio_ext.script import Manager
+from invenio_records.models import RecordMetadata
+from invenio_records.signals import before_record_index
+from invenio_records.tasks.index import get_record_index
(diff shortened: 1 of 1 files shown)


kaplun:
cc @jalavik

inspirehep/manage.py
```diff
@@ -64,7 +65,19 @@ def reindexhp():
 def reindex():
     """Reindex all records."""
     from invenio_records.models import RecordMetadata
-    from invenio_records.tasks.index import index_record
+    from invenio_records.tasks.index import get_record_index
+    from invenio_ext.es import es
+    from invenio_records.signals import before_record_index
+    from invenio_base.globals import cfg
```
jacquerie: Import inside of functions hinder testability, since they are harder to mock. I'd say that we should do this only when we would get circular imports (but that should be a reason to refactor).

Can you try putting as many of these in the outermost scope as possible?

kaplun: Good point

jalavik:
@kaplun Cool! Do you notice a big speedup?

kaplun:
Ahem... Thanks for asking... it's actually... slower!

kaplun:
Although the bottleneck is the computation of citations, that can also be done in bulk at the end. I will do it. as well.

kaplun:
To be redone (if needed on Invenio3)

PR #850, opened 2016-02-09
travis: parallel testing

kaplun:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,11 @@ language: python
 python:
   - 2.7
 
+env:
+  - PYTEST="tests/test_views.py"
+  - PYTEST="tests/test_workflows.py"
+  - PYTEST="--ignore tests/test_views.py --ignore tests/test_workflows.py"
+
 before_install:
   - "travis_retry pip install --upgrade pip"
   - "travis_retry pip install check-manifest mock twine wheel coveralls"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -31,6 +31,11 @@ language: python
 python:
   - 2.7
 
+env:
+  - PYTEST="tests/test_views.py"
+  - PYTEST="tests/test_workflows.py"
+  - PYTEST="--ignore tests/test_views.py --ignore tests/test_workflows.py"
```
jacquerie: I don't like the harcoded test filenames. A better solution (IMO) is to separate the tests in `integration` and `unit` folders, and run those separately.

kaplun: Ouch :dancer: I have just seen it after I clicked merge...

jacquerie: No worries! I will fix this eventually, as long as `Bibtex` doesn't kill me first...

kaplun: BTW: sure. This was just a quick thing to gain 5 minutes in all the next testing. On the other hand, Travis testing is as slow as the slowest of its matrix component, so if you break up unit testing and integration testing only, the bottle neck will still be integration. We would still need to break it up a bit.

jacquerie: Sure! Just for laughs, the problem of splitting them in the optimal way is NP-complete: https://en.wikipedia.org/wiki/Partition_problem, so maybe we shouldn't sweat it...

PR #874, opened 2016-02-10
templates: strip plot caption

### LINKED ISSUE #870: Remove leading numbers from plot captions
All plot captions have a trailing number:

screen shot 2016-02-09 at 5 27 41 pm

Does this have any meaning? Shall we remove it on display or at the source?

jmartinm:
* Removes leading internal number in plot captions. (closes #870)

### DIFF
--- a/inspirehep/base/templates/format/record/Inspire_HTML_detailed_macros.tpl
+++ b/inspirehep/base/templates/format/record/Inspire_HTML_detailed_macros.tpl
@@ -333,7 +333,7 @@
               {% if url.get('url') %}
                 {% if url.get('url').endswith(".png") or url.get('url').endswith(".jpg") %}
                 <div id="slide-content-{{ loop.index0 }}">
-                  <span>{{ url.get('description') }}</span>
+                  <span>{{ url.get('description')|strip_leading_number_plot_caption }}</span>
                 </div>
                 {% endif %}
               {% endif %}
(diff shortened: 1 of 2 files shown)


inspirehep/base/templates/format/record/Inspire_HTML_detailed_macros.tpl
```diff
@@ -333,7 +333,7 @@
               {% if url.get('url') %}
                 {% if url.get('url').endswith(".png") or url.get('url').endswith(".jpg") %}
                 <div id="slide-content-{{ loop.index0 }}">
-                  <span>{{ url.get('description') }}</span>
+                  <span>{{ url.get('description')|strip_leading_number_plot_caption() }}</span>
```
jalavik: As in the other PR I wonder if filters are okay to not call?

inspirehep/ext/jinja_filters/record.py
```diff
@@ -424,6 +424,11 @@ def long_timeout(timeout):
         return 5184000  # 60 days
 
     @app.template_filter()
+    def strip_leading_number_plot_caption(text):
+        import re
```
jalavik: `re` is already imported at the top of the file.

jmartinm: Oh no! Good catch, will fix that.

PR #876, opened 2016-02-10
dojson: guarantee order when removing duplicates

jacquerie:
Introduces two new utils, `remove_duplicates_from_list` and `remove_duplicates_from_list_of_dicts`. They do as it says on the tin, except they also preserve the original ordering, something which is useful to have during testing.

Not having that was causing failures such as https://travis-ci.org/inspirehep/inspire-next/jobs/108116178#L1952. Testing that this bug won't occur again is outside of the scope of this PR; maybe in a future PR?

### DIFF
--- a/inspirehep/dojson/utils.py
+++ b/inspirehep/dojson/utils.py
@@ -104,3 +104,35 @@ def strip_empty_values(obj):
         return type(obj)(new_obj)
     else:
         return obj
+
+
+def remove_duplicates_from_list(l):
+    """Remove duplicates from a list preserving the order.
+
+    We might be tempted to use the list(set(l)) idiom,
+    but it doesn't preserve the order, which hinders
+    testability."""
+    result = []
+
+    for el in l:
+        if el not in result:
+            result.append(el)
+
+    return result
+
+
+def remove_duplicates_from_list_of_dicts(ld):
+    """Remove duplicates from a list of dictionaries preserving the order.
+
+    We can't use the generic list helper because a dictionary isn't
+    hashable. Taken from http://stackoverflow.com/a/9427216/374865."""
+    result = []
+    seen = set()
+
+    for d in ld:
+        t = tuple(d.items())
+        if t not in seen:
+            result.append(d)
+            seen.add(t)
+
+    return result
--- a/inspirehep/dojson/hep/fields/bd01x09x.py
+++ b/inspirehep/dojson/hep/fields/bd01x09x.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of INSPIRE.
-# Copyright (C) 2014, 2015 CERN.
+# Copyright (C) 2014, 2015, 2016 CERN.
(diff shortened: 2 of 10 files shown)


jacquerie:
I forgot to mention that @ioannistsanaktsidis pointed out that https://github.com/inspirehep/inspire-next/blob/b2c7bd2b3e8e565f252a6097aa50d3b4fcb8f93d/inspirehep/ext/jinja_filters/record.py#L94-L102 contains a similar helper. If anybody else needs this helper somewhere else maybe it's time to move it in the outermost possible "scope" (https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

inspirehep/dojson/utils.py
```diff
@@ -104,3 +104,38 @@ def strip_empty_values(obj):
...
         return obj
+
+
+def remove_duplicates_from_list(l):
+    """Remove duplicates from a list preserving the order.
+
+    We might be tempted to use the list(set(l)) idiom,
+    but it doesn't preserve the order, which hinders
+    testability."""
+    seen = set()
```
jalavik: Do you need the set here at all?

```python
In [1]: l = ['foo', 'baz', 'foo']

In [2]: clean = []

In [3]: for s in l:
   ...:     if s not in clean:
   ...:         clean.append(s)
   ...:         

In [4]: clean
Out[4]: ['foo', 'baz']
```

ioannistsanaktsidis: @jalavik this guy here has the same question :P 
https://pythonadventures.wordpress.com/2013/10/31/remove-duplicates-from-a-list-and-keep-the-original-order-of-the-elements/

jalavik: I get the speed issue, but here we are usually dealing with small lists, **unless** we are talking about authors.

jacquerie: And @jalavik is right! I was thinking _performance_, but it's actually **worse**:

```
In [1]: import random

In [2]: def test_with_set():
   ...:     l = random.sample(range(5), 3)
   ...:     seen = set()
   ...:     result = []
   ...:     for el in l:
   ...:         if el not in seen:
   ...:             result.append(el)
   ...:             seen.add(el)
   ...:     return result
   ...:

In [3]: def test_without_set():
    l = random.sample(range(5), 3)
    result = []
    for el in l:
        if el not in result:
            result.append(el)
    return result
   ...:

In [4]: %timeit test_with_set()
The slowest run took 5.11 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 4.11 µs per loop

In [5]: %timeit test_without_set()
The slowest run took 4.30 times longer than the fastest. This could mean that an intermediate result is being cached.
100000 loops, best of 3: 3.72 µs per loop
```

(yay microbenchmarks!)

jalavik: The set checking probably wins if you up the number of samples to the tens of thousands, but yeah.

PR #878, opened 2016-02-10
Workflows and OAI harvester adjustments

jalavik:
Adjustments and bug fixes after new versions of workflows and oaiharvester.


kaplun:
Is this for this milestone?

jalavik:
@kaplun Yeah it is required for us to deploy current state.

jalavik:
There is a circular dependency in the testing now that also causes #882 to fail. I will just merge them now as it is pretty clear that those are the only failing parts.

PR #881, opened 2016-02-11
Mobile display improvements

nikpap:
- [x] Collections links moved to mobile menu.
- [x] Feedback button hidden from mobile.
- [x] Exporting from search results hidden from mobile.
- [x] Full width controls in top of search results.
- [x] Full height mobile menu - no scroll bar.
- [x] No scroll when mobile menu is open.
- [x] Mobile menu icons.
- [x] Resources not working correctly in mobile.
- [x] Detailed view references.
- [x] Cite modal arrow turned wrong.

capture d ecran 2016-02-19 a 15 47 29


jmartinm:
screen shot 2016-02-12 at 1 59 27 pm

All the export functionality can be hidden in mobile.

That would avoid so much whitespace and have more compact results.

jmartinm:
Feedback should move out of the way

screen shot 2016-02-12 at 2 00 42 pm

Either we kill it completely or it moves to the footer of the page.

jmartinm:
For the top part, the Cite button can be hidden.

screen shot 2016-02-12 at 2 01 54 pm

The rest of the controls should take full width instead of being compressed in a column.

jmartinm:
I think it will be better if the whole top menu opens up and there is no scroll inside.

screen shot 2016-02-12 at 2 05 17 pm

jmartinm:
In detailed view, the references box will need a bit of love

screen shot 2016-02-12 at 2 07 28 pm

The 'Show X entries' and the search could be side by side.

Avoid the pagination being cut.

If possible reduce column width for citation count.

jmartinm:
screen shot 2016-02-12 at 2 09 26 pm

It seems like in mobile view, some CSS selector is making all arrows turn.

That should not affect the cite modal for example.

inspirehep/base/templates/header.html
```diff
@@ -28,6 +28,8 @@
 ## structure: (endpoint, values, blueprint, caption)
 #}
 {% set navigation_bar = [
+  ('/tools/', {}, '', 'Tools'),
+  ('/links-stats/', {}, '', 'Links and stats'),
```
nikpap: The idea is that tools, links & stats will be moved to their own pages.

nikpap:
Moved to Invenio3: https://github.com/inspirehep/inspire-next/pull/902

PR #885, opened 2016-02-11
records: impact graph API addition

jmartinm:
* Adds API to retrieve information regarding references and citations
  for a given record. Needed by impact graph visualization.

* Adds new record utils for record-related utilities.

* Adds get_title util.

* Adds appropriate tests for the added functionality.

### DIFF
--- a/inspirehep/modules/records/views.py
+++ b/inspirehep/modules/records/views.py
@@ -1,0 +1,105 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from flask import Blueprint, jsonify
+
+from invenio_base.decorators import wash_arguments
+from invenio_ext.es import es
+from invenio_search.api import Query
+
+from inspirehep.utils.record import get_title
+
+
+blueprint = Blueprint(
+    'inspire_records',
+    __name__,
+    url_prefix='/record',
+    template_folder='templates',
+    static_folder="static",
+)
+
+
+@blueprint.route('/api/impactgraph', methods=['GET'])
+@wash_arguments({'recid': (unicode, "")})
+def impactgraph(recid):
+    """Handler for impact graph information retrieval."""
+    record = es.get_source(index='hep', id=recid, doc_type='record')
+
+    out = {}
+
+    # Add information about current record
+    out['inspire_id'] = record['control_number']
+    out['title'] = get_title(record)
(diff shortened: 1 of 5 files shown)


inspirehep/modules/records/views.py
```diff
@@ -0,0 +1,103 @@
...
+
+    # Get citations
+    citations = []
+
+    es_query = Query('refersto:' + recid).search()
+    es_query.body.update(
+        {
+            'size': 9999
+        }
+    )
```
jacquerie: Mh, I've seen this pattern used so many times that maybe it's time to create a helper.

jmartinm: Something to check, not sure at the moment where would be a good place for such utility :)

tests/test_impactgraph_api.py
```diff
@@ -0,0 +1,47 @@
...
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Tests for Impact Graph API."""
+
+import json
+import os
+import pkg_resources
+
+from invenio.testsuite import InvenioTestCase
```
jacquerie: invenio_testing?

inspirehep/modules/records/views.py
jalavik: As it is technically a lower level API detail, shouldn't `invenio-search` API already take care of this limitation instead of the caller?

kaplun: But Invenio-search doesn't know about references... Do you mean just the possibility to override the size?

jalavik: @kaplun Exactly. Just the `size` part :)

tests/test_impactgraph_api.py
jmartinm:
```bash
$ git grep invenio_testing | wc -l
8
```

```bash
$ git grep invenio.testsuite | wc -l
      25
```

Going with the masses.

inspirehep/modules/records/views.py
jmartinm: We leave it for Invenio 3? :)

tests/test_impactgraph_api.py
jalavik: Both works, `invenio.testsuite` is aliased to `invenio_testsuite`.

inspirehep/modules/records/views.py
jalavik: Sure!

tests/test_impactgraph_api.py
jmartinm: > Both works, invenio.testsuite is aliased to invenio_testsuite.

You mean `invenio_testing`?

I remember in the past `invenio_testing` did not have all the functionality ported so we were using `invenio.testuite` instead - which is this one https://github.com/inspirehep/invenio/tree/labs/invenio/testsuite

jmartinm:
A couple of issues that need fixing in this PR (and tests):

* Some records do not have `earliest_date` field (so accessing a citation will give keyError)
* If there are no references, [this line](https://github.com/inspirehep/inspire-next/pull/885/files#diff-914a66753d90550c09fb9280e499ec23R79) will fail. Iterating None.

jmartinm:
Also check in QA records https://qa.inspirehep.net/record/api/impactgraph?recid=554898, https://qa.inspirehep.net/record/api/impactgraph?recid= 1374543

inspirehep/modules/records/views.py
```diff
@@ -0,0 +1,105 @@
...
+            "inspire_id": citation['control_number'],
+            "citation_count": citation['citation_count'],
+            "title": get_title(citation),
+            "year": citation['earliest_date'].split('-')[0]
+        })
+
+    out['citations'] = citations
+
+    # Get references
+    record_references = record.get('references')
```
jacquerie: `record.get('references', [])`, then?

jacquerie:
Also, the impact graph data generation currently takes ~20 seconds on this record: https://qa.inspirehep.net/record/api/impactgraph?recid=4328. This could be a consequence of the hardware limitations of the QA machine, though.

kaplun:
@jmartinm I am closing this as superseded by #1044. If not the case please reopen it.

PR #887, opened 2016-02-12
dojson: fixes data model validation errors

### LINKED ISSUE #872: Data model validation errors
After #869 several validation errors is visible within our demo records:

- [ ]  not `uniqueItems` in `titles`
```
2016-02-09 15:08:52,242 WARNING: ValidationError: Record 1319638: [{'source': 'arXiv', 'title': 'Precision luminosity measurements at LHCb'}, {'source': 'arXiv', 'title': 'Precision luminosity measurements at LHCb'}] has non-unique elements

Failed validating u'uniqueItems' in schema[u'properties'][u'titles']:
    {u'items': {u'properties': {u'source': {u'type': u'string'},
                                u'subtitle': {u'type': u'string'},
                                u'title': {u'type': u'string'}},
                u'type': u'object'},
     u'type': u'array',
     u'uniqueItems': True}

On instance[u'titles']:
    [{'source': 'arXiv',
      'title': 'Precision luminosity measurements at LHCb'},
     {'source': 'arXiv',
      'title': 'Precision luminosity measurements at LHCb'}]
```

- [ ]  not integer in `references.number`
```
2016-02-09 15:08:46,662 WARNING: ValidationError: Record 1317658: '1' is not of type u'integer'

Failed validating u'type' in schema[u'properties'][u'references'][u'items'][u'properties'][u'number']:
    {u'type': u'integer'}

On instance[u'references'][0][u'number']:
    '1' [in /Users/jlavik/Envs/next/src/inspire/inspirehep/modules/migrator/tasks.py:356]
```

- [ ]  bad data from 999C6, `array` instead of `string`.
```
2016-02-09 15:08:39,800 WARNING: ValidationError: Record 1313115: ['Invenio/1.1.2.1260-aa76f refextract/1.5.44', 'content.pdf;1'] is not of type u'string'

Failed validating u'type' in schema[u'properties'][u'refextract'][u'items'][u'properties'][u'version']:
    {u'type': u'string'}

On instance[u'refextract'][0][u'version']:
    ['Invenio/1.1.2.1260-aa76f refextract/1.5.44', 'content.pdf;1'] [in /Users/jlavik/Envs/next/src/inspire/inspirehep/modules/migrator/tasks.py:356]
```

There are more as well to add as we go.

ioannistsanaktsidis:
* Closes #872.

### DIFF
--- a/inspirehep/dojson/hep/fields/bd01x09x.py
+++ b/inspirehep/dojson/hep/fields/bd01x09x.py
@@ -130,15 +130,23 @@ def dois2marc(self, key, value):
 
 
 @hep.over('external_system_numbers', '^035..')
-@utils.for_each_value
-@utils.filter_values
 def external_system_numbers(self, key, value):
     """System Control Number."""
-    return {
-        'value': value.get('a'),
-        'institute': value.get('9'),
-        'obsolete': value.get('z'),
-    }
+    value = utils.force_list(value)
+
+    def get_value(value):
+        return {
+            'value': value.get('a'),
+            'institute': value.get('9'),
+            'obsolete': bool(value.get('z')),
+        }
+
+    external_system_numbers = self.get('external_system_numbers', [])
+
+    for val in value:
+        external_system_numbers.append(get_value(val))
+    return inspire_dojson_utils.remove_duplicates_from_list_of_dicts(
+        external_system_numbers)
 
 
 @hep2marc.over('035', 'external_system_numbers')
@@ -156,6 +164,8 @@ def external_system_numbers2marc(self, key, value):
 @hep.over('report_numbers', '^037..')
 def report_numbers(self, key, value):
     """Source of Acquisition."""
+    value = utils.force_list(value)
+
     def get_value(value):
         return {
             'source': value.get('9'),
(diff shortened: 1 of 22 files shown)


inspirehep/dojson/hep/fields/bd90x99x.py
```diff
@@ -32,15 +32,36 @@
...
 def references(self, key, value):
     """Produce list of references."""
+    recid = ''
+    number = ''
+    year = ''
+    if '0' in value:
+        try:
+            recid = int(value.get('0'))
+        except:
+            # Some crap in they recid :-(
```
jacquerie: Err...

ioannistsanaktsidis: Still in WIP, copy paste line, it will be replaced.

jmartinm:
:+1: Go @ioannistsanaktsidis !

inspirehep/dojson/hep/fields/bd90x99x.py
kaplun: Right, can you amend it to use the emoji :crap:. It's definitively sweeter :smile:.

(my fault sorry. but the input is really bad sometimes...)

ioannistsanaktsidis:
Some records fail because the abstract field does not have source, which we set as required in the schema. Is it a problem if I remove the source from required ?

jmartinm:
> Some records fail because the abstract field does not have source, which we set as required in the schema. Is it a problem if I remove the source from required ?

I think you can go ahead.

annetteholtkamp:
The source is definitely not mandatory

- Annette

ioannistsanaktsidis:
There is only one small thing left and we are good to go :+1:

ioannistsanaktsidis:
Waiting to become green and we are good to go.

inspirehep/dojson/hep/fields/bd01x09x.py
```diff
@@ -130,15 +130,24 @@ def get_value(val):
...
+    def get_value(value):
+        return {
+            'value': value.get('a'),
+            'institute': value.get('9'),
+            'obsolete': bool(value.get('z')),
+        }
+
+    external_system_numbers = self.get('external_system_numbers', [])
+
+    if isinstance(value, list):
```
jalavik: To avoid duplication of errors I think we can avoid the `isinstance` pattern by just using `utils.force_list` and always treat as list. WDYT?

ioannistsanaktsidis: Sounds good. I will check it out and amend the commit.

ioannistsanaktsidis:
Good to go!

jalavik:
Thanks @ioannistsanaktsidis !

PR #888, opened 2016-02-12
less: make modal headers consistent

jacquerie:
As mentioned in https://github.com/inspirehep/inspire-next/pull/599#issuecomment-183371344, here's a PR to make all modal headers consistently blue.

BTW, @nikpap, can I remove this `10px` margin? https://github.com/inspirehep/inspire-next/blob/287381506bdd46bc0809c701b2a194be1ccff697/inspirehep/base/static/less/format/detailed-record.less#L129-L131 It's the only inconsistency I found in the headers I checked.

### DIFF
--- a/inspirehep/base/static/less/base/modals.less
+++ b/inspirehep/base/static/less/base/modals.less
@@ -1,8 +1,6 @@
 /**
- * -*- mode: text; coding: utf-8; -*-
- *
  * This file is part of INSPIRE.
- * Copyright (C) 2014, 2016 CERN.
+ * Copyright (C) 2016 CERN.
 
  * INSPIRE is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


jmartinm:
> BTW, @nikpap, can I remove this 10px margin? https://github.com/inspirehep/inspire-next/blob/287381506bdd46bc0809c701b2a194be1ccff697/inspirehep/base/static/less/format/detailed-record.less#L129-L131 It's the only inconsistency I found in the headers I checked.

You can go ahead and remove it. The rule is not being applied anyway :)

screen shot 2016-02-15 at 9 28 39 am

kaplun:
Is this PR still needed, or can it be closed?

jacquerie:
This would need to be rewritten to SASS anyway (which is an easy rewrite!), so I guess this can be closed.

PR #889, opened 2016-02-15
authors: fix full name in bief view

jacenkow:
* Updates record.get('name.value') since not every record
  has a preferred_name field.


jacenkow:
@jmartinm

jmartinm:
@jacenkow as we spoke today, we can leave this and the final template will be done in the `invenio3` branch.

PR #892, opened 2016-02-18
receivers: fixes document type bug

### LINKED ISSUE #891: document type bug
When a record contains the collection 'Thesis' the document type should be `thesis`. Currently it is peer reviewed (https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/receivers.py#L85) which is wrong.

In the case of `Published` we will rename it to `peer reviewed`

ioannistsanaktsidis:
* Closes #891.


ioannistsanaktsidis:
cc @jmartinm

PR #839, opened 2016-02-07
tests: unit test utils

jacquerie:
The objective of this PR is to have 100% unit test coverage for the files in the `inspirehep/utils` folder.

This entails a complete code review of these files, which will result in just comments: future PRs will address (or delete) these. The only changes that will happen to the code are changes to improve its testability: moving imports in the outer scope, injecting dependencies, and so on.

I'm currently testing too much: a lot of these tests will be deleted, so please don't start reviewing.

PR #903, opened 2016-02-23
export: ports formatter in invenio3

ioannistsanaktsidis:

### DIFF
--- a/inspirehep/modules/formatter/config.py
+++ b/inspirehep/modules/formatter/config.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+FORMATTER_EXPORT_LIMIT = 5000
+"""Formatter prefix added before all variables."""
--- a/inspirehep/modules/formatter/views.py
+++ b/inspirehep/modules/formatter/views.py
@@ -44,7 +44,7 @@ blueprint = Blueprint(
 @blueprint.route('/bibtex', methods=['GET', ])
 def get_bibtex():
     recid = request.values.get('recid', 0, type=int)
-    pid = PersistentIdentifier.get('recid', str(recid))
+    pid = PersistentIdentifier.get('literature', str(recid))
     record = es.get_source(
         index='records-hep', id=str(pid.object_uuid), doc_type='hep')
     bibtex = Bibtex(record).format()
@@ -54,7 +54,7 @@ def get_bibtex():
 @blueprint.route('/latex', methods=['GET', ])
 def get_latex():
     recid = request.values.get('recid', 0, type=int)
-    pid = PersistentIdentifier.get('recid', str(recid))
+    pid = PersistentIdentifier.get('literature', str(recid))
     latex_format = request.values.get('latex_format', '', type=unicode)
     record = es.get_source(
         index='records-hep', id=str(pid.object_uuid), doc_type='hep')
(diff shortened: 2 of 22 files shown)


jmartinm:
When you port the formats from `utils` like `Bibtex` remember to port as well the tests to make sure they are working correctly.

ioannistsanaktsidis:
:+1:

inspirehep/modules/formatter/config.py
```diff
@@ -0,0 +1,23 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+FORMATTER_EXPORT_LIMIT = 5000
```
jmartinm: We should add a docstring below every config variable to explain what they are used for. Like

```python
COLLECTIONS_CACHE_KEY = 'DynamicCollections::'
"""Key prefix added before all keys in cache server."""
```

inspirehep/modules/formatter/views.py
```diff
@@ -0,0 +1,127 @@
...
+
+from invenio_ext.es import es
+
+from invenio_base.decorators import wash_arguments
+
+from six import text_type
+
+from invenio_base.globals import cfg
+
+import itertools
```
jmartinm: Check this https://pypi.python.org/pypi/isort in order to sort the imports properly.

inspirehep/modules/formatter/views.py
```diff
@@ -0,0 +1,127 @@
...
+from inspirehep.utils.cv_latex import Cv_latex
+from inspirehep.utils.cv_latex_html_text import Cv_latex_html_text
+
+from invenio_ext.es import es
+
+from invenio_base.decorators import wash_arguments
+
+from six import text_type
+
+from invenio_base.globals import cfg
```
jmartinm: this is now `current_app.config`

jacquerie:
> When you port the formats from utils like Bibtex remember to port as well the tests to make sure they are working correctly.

Uhm, maybe this can be avoided? I could resubmit #839 rebased on top of `invenio3` + @ioannistsanaktsidis's branch instead.

ioannistsanaktsidis:
Amended the latest changes. Note that after fixing the of= parameter and the invenio_knowledge there are some parts in the code that we need to uncomment. I have placed a note there.

ioannistsanaktsidis:
About the output formats, I noticed that some of them are really outdated, e.g hca.yml and some others. They are still  using .bft files . I think of keeping them out of the new. Suggestions ?

ioannistsanaktsidis:
@jmartinm this one contains both my changes as well as the test from @jacquerie . Thats why we have 2 commits to be merged. Cheers.

jmartinm:
For all these export formats, I think the best is going to be to use the `RECORDS_REST_ENDPOINTS` [configuration](https://github.com/inspirehep/inspire-next/blob/invenio3/inspirehep/config.py#L83).

There, one can define for each export type a serializer:

```
RECORDS_REST_ENDPOINTS = dict(
    literature=dict(
        pid_type='literature',
        pid_minter='inspire_recid_minter',
        pid_fetcher='inspire_recid_fetcher',
        search_index='records-hep',
        search_type='hep',
        record_serializers={
            'application/json': ('invenio_records_rest.serializers'
                                 ':json_v1_response'),
            'application/marcxml+xml': (
                'inspirehep.modules.records.serializers.marcxml_v1_response'),
            'application/x-cv+text': (
                'inspirehep.modules.records.serializers.cv_text_v1_response'),
            'application/x-bibtex': (
                'inspirehep.modules.records.serializers.bibtex_v1_response'),

        },
        search_serializers={
            'application/json': ('invenio_records_rest.serializers'
                                 ':json_v1_search'),
            'application/vnd+inspire.brief+json': ('inspirehep.modules.records.serializers'
                                                   ':json_brief_v1_search'),
        },
        list_route='/literature/',
        item_route='/literature/<pid_value>',
        default_media_type='application/json',
        max_result_window=10000,
        query_factory_imp='inspirehep.modules.search.query:inspire_query_factory',
    ),
```

The logic of the exports (bibtex and so on) can be reused.

By the way, this PR also ports the `yml` files. This was part of the `invenio-formatter` which does not exists anymore so we don't need to port them.

Shall we open a new issue for creating the export formats in the Invenio 3 way and close this one in the meantime?

jmartinm:
There is this issue already actually https://github.com/inspirehep/inspire-next/pull/903

ioannistsanaktsidis:
We can do that. So should I open a new one that contains only the export of the formats or I can remove the unnecessary files from this one?

jmartinm:
The starting point would be to create a PR where you add just the serializers so that if you do a request to `/api/literature` (for a general search) or `/api/literature/123` for a given record with content type `application/x-bibtex` the bibtex is returned.

You can create a PR for each output format.

Then we will use that API from places like detailed record, angularjs search page and so on to retrieve the export formats when needed.

ioannistsanaktsidis:
:+1:

ioannistsanaktsidis:
I will keep this one open as a reference, because I started porting the formats in the new brief results using angular js. In the end this one will be closed and I will open a new one, adding also @jacquerie 's tests.

jmartinm:
Superseded by https://github.com/inspirehep/inspire-next/pull/1024 , right @ioannistsanaktsidis ?

ioannistsanaktsidis:
@jmartinm :+1:

ioannistsanaktsidis:
We can close this one as we handiling the exports in a different PR.

PR #904, opened 2016-02-23
theme: hogan dependency fix

jalavik:
* Fixes hogan npm dependency to `hogan.js` instead of `hogan`.

* Adds JS bundles to be excluded from other bundles to config
  variable `THEME_BASE_BUNDLES_EXCLUDE_JS`.

### DIFF
--- a/inspirehep/modules/theme/ext.py
+++ b/inspirehep/modules/theme/ext.py
@@ -67,4 +67,7 @@ class INSPIRETheme(object):
 
     def init_config(self, config):
         """Initialize configuration."""
-        pass
+        from .bundles import js
+        # Set JS bundles to exclude for purpose of avoiding double jQuery etc.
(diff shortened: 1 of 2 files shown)


inspirehep/modules/theme/ext.py
```diff
@@ -67,4 +67,7 @@ def init_app(self, app, assets=None, **kwargs):
 
     def init_config(self, config):
         """Initialize configuration."""
-        pass
+        from .bundles import js
+        # Set JS bundles to exclude for purpose of avoiding double jQuery etc.
+        # when other modules are building their JS bundles.
+        config.setdefault("THEME_BASE_BUNDLES_EXCLUDE_JS", js.contents)
```
jmartinm: Where is this config variable taken into account?

PR #906, opened 2016-02-24
Restore of homepage design

nikpap:


jmartinm:
Can you remove as well the Tools and resources ?

As we have there links to internal searches.

PR #909, opened 2016-02-24
global: authors and forms modules

jmartinm:
* Introduces authors module.

* Introduces forms module. This module is used to build forms using WTForms.

* FIXME authors addition form needs to be added and workflows added.


jacquerie:
Uh, this didn't go through Travis? Not a big deal, but I wonder why.

PR #912, opened 2016-02-25
authors: workflows porting

jmartinm:


jmartinm:
Ongoing porting of author addition and update forms. Now with basic workflows integration in update form.

Still missing porting generic workflows tasks for sending RT tickets and robotuploads, among others.

Also the author addition form needs to be ported.

PR #914, opened 2016-02-25
RFC git: gitignore update

jacquerie:
When I run
```shell
$ python setup.py test
```
on my machine I get an `.eggs` folder which then I have to avoid committing/delete. Since I got tired of doing that, I decided to update `.gitignore`: it pulls all the latest changes from the one recommended by GitHub (https://github.com/github/gitignore/blob/master/Python.gitignore), and retains all our modifications.

I only removed `bin/`, since we are not producing binaries anymore (we use the instance's CLI), `bower.json`, since we are not using Bower anymore, `**/static/vendors/**` and `**/static/gen/**`, since assets generation now happens in the instance folder.

Have a look if you want something added/removed, in case it conflicts with your development environment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,20 @@
+# Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
+*$py.class
+
+# C extensions
 *.so
-bin/
+
+# Distribution / packaging
+.Python
+env/
 build/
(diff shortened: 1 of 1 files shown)


jmartinm:
Looks good to me. We might not need the line about Django though.

PR #897, opened 2016-02-22
theme: add postfeedback handler

jacquerie:
Also adds the `app` fixture, which can be injected in the tests that need access to the application context.

Similar fixtures for DB, ES... will be ported from Zenodo (https://github.com/zenodo/zenodo/blob/ca7f630c45ff01ccf0cb84587407e23a060a7879/tests/unit/conftest.py) as soon as they are needed.

PR #917, opened 2016-02-25
utils: make_user_agent_string addition

jmartinm:
* Adds conftest.py with app fixture.

### DIFF
--- a/inspirehep/utils/url.py
+++ b/inspirehep/utils/url.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Helpers for handling with http requests and URL handling."""
+
+from __future__ import absolute_import, print_function
+
+from flask import current_app
+
+from inspirehep import __version__
+
+
+def make_user_agent_string(component=""):
+    """Return a nice and uniform user-agent string to be used by INSPIRE."""
+    ret = "InspireHEP-{0} (+{1};)".format(
+        __version__,
+        current_app.config.get('SERVER_NAME', ''),
+    )
+    if component:
+        ret += " [{}]".format(component)
+    return ret
--- a/tests/test_url.py
+++ b/tests/test_url.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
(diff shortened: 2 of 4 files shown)


jmartinm:
@jacquerie this includes the `conftest.py`

jmartinm:
Oh, and it will conflict with https://github.com/inspirehep/inspire-next/pull/897/files#diff-484462fced51d1a06b1d93b4a44dd535R1 so I can merge that one first and then apply my changes.

inspirehep/utils/url.py
```diff
@@ -0,0 +1,40 @@
...
+"""Helpers for handling with http requests and URL handling."""
+
+from __future__ import absolute_import, print_function
+
+from flask import current_app
+
+
+def make_user_agent_string(component=""):
+    """Return a nice and uniform user-agent string to be used by INSPIRE."""
+    from inspirehep import __version__
```
jacquerie: Is it possible to move it to the outer scope?

tests/test_url.py
```diff
@@ -0,0 +1,30 @@
...
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+from inspirehep.utils.url import make_user_agent_string
+
+
+def test_user_agent_string(app):
+    """Test that user agent is created."""
+    user_agent = make_user_agent_string()
+    assert "InspireHEP" in user_agent
```
jacquerie: I know this sounds overkill, but could you mock a realistic `__version__` and `SERVER_NAME` and assert exact equality? So that we have an example of realistic output.

tests/test_url.py
```diff
@@ -0,0 +1,30 @@
...
+
+    user_agent_with_component = make_user_agent_string("mycomponent")
+    assert "InspireHEP" in user_agent_with_component
+    assert "mycomponent" in user_agent_with_component
```
jacquerie: Same here: could you use a realistic component name and assert exact equality?

inspirehep/utils/url.py
```diff
@@ -0,0 +1,40 @@
...
+
+    ret = "InspireHEP-{0} (+{1};)".format(
+        __version__,
+        current_app.config.get('SERVER_NAME', ''),
+    )
+    if component:
+        ret += " %s" % component
```
jacquerie: ...maybe some square brackets around the component, (upper/lower)cased?

jacquerie:
Apologies for the very nitpicky comments!

jmartinm:
@jacquerie I actually did not find a way to mock the `__version__` used here https://github.com/inspirehep/inspire-next/pull/917/files#diff-333edcbb5824c49454f5a725d5b325dcR35 any suggestion?

jacquerie:
> any suggestion?

This should work:
```python
import mock

@mock.patch('inspirehep.utils.url.__version__', '0.1.0dev0')
def test_make_user_agent_string(app):
    ...
```

It's a little tricky to figure out _where_ to mock. This post helped me: http://alexmarandon.com/articles/python_mock_gotchas/#patching-in-the-wrong-place

tests/test_url.py
```diff
@@ -0,0 +1,35 @@
...
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+from inspirehep.utils.url import make_user_agent_string
+
+
+def test_user_agent_string(app):
+    """Test that user agent is created."""
+    # Set SERVER_NAME that will be used in user-agent
```
jacquerie: You need to create an application context to modify its configuration! The fixture doesn't put you in one.

inspirehep/utils/url.py
```diff
@@ -0,0 +1,40 @@
...
+"""Helpers for handling with http requests and URL handling."""
+
+from __future__ import absolute_import, print_function
+
+from flask import current_app
+
+from inspirehep import __version__
+
+
+def make_user_agent_string(component=""):
```
jacquerie: Oh, now I realize! If we have a mutable as a default argument then it is initialized only once, so `component` will be whatever we put there the first time.

So we can't do this. We'd have to pass `component=None`, in fact, since `if component` will evaluate to `False`.

tests/test_url.py
jmartinm: It works for me actually like this. I am getting the expected result.

jmartinm: Nevermind, it's true it does not work like this ;)

jmartinm:
@jacquerie this is more a testing exercise than a feature :)

jacquerie:
> @jacquerie this is more a testing exercise than a feature :)

:+1: 

One last comment and I think we're good.

PR #933, opened 2016-02-26
schema: schema fix

Panos512:
* Alters schema to make records importable.

### DIFF
--- a/inspirehep/modules/records/jsonschemas/records/hep.json
+++ b/inspirehep/modules/records/jsonschemas/records/hep.json
@@ -683,9 +683,6 @@
                     "id": {
                         "type": "string"
                     },
-                    "set": {
-                        "type": "array"
-                    },
                     "previous_set": {
                         "type": "string"
                     }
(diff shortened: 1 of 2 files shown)


inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -683,9 +683,6 @@
                     "id": {
                         "type": "string"
                     },
-                    "set": {
```
jmartinm: I think this field is in the `dojson` definition so being imported?

If the problem is that the value is an array or viceversa maybe we can do  `"type": ["array", "string"]`

Ideally it can be fixed on import time https://github.com/inspirehep/inspire-next/blob/invenio3/inspirehep/dojson/common/base.py#L109 to make sure the field has a string inside (or an array) - @kaplun do you know the type of this field?

kaplun: Could be multiple. Actually I wonder if this has to be ported at all... This is used by OAI-PMH server to decide which records to expose for which set. I think this is going to be overhauled by `inveniosoftware/invenio-oaiserver`. 

What about dropping the `oai_pmh` field entirely everywhere?

Panos512: I didn't know we could have multiple types! Yeah, this will solve the problem and it will leave `set` there.

Issue #918, opened 2016-02-26
Porting testsuite


jmartinm:
All the tests in `master` branch should be ported and pass on Travis.

* [x] test_author_bibfield.py
* [ ] test_author_workflow.py
* [ ] test_authors.py
* [x] test_bai.py (#1088)
* [x] test_bibfield.py
* [x] test_bibtex.py (#1024)
* [ ] test_cache_manager.py
* [x] test_citation_counts.py (#1036)
* [x] test_cvlatex.py (#1024)
* [x] test_cvlatex_html_text.py (#1024)
* [x] test_date_utils.py
* [x] test_datefilter.py
* [x] test_dojson_receivers.py (#1088)
* [x] test_export.py (#1024)
* [ ] test_formulas.py (depends on #1165)
* [x] test_hep.py
* [ ] test_hep_submission.py
* [x] test_hepnames.py
* [ ] test_holdingpen.py
* [x] test_icu.py (#1088)
* [ ] test_indexing.py
* [x] test_latex.py (#1024)
* [ ] test_matching.py
* [ ] test_migrator_utils.py
* [ ] ~~test_publication_info.py~~ (Done in #967 with what I think is a better approach)
* [ ] test_records.py
* [ ] test_reindexing.py
* [x] test_robotupload.py
* [x] test_text.py
* [ ] ~~test_utils.py~~ (IMO test belongs [here](https://github.com/inspirehep/invenio-utils/pull/2))
* [ ] ~~test_version.py~~ (Logic to be tested is no longer there, and isn't worth the effort to port)
* [x] test_views.py
* [ ] test_workflows.py

jacquerie:
You can assign this to me, so (I think) I can start ticking off boxes : )

kaplun:
@jacquerie are there one or more PR associated with this issue?

jacquerie:
Definitely #1024 and #1036, but probably more. I haven't been very diligent in ticking off boxes : /

kaplun:
D'oh! Can you refer to this issue then in the commit log of the respective PRs?

jmartinm:
I think this issue is not relevant any more. Tests have been ported as needed and new ones have been created. @jacquerie can you confirm and close?

PR #934, opened 2016-02-26
authors: name variations

Panos512:
* Adds name_variations field to the records.

### DIFF
--- a/inspirehep/modules/authors/receivers.py
+++ b/inspirehep/modules/authors/receivers.py
@@ -22,12 +22,23 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
-"""INSPIRE authors."""
 
-from __future__ import absolute_import, print_function
+from invenio_indexer.signals import before_record_index
 
-from .ext import INSPIREAuthors
+from .utils import author_tokenize
 
-from .receivers import *
 
-__all__ = ('INSPIREAuthors', )
+@before_record_index.connect
+def generate_name_variatons(recid, json, *args, **kwargs):
+    """Populate a json record before indexing it.
+    Adds a field for all the possible variations of an authors name
+    :param recid: The id of the record that is going to be indexed.
+    :param json: The json representation of the record that is going to be
+                 indexed.
+    """
+    authors = json.get("authors")
+    if authors:
+        for author in authors:
+            name = author.get("full_name")
+            if name:
+                author.update({"name_variations": author_tokenize(name)})
--- a/inspirehep/modules/authors/utils.py
+++ b/inspirehep/modules/authors/utils.py
@@ -3,25 +3,27 @@
 # This file is part of INSPIRE.
 # Copyright (C) 2015, 2016 CERN.
 #
-# INSPIRE is free software; you can redistribute it and/or
-# modify it under the terms of the GNU General Public License as
-# published by the Free Software Foundation; either version 2 of the
-# License, or (at your option) any later version.
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
 #
-# INSPIRE is distributed in the hope that it will be useful, but
-# WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-# General Public License for more details.
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


inspirehep/modules/authors/receivers.py
```diff
@@ -0,0 +1,43 @@
...
+
+@before_record_index.connect
+def generate_name_variatons(recid, json, *args, **kwargs):
+    """Populate a json record before indexing it.
+    Adds a field for all the possible variations of an authors name
+    :param recid: The id of the record that is going to be indexed.
+    :param json: The json representation of the record that is going to be
+                 indexed.
+    """
+    from inspirehep.modules.authors.utils import author_tokenize
```
kaplun: I think you can move this import outside.

kaplun:
Can you recover the corresponding tests from the master branch (or add some if missing?)

Panos512:
@jacquerie has already ported the corresponding test on his localhost so it will be up soon.

inspirehep/modules/authors/receivers.py
Panos512: done!

kaplun:
Uh!? So the tests would arrive afterwards?

Panos512:
Yeah because Jacopo is constructing the new tests structure and has already ported some tests from the invenio2 version.

tests/unit/authors/authors_test_variations.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
```
jacquerie: Filename should be `tests/unit/authors/test_authors_utils.py`.

tests/unit/authors/authors_test_variations.py
```diff
@@ -0,0 +1,53 @@
...
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
```
jacquerie: Please add CERN paragraph.

tests/unit/authors/authors_test_variations.py
```diff
@@ -0,0 +1,53 @@
...
+
+from inspirehep.modules.authors.utils import author_tokenize
+
+
+def test_name_variations():
+    name = "John Richard Ellis"
```
jacquerie: This sounds a little undertested. "John Richard Ellis" is quite standard, as far as names go. Maybe @kaplun can provide more interesting test cases?

jacquerie:
The main problem of this PR is that it depends on Beard, for which we have no clean installation mechanism now (we need to build the Docker image). So this will have to wait : /

kaplun:
Wait: why it depends on beard?

kaplun:
Ah! For the little: `from beard.utils.strings import asciify` in `authors.utils`? Well that function could be easily factored out.

jacquerie:
> For the little: `from beard.utils.strings import asciify` in `authors.utils`?

Yep. We could make that import local in `bai` with a big `FIXME`, though.

kaplun:
But actually we should really factor out all  the beard-utiliy into a separate package that doesn't have any major dependency... (OK a bit OT here). So for the time being I guess @Panos512 can indeed temporary factor out this specific function in this PR

jacquerie:
This has tests for the function that is called, but could benefit from some unit tests to the receiver. Take some ispiration by @mihaibivol's tests here: https://github.com/inspirehep/inspire-next/blob/19921ac71621a8e1ff89ae114fc6864d0193c811/tests/unit/records/test_records_receivers.py

kaplun:
@jacquerie, @Panos512 what is the status of this PR?

inspirehep/modules/authors/utils.py
```diff
@@ -66,6 +68,9 @@ def bai(name):
         last_name = names[0]
         initials = []
 
+    # FIXME: This should be loaded on the outer level when travis is updated.
+    from beard.utils.strings import asciify
```
kaplun: @Panos512 can you cut & paste from `beard` the `asciify` function into `utils.py` as in:

```python
try:
    from beard.utils.strings import asciify
except ImportError:
    def asciify(...):
         ....
```

In this way we don't depend necessarily on `beard` but if it's there we use it :-)

jacquerie: Yeah, this solution is much better than moving the import as I suggested. :+1:

jacquerie: Actually, why don't you put it into `inspirehep/modules/authors/utils.py`?

jacquerie:
Let's raise this PR from the dead, since I need it for #918. Will close and refile against `master`.

Issue #927, opened 2016-02-26
Port invenio-knowledge module


jmartinm:
We currently use `invenio_knowledge` in different components:

https://github.com/inspirehep/inspire-next/search?utf8=%E2%9C%93&q=invenio_knowledge

The module can be made Invenio 3 ready so that we can use it again.

kaplun:
* [ ] Port `CODENS` KB to use the journal collection
* [ ] Port `DEPOSIT_INSPIRE_SUBJECTS_KB` to a file resource
* [ ] Port `DEPOSIT_INSPIRE_DEGREE_KB` to a file resource
* [ ] Port `DEPOSIT_INSPIRE_JOURNALS_KB` to the journal collection (this is going to be done also on Legacy!)

kaplun:
cc: @ioannistsanaktsidis

kaplun:
@jacenkow @ioannistsanaktsidis has someone worked on it?

kaplun:
Actually: `DEPOSIT_INSPIRE_SUBJECTS_KB` and `DEPOSIT_INSPIRE_DEGREE_KB` is going to be moved by @tomaszgy into sub jsonschemas.

ioannistsanaktsidis:
I started implementing some parts which I needed for Bibtex, latex , but in the end I realised that I didn't need them so I dropped it.

kaplun:
@ioannistsanaktsidis `CODENS` where needed in BibTeX and LaTeX conversion. How are you doing at the moment then?

ioannistsanaktsidis:
Here it is https://github.com/ioannistsanaktsidis/inspire-next/blob/invenio3-formatter-with-angular/inspirehep/utils/bibtex.py#L668-L675 . Stiil in an open PR through. I will need @jacquerie tests to confirm 100% it is working.

kaplun:
OK I guess we can close this particular issue as we are anyway not porting KBs, rather reimplementing each time a particular solution.

PR #936, opened 2016-02-26
format: HEP detailed record external links

nikpap:
- [x] Resolve issue with strings vs list of dictionaries in url's - check recid 1393760.
- [x] URL's
- Articles from SCOAP3: eg 1393760
- [x] `external_system_numbers` fields: ids from other services
- check 1280200, 1388555
- [x] Crosscheck with current production code
- [x] Comma-separate values
- [x] Resolve abbreviations to strings eg `description`
- Check 1230937
- [x] Fallback ADS link via arXiv:e-print
- [x] File parsing alternative
- [x] Call function only once
- [x] Jinja2 filters tests
- [x] Removed obsolete `Access restricted` from weblinks

### DIFF
--- a/inspirehep/kbs/weblinks.kb
+++ b/inspirehep/kbs/weblinks.kb
@@ -1,0 +1,523 @@
+00070---International Journal of Mathematics Server
+00108---J.Knot Theor.Ramifications
+00140---Heavy Ion Physics Server
+00145---Int. J. Math. Math. Sci. Server
+00154---J. Nonlin. Math. Phys. Server
+00181---Spektrum der Wissenschaft Server
+00182---SLAC Beam Line Server
+00201---ICFA Instrum.Bull. Server
+00203---Adv. Theor. Math. Phys. Server
+00211---KFZ Nachrichten Server
+00222---Living Reviews in Relativity Server
+00228---Nonlin. Phenom. Complex Syst. Server
+00238---Math. Phys. Anal. Geom. Server
+00256---Geometry and Topology Server
+00257---Electron.J.Diff.Eq. Server
+00264---Entropy Server
+00286---HEP Lib.Web. Server
+00327---World Sci.Lect.Notes Phys. Server
+00357---Cent. Eur. J. Phys. Server
+00372---Romanian Astronomical Journal Server
+00376---ICFA Server
+00411---Les Houches Lect. Notes Server
+00436---Int. J. Geom. Meth. Mod. Phys. Server
+00438---Phys.Part.Nucl.Lett. Server
+00462---Journal of Physics Conference Series Server
+00466---Prog.Phys. Server
+00480---SIGMA Server
+00484---Electron. J. Theor. Phys. Server
+AAEJA---Astronomy and Astrophysics Server
+AASRE---Astronomy and Astrophysics Review Server
+ADSABS---ADS Abstract Service
+ADSABS2---ADS Abstract Service
+AFLBD---Annales De La Fondation Louis De Broglie Server
+AHPAA---Ann. Henri Poincare Server
+AIPCONF---AIP Conference Server
+AJPIA---Am.J.Phys. Server
+ALETE---Astronomy Letters Server
+ANJOA---Astron.J. Server
+ANPYA---Annalen der Physik Server
+APCPC---AIP Conference Server
+APHYE---Astroparticle Physics Server
+APNYA---Ann. Phys. Server
+APPLA---Applied Physics Letter Server
+APPOA---Acta Physica Polonica Server
+APSSB---Astrophys. Space Sci. Server
+APSVC---Acta Physica Slovaca Server
+ARAAA---Annual Review of Astronomy and Astrophysics Server
+ARNUA---Annual Review of Nuclear and Particle Science
+ARTHF---Algebras and Representation Theory Server
+ARXERR---Erratum from arXiv.org server
+ASJOA---Astrophysical Journal Server
+ATLAS---ATLAS Notes from CERN server
+ATMP---ATMP Server
+ATPYA---Astrophysics Server
+ATROE---Astronomy Reports Server
+AUJPA---Austral.J.Phys. Server
+BABAR---Postscript from BABAR [BABAR Collaborators Only]
+BABARWEB---BaBar Publications Database
+BABARWEBP---BaBar Password Protected Publications Database
+BJPHE---Braz. J. Phys. Server
+BNL---Brookhaven Document Server
+BOOK-REVIEW---Review of this book
+BOOKCL---SLAC BOOKS Database
+BOOKREVIEW---Book review/article
+BOOKS---SLAC BOOKS Database
+BROOKHAVEN---Brookhaven Document Server
+C00-02-07---CHEP2000 Server
+C00-06-26.2---EPAC 2000 Server
+C00-09-14---QFTHEP00 Server
+C00-10-09.4---PCAPAC00 Server
+C01-01-20.1---Moriond Server
+C01-06-18.1---PAC01 Server
+C01-06-25.8---BeamBeam01 Server
+C01-07-26---MENU2001 Server
+C01-08-07---ICRC01 Server
+C01-08-20.1---FEL 2001 Server
+C01-09-06.1---QFTHEP01 Server
+C01-10-02.1---DESY Conference Server
+C02-03-18.1---Durham Conference Server
+C02-06-17.2---SUSY02 Server
+C02-09-22---Hadron Structure 2002 Server
+C02-10-15.1---Aguas de Lindoia 2002 Server
+C03-07-31---ICRC Server
+C04-04-14---DIS04 Server
+C04-04-19.4---ECLOUD04 Server
+C04-11-08.2---HHH-2004 Server
+C05-03-22---Phystat05 Conference Server
+C05-09-12---Phystat05 Conference Server
+C05-09-12.10---CERN Conference Server
+C05-10-17---ICATPP9 Conference Server
+C06-09-16.1---Yalta06 Conference Server
+C07-04-16---DIS2007 Conference Server
+C07-05-21.2---Blois07 Server
+C91-08-05---SLAC Summer Inst. 1991 Server
+C92-07-13.1---SLAC Summer Inst. 1992 Server
+C93-07-26---SLAC Summer Inst. 1993 Server
+C95-07-10.2---SLAC Summer Inst 1995 Server
+C95-10-29---ICALEPCS'95 Server
+C97-09-04.1---QFTHEP'97 Server
+C97-10-13---IWAA'97 Server
+C98-06-22---EPAC98 Server
+C99-03-29---JACoW PAC99 Server
+C99-04-06---ANL Conference Server
+C99-04-12.3---LHC 99 JACoW Server
+C99-08-17---Conference Server
+C99-10-04---ICALEPCS99 Server
+CBPF---Rio de Janeiro, CBPF Server
+CDF93---CDF Document Server
+CDF94---CDF Document Server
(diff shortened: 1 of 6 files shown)


nikpap:
@jmartinm Have (some?) URL's been moved to the `external_system_numbers` field?

nikpap:
It would appear that some URL's are simple strings eg `u'urls': [u'http://www.physik.rwth-aachen.de/']` (record id: 902624), whereas others are a list of dictionaries as in

```
u'urls': [{u'url': u'http://dx.doi.org/10.1103/PhysRevD.66.043003', u'doc_string': u'10.1103/PhysRevD.66.043003', u'description': u'DOI'}, {u'url': u'http://link.aps.org/abstract/PRD/V66/E043003/', u'doc_string': u'V66/E043003/', u'description': u'PHRVA-D'}]
```
(record id: 586817)

I was wondering if that has been a decision or a bug that needs to be resolved. @jmartinm @jacquerie

jalavik:
@nikpap Just FYI, for questions regarding the data model, you can also refer to the JSON schemas which have been recently updated to reflect reality. E.g. https://github.com/inspirehep/inspire-next/blob/invenio3/inspirehep/modules/records/jsonschemas/records/hep.json#L1043

So anything not following that schema is wrong (or need to be put up for discussion like you did :+1: ).

As for `urls`, I think the schema way of list of dictionaries is the good one?

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -668,3 +669,28 @@ def format_date(datetext):
...
+    """Checks if given url starts with the following."""
+    if institute.lower() in ['kekscan', 'ads', 'cds', 'hal', 'msnet', 'msnet']:
+        return True
+    return False
+
+
+@blueprint.app_template_filter()
+def weblinks(description):
+    """..."""
+    weblinks = open(os.path.join('inspirehep', 'weblinks.kb'), 'r')
```
nikpap: @jmartinm @jalavik Is there an alternative/better way to parse this file?

jalavik: You are right to doubt this method as it would read and parse the file upon every call to this filter. Not great for templating speed. So we'd want to avoid relying on reading arbitrary files every time we render templates. 

Now, since invenio-knowledge is probably not gonna be ported as is. One suggestion is to read the file upon extension initialization in (E.g. `theme/ext.py`) and keeping it in memory as it is only ~500 entries. See one example of states here: https://github.com/inveniosoftware/invenio-access/blob/master/invenio_access/ext.py or https://github.com/inspirehep/invenio-workflows/blob/invenio3/invenio_workflows/ext.py

However, we might have more files like this so we should see how to best implement it see e.g. #927

nikpap:
I've updated this PR with (hopefully) a better way to parse the `weblinks.kb` file. @jalavik might be best to have a look. Also, think we need to decide where to place the file - would `theme/static` be a good option?

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -668,3 +669,26 @@ def format_date(datetext):
...
+    """Checks if given url starts with the following."""
+    if institute.lower() in ['kekscan', 'ads', 'cds', 'hal', 'msnet', 'msnet']:
+        return True
+    return False
+
+
+@blueprint.app_template_filter()
+def weblinks(description):
+    """Renames external links based on the description given."""
+    theme = INSPIRETheme()
```
jalavik: You do not need to reinstantiate the extension again as it is done upon loading of the entire application. Otherwise we would still always re-read the file upon constructing this object. To get the reference to it use the following:
```python
current_app.extensions.get('inspire-theme').weblinks.get(description)
```

inspirehep/modules/theme/ext.py
```diff
@@ -89,3 +92,12 @@ def setup_app(self, app):
                             force_lower=False)
         del gravatar
         return app
+
+    def init_weblinks_dictionary(self):
+
```
jalavik: Add a docstring explaining short what this function is doing.

inspirehep/modules/theme/ext.py
```diff
@@ -89,3 +92,12 @@ def setup_app(self, app):
...
+        weblinks_dictionary = {}
+        with open(os.path.join('inspirehep', 'weblinks.kb'), 'r') as file:
```
jalavik: This will not work correctly as it generates the relative path `inspirehep/weblinks.kb`. So running this on a webserver or anywhere other than from `$(VIRTUAL_ENV)/src/inspirehep` will fail. To get the right path reliably, take a look at this approach: https://pythonhosted.org/setuptools/pkg_resources.html#resource-extraction

```pkg_resources.resource_filename('inspirehep', 'weblinks.kb')```

File: inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_detailed.tpl [hunk 5]
```diff
@@ -57,7 +57,7 @@
             {% endif %}
           </div>
         {% endif %}
-        {% if record.get('urls') %}
+        {% if record.get('urls') or record.get('external_system_numbers') %}
```
jalavik: Just FYI, I think this will also work:
```
{% if record.urls or record.external_system_numbers %}
```

jalavik:
@nikpap Thanks! Some few things to look at still. And I think we need to put the `weblinks.kb` file under a specific folder. Perhaps `inspirehep/kbs/`? @jmartinm

jmartinm:
> @nikpap Thanks! Some few things to look at still. And I think we need to put the weblinks.kb file under a specific folder. Perhaps inspirehep/kbs/?

sounds good to me.

nikpap:
@jalavik I've incorporated all the changes, thanks a ton for the feedback!
@jmartinm You can have a look at various ID's, notably 1230937 or 1280200.

inspirehep/modules/theme/ext.py
```diff
@@ -89,3 +92,12 @@ def setup_app(self, app):
                             force_lower=False)
         del gravatar
         return app
+
+    def init_weblinks_dictionary(self):
+        """Initialize the dictionary for HEP detailed record external links."""
+        weblinks_dictionary = {}
+        with open(resource_filename('inspirehep', 'kbs/weblinks.kb'), 'r') as file:
```
jalavik: file is a Python keyword, so you can change it to `kb_file`.

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -668,3 +668,26 @@ def format_date(datetext):
...
+    if institute.lower() in ['kekscan', 'ads', 'cds', 'hal', 'msnet', 'msnet']:
+        return True
+    return False
+
+
+@blueprint.app_template_filter()
+def weblinks(description):
+    """Renames external links based on the description given."""
+    value = current_app.extensions.get('inspire-theme').weblinks.get(description)
+    if (value):
```
jalavik: You can remove the parentheses.

jalavik:
@nikpap :+1: 

For the future, let's keep on eye on more cases like this as we can then make the KB loading more generic and separated out of themes.

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -677,3 +677,28 @@ def format_date(datetext):
...
+    """Checks if given url starts with the following."""
+    return not url.endswith(('.pdf', '.png', '.jpg', '.jpeg'))
+
+
+@blueprint.app_template_filter()
+def is_institute(institute):
+    """Checks if given url starts with the following."""
+    if institute.lower() in ['kekscan', 'ads', 'cds', 'hal', 'msnet', 'msnet']:
+        return True
+    return False
```
jacquerie: This can just be `return institute.lower() in ['kekscan', 'ads', 'cds', 'hal', 'msnet', 'msnet']`.

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -677,3 +677,28 @@ def format_date(datetext):
             return date
         elif len(datestruct) == 1:
             return datestruct[0]
+
+
+@blueprint.app_template_filter()
+def is_external_link(url):
+    """Checks if given url starts with the following."""
+    return not url.endswith(('.pdf', '.png', '.jpg', '.jpeg'))
```
jacquerie: Incorrect docstring and weird implementation, this being `is_external_link`...

inspirehep/modules/theme/ext.py
```diff
@@ -41,6 +43,7 @@ def __init__(self, app=None, **kwargs):
         self.menu_ext = Menu()
         self.menu = None
         self.breadcrumbs = Breadcrumbs()
+        self.weblinks = self.init_weblinks_dictionary()
```
jacquerie: Design: why not make this an object creation using a `KB` class that wraps the file accessing? This could become our guerrilla substitute for Invenio-Knowledge.

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -677,3 +677,28 @@ def format_date(datetext):
...
+
+
+@blueprint.app_template_filter()
+def is_institute(institute):
+    """Checks if given url starts with the following."""
```
jacquerie: Similarly, incorrect docstring.

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -677,3 +677,28 @@ def format_date(datetext):
...
+
+@blueprint.app_template_filter()
+def weblinks(description):
+    """Renames external links based on the description given."""
+    value = current_app.extensions.get('inspire-theme').weblinks.get(description)
+    if value:
+        return value
+    if description:
+        return 'Link to ' + description
+    return 'Link to fulltext'
```
jacquerie: BTW, can you add simple tests for these filters in `tests/unit/theme/test_jinja2filters`?

jacquerie:
I took the completed checklist as a sign that this is ready and needs review, if this is still WIP many apologies @nikpap :dancer:

inspirehep/kbs/weblinks.kb
```diff
@@ -0,0 +1,523 @@
...
+IOPLETT---IOP Phys.Express Lett. Server
+IRNPE---Int.Rev.Nucl.Phys. Server
+JACOW---Full-text at JACoW Server
+JACOWS---Slides on JACoW Server
+JAUMA---J.Austral.Math.Soc. Server
+JCAPA---JCAP Electronic Journal Server
+JCTPA---J. Comput. Phys. Server
+JDGEA---J. Diff. Geom. Server
+JETP---J. Exp. Theor. Phys. Server (Access may be <a href=/find/restrict.html> restricted )
+JETPL---J. Exp. Theor. Phys. Lett. Server (Access may be <a href=/find/restrict.html> restricted )
```
jacquerie: @kaplun Does this link make sense anymore? I see that it's not on INSPIRE now: https://inspirehep.net/find/restrict.html

kaplun: Oih. We might have forgotten it when migrating away from the old balancer maybe. Let me check.

kaplun: Can't find any traces. I guess the A tag can be dropped.

kaplun: @annetteholtkamp also mention this KB is a bit old style. We can really drop all the `(Access may be...)` since anyway our users don't expect INSPIRE will magically grant them access to papers. Also more and more we should use DOIs to redirect user to papers.

nikpap:
@jacquerie @eamonnmag this should be ready.

jacquerie:
Will review later today (5pm).

eamonnmag:
All checked. Looks fine to merge.

PR #939, opened 2016-02-29
WIP 	search: suggestion messages

Panos512:
* Adds suggestion messages on search.

* Ports #802 on invenio3 version.

### DIFF
--- a/inspirehep/modules/search/__init__.py
+++ b/inspirehep/modules/search/__init__.py
@@ -28,4 +28,6 @@ from __future__ import absolute_import, print_function
 
 from .ext import INSPIRESearch
 
+from .receivers import *
+
 __all__ = ('INSPIRESearch', )
--- a/inspirehep/modules/search/receivers.py
+++ b/inspirehep/modules/search/receivers.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015, 2016 CERN.
+#
+# Invenio is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+
+from flask import flash
+
+from .signals import extra_keywords, malformed_query, unsupported_keyword
+
+
+@unsupported_keyword.connect
+def unsupported_keyword_message(sender, keyword, *args, **kwargs):
+    flash(str('<b>' + keyword +
+              "</b> keyword is currently unsupported."), "query_suggestion_with_redirect")
+
+
+@malformed_query.connect
+def malformed_query_message(sender):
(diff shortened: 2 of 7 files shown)


Panos512:
WIP Preview:

screen shot 2016-02-29 at 11 21 37

kaplun:
Hi @Panos512, Nice! I would though try to improve readability (e.g. more contrast between font color and background, and removing the transparency).

Panos512:
Yeah, also as I was discussing with @jmartinm a long time ago we could also change the color (to yellow maybe?)

inspirehep/modules/search/query_suggestions.py
```diff
@@ -0,0 +1,47 @@
+
```
kaplun: Shouldn't this file be split between signals and receivers? Ideally signals could be moved into `invenio-query-parser` then.

Panos512: Yeah it makes sense to split the files. But since we use the walkers from the search module to give suggestions and not from the query-parser I am not sure if the signals should be in query-parser or the module.

Panos512:
WIP Preview v2:

screen shot 2016-02-29 at 14 49 34

nikpap:
I was wondering if the background could match the colour of each collection?

kaplun:
@nikpap but then e.g. for Authors, Data, Institution it wouldn't convey the conventional "Warning" meaning, provided by using yellow/orange colors. (indeed our HEP collection is a bit a warning :smile:).

jalavik:
@Panos512 @nikpap  Still WIP or ready to go?

Panos512:
@jalavik not yet. We need to find a new way to pass messages from backend now that we'll replace jinja with angular.

kaplun:
This Angular thing is getting out of hand :tada:

jmartinm:
In order to pass the error messages from the REST API to the AngularJS client you can have a look at the existing examples:

1) Raise an custom exception anywhere during the parsing phase. See the [exception being raised](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L3660) and the [custom exception](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/errors.py#L32). The `description` field will be part of the response to AngularJS.

1.1) Note that on INSPIRE we overwrite certain parts of the `invenio-records-rest` view function through the [factories provided](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L371). You can see INSPIRE's addition [here](https://github.com/inspirehep/inspire-next/blob/invenio3/inspirehep/modules/search/query.py). You can create some custom Exceptions and raise them there.

2) On AngularJS `invenio-search-js` an [event is emitted](https://github.com/inveniosoftware/invenio-search-js/blob/master/src/invenio-search-js/invenioSearch.module.js#L135) when the request errors.

2.1) `invenio-search-js` has a [default handler for the exception above](https://github.com/inveniosoftware/invenio-search-js/blob/master/src/invenio-search-js/invenioSearch.module.js#L202) that modifies the [controller](https://docs.angularjs.org/guide/controller).

2.2) When the controller is modified, our template [knows about it](https://github.com/inspirehep/inspirehep-search-js/blob/master/src/inspirehep-search-js/templates/error.html#L1) and puts a message in the page (with `ng-if`). That template is used in our [search results template](https://github.com/inspirehep/inspire-next/blob/invenio3/inspirehep/modules/search/templates/search/search.html#L92).


So two possible solutions (for the client side part) are to listen to the event mentioned in 2) and modify the interface, or to create a template like the one in 2.2) that will react when the controller is modified and show a message.

kaplun:
@jmartinm, @Panos512 this was getting ready to be integrated, correct?

Panos512:
@kaplun it's here https://github.com/inspirehep/inspirehep-search-js/pull/5 , but it still needs a test and some visual improvements.

kaplun:
@Panos512, this needs to be closed and reopened against `master`.

PR #943, opened 2016-02-29
format: HEP record DOI

nikpap:
* Correctly displays DOI in search results and detailed records.

* Consolidates the two macros in one generic macro.

### DIFF
--- a/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_brief.tpl
+++ b/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_brief.tpl
@@ -19,7 +19,7 @@
 
 {% from "inspirehep_theme/format/record/Inspire_Default_HTML_general_macros.tpl" import render_record_title, render_record_authors, record_abstract, record_arxiv, record_report_numbers, record_cite_modal with context %}
 
-{% from "inspirehep_theme/format/record/Inspire_Default_HTML_brief_macros.tpl" import record_journal_info, render_doi, record_journal_info_and_doi with context %}
+{% from "inspirehep_theme/format/record/Inspire_Default_HTML_brief_macros.tpl" import record_journal_info, record_journal_info_and_doi with context %}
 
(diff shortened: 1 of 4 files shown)


inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -17,9 +17,9 @@
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
```
jalavik: @jmartinm Do we still need the brief template?

jmartinm: Nope it will disappear. @nikpap you can remove the brief parts. 

Depending on how complex the logic is, we can either implement it for the brief format here (https://github.com/inspirehep/inspirehep-search-js/blob/master/src/inspirehep-search-js/filters/doi.js) or we do a jinja filter that renders the DOI field and we return the pre-formatted value in the serializer (https://github.com/inspirehep/inspire-next/blob/invenio3/inspirehep/modules/records/serializers/json.py#L49)

nikpap:
@jmartinm updated this PR. Regarding the brief format, the only custom logic being applied in this case is using the `remove_duplicates_from_dict` filter to remove duplicate DOI's from being shown.

ioannistsanaktsidis:
@nikpap note that this problem will be solved after we merge this:
https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/receivers.py#L176-L186
After that the ```remove_duplicates``` filters will not be needed.

nikpap:
@jacquerie @eamonnmag It seems that in some cases the `DOI` field is not present in the record json data. For example at http://inspirehep.net/record/1408310/export/xm there is a DOI field, but at http://localhost:5000/literature/1408310 no DOI can be found in the data of the record.

eamonnmag:
This is pretty likely. In this case, you just skip the field?

PR #945, opened 2016-03-01
records: use different pid_type for different record types

jmartinm:
* Different record types now use different pid_types.

* NOTE: Each document type now has a specific endpoint for detailed
  views. E.g. invenio_records_ui.literature.

### DIFF
--- a/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_brief.tpl
+++ b/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_brief.tpl
@@ -37,7 +37,7 @@
               </div>
               <div class="col-md-11">
                 <h4 class="custom-h">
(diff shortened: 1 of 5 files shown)


inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_brief.tpl
```diff
@@ -37,7 +37,7 @@
               </div>
               <div class="col-md-11">
                 <h4 class="custom-h">
-                    <a class="title" href="{{ url_for('invenio_records_ui.recid', pid_value=record['control_number']) }}">
+                    <a class="title" href="{{ url_for('invenio_records_ui.literature', pid_value=record['control_number']) }}">
```
jalavik: This will eventually cause troubles for other collections but we can add that later.

PR #951, opened 2016-03-01
search: initial AngularJS app

jmartinm:
* Uses AngularJS app to show search results.


jmartinm:
Already merged.

PR #953, opened 2016-03-02
RFC scripts: introducing scripts

jacquerie:
I like the idea of using the shell for shell things. This should be the place for all your convenience scripts that you wrote to ease the pain! Like @dset0x had in https://github.com/dset0x/invinipain.


jalavik:
:+1:

jacquerie:
Finally green, but I'm so sad that I couldn't make ES work as intended on Travis. We will still have to do Docker builds, I guess : (

jmartinm:
:+1:

scripts/recreate_records
```diff
@@ -0,0 +1,46 @@
...
+  >&2 echo "You must have the celery command to use this script."
+  exit 1
+fi
+
+inspirehep db drop --yes-i-know
+inspirehep db create
+inspirehep index destroy --force --yes-i-know
+inspirehep index init
+
+celery 2>/dev/null worker -E -A inspirehep.celery --workdir="${VIRTUALENV}" --purge &
```
jalavik: I am wondering if it is "safer" here to process id instead of killing all celery worker instances. 

```
celery 2>/dev/null worker -E -A inspirehep.celery --workdir="${VIRTUALENV}" --purge &
CELERY_PID=$!
inspirehep migrator populate -f inspirehep/demosite/data/demo-records.xml.gz --wait=true
kill $CELERY_PID
```

jacquerie: Yes, your approach looks nicer. Mine was lifted straight from the documentation: http://docs.celeryproject.org/en/latest/userguide/workers.html#stopping-the-worker

PR #954, opened 2016-03-02
robots.txt: Disallow all search engine crawling

nikpap:
Some of the QA pages can be currently found in search engines results. This update would tell all robots to not crawl any pages on the QA website.

The "User-agent: *" means this section applies to all robots.
The "Disallow: /" tells the robot that it should not visit any pages on the site.


jalavik:
@nikpap Thanks. This should be the default one everywhere (so it belongs in master) and then on the operations side, we'll have a "real" robots.txt (e.g. via puppet etc.)

jmartinm:
Yep, I see that quite a lot of pages are indexed https://www.google.com/?q=site:qa.inspirehep.net

We can also remove them from the Google webmaster tools @kaplun

jalavik:
Merged as 1dbd6e9

PR #961, opened 2016-03-03
search: enables wildcards

### LINKED ISSUE #831: Search wildcards
Currently on inspirehep.net we support wildcards. 
E.g. someone can query for `muon*` or `find r atlas-conf-*` . 

We can have this on labs but the problem is that a `*` character can also be used as the multiplication symbol. What we could do is have the users use a special character to "wrap" a `wildacrad query` (as `/` is used for a regex query e.g. `title:/^E.*s$/`).

Panos512:
* Enables searching with wildcards (only used in the end of a word).

* Should be merged with inspirehep/invenio-query-parser#14.

* Should be merged with  inspirehep/invenio-query-parser#18.

* Closes box no.10 of #817.

* Closes #831.

### DIFF
--- a/inspirehep/modules/search/walkers/elasticsearch.py
+++ b/inspirehep/modules/search/walkers/elasticsearch.py
@@ -29,7 +29,7 @@ from invenio_query_parser.ast import (AndOp, DoubleQuotedValue, EmptyQuery,
                                       GreaterEqualOp, GreaterOp, Keyword,
                                       KeywordOp, LowerEqualOp, LowerOp, NotOp,
                                       OrOp, RangeOp, RegexValue,
-                                      SingleQuotedValue, Value, ValueQuery)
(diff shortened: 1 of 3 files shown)


jacquerie:
You might want to take another stab at this, now that we have some testing infrastructure in place? BTW, maybe it's time to support a query such as `find a chkv#` as a `WildcardQuery`.

kaplun:
@jacquerie where is it documented to use # as a wildcard? was it in the SPIRES syntax?

kaplun:
@Panos512 can you try rebasing and make it green?

kaplun:
Can you add to commit log: `closes #831`?

Panos512:
@kaplun done! Btw this fails because  inspirehep/invenio-query-parser#18 is not yet merged.

PR #976, opened 2016-03-08
search: type_code search

### LINKED ISSUE #970: find tc
See: https://inspirehep.net/info/hep/search-tips#tc

One should be able to search by: `find tc b` and `find tc book`. This could be implemented via an enhancer that read the `document_type` field (or was it collections :smile:?) and the `CORE` field.

Panos512:
* Enables Searching by Type-Code (tc).

* Closes #970 .

* Need to be merged with inspirehep/invenio-query-parser#17 .

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -147,6 +147,7 @@ SEARCH_ELASTIC_KEYWORD_MAPPING = {
                     ],
     "abstract": ["abstracts.value"],
     "collaboration": ["collaboration.value", "collaboration.raw^2"],
+    "tc": ["collection"],
     "collection": ["collections.primary"],
     "doi": ["dois.value"],
     "doc_type": ["facet_inspire_doc_type"],
--- a/inspirehep/modules/records/mappings/records/hep.json
+++ b/inspirehep/modules/records/mappings/records/hep.json
@@ -20,7 +20,20 @@
                     "type": "pattern_replace",
                     "pattern": "[ \t]{2,}",
                     "replacement": " "
-                }
+                },
+                "tc_synonyms" : {
+                    "type" : "synonym",
+                    "synonyms" : [
+                        "b => book",
+                        "c => conferencepaper",
+                        "experimental note => note",
+                        "p => published",
+                        "t => thesis",
+                        "i => introductory",
+                        "l => lectures",
+                        "r => review"
+                      ]
+                    }
             },
             "analyzer": {
                 "basic_analyzer": {
@@ -42,6 +55,11 @@
                         "whitespace_remove",
                         "trim"
                     ]
+                },
+                "tc_synonyms_analyzer": {
+                  "type": "custom",
+                  "tokenizer": "keyword",
+                  "filter": ["tc_synonyms", "lowercase"]
                 }
             }
         },
@@ -232,7 +250,7 @@
                 },
                 "collection": {
                     "type": "string",
-                    "index": "not_analyzed"
+                    "analyzer": "tc_synonyms_analyzer"
                 },
                 "exactauthor": {
                     "type": "string",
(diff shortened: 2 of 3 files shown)


inspirehep/config.py
```diff
@@ -363,9 +363,9 @@ def _(x):
                     ],
     "abstract": ["abstracts.value"],
     "collaboration": ["collaboration.value", "collaboration.raw^2"],
-    "collection": ["collections.primary"],
```
jalavik: Is collection no longer used?

Panos512: Oops I missed that. Although it doesn't seem to be used anywhere.

jalavik: I think we still need it though. Keep it for now and then we'll see.

Panos512: Sure, I added it again :dancers:

jalavik:
It looks sound. I will have a check locally first, but soon we'll hopefully have a test suite for this @mihaibivol @jacquerie

Panos512:
This is failing because of inspirehep/invenio-query-parser#15 (uses wrong keyword-mapping) .

Panos512:
This is failing because of inspirehep/invenio-query-parser#17 (uses wrong keyword-mapping) .

jacquerie:
> but soon we'll hopefully have a test suite for this

Partially done, for now we unit test the query sent to ES. Soon, the results.

tests/unit/search/test_search_query.py
```diff
@@ -807,11 +807,19 @@ def test_find_report():
...
+    assert expected == result
+
+
+def test_type_code_colon():
+    query = InspireQuery('tc: l')
+
+    expected = {'query': {'multi_match': {'query': 'l', 'fields': ['collection']}}}
     result = query.body
 
     assert expected == result
```
jacquerie: As I anticipated, my test ordering isn't easily grokkable :dancer:

Shall we try after lunch try to come up with a better ordering together?

tests/unit/search/test_search_query.py
```diff
@@ -807,11 +807,19 @@ def test_find_report():
...
 
-    expected = {}
+    expected = {'query': {'multi_match': {'query': 'book', 'fields': ['collection']}}}
+    result = query.body
+
+    assert expected == result
+
+
+def test_type_code_colon():
+    query = InspireQuery('tc: l')
```
jacquerie: May I ask for a proper value, here? Something that would return values when the generated query is sent to ES.

tests/unit/search/test_search_query.py
```diff
@@ -781,7 +781,7 @@ def test_find_country_code():
 def test_find_date():
     query = InspireQuery('fin date > today')
 
-    expected =  {}
```
jacquerie: Whoops, thanks.

inspirehep/modules/records/mappings/records/hep.json
```diff
@@ -20,7 +20,20 @@
...
+                        "b => book",
+                        "c => conferencepaper",
+                        "experimental note => note",
+                        "p => published",
+                        "t => thesis",
+                        "i => introductory",
+                        "l => lectures",
+                        "r => review"
+                      ]
+                    }
```
jacquerie: :children_crossing: **warning** :children_crossing:

Since we are touching the mappings, we _might_ have to reindex when this is merged.

tests/unit/search/test_search_query.py
Panos512: `l` stands for `lectures` and it also makes use of ES synonyms.
You think it's not clear enough for someone that reads the test?

tests/unit/search/test_search_query.py
Panos512: Yeah ofc :dancer:

tests/unit/search/test_search_query.py
jacquerie: Yeah, now I see. I wasn't seeing any results in Sense, so I thought the query was malformed. It can stay as it is.

PR #932, opened 2016-02-26
filters: ports publication info filter in invenio3

ioannistsanaktsidis:
Signed-off-by: Ioannis Tsanaktsidis <ioannis.tsanaktsidis@cern.ch>

PR #967, opened 2016-03-04
Porting schemas to use JSONref

mihaibivol:
* [x] In order to reduce boilerplate in other schema definitions add `json_reference.json` schema as defined above to be served.

* [x] Amend jsoschemas to use `smth_record: {"$ref": "json_reference.json"}`

* [x] Amend ES Mappings (If actually needed)

* [x] Amend Dojson
  * [x] Add forward and backward $ref generation
  * [x] + Unit test
  * [x] Change __all__ fields that use recids
  * [x] + Test against __existing__ fixtures

* [x] Amend Receivers to put back the recid to be indexed in ES
  * [x] Unit Test (Integration tests not needed)

* [x] Fix broken views
* [x] Test thanks @jacquerie :+1: 

* [x] ~~Integration test the process~~ Wait for better framework to do this
LE: Actually the base of any integration test will be reindexing everything and asserting that all the marc records were indexed as json correctly. So any integration testing framework will always use this code.

* Closes #958

PR #978, opened 2016-03-08
theme: implement or fix xfailed tests

jacquerie:
Not yet fulfilling the promise in https://github.com/inspirehep/inspire-next/pull/932#issuecomment-189243461.

### DIFF
--- a/inspirehep/modules/theme/jinja2filters.py
+++ b/inspirehep/modules/theme/jinja2filters.py
@@ -39,6 +39,7 @@ from inspirehep.utils.date import (
 )
 
 from invenio_search.api import Query
+from invenio_search import current_search_client as es
 
 from .views import blueprint
 from inspirehep.utils.template import render_macro_from_template
@@ -75,6 +76,17 @@ def apply_template_on_array(array, template_path, **common_context):
             )
 
     return rendered
+
+
+def collection_to_index(collection_name):
+    """Translates a collection name to the corresponding index."""
+    try:
+        mapping = current_app.config['SEARCH_ELASTIC_COLLECTION_INDEX_MAPPING']
+        index = mapping[collection_name.lower()]
+    except KeyError:
+        index = 'records-hep'
+
+    return index
 
 
 @blueprint.app_template_filter()
@@ -187,16 +199,19 @@ def count_words(value):
 
 
 @blueprint.app_template_filter()
-def is_intbit_set(value):
-    from intbitset import intbitset
-    if isinstance(value, intbitset):
-        value = value.tolist()
-    return value
-
-
-@blueprint.app_template_filter()
 def remove_duplicates_from_dict(value):
-    return [dict(t) for t in set([tuple(d.items()) for d in value])]
+    """Actually removed duplicates from a list.
+
+    Kept because other PRs might depend on it.
+    FIXME(jacquerie): remove or rename.
+    """
+    result = []
+
+    for d in value:
+        if d not in result:
+            result.append(d)
+
+    return result
 
 
 @blueprint.app_template_filter()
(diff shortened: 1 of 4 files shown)


tests/unit/theme/test_theme_jinja2filters.py
```diff
@@ -0,0 +1,670 @@
...
+    email_links, epoch_to_year_format, experiment_date, experiment_link,
+    format_cnum_with_hyphons, format_cnum_with_slash, format_date, has_space,
+    institutes_links, is_intbit_set, is_list, is_upper, join_array,
+    join_nested_lists, latex, limit_facet_elements, link_to_hep_affiliation,
+    long_timeout, new_line_after, number_of_records, number_of_search_results,
+    proceedings_link, publication_info, remove_duplicates,
+    remove_duplicates_from_dict, sanitize_arxiv_pdf, sanitize_collection_name,
+    search_for_experiments, sort_list_by_dict_val, split_author_name,
+    strip_leading_number_plot_caption, url_links, words, words_to_end
+)
```
jacquerie: Should probably be replaced with `*`.

jacquerie:
Now fulfilling the promise in https://github.com/inspirehep/inspire-next/pull/932#issuecomment-189243461. The bugs I found are documented with an `xfailed` test, which can be run with `pytest --runxfail tests/unit` to see the error. I also provided a succinct `reason` parameter to help understanding what the error is; if it's not enough, ping me somewhere and I'll try to remember what was happening : )

jalavik:
Thanks for this. When will the xfailed bugs be addressed? Is that already in the works?

I am asking because some of them seem trivial to fix and could potentially be part of this PR unless it steps on someone else's work. E.g. intbitset we no longer need to use I think @kaplun  And the new search API just need to be adjusted etc.

jacquerie:
> Is that already in the works?

Not by me, because in fact I wasn't sure if @ioannistsanaktsidis was already working on this file. But yes, some are trivial to fix.

kaplun:
Indeed `intbitset`is not needed anymore.

jacquerie:
Tests got submitted in another PR: https://github.com/inspirehep/inspire-next/pull/967. I will reuse this PR to do the proposed changes (`Query` API change, other minor stuff).

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -284,13 +287,16 @@ def proceedings_link(record):
     out = ''
     if not cnum:
         return out
-    search_result = Query("cnum:%s and 980__a:proceedings" % (cnum,)).\
-        search()
-    if len(search_result):
-        if len(search_result) > 1:
+
+    query = Query("cnum:%s and 980__a:proceedings" % (cnum,))
+    result = es.search(index='records-hep', body=query.body)
```
jacquerie: Maybe I should remove this `index` argument.

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -399,9 +406,11 @@ def collection_select_current(collection_name, current_collection):
 @blueprint.app_template_filter()
 def number_of_records(collection_name):
     """Returns number of records for the collection."""
-    # return len(Query("").
-    #            search(collection=collection_name))
-    return 0
+    query = Query()
+    index = collection_to_index(collection_name)
+    result = es.count(index=index, body=query.body)
```
jacquerie: This could be abstracted in a method.

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -510,15 +519,17 @@ def number_of_search_results(query, collection_name):
...
                 "number_of_hits"
             ]
-    # Replay the Query to get total number of hits
-    return len(Query('{query} AND collection:"{collection}"'.format(
-        {'query': query, 'collection': collection_name}
-    )))
+
+    query = Query(query)
+    index = collection_to_index(collection_name)
+    result = es.count(index=index, body=query.body)
```
jacquerie: Similarly, this could be abstracted.

jacquerie:
Most xfailed tests are now green, except two, who required more effort and were kept xfailed. I also decided to keep the querying concrete, so that we can abstract them when other people need that logic, so that the API will be as flexible as possible.

jacquerie:
@jalavik Now incorporates your oral comments of yesterday.

PR #980, opened 2016-03-08
search: create collection search pages

jmartinm:
* Creates a different template for each collection search results page.

* Adapts search view so that a different template is applied depending
  on the collection being searched on (using cc parameter).


jmartinm:
At the moment, the same serializer is being used for all collections

```
'application/vnd+inspire.brief+json': (
    'inspirehep.modules.records.serializers'
    ':json_brief_v1_search'
```

This needs to be changed later to have a different one per collection.

Search results templates currently just inherit from `search.html`, later they should overwrite some blocks to have, for example, their own brief format:

```
{%- block search_results %}
    <invenio-search-results
     template="{{ url_for('static', filename='node_modules/inspirehep-search-js/dist/templates/literature/results.html') }}">
    </invenio-search-results>
{%- endblock search_results %}
```

jmartinm:
One view `/search` is at the moment being shared (with different template being rendered depending on `cc` parameter) - https://github.com/jmartinm/inspire-next/blob/collection-search-pages/inspirehep/modules/search/views.py#L41 this seems to be working fine, any other solutions?

jmartinm:
@jalavik @kaplun @jacquerie let me know if it looks ok for you.

jalavik:
Looks good :+1:

PR #982, opened 2016-03-09
records: adds display fields

ioannistsanaktsidis:

### DIFF
--- a/inspirehep/modules/theme/bundles.py
+++ b/inspirehep/modules/theme/bundles.py
@@ -33,7 +33,7 @@ almondjs = NpmBundle(
     npm={
         "almond": "~0.3.1",
         "hogan.js": "~3.0.2",
-        "requirejs-hogan-plugin": "~0.3.1",
+        "requirejs-hogan-plugin": "~0.3.1"
     }
 )
 
@@ -65,7 +65,6 @@ js = NpmBundle(
         "flightjs": "~1.5.0",
         "angular": "~1.4.8",
         "inspirehep-js": "git+https://git@github.com/inspirehep/inspirehep-js.git"
-
     }
 )
 
--- a/inspirehep/modules/theme/static/js/settings.js
+++ b/inspirehep/modules/theme/static/js/settings.js
@@ -69,7 +69,7 @@ require.config({
       deps: ['angular', 'angular-ui-bootstrap']
     },
     'inspirehep-search': {
-        deps: ['angular', 'angular-sanitize']
+        deps: ['angular', 'angular-sanitize', 'angular-ui-bootstrap']
     },
     'invenio-search': {
         deps: ['angular']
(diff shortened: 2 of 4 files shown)


inspirehep/modules/theme/jinja2filters.py
```diff
@@ -613,18 +619,19 @@ def publication_info(record):
...
-                    'conferences', str(pub_info['parent_recid']))
-                parent_rec = es.get_source(
-                    index='records-hep',
-                    id=str(pid.object_uuid),
-                    doc_type='hep',
-                    ignore=404)
+                    'conference_recid' not in pub_info:                
                 try:
+                    pid = PersistentIdentifier.get(
+                        'conferences', str(pub_info['parent_recid']))
```
jacquerie: The `parent_recid` is probably not in `conferences`. It might be in `literature`, for the case I tested: https://github.com/jacquerie/inspire-next/commit/007fe31b441faf88a5593489a6f46e341ea522c4#diff-cc4ec36bfdca525994bc68c9082219e3R951

ioannistsanaktsidis: Hmmm I am not quite sure about this. @jmartinm give us your lights here.
https://github.com/inspirehep/inspire-next/issues/541

ioannistsanaktsidis:
@jmartinm  this one is ready.

inspirehep/modules/search/static/js/search/app.js
```diff
@@ -24,8 +24,9 @@
 require([
     'angular',
     'angular-sanitize',
+    'angular-ui-bootstrap',
```
jmartinm: this will not be needed after rebase.

inspirehep/modules/theme/bundles.py
```diff
@@ -65,7 +66,8 @@
         "clipboard": "~1.5.8",
         "flightjs": "~1.5.0",
         "angular": "~1.4.9",
-        "angular-sanitize": "~1.4.9"
+        "angular-sanitize": "~1.4.9",
+        "angular-ui-bootstrap": "~1.2.4"
```
jmartinm: I think I have missed adding this dependency. So when you rebase you can add it.

jmartinm: Checking now we don't need to add it. As our library `inspirehep-js` depends on it (has it on its `package.json`) it already gets installed. And it is better since we can control the version from that package.

ioannistsanaktsidis: Cool I will remove it then ;)

inspirehep/modules/theme/static/js/settings.js
```diff
@@ -51,8 +51,8 @@ require.config({
     'angular-sanitize': {
       deps: ['angular']
     },
-    'angular-ui-bootstrap': {
-      deps: ['angular']
+    "angular-ui-bootstrap": {
```
jmartinm: Can you put back single quotes in this file?

PR #986, opened 2016-03-10
Make scripts/recreate_records work with honcho start

mihaibivol:
* `scripts/recreate_records` did not cleanup any workers upon cancellation -> fix that using trap
* `scripts/recreate_records` started new workers even if others were started via `honcho start` -> fix using a common pidfile
* the first argument is the celery loglevel. Makes things easy to debug when breaking `recreate_records`

TODO
* ~~test `--autoreload` (having the server started with `honcho start` would not autoreload on code changes before, but the webapp did)~~ Does not work
* ~~make everything work with ES started as a service (`service elasticsearch start`)~~

LE:
`scripts/recreate_records` doesn't care about how ES is started so it might be left like this for now. On the other hand I am not sure on how `honcho start` works with ES installed as a service.
LE2:
Managing ES will remain the dev's responsibility
LE3:
Autoreload doesn't work :(

cc @kaplun, @jacquerie

### DIFF
--- a/scripts/recreate_records
+++ b/scripts/recreate_records
@@ -36,14 +36,30 @@ if [ ! -x "$(command -v celery)" ]; then
   exit 1
 fi
 
+CELERY_LOGLEVEL=$1
+if [ -z $1 ]; then
+	CELERY_LOGLEVEL="ERROR"
+fi
+
+cleanup() {
+	echo "----> Killing Celery Worker ${CELERY_PID}"
+	kill "${CELERY_PID}"
+}
+
+CELERY_PIDFILE="${VIRTUAL_ENV}/worker.pid"
+if [ -f "${CELERY_PIDFILE}" ]; then
+	echo "----> Purging all pending messages"
+	celery purge -f -A inspirehep.celery --workdir="${VIRTUAL_ENV}"
+else
+	celery worker -E -A inspirehep.celery --loglevel=${CELERY_LOGLEVEL} --workdir="${VIRTUAL_ENV}" --autoreload --pidfile="${CELERY_PIDFILE}" --purge &
+	CELERY_PID=$!
+	trap 'cleanup' 0 1 2 3 6
+fi
+
 inspirehep db drop --yes-i-know
 inspirehep db create
 inspirehep index destroy --force --yes-i-know
 inspirehep index init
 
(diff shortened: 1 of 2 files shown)


jmartinm:
The `--autoreload` seems like a nice feature! So you added it as a parameter but did not try if it is doing the right thing and restarting the workers on code change? Would be very nice if it works

scripts/recreate_records
```diff
@@ -36,14 +36,30 @@ if [ ! -x "$(command -v celery)" ]; then
...
+}
+
+CELERY_PIDFILE="${VIRTUAL_ENV}/worker.pid"
+if [ -f "${CELERY_PIDFILE}" ]; then
+	echo "----> Purging all pending messages"
+	celery purge -f -A inspirehep.celery --workdir="${VIRTUAL_ENV}"
+else
+	celery worker -E -A inspirehep.celery --loglevel=${CELERY_LOGLEVEL} --workdir="${VIRTUAL_ENV}" --autoreload --pidfile="${CELERY_PIDFILE}" --purge &
+	CELERY_PID=$!
+	trap 'cleanup' 0 1 2 3 6
```
jacquerie: Personal education: what are those numbers? I would have expected to have 9 and 15 there, or `SIGTERM` and `SIGKILL`, but not those.

mihaibivol:
I'll have to play a bit with it to see its behavior.

scripts/recreate_records
mihaibivol: SIGTERM is not trappable. I took the signals from here, and yeah, 15 is missing. http://steve-parker.org/sh/trap.shtml

jacquerie: > SIGTERM is not trappable.

Silly me! Yes, of course.

mihaibivol:
Updated the issue and I'm not sure about the behavior of `honcho start` when ES is already running on the machine. Is anyone using both?

jacquerie:
> Updated the issue and I'm not sure about the behavior of honcho start when ES is already running on the machine. Is anyone using both?

Yes: the two ES instances join each other in a single cluster, which could lead to subtle bugs. In fact, I usually comment out the `indexer` line in the `Procfile`.

jacquerie:
LGTM!

PR #989, opened 2016-03-10
JSON Reference Resolver that bypasses HTTP requests for known resources.

mihaibivol:
* Adds JSON reference loaders that first try to resolve recird URIS
  against the database or Elasticsearch before making HTTP requests.

usage example
```python
>>> from inspirehep.modules.records import json_ref_loader
>>> json_ref_loader.replace_refs({'$ref': 'http://localhost:5000/api/literature/1385669'})
{'samething': 'from db'}
>>> json_ref_loader.replace_refs({'$ref': 'http://localhost:5000/api/literature/1385669'}, 'es')
{'samething': 'from elastic'}
>>> json_ref_loader.replace_refs({'$ref': 'http://localhost:5000/api/literature/1385669'}, 'http')
{'samething': 'using GET'}
```

### DIFF
--- a/inspirehep/modules/records/json_ref_loader.py
+++ b/inspirehep/modules/records/json_ref_loader.py
@@ -1,0 +1,115 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Resource-aware json reference loaders to be used with jsonref."""
+
+from jsonref import JsonLoader, JsonRef
+from werkzeug.urls import url_parse
+
+from flask import current_app
+
+from inspirehep.utils import record_getter
+
+
+class AbstractRecordLoader(JsonLoader):
+    """Base for resource-aware record loaders.
+
+    Resolves the refered resource by the given uri by first checking against
+    local resources.
+    """
+
+    def get_record(self, record_type, recid):
+        raise NotImplementedError()
+
+    def get_remote_json(self, uri, **kwargs):
+        parsed_uri = url_parse(uri)
+        # Normalize optional 'http://' protocol in the config.
+        server_name = current_app.config.get('SERVER_NAME')
+        if not server_name.startswith('http://'):
+            server_name = 'http://{}'.format(server_name)
+        parsed_server = url_parse(server_name)
+
+        if parsed_uri.netloc and parsed_uri.netloc != parsed_server.netloc:
+            return super(AbstractRecordLoader, self).get_remote_json(uri,
+                                                                     **kwargs)
+        path_parts = parsed_uri.path.strip('/').split('/')
+        if len(path_parts) < 2:
+            current_app.logger.error('Bad JSONref URI: {0}'.format(uri))
+            return None
+
+        record_type = path_parts[-2]
+        recid = path_parts[-1]
+        res = self.get_record(record_type, recid)
+        return res
+
+
+class ESJsonLoader(AbstractRecordLoader):
+    """Resolve resources by retrieving them from Elasticsearch."""
+
+    def get_record(self, record_type, recid):
+        try:
+            return record_getter.get_es_record(record_type, recid)
+        except record_getter.RecordGetterError:
+            return None
+
+
+class DatabaseJsonLoader(AbstractRecordLoader):
+
+    def get_record(self, record_type, recid):
+        try:
+            return record_getter.get_db_record(record_type, recid)
+        except record_getter.RecordGetterError:
+            return None
+
+
+es_record_loader = ESJsonLoader()
+db_record_loader = DatabaseJsonLoader()
+
+
+def replace_refs(obj, source='db'):
+    """Replaces record refs in obj by bypassing HTTP requests.
+
+    Any reference URI that comes from the same server and references a resource
+    will be resolved directly either from the database or from Elasticsearch.
+
+    :param obj:
+        Dict-like object for which '$ref' fields are recursively replaced.
+    :param source:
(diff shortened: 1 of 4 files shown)


jmartinm:
Looks good to me.

inspirehep/modules/records/json_ref_loader.py
```diff
@@ -0,0 +1,122 @@
...
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
```
jacquerie: Please add the paragraph about CERN being an international organization, following http://legal.web.cern.ch/licensing/software-copyright and http://cds.cern.ch/record/1482206?ln=en.

inspirehep/modules/records/json_ref_loader.py
```diff
@@ -0,0 +1,122 @@
...
+                                     'retrieving record [{1}:{2}]'.format(
+                                         e, record_type, recid))
+        finally:
+            return res
+
+
+class ESJsonLoader(AbstractRecordLoader):
+    """Resolve resources by retrieving them from Elasticsearch."""
+
+    def get_record(self, record_type, recid):
```
jacquerie: This looks useful in general. Maybe its implementation should be promoted to `utils`?

inspirehep/modules/records/json_ref_loader.py
```diff
@@ -0,0 +1,122 @@
...
+
+        return es.get_source(
+            index=search_conf['search_index'],
+            id=str(pid.object_uuid),
+            doc_type=search_conf['search_type'])
+
+
+class DatabaseJsonLoader(AbstractRecordLoader):
+
+    def get_record(self, record_type, recid):
```
jacquerie: Similarly, I wonder if this is more generally useful.

tests/unit/records/test_records_json_ref_loader.py
```diff
@@ -0,0 +1,138 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+import mock
+import pytest
```
jacquerie: Unused.

inspirehep/modules/records/json_ref_loader.py
mihaibivol: @kaplun, Regarding the silencing of errors and moving this logic to utils:
* I this code will be used in many user views and uncaught exceptions will transform into 500s which isn't nice.
Therefore, I'd rather silence the errors, serve the response with None and alert Sentry

If I am to port the logic into a separate module, it should be something that reduces the boilerplate of having everytime the function is called:
```
try:
    utils.get_db_record(recid)
except PIDDoesNotExistError:
     do something
except TransportError:
      do another thing
```

Should I add a new module somewhere that catches these two errors and raises a single type of error instead of:
* PIDDoesNotExistError
* TransportError (elasticsearch error)
* NoResultFound (ORM error)

IMO the usecase for jsonref is
if something goes wrong, return None, because any caller will just try to replace refs and in case of exception would just set the result to none and silence it anyway.

The usecase for an almost "core" library with many calls would be
```
try:
    utils.get_record_from_db(recid)
except RecordUtilsError as e:
    do something
```

Plus, if the library raises a custom Error the code catching the API specific errors could automatically log a custom message that would alert in case of data corruption.

mihaibivol: TL;DR

where should I put a record_getter.py file that has the following methods and types exposed and is there any reason of not collecting all specific DB exceptions into a single one?:
* RecordGetterError(Exception)
* get_record_from_db
* get_record_from_es

jacquerie:
Nevermind what I said about Sentry: its configuration is outdated anyway, so we shouldn't worry about feeding it the right data. I added an item to the DevOps backlog.

inspirehep/modules/records/json_ref_loader.py
kaplun: I am not the best in suggesting how to structure our overlay. @jalavik, @jmartinm what do you think? I guess you can put this `record_getter.py` file simply under utils.

mihaibivol: @kaplun In this PR it's already under utils :)

tests/unit/utils/test_utils_record_getter.py
```diff
@@ -0,0 +1,81 @@
...
+    @record_getter.raise_record_getter_error_and_log
+    def badfn(record_type, recid):
+        raise Exception('message')
+
+    try:
+        badfn(None, None)
+    except record_getter.RecordGetterError as e:
+        pass
+    else:
+        assert False, 'Should raise RecordGetterError'
```
jacquerie: You can use `pytest.raises` for this: https://pytest.org/latest/assert.html#assertions-about-expected-exceptions

tests/unit/utils/test_utils_record_getter.py
```diff
@@ -0,0 +1,81 @@
...
+
+class _MockPID(object):
+
+    def __init__(self, object_uuid):
+        self.object_uuid = object_uuid
+
+
+@mock.patch('inspirehep.utils.record_getter.PersistentIdentifier')
+@mock.patch('inspirehep.utils.record_getter.es')
+def test_get_es_record(es, pid):
```
jacquerie: I'm not a fan of this kind of unit test. If the purpose of a function is _only_ to interact with DB/ES then I think it should only be tested at the integration level. See: http://googletesting.blogspot.ch/2015/01/testing-on-toilet-change-detector-tests.html

tests/unit/utils/test_utils_record_getter.py
```diff
@@ -0,0 +1,81 @@
...
+    assert with_es == rec
+    es.get_source.assert_called_with(
+        index='records-hep',
+        doc_type='hep',
+        id='myuid')
+
+
+@mock.patch('inspirehep.utils.record_getter.PersistentIdentifier')
+@mock.patch('inspirehep.utils.record_getter.Record')
+def test_get_db_record(record, pid):
```
jacquerie: Same observation. IMO this whole file can be deleted. Sorry for the wasted effort : (

jacquerie:
Apart from the over-zealous testing, LGTM.

inspirehep/modules/records/json_ref_loader.py
kaplun: :+1:

mihaibivol:
Kept only the decorator test as I actually found a bug in it because I was forgetting to return in the initial code. Surely everything should break if these methods will be used and are flawed.

jacquerie:
>  Kept only the decorator test as I actually found a bug in it because I was forgetting to return in the initial code.

Fair enough. :+1:

PR #995, opened 2016-03-14
After deployment fixes

jmartinm:


jmartinm:
@jalavik can you have a quick look?

inspirehep/modules/deposit/dojson/fields/literature.py
```diff
@@ -221,7 +221,7 @@ def languages(self, key, value):
     if value not in ('en', 'oth'):
         return [unicode(dict(languages).get(value))]
     else:
-        return [value]
+        return []
```
kaplun: I believe the problem is really that there is the language mentioned in the title (rather than as a field)

kaplun: > Jan, all submissions seem to get (ln en) behind the title.
Neutrino Factory and Muon Collider Collaboration R&D Activities (In en)
https://inspirehep.net/record/1427157
-- Annette

jmartinm: Yes, that comes from the format in production. If there is a language field it is displayed close to the title. For English we don't put anything in the field so nothing is displayed.

kaplun: Oh sorry, then :)

jalavik:
@jmartinm Looks fine. Shame CI is busted tho for the branch. Will check on that when I am back.

jmartinm:
Deployed.

kaplun:
CC: @annetteholtkamp

PR #1000, opened 2016-03-15
global: security loading workaround

mihaibivol:
* Changes config so that flask_security is not needed before being
  loaded.

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -104,7 +104,9 @@ SECURITY_RESET_SALT = "CHANGE_ME"
 
 # User profile
 # ============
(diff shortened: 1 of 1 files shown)


inspirehep/config.py
```diff
@@ -104,7 +104,7 @@ def _(x):
 
 # User profile
 # ============
-USERPROFILES_EXTEND_SECURITY_FORMS = True
+USERPROFILES_EXTEND_SECURITY_FORMS = False
```
kaplun: Can you add a FIXME to remember that this should be investigated at some point?

mihaibivol: Done :+1:

PR #1003, opened 2016-03-15
cite: adds clipboard and adjusts settings 

ioannistsanaktsidis:


inspirehep/modules/search/static/js/search/app.js
```diff
@@ -22,9 +22,13 @@
  */
 
 require([
+    'angular',
+    'angular-sanitize',
+    'angular-ui-bootstrap',
```
jmartinm: This first 3 are not needed, since `settings.js` already defines the `deps` so they will be correctly loaded.

inspirehep/modules/search/static/js/search/app.js
```diff
@@ -22,9 +22,13 @@
...
     'invenio-search',
     'inspirehep-search',
-    'inspirehep'
+    'inspirehep',
+    'ngclipboard'
```
jmartinm: Instead of adding it here, it is better if we add it in the `settings.js` close to the library that really needs it as a dependency - https://github.com/inspirehep/inspire-next/blob/invenio3/inspirehep/modules/theme/static/js/settings.js#L69

inspirehep/modules/theme/bundles.py
```diff
@@ -63,9 +63,10 @@
         "toastr": "~2.1.2",
         "clipboard": "~1.5.8",
         "flightjs": "~1.5.0",
-        "angular": "~1.4.8",
-        "inspirehep-js": "git+https://git@github.com/inspirehep/inspirehep-js.git"
-
+        "angular": "~1.4.9",
+        "angular-sanitize": "~1.4.9",
+        "inspirehep-js": "git+https://git@github.com/inspirehep/inspirehep-js.git",
+        "ngclipboard": "~1.1.1"
```
jmartinm: This file does not need to be modified. `ngclipboard ` is a dependency of `inspirehep-js` so it should be defined in the [package.json](https://github.com/inspirehep/inspirehep-js/blob/master/package.json). Then it will get installed correctly.

inspirehep/modules/theme/static/js/settings.js
```diff
@@ -35,7 +35,9 @@ require.config({
     flight: 'node_modules/flightjs/build/flight',
     hgn: 'node_modules/requirejs-hogan-plugin/hgn',
     hogan: 'node_modules/hogan.js/web/builds/3.0.2/hogan-3.0.2.amd',
+    "ngclipboard": "node_modules/ngclipboard/src/ngclipboard",
```
jmartinm: please, keep all in single quotes for consistency in the file.

inspirehep/modules/theme/static/js/settings.js
```diff
@@ -35,7 +35,9 @@ require.config({
...
     inspirehep: 'node_modules/inspirehep-js/dist/inspirehep',
+    "inspire-clipboard": "js/inspire_clipboard",
```
jmartinm: lets call everything `inspirehep` instead of `inspire`

inspirehep/modules/theme/static/js/settings.js
jmartinm: They are now listed alphabetically, can you keep the order? :)

inspirehep/modules/theme/static/scss/format/brief-results.scss
```diff
@@ -508,4 +508,17 @@ i.fa-chevron-right.facet-slider {
 
 .brief-results-sortby {
 	color: #737373;
+}
+
+div > .modal-dialog {
+    width: 900px;
```
jmartinm: This should be tested in all screens. Instead of absolute pixels, it should be a percentage so that it displays correctly in all viewports.

inspirehep/modules/theme/static/scss/format/brief-results.scss
```diff
@@ -508,4 +508,17 @@ i.fa-chevron-right.facet-slider {
...
+}
+
+#textarea-cite {
+    resize: none;
+    width: 868px;
```
jmartinm: Same as before.

inspirehep/modules/theme/static/scss/format/brief-results.scss
```diff
@@ -508,4 +508,17 @@ i.fa-chevron-right.facet-slider {
...
+    min-height: 237px;
+    background-color: #f5f5f5;
```
jmartinm: What is this color? Is it something we use in other places and we can add as a [variable](https://github.com/inspirehep/inspire-next/blob/invenio3/inspirehep/modules/theme/static/scss/base/variables.scss) ?

ioannistsanaktsidis: this is the grey that was used in the <pre> tag that now is a <textarea>. I don't think we are using it anywhere else. Should I keep it like this?

inspirehep/modules/theme/static/scss/format/brief-results.scss
```diff
@@ -508,4 +508,17 @@ i.fa-chevron-right.facet-slider {
...
+
+div > .modal-dialog {
+    width: 63%;
+}
+
+#textarea-cite {
+    resize: none;
+    width: 100%;
+    min-height: 237px;
+    background-color: #f5f5f5;
```
jmartinm: better use variable names here. see for example https://github.com/inspirehep/inspire-next/commit/e31d1bfcc2d7c10906201cdb14715b0ac81f2e75

jmartinm:
The commits are the same as in https://github.com/inspirehep/inspire-next/pull/1024/commits but with different file modifications.

ioannistsanaktsidis:
@jmartinm this one can also go in, it is for the clipboard and the settings  ;)

ioannistsanaktsidis:
Actually I will close this PR. This one https://github.com/inspirehep/inspire-next/pull/1024/ applies the same changes. We don't this one anymore.

PR #1009, opened 2016-03-16
Fix for prod https:// SERVER_NAME in dojson and ref_loader

mihaibivol:
* Fixes double http:// prepending when SERVERN_NAME is
  http://inspirehep.net

cc @kaplun Almost forgot to fix this :)

### DIFF
--- a/inspirehep/dojson/utils.py
+++ b/inspirehep/dojson/utils.py
@@ -19,8 +19,8 @@
 
 """DoJSON related utilities."""
 
-import pkg_resources
 import six
+import re
 
 try:
     from flask import current_app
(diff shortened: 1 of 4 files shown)


jacquerie:
Can you add an assertion for this in `tests/unit/dojson/test_dojson_utils.py`?

jacquerie:
Nitpick: s/SERVERN_NAME/SERVER_NAME/ in the commit.

inspirehep/dojson/utils.py
```diff
@@ -158,7 +158,10 @@ def get_record_ref(recid, record_type='record'):
         server = current_app.config.get('SERVER_NAME', default_server)
     else:
         server = default_server
-    return {'$ref': 'http://{}/api/{}/{}'.format(server, record_type, recid)}
+    # This config might also be http://inspirehep.net.
+    if not server.startswith('http://'):
```
kaplun: What about HTTPS? Invenio is moving to be full HTTPS only (e.g. in default Dockerfile recommendations etc.)

mihaibivol: Is then SERVER_NAME set up to be https:// ?

mihaibivol: If so, then I have the same error also in the json ref resolver. I didn't want it to depend on dojson as dojson has a narrow use-case.

kaplun: SERVER_NAME could be indeed either http or https depending on the specific installation.

mihaibivol:
Now it should work for all `SERVER_NAME`

PR #1011, opened 2016-03-17
theme: jsonref in jinja2filters

mihaibivol:
* Uses replace_refs in publication_info filter.

### DIFF
--- a/tests/unit/theme/test_theme_jinja2filters.py
+++ b/tests/unit/theme/test_theme_jinja2filters.py
@@ -46,6 +46,15 @@ def jinja_mock_env():
     return JinjaMockEnv()
 
 
+@pytest.fixture
+def mock_replace_refs():
+    def get_replace_refs_mock(title, control_numbers):
+        control_numbers_map = {c[0]['$ref']: c[1] for c in control_numbers}
+        return lambda o, s: {'title': title,
+                             'control_number': control_numbers_map[o['$ref']]}
+    return get_replace_refs_mock
+
+
 @mock.patch('inspirehep.modules.theme.jinja2filters.current_app.jinja_env.get_template')
 def test_apply_template_on_array_returns_empty_list_on_empty_list(g_t, jinja_env):
     g_t.return_value = jinja_env.from_string('{{ content }}')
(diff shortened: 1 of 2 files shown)


jacquerie:
Were tests failing locally? BTW, can you also add another commit reusing your `inspirehep/utils/record_getter.py` where relevant in this file?

mihaibivol:
Forgot to fix the tests before commiting. Only checked in the interface. Now they should pass with a kinda ugly workaround

mihaibivol:
I don't see anything else in `jinja2filters.py` that can be refactored to use `record_getter`

jacquerie:
Whoops, you're right. I got it confused with another file.

mihaibivol:
I'm not sure about the `control_number` solution to get recids instead of parsing the $ref. Also I don't know how to make the tests cleaner when populating `control_number`

tests/unit/theme/test_theme_jinja2filters.py
```diff
@@ -910,14 +913,15 @@ def test_publication_info_with_pub_info_and_conf_info(g_s, g):
...
+@mock.patch('inspirehep.modules.theme.jinja2filters.replace_refs')
+def test_publication_info_from_conference_recid_and_not_parent_recid(r_r):
     with_title = Record({
         'title': '20th International Workshop on Deep-Inelastic Scattering and Related Subjects'
     })
 
-    g_s.return_value = with_title
+    r_r.side_effect = lambda o, s: dict([('control_number',
+                                          o['$ref'].split('/')[-1])] +
+                                        with_title.items())
```
jacquerie: To clean things up a bit, since you are repeating this `lambda` four times you could make it a pytest fixture local to this file, inject it explicitly in the function that need it as `mock_r_r`, then write `r_r.side_effect = mock_r_r`.

jacquerie: That said, this is a little more logic in a mock than what I'm comfortable with: http://googletesting.blogspot.ch/2014/07/testing-on-toilet-dont-put-logic-in.html. Can you rewrite this more explicitly, maybe at the cost of boilerplate/repetition?

mihaibivol: I still have a bit of logic that decides which `control_number` is returned by which ref object. I'm not sure I can make it more explicit since I don't want to assert any order in the calls.

mihaibivol:
I still have some logic, but at least it isn't parsing objects and just recieves a mapping for the returned `control_number` for each ref object.

PR #1015, opened 2016-03-18
Docker compose setup for dev and integration tests

mihaibivol:
## Howto:

Get the latest Docker setup.

* If you are using Mac, please build a simple box with `docker-engine` above `1.10` and `docker-compose` above `1.6.0`.

Make sure you can run docker without sudo.
* `id $USER`
If you are not in the `docker` group, run the following command and then restart `docker`. If this doesn't work, just restart your machine :)
*  `sudo usermod -a -G docker $USER`

Get docker-compose:
```shell
sudo pip install docker-compose
```

* Add `DOCKER_DATA` env variable in your `.bashrc` or `.zshrc`. In this directory you will have all the persistent data between Docker runs.
```
export DOCKER_DATA=~/inspirehep_docker_data/
```
By default the virtualenv and everything else will be kept on `/tmp` and they will be available only until the next reboot.

* Install a host persistent venv and build assets
```shell
docker-compose pull
docker-compose -f docker-compose.deps.yml run --rm pip
docker-compose -f docker-compose.deps.yml run --rm assets
```

* Run the service locally
```shell
docker-compose up
```
Go to `localhost:5000`

* Populate database

```shell
docker-compose run --rm web scripts/recreate_records
```

* Run tests in an __isolated__ environment:

```shell
docker-compose -f docker-compose.test.yml run --rm test
docker-compose -f docker-compose.test.yml kill
docker-compose -f docker-compose.test.yml rm
```

__The kill command is mandatory, otherwise you will use the test DB instead of the correct one in dev.__

I don't know how to kill all the other services just after the run command exited.

## Extra useful tips

* Run a random shell

After `docker-compose up` just run
```
docker-compose run --rm web inspirehep shell
```

* Reload code in a worker

With `docker-compose up` just run:
```
docker-compose restart worker
```

* Quick and safe reindex

With `docker-compose up` just run:
```
docker-compose restart worker && docker-compose run --rm web scripts/recreate_records
```
## TODOs:

* [x] Fix indexing
* [x] Add test compose file
* [x] Make data persistent for dev

## Future Stuff:

* Use build arg to actually create a `staging` or `prod` image.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,50 +17,53 @@
 # along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 
-addons:
-  postgresql: "9.4"
-  apt:
-    packages:
-      - nodejs
-
 notifications:
   email: false
 
+# Not needed at the moment since we use Docker.
+# If this kicks in we can try to use the cache
 sudo: false
 
 language: python
 
-cache:
-  directories:
-    - $HOME/.cache/pip
-    - $HOME/.nvm
-
 services:
-  - postgresql
-  - elasticsearch
-  - rabbitmq
-  - redis
+  - docker
 
 env:
-  - SQLALCHEMY_DATABASE_URI="postgresql+psycopg2://postgres@localhost:5432/inspirehep"
+  global:
+    - DOCKER_DATA="$HOME/docker_data"
+    - DOCKER_VERSION=1.10.1-0~trusty
+    - DOCKER_COMPOSE_VERSION=1.6.0
 
 python:
   - "2.7"
 
 before_install:
-  - "travis_retry pip install --upgrade pip setuptools py mock pytest"
-  - "travis_retry pip install twine wheel coveralls"
-  - "npm update && npm install --silent -g node-sass clean-css uglify-js requirejs"
+  - travis_retry pip install coveralls
+  # List available docker versions.
+  - apt-cache madison docker-engine
+  # Update Docker to 1.10.1 https://graysonkoonce.com/managing-docker-and-docker-compose-versions-on-travis-ci/.
+  - sudo apt-get -o Dpkg::Options::="--force-confnew" install -y docker-engine=${DOCKER_VERSION}
+  # Add docker-compose at the version specified in ENV.
+  - sudo rm /usr/local/bin/docker-compose
+  - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+  - chmod +x docker-compose
+  - sudo mv docker-compose /usr/local/bin
 
 install:
-  - "travis_retry pip install -r requirements.txt --pre"
-  - "scripts/clean_assets"
-
-before_script:
-  - "sleep 10"  # Allow ES to start.
+  - "travis_retry docker-compose pull"
+  - "travis_retry docker-compose build"
+  - "travis_retry docker-compose -f docker-compose.deps.yml run --rm pip"
+  - "docker-compose -f docker-compose.deps.yml run --rm assets"
 
 script:
-  - "python setup.py test"
+  - "docker-compose -f docker-compose.test.yml run --rm test"
+
+after_script:
+  - "docker-compose -f docker-compose.test.yml kill"
+  - "docker-compose -f docker-compose.test.yml rm -f"
 
 after_success:
+  # Coverage report contains Docker paths. We replace them, so that we can run Coveralls.
+  - sed -i 's@\"/code/@'"\"$(pwd)/"'@g' .coverage
   - coveralls
--- a/docker-compose.deps.yml
+++ b/docker-compose.deps.yml
@@ -1,3 +1,25 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
 version: '2'
 
 services:
--- a/docker-compose.deps.yml
+++ b/docker-compose.deps.yml
@@ -1,0 +1,21 @@
+version: '2'
+
+services:
+  pip:
+    extends:
+      file: services.yml
+      service: base
+    command: pip-accel install -r requirements.txt --pre
+    volumes_from:
+      - static
+  assets:
+    extends:
+      file: services.yml
+      service: base
+    command: scripts/clean_assets
+    volumes_from:
+      - static
+  static:
+    extends:
+      file: services.yml
+      service: static
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -1,0 +1,55 @@
+version: '2'
+
+services:
+  service_base:
+    # Overrides default inspirehep config.
+    extends:
+      file: services.yml
+      service: base
+    environment:
+      - SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://inspirehep:dbpass123@database:5432/inspirehep
+      - BROKER_URL=amqp://guest:guest@rabbitmq:5672//
+      - CELERY_RESULT_BACKEND=amqp://guest:guest@rabbitmq:5672//
+      - CACHE_REDIS_URL=redis://redis:6379/0
+      - SEARCH_ELASTIC_HOSTS=indexer
+
+  test:
+    extends:
+      service: service_base
+    command: python setup.py test
+    volumes_from:
+      - static
+    links:
+      - database
+      - indexer
+      - rabbitmq
+      - redis
+    depends_on:
+      - worker
+  worker:
+    extends:
+      service: service_base
+    command: celery worker -E -A inspirehep.celery --loglevel=INFO --purge
+    volumes_from:
+      - static
+    links:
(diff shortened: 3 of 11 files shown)


mihaibivol:
@inspirehep/inspire-dev Feel free to comment on how these commands can be further simplified. e.g. I'm not sure how to drop the lengthy `docker-compose -f docker-compose.build.yml`

services.yml
```diff
@@ -0,0 +1,19 @@
+version: '2'
+
+services:
+  base:
+    build: .
+    image: inspirehep/inspire-next
+  indexer:
+    image: elasticsearch
+    command: elasticsearch -Dcluster.name="inspire"
```
kaplun: Is there some preparatory steps where one can install icu and reindexing plugin?

docker-compose.deps.yml
```diff
@@ -0,0 +1,22 @@
...
+     volumes_from:
+       - static
+   assets:
+     extends:
+       file: services.yml
+       service: base
+     command: scripts/clean_assets
+     volumes_from:
+       - static
+  static:
```
jacquerie: Here I get an error about indentation
```shell
ERROR: yaml.parser.ParserError: while parsing a block mapping
  in "./docker-compose.deps.yml", line 1, column 1
expected <block end>, but found '<block mapping start>'
  in "./docker-compose.deps.yml", line 18, column 3
```
which is easy to fix, but I wonder why your docker wasn't complaining.
```shell
$ docker --version
Docker version 1.10.3, build 20f81dd
```

jacquerie:
I almost got this working. One problem under Mac OS is that ports aren't forwarded, so one has to type `UGLY_URL:5000` to access the web app, or set up port forwarding in Virtualbox, or forward all ports using a local nginx (http://stackoverflow.com/questions/25327012/access-docker-from-external-machine-in-network).

jacquerie:
Actually, one **has** to enable one of these solutions, otherwise the application will return a 404 (I have no idea why...).

Other than that, I measured that this solution runs tests 5X slower than before, but achieves better isolation and reproducibility, a tradeoff which I think we can afford (it's 12 seconds vs 1 minute). Moreover, disk usage is just 300MB, probably thanks to the `--rm` option.

So it's a :+1: for me. @kaplun, if we decide to go forward on this (it's your call anyway!) I think we should have a Monday meeting to explain the new developer workflow that @mihaibivol is building.

kaplun:
Eheh, Indeed I'd like to understand it as well... What do you think to do when we introduce beard and the thousands of compiled dependencies?

mihaibivol:
Currently everything that we "build" is cached in /virtualenv in the `base` service. This means that for dev at least you only have to install the requirements once and update them exactly as you would do when using a local venv. I am still not sure what's the best take for running this setup on Travis.

.travis.yml
```diff
@@ -17,50 +17,47 @@
...
-  apt:
-    packages:
-      - nodejs
-
 notifications:
   email: false
 
+# Not needed at the moment since we use Docker.
+# If this kicks in we can try to use the cache
 sudo: false
```
jacquerie: Maybe we can give up and have `sudo: required`?

mihaibivol: The builds should start failing when they implement this in Docker (probably not too soon) and we'll know to update our build

.travis.yml
```diff
@@ -17,50 +17,47 @@
...
+    - DOCKER_VERSION=1.10.1-0~trusty
 
 python:
   - "2.7"
 
 before_install:
-  - "travis_retry pip install --upgrade pip setuptools py mock pytest"
-  - "travis_retry pip install twine wheel coveralls"
-  - "npm update && npm install --silent -g node-sass clean-css uglify-js requirejs"
+  - pip install docker-compose
```
jacquerie: What's the reason we don't use the linked post's solution for fixing `docker-compose` version?

docker-compose.test.yml
```diff
@@ -0,0 +1,55 @@
...
+    extends:
+      service: service_base
+    command: celery worker -E -A inspirehep.celery --loglevel=INFO --purge
+    volumes_from:
+      - static
+    links:
+      - database
+      - indexer
+      - rabbitmq
+      - redis
```
jacquerie: Aren't we missing `workermon`? It's seldom useful, but it's invaluable in those cases.

dockerfiles/docker_entrypoint.sh
```diff
@@ -0,0 +1,16 @@
...
+    virtualenv /virtualenv
+    source /virtualenv/bin/activate
+    pip install --upgrade pip-accel setuptools
+else
+    source /virtualenv/bin/activate
+fi
+
+clear_pycache() {
+    find \( -name __pycache__ -o -name '*.pyc' \) -delete
+}
```
jacquerie: This is just defined, not executed?

scripts/clear_pycache
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sudo find \( -name __pycache__ -o -name '*.pyc' \) -delete
```
jacquerie: As I was saying the other day, not cross-platform. On Mac OS X:
```
% find \( -name __pycache__ -o -name '*.pyc' \) -delete
usage: find [-H | -L | -P] [-EXdsx] [-f path] path ... [expression]
       find [-H | -L | -P] [-EXdsx] -f path [path ...] [expression]
```

scripts/recreate_records
```diff
@@ -36,26 +36,6 @@ if [ ! -x "$(command -v celery)" ]; then
...
-CELERY_PIDFILE="${VIRTUAL_ENV}/worker.pid"
-if [ -f "${CELERY_PIDFILE}" ]; then
-	echo "----> Purging all pending messages"
-	celery purge -f -A inspirehep.celery --workdir="${VIRTUAL_ENV}"
-else
-	celery worker -E -A inspirehep.celery --loglevel=${CELERY_LOGLEVEL} --workdir="${VIRTUAL_ENV}" --autoreload --pidfile="${CELERY_PIDFILE}" --purge &
-	CELERY_PID=$!
-	trap 'cleanup' 0 1 2 3 6
-fi
-
```
jacquerie: As a general indication, but not something that should be done now, @jalavik suggests that we keep the doors open for the developers that don't want to ride the Docker train. So this script will have to be split in two cases, one run by docker, one run by non-Docker users.

docker-compose.test.yml
mihaibivol: can you add it and test it? I'm not sure how it exactly works

.travis.yml
mihaibivol: This has the same version and makes .travis.yml less verbose. Also it's the same as I did locally.

jacquerie: Yeah, but isn't his approach more robust? For example, let's say that the version on pip updates to 2.0.0, which is incompatible with the Docker version we require.

docker-compose.test.yml
jacquerie: Sure. Its actual name is `flower`, by the way.

services.yml
```diff
@@ -0,0 +1,19 @@
+version: '2'
+
+services:
+  base:
+    image: mihaibivol/python-base
```
jalavik: I added you to the inspirehep dockerhub organization, so can you push these there?

jalavik:
Awesome stuff @mihaibivol. I will give this a closer look and test.

docker-compose.test.yml
jacquerie: Added [here](https://github.com/jacquerie/inspire-next/commit/f321efc7fee525838108621f5687055051030043) and tested working on my machine.

jacquerie:
If Mac OS X developers want to test the Docker way here's a Vagrantfile to help them do so: https://github.com/jacquerie/inspire-next/commit/7ddb99c8779d6374be46f9808df81040fe927a1d.
 Unfortunately I wasn't able to have docker-machine running + ES (known issue of mismatched UIDs), so for now Vagrant is the way.

jacquerie:
...you're going to hate me a bit, but here's another commit you should cherry-pick and squash: https://github.com/jacquerie/inspire-next/commit/071ce8543b59afbffe31828887bf37d101bf18f6

This one adds the license headers to the files you introduced.

.travis.yml
```diff
@@ -17,50 +17,56 @@
...
+  - apt-cache madison docker-engine
+  # Update Docker to 1.10.1 https://graysonkoonce.com/managing-docker-and-docker-compose-versions-on-travis-ci/.
+  - sudo apt-get -o Dpkg::Options::="--force-confnew" install -y docker-engine=${DOCKER_VERSION}
+  # Add docker-compose at the version specified in ENV.
+  - sudo rm /usr/local/bin/docker-compose
+  - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
+  - chmod +x docker-compose
+  - sudo mv docker-compose /usr/local/bin
+  # Put coveralls
+  - pip install coveralls
```
jacquerie: `travis_retry` to avoid trivial build errors?

jacquerie: And maybe remove the comment and put it before everything, where it was. Or does that break stuff?

.travis.yml
```diff
@@ -17,50 +17,56 @@
...
-  - "python setup.py test"
+  - "docker-compose -f docker-compose.test.yml run --rm test"
+
+after_script:
+  - "docker-compose -f docker-compose.test.yml kill"
+  - "docker-compose -f docker-compose.test.yml rm -f"
 
 after_success:
+  # Coverage report contains Docker paths. Replace so
+  # we can run coveralls from Travis
```
jacquerie: Nit: `# Coverage report contains Docker paths. We replace them, so that we can run Coveralls.`

scripts/clear_pycache
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sudo find . \( -name __pycache__ -o -name '*.pyc' \) | xargs rm -rf
```
jacquerie: I wonder if we can be a little smarter and use `sudo` only when it's actually needed. But it's not a blocker for this PR, just a random thought.

jacquerie:
LGTM. Cherry-pick those two commits, squash, incorporate those nits and :shipit:

kaplun:
Guys you have my full trust on this branch :-) Don't know if @jalavik wanted to further comment...

jacquerie:
:ship: :ship: :ship:

jalavik:
Great stuff guys. Note that it is still optional if one wants to develop using docker. This branch does not enforce it.

jacquerie:
BTW we kept the Vagrantfile out of this PR because we are waiting to see if Docker for Mac could be an alternative: https://blog.docker.com/2016/03/docker-for-mac-windows-beta/.

rikirenz:
I have tried to run the command `docker-compose -f docker-compose.test.yml run --rm test` but an error is triggered. Is it my fault? am I missing something?

PR #1018, opened 2016-03-21
tests: test missing postfeedback branch

jacquerie:


tests/unit/theme/test_theme_views.py
```diff
@@ -90,12 +107,9 @@ def test_postfeedback_provided_email(email_app):
         assert response.status_code == 200
 
 
-@mock.patch('inspirehep.modules.theme.views.current_user')
-def test_postfeedback_logged_in_user(current_user, email_app):
+@mock.patch('inspirehep.modules.theme.views.current_user', user_with_email)
+def test_postfeedback_logged_in_user(email_app):
     """Falls back to the email of the logged in user."""
-    current_user.is_anonymous = False
```
jacquerie: @eamonnmag This line was causing dependencies in tests, since it's modifying an application global.

jacquerie:
This should be straightforward, but if somebody wants to review they're welcome.

PR #1020, opened 2016-03-23
WIP search: google like search fields

Panos512:
* Searches on different fields based on the index we are searching on.

### DIFF
--- a/inspirehep/modules/search/query.py
+++ b/inspirehep/modules/search/query.py
@@ -24,14 +24,14 @@
 
 """INSPIRE Query class to wrap a query object from invenio_search."""
 
-from flask import request, current_app
+from flask import current_app, request
 
-from invenio_search.api import Query
-from invenio_search.utils import query_enhancers, search_walkers
+from invenio_query_parser.ast import MalformedQuery
 
 from invenio_records_rest.errors import InvalidQueryRESTError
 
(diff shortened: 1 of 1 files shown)


inspirehep/modules/search/query.py
```diff
@@ -71,16 +93,9 @@ def build(self, **kwargs):
...
-                            "title^3",
-                            "title.raw^10",
-                            "abstract^2",
-                            "abstract.raw^4",
-                            "author^10",
-                            "author.raw^15",
-                            "reportnumber^10",
-                            "eprint^10",
-                            "doi^10"]}}
+                        "fields": google_search_fields[self.index]
```
jacquerie: Design: this could even be a method call instead of a lookup, so we can inject logic if/when we A/B test search.

kaplun: @jacquerie you mean:
```python
"fields": self.google_search_fields()
```
?

jacquerie: Probably not on `self`, but yes.

kaplun: Jacopo where are you?

kaplun:
@Panos512 this PR needs to be closed and reopened against `master` branch.

I can take care of it :)

kaplun:
Superseded by #1098

PR #1021, opened 2016-03-24
authors: submission fixes

jalavik:


inspirehep/modules/authors/views.py
```diff
@@ -42,20 +42,20 @@
     url_for, \
     redirect
 from flask_breadcrumbs import register_breadcrumb
+from flask.ext.login import current_user
+from flask_login import login_required
```
jacquerie: Can't this be both `from flask_login`, or am I missing something?

jalavik: Yeah, it can! Thanks!

inspirehep/modules/authors/views.py
```diff
@@ -290,11 +291,13 @@ def submitnew():
     visitor = DataExporter()
     visitor.visit(form)
 
-    myobj = DbWorkflowObject.create_object(id_user=current_user.get_id())
-    myobj.set_data(visitor.data)
+    workflow_object = DbWorkflowObject.create_object(id_user=current_user.get_id())
+    workflow_object.data = visitor.data
+    workflow_object.save()
+    db.session.commit()
```
kaplun: Shouldn't this be part of a begin_nested?

jalavik: I don't think it is strictly required as inside the create_object and save calls there are nested transactions and as this is at top-level before the request ends we need the specific commit call.

jalavik:
WIP awaiting new updates to invenio-workflows

PR #1022, opened 2016-03-29
theme: add MathJax

jacenkow:
* Extends JS dependencies in order to render mathematical notation
  using MathJax. ~~NOTE: works only in ASSETS_DEBUG = True.~~

### DIFF
--- a/inspirehep/modules/search/templates/search/search.html
+++ b/inspirehep/modules/search/templates/search/search.html
@@ -21,6 +21,8 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 #}
 
+{% from "inspirehep_theme/format/record/Inspire_Default_HTML_general_macros.tpl" import mathjax %}
+
 {% extends config.SEARCH_UI_BASE_TEMPLATE %}
 
 {%- block css %}
@@ -30,6 +32,7 @@
 
 {%- block javascript %}
   {{ super() }}
+  {{ mathjax() | safe }}
   {% assets "invenio_search_ui_search_js" %}<script src="{{ ASSET_URL }}"></script>{% endassets %}
 {%- endblock javascript %}
 
--- a/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_detailed.tpl
+++ b/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_detailed.tpl
@@ -104,6 +104,7 @@
 
 {% block javascript %}
   {{ super() }}
+  {{ mathjax() | safe }}
   {%- assets "inspirehep_detailed_js" %}
     <script src="{{ ASSET_URL }}"></script>
   {%- endassets %}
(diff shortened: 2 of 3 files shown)


jacenkow:
screen shot 2016-03-29 at 13 03 39

CC: @kaplun

jacenkow:
Unfortunately works only with ```ASSETS_DEBUG = True```, ie. when it's **not** processed by ```almond```.

inspirehep/modules/theme/static/js/settings.js
```diff
@@ -39,6 +39,7 @@ require.config({
     'inspirehep-search': 'node_modules/inspirehep-search-js/dist/inspirehep-search',
     'invenio-search': 'node_modules/invenio-search-js/dist/invenio-search-js',
     jquery: 'node_modules/jquery/jquery',
+    'mathjax': 'http://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS-MML_HTMLorMML&amp;delayStartupUntil=configured',
```
kaplun: Maybe put here `'//cdn.' so that it works with https too.

kaplun:
@eamonnmag how do you do in HepData?

eamonnmag:
I don't use requirejs since it's terrible. I just add the mathjax cdn script on the page templates where it's required.

jacenkow:
@kaplun done.

kaplun:
Good! Can you squash the 2 commits?

jacenkow:
@kaplun ba da bing ba da boom.

jacenkow:
@jmartinm, note that all `AngularJS` apps will need a special directive with sth like `mathjax.Hub.Queue(["Typeset", mathjax.Hub])`

jalavik:
@jacenkow Actually, we already have MathJax inclusion available here: https://github.com/inspirehep/inspire-next/blob/invenio3/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_general_macros.tpl#L269-L288

So we can use this on the pages you need to. However, the version should be upgraded to 2.6 release series.

jacenkow:
@jalavik I see, however it makes sense to me of putting `MathJax` once in `javascript.html`, since we use it (almost) in every page instead of calling macro in every template page.

eamonnmag:
It's not used in every page. And since it's a fairly heavy dependency, it's better to load it only when needed.

jacenkow:
@jalavik done. 

Since we use templates to directly serve different endpoints (`literature`, `authors`, etc.) we are not extending `inspirehep_theme/records/base.html` by these ones anymore. Just leaving a note about the template that is still used by `metadata.html`, `helpers.html`, `tips.html` and `inspire_approval.html`.

jalavik:
Thanks @jacenkow!

jacenkow:
@jmartinm regarding `AngularJS` apps, 
```js
$timeout(function() {
  MathJax.Hub.Queue(["Typeset", MathJax.Hub]);
}, 0);
```
works for me.

PR #1023, opened 2016-03-29
search: enables refersto:recid:xxx search

Panos512:
* When `refersto` is followed by `recid` the search will work as intended.

### DIFF
--- a/inspirehep/modules/search/walkers/elasticsearch.py
+++ b/inspirehep/modules/search/walkers/elasticsearch.py
@@ -90,6 +90,11 @@ class ElasticSearchDSL(object):
     @visitor(Value)
     def visit(self, node):
         def query(keyword):
+            # FIXME: This is a temporary hack that should be removed when
+            # nested keywords search is implemented.
+            if str(node.value).startswith('recid:'):
+                node.value = node.value[len('recid:'):]
+
             fields = self.get_fields_for_keyword(keyword, mode='a')
             if fields == ['authors.full_name', 'authors.alternative_name']:
                 return {"bool":
--- a/tests/unit/search/test_search_query.py
+++ b/tests/unit/search/test_search_query.py
@@ -465,7 +465,6 @@ def test_journal_colon():
     assert expected == result
 
 
-@pytest.mark.xfail(reason='extra "recid:" added to the ES query')
 def test_refersto_colon_recid_colon():
     query = InspireQuery('refersto:recid:1286113')
 


inspirehep/modules/search/walkers/elasticsearch.py
```diff
@@ -90,6 +90,12 @@ def visit(self, node):
     @visitor(Value)
     def visit(self, node):
         def query(keyword):
+
+            # FIXME: This is a temporary hack that should be removed when
+            # nested keywords search is implemented
```
jacquerie: Nit: you can remove the space between the function declaration and the comment. Maybe indent the text and add a dot at the end:
```
# FIXME: This is a temporary hack that should be removed when
#              we implement proper nested keywords search.
```

inspirehep/modules/search/walkers/elasticsearch.py
```diff
@@ -90,6 +90,12 @@ def visit(self, node):
...
+            if str(node.value).startswith('recid:'):
+                node.value = node.value[len('recid:'):]
```
jacquerie: Were you as surprised as me to see that there's no method in the standard library for this? http://stackoverflow.com/a/16891418/374865

jacquerie:
Actually I forgot the most important thing! Please make sure that [this test](https://github.com/inspirehep/inspire-next/blob/a434fa5c04ea3799909510d08b170549b023e503/tests/unit/search/test_search_query.py#L468-L484) now works as expected, and delete the `xfail` decorator if that's the case.

inspirehep/modules/search/walkers/elasticsearch.py
Panos512: Totaly. It's the first time Python failed me on something like this tbh :cry:

jacquerie:
Gah, sorry to be so annoying: please remove the backticks in the commit title.

Panos512:
@jacquerie done :)

jacquerie:
Will merge as soon as Travis is done.

inspirehep/modules/search/walkers/elasticsearch.py
```diff
@@ -90,6 +90,11 @@ def visit(self, node):
     @visitor(Value)
     def visit(self, node):
         def query(keyword):
+            # FIXME: This is a temporary hack that should be removed when
+            # nested keywords search is implemented.
+            if str(node.value).startswith('recid:'):
+                node.value = node.value[len('recid:'):]
+
```
kaplun: We will have to drop this specific behavior once we enable second order operators...

Panos512: Yes of course. But since we have the functionality that enables us to search references based on recids why not use it? 

Also when we introduce second order operations this will just be ignored (until someone sees the fixme tag and actually remove it) since there will be no values starting with `recid:` . So we don't risk braking anything in the future (I hope :smiley: ).

kaplun: :+1:

kaplun: Actually I haven't seen the nice FIXME right above :dancer:

PR #1024, opened 2016-03-29
Invenio3 formatter with angular

ioannistsanaktsidis:


inspirehep/modules/records/serializers/schemas/latexus_schema.py
```diff
@@ -0,0 +1,35 @@
...
+
+from __future__ import absolute_import, print_function
+
+from marshmallow import Schema, fields
+
+
+class RecordSchemaLATEXUSV1(Schema):
+    """Schema for latexus v1 in JSON."""
+
+    latexus = fields.Str()
```
kaplun: Is this a stub?

ioannistsanaktsidis: Nope, it is not. This  """Marshmallow bibtex schema."""  should be  """Marshmallow latexus schema."""

inspirehep/modules/formatter/views.py
```diff
@@ -0,0 +1,125 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+import itertools
```
jmartinm: All of these will disappear since we use the `invenio-records-rest` API right?

By passing the `Accept` header with the format we want it will return bibtex, latex and so on.

ioannistsanaktsidis: Yes exactly, all of them will disappear. I just keep them as reference now to check the logic sometimes. I will remove them when it is ready for merging.

jmartinm: And if it is not needed we can probably remove the whole `formatter` module.

ioannistsanaktsidis: Could be the case, everything now works with the serializers. Formatter will not be needed anymore in the end of the day.

inspirehep/modules/records/serializers/bibtex_serializer.py
```diff
@@ -0,0 +1,43 @@
...
+    """JSON bibtex format serializer."""
+
+    def serialize(self, pid, record, links_factory=None):
+        """Serialize a single bibtex from a record.
+        :param pid: Persistent identifier instance.
+        :param record: Record instance.
+        :param links_factory: Factory function for the link generation,
+                              which are added to the response.
+        """
+        return Bibtex(record).format()
```
jmartinm: Did you check that this works? As the schema has the field `bibtex` - https://github.com/inspirehep/inspire-next/pull/1024/files#diff-cc384d6e061b5ef5cce6d004dfa7984aR35 I am wondering if you need to return a structure with such a field inside. (like we do for brief response https://github.com/inspirehep/inspire-next/blob/invenio3/inspirehep/modules/records/serializers/json.py#L79

ioannistsanaktsidis: I don't think it is necessary. I tried it like this and seems to work good.

inspirehep/utils/bibtex.py
```diff
@@ -0,0 +1,679 @@
...
+                                dashpos = page.find('-')
+                                break
+                        else:
+                            dashpos = pages.find('-')
+                        if dashpos > -1:
+                            pages = pages.split('-')[0]
+                try:
+                    if journal and (volume != '' or pages != ''):
+                        recid = self.record['control_number', '']
+                        record = get_es_record_from_doc_type('literature',
```
jmartinm: How does this work? Can it be `literature` and `journals` ?

inspirehep/utils/record_getter.py
```diff
@@ -73,6 +73,17 @@ def get_es_record(record_type, recid):
 
 
 @raise_record_getter_error_and_log
+def get_es_record_from_doc_type(record_type, recid, collection):
```
jmartinm: What is the difference between this one and `get_es_record` ?

jmartinm:
There is a bit of a mixture of changes in CSS, JS, serializers and so on. Can you rebase and clean it up to just take care of the serializers?

inspirehep/utils/bibtex.py
ioannistsanaktsidis: Well when I ported it for invenio3 I had it like this:
```
if journal and (volume != '' or pages != ''):
    recid = self.record['control_number', '']
    pid = PersistentIdentifier.get(
                            'literature', str(recid))
                        record = es.get_source(
                            index='records-journals',
                            id=str(pid.object_uuid),
                            doc_type='journals')
```
I introduced this 
```
def get_es_record_from_doc_type(record_type, recid, collection):
```
so as to do the same job.
Makes sense?

ioannistsanaktsidis:
This PR should be merged together with this one https://github.com/inspirehep/inspire-next/pull/1003

ioannistsanaktsidis:
So this PR is failing because of the clipboard dependency. We have to merge this one https://github.com/inspirehep/inspirehep-js/pull/3 and I will push force again. After that the PR should not be failing.

inspirehep/utils/bibtex.py
```diff
@@ -0,0 +1,679 @@
...
+                                                             recid,
+                                                             'journals')
+                        coden = ','.join(
+                            [record['coden'][0], volume, pages])
+                        return coden
+                except:
+                    return ''
+        else:
+            return ''
```
kaplun: @ioannistsanaktsidis there is a problem with this function: currently it will return only the `coden` when available, otherwise it will only return `"". As it is currently written will never return the pubnote.

ioannistsanaktsidis: Thats why I needed the knowledge module, so as to use this https://github.com/inspirehep/inspire/blob/master/bibformat/format_elements/bfe_INSPIRE_publi_coden.py#L57

kaplun: OK, but what I mean is that this function is not returning a pubnote at all :smile: should it be rather called `_get_coden`?

kaplun: Ah no sorry. Now I see: I was misleaded by:
```python
coden = ','.join(['coden'][0], volume, pages])
```

that variable should actually be called pubnote (since it's the union of a journal title/coden, a volume and pages)

ioannistsanaktsidis: Cool :+1:

ioannistsanaktsidis:
@jmartinm  this one is ready, I just need to wait the tests from @jacquerie  and then we can port it.

inspirehep/modules/search/static/js/search/app.js
```diff
@@ -22,9 +22,6 @@
  */
 
 require([
-    'angular',
```
jmartinm: are you sure about removing this?

inspirehep/modules/theme/bundles.py
```diff
@@ -63,7 +63,8 @@
         "toastr": "~2.1.2",
         "clipboard": "~1.5.8",
         "flightjs": "~1.5.0",
-        "angular": "~1.4.8",
+        "angular": "~1.4.9",
+        "angular-sanitize": "~1.4.9",
```
jmartinm: In principle I don't think we need to add this. `inspirehep-search-js` already depends on that package so it gets installed.

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -45,6 +45,10 @@
 from .views import blueprint
 from inspirehep.utils.template import render_macro_from_template
 from inspirehep.modules.records.json_ref_loader import replace_refs
+from inspirehep.utils.bibtex import Bibtex
```
jmartinm: can you keep the order of imports? (place these under `inspirehep.utils`)

inspirehep/utils/record_getter.py
jmartinm: ping @ioannistsanaktsidis ;)

jmartinm:
Looks good overall, have not checked in detail the actual export functions as they are the same we had in invenio 2.x and waiting for @jacquerie 's tests

inspirehep/utils/record_getter.py
ioannistsanaktsidis: I introduced this one so as to take care of cases like this :
```
 recid = self.record['control_number', '']
    pid = PersistentIdentifier.get(
                            'literature', str(recid))
                        record = es.get_source(
                            index='records-journals',
                            id=str(pid.object_uuid),
                            doc_type='journals')
```
Makes sense?

inspirehep/utils/bibtex.py
jmartinm: I am not sure in that example. Maybe it would make more sense to me if the code was:

```python
pid = PersistentIdentifier.get(
                            'journals', str(recid))
```

So that everything is related to `journals`. 

Otherwise I am not sure how the document will be found in the ES index.

ioannistsanaktsidis: Ok, will check it again then when I apply @jacquerie  tests.

inspirehep/modules/search/static/js/search/app.js
ioannistsanaktsidis: I can't see anything strange happening when I remove them...Do you think that they should be there so as to be initialized when the app is created ?

jmartinm: Should be fine as the dependencies are defined in `settings.js` so I think they can be removed.

jacquerie:
Question: the new approach using serializers means we can get rid of the Jinja2 filters? https://github.com/inspirehep/inspire-next/blob/df77c34f7062553c9f3d1041db35a20cdc76730c/inspirehep/modules/theme/jinja2filters.py#L217-L234

ioannistsanaktsidis:
I guess yeah, there is no need of them. I will remove them from the PR.

jacquerie:
Good, four less lines of untested code :dancer:

ioannistsanaktsidis:
Cool :+1:

inspirehep/utils/export.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+class MissingRequiredFieldError(LookupError):
+
+    """Base class for exceptions in this module.
+    The exception should be raised when the specific,
+    required field doesn't exist in the record.
+    """
+
+    def _init_(self, field):
```
jacquerie: Double underscores here...

inspirehep/utils/export.py
```diff
@@ -0,0 +1,149 @@
...
+        self.field = field
+
+    def _str_(self):
```
jacquerie: ...and here.

kaplun:
> Good, four less lines of untested code :dancer:

> ... Coverage decreased (-8.4%) to 48.914% when 

Mmh :confused:

jacquerie:
> Mmh :confused:

?

> Coverage decreased (-0.5%) to 56.861%

(and, by the end of tomorrow, that number will be positive)

kaplun:
Was just joking on: less lines to test, and the next line from coveralls was about the fact that coverage decreased :dancer:

jacquerie:
I don't think this is WIP anymore :tada:

This implements a core feature of INSPIRE, so I think we should merge this as soon as possible. Truth to be told, I'm not satisfied by the architecture of some parts of the code, but @ioannistsanaktsidis has not a lot of time left with us to fix those. The extensive unit tests should help us refactor anyway.

BTW, a few bugs I found are recorded as `xfailed` tests. We might want to fix at least those before AB.

ioannistsanaktsidis:
I need to add another function to the serializers for importing multiple
formats in one go. Will add it on Monday and then it can go on ;)

ioannistsanaktsidis:
So :) I added the functionality for exporting multiple formats at once. This one will reduce our JS code by a lot as some staff that I implemented in Angular are not needed anymore. Cheers 💃

inspirehep/modules/records/serializers/bibtex_serializer.py
```diff
@@ -0,0 +1,55 @@
...
+        """Serialize a search result.
+        :param pid_fetcher: Persistent identifier fetcher.
+        :param search_result: Elasticsearch search result.
+        :param links: Dictionary of links to add to response.
+        """
+        records = []
+        for hit in search_result['hits']['hits']:
+            records.append(Bibtex(record=hit['_source']).format())
+
+        return "\n".join(records)
```
jmartinm: Isn't it better if from the API it comes back as a list and then on the UI we iterate and display them?

PR #1028, opened 2016-03-31
Fixed installation link

labkode:
This pull request fixed the installation link.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,7 +32,7 @@ Installation
 
 See the `install guide`_ here.
 
-.. _install guide: docs/installation.rst
+.. _install guide: ./INSTALL.rst
 
 ==============
 Happy hacking!


jacquerie:
Hi @labkode, thanks for the PR. I will merge it because it's obviously correct, but I think `docs/installation.rst` would probably be a better place for that document. Sadly, we don't have a very good pipeline for documentation now, since things are very much in flux : (

jacquerie:
Whoops, I forgot to mention that we have some conventions when structuring the commit messages. We sign them with a `Signed-off-by` and we add a tag to the title that specifies the affected part of INSPIRE. We also prefer to use imperatives in titles.

In your case here's what it should look like:
```
docs: update installation link

Signed-off-by: Hugo Gonzalez Labrador <youremail@example.com>
```

jacquerie:
Thanks again, @labkode!

PR #1031, opened 2016-04-04
migrator: restore citation_count from Invenio2

kaplun:

### DIFF
--- a/inspirehep/modules/migrator/tasks.py
+++ b/inspirehep/modules/migrator/tasks.py
@@ -26,10 +26,16 @@ import gzip
 import re
 import traceback
 import zlib
+import click
 
 from celery.utils.log import get_task_logger
 
+from collections import Counter
+
 from dojson.contrib.marc21.utils import create_record as marc_create_record
+
+from elasticsearch.helpers import bulk as es_bulk
+from elasticsearch.helpers import scan as es_scan
 
 from flask import current_app, url_for
 
@@ -49,7 +55,10 @@ from celery import shared_task
 # from invenio_ext.es import es
 from invenio_db import db
 
+from invenio_pidstore.models import PersistentIdentifier, NoResultFound
 from invenio_records import Record
+from invenio_search import current_search_client
+from invenio_search.utils import schema_to_index
 
 from six import string_types, text_type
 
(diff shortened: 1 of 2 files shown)


kaplun:
@jacquerie what do you think it's the best way to test this nowadays?

inspirehep/modules/migrator/tasks.py
```diff
@@ -288,26 +297,28 @@ def migrate_chunk(chunk, broken_output=None, dry_run=False):
...
-            yield {'_op_type': 'update',
-                   '_index': 'hep',
-                   '_type': 'record',
-                   '_id': recid,
-                   'doc': {'citation_count': citation_count}
-                   }
-
-    logger.info("Extracting all citations...")
+            try:
+                uuid = PersistentIdentifier.get('literature', recid).object_uuid
```
kaplun: @jalavik, @jacquerie  here I am forcing `literature` namespace.

jalavik: We could get around it this way perhaps:

```python
In [1]: from invenio_pidstore.models import PersistentIdentifier

In [2]: PersistentIdentifier.query.filter(PersistentIdentifier.object_type == "rec", PersistentIdentifier.pid_value == "4328").all()
Out[2]: [<PersistentIdentifier literature:4328 / rec:05bda190-b149-425c-8825-1d06281634e7 (R)>]

In [3]: PersistentIdentifier.query.filter(PersistentIdentifier.object_type == "rec", PersistentIdentifier.pid_value == "1028379").all()
Out[3]: [<PersistentIdentifier authors:1028379 / rec:f598e4d1-88d7-43c4-a9cb-32c51fa136f1 (R)>]
```

kaplun: I wonder if this has any performance implication... but otherwise sounds great :)

jalavik: As long as there are indices on these we should be good. It is actually one function call less as `.get` calls `.filter`.

jacquerie:
> what do you think it's the best way to test this nowadays?

It's time to add the first integration test! It should be possible to split the build in unit and integration using the trick in https://github.com/jacquerie/inspire-next/commit/637dceb58b12988b9af0c8a0a884dc7768e3ce19#diff-354f30a63fb0907d4ad57269548329e3R45, then I would create a `tests/integration/test_citation_counts.py` that checks that the citation counts of all demo records (or an interesting subset of them) are what they are supposed to be.

kaplun:
@jacquerie OK! I guess I can meanwhile merge this branch.

PR #1036, opened 2016-04-07
tests: test add_citation_counts

jacquerie:
First example of an integration test.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -49,6 +49,7 @@ services:
     depends_on:
       - worker
   worker:
+    container_name: inspirehep-test-worker
     extends:
       service: service_base
     command: celery worker -E -A inspirehep.celery --loglevel=INFO --purge
@@ -60,18 +61,23 @@ services:
       - rabbitmq
       - redis
   redis:
+    container_name: inspirehep-test-redis
     image: redis
   indexer:
+    container_name: inspirehep-test-indexer
     extends:
       file: services.yml
       service: indexer
   rabbitmq:
+    container_name: inspirehep-test-rabbitmq
     image: rabbitmq
   database:
+    container_name: inspirehep-test-database
     extends:
       file: services.yml
       service: database
   static:
+    container_name: inspirehep-test-static
     extends:
       file: services.yml
       service: static
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -35,10 +35,16 @@ services:
       - CACHE_REDIS_URL=redis://redis:6379/0
       - SEARCH_ELASTIC_HOSTS=indexer
 
-  test:
+  unit:
     extends:
       service: service_base
-    command: python setup.py test
+    volumes_from:
+      - static
+    command: ./run-tests.sh unit
+  integration:
+    extends:
+      service: service_base
+    command: ./run-tests.sh integration
     volumes_from:
       - static
     links:
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -1,0 +1,31 @@
+#!/bin/sh
+set -e
+
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+case $1 in
+    integration)
+        python setup.py test --pytest="--ignore tests/unit"
+        ;;
+    *)
+        python setup.py test --pytest="--ignore tests/integration"
+        ;;
+esac
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,0 +1,67 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Pytest configuration for integration tests."""
+
+from time import sleep
+
+import pytest
+
+from invenio_db import db
+from invenio_search import current_search_client as es
+
+from inspirehep.factory import create_app
+
+
+@pytest.yield_fixture(scope='session', autouse=True)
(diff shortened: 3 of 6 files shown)


tests/integration/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@pytest.yield_fixture(scope='session', autouse=True)
+def app(request):
+    """Flask application fixture."""
+    app = create_app()
+
+    with app.app_context():
+        # XXX(jacquerie): import must be local, otherwise task defaults to pickle serializer.
+	from inspirehep.modules.migrator.tasks import add_citation_counts, migrate
```
jmartinm: is the indendation correct here?

tests/integration/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+
+import pytest
+
+from invenio_search import current_search_client as es
+
+from inspirehep.factory import create_app
+
+
+@pytest.yield_fixture(scope='session', autouse=True)
+def app(request):
```
jmartinm: This code gets executed once or once per integration test? Wondering about the performance if it is executed several times.

jacquerie: Just once, because this is a `session` fixture: https://pytest.org/latest/fixture.html#sharing-a-fixture-across-tests-in-a-module-or-class-session

tests/integration/conftest.py
jacquerie: Yeah, an annoying consequence of editing code inside of Vagrant, where Vim isn't configured the way I like it : /

tests/integration/conftest.py
jmartinm: :+1:

docker-compose.test.yml
```diff
@@ -60,18 +67,25 @@ services:
...
   redis:
+    container_name: inspirehep-test-redis
     image: redis
   indexer:
+    container_name: inspirehep-test-indexer
     extends:
       file: services.yml
       service: indexer
+    ports:
+      - 9200:9200
```
jacquerie: Whoops, leftover.

tests/integration/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+        # XXX(jacquerie): makes sure that ES is up.
+        sleep(10)
+
+        _es = app.extensions['invenio-search']
+        list(_es.create(ignore=[400]))
+
+        migrate('./inspirehep/demosite/data/demo-records.xml.gz', wait_for_results=True)
+
+        # XXX(jacquerie): makes sure that all HEP records were migrated.
+        sleep(10)
```
jacquerie: This could become a `refresh` call, I think.

jacquerie:
Travis issue slayed (see attached picture). It was in fact due the version, as suggested in the standup.

Now in a Docker build you can run the unit tests with:
```
$ docker-compose -f docker-compose.test.yml run --rm unit
```
and the integration tests with:
```
$ docker-compose -f docker-compose.tests.yml run --rm integration
```
but, in the second case you'll have to remember to
```
$ docker-compose -f docker-compose.tests.yml kill
$ docker-compose -f docker-compose.tests.yml rm -f
```
to remove the containers containing the past test data. The fact that they remain lying around is actually useful, because then you can go into them and inspect the state of the data.

Non-Docker dev environments should do
```
python setup.py test --pytest=tests/unit
```
and
```
python setup.py test --pytest=tests/integration
```
respectively.

paolo-uccello-san-giorgio-e-il-drago-1024x783

jacquerie:
@nikpap, after this is merged (tomorrow at most) we support integration tests on demo records. BTW, most view unit tests should be made integration tests.

jacquerie:
Ready for review.

kaplun:
LGTM

tests/integration/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # XXX(jacquerie): makes sure that ES is up.
+        sleep(10)
+
+        _es = app.extensions['invenio-search']
+        list(_es.create(ignore=[400]))
+
+        migrate('./inspirehep/demosite/data/demo-records.xml.gz', wait_for_results=True)
+
+        # XXX(jacquerie): makes sure that all HEP records were migrated.
```
jalavik: Small nitpick. Does the comment syntax need to be associated to a name if e.g. `git blame` or `git log`  can be used for tracking down a certain person? Just in interest of slightly less text:

```
# makes sure that all HEP records were migrated.
```

jalavik:
LGTM, nitpick is not a blocker for me either.

tests/integration/conftest.py
jacquerie: Ah, yeah, that is a notation that I use when developing, no need to commit that. The comment is enough.

jacquerie:
A problem is the fact that `python setup.py test --pytest=tests/unit` won't run PEP8 locally. Having Travis reject a build for PEP8 issues is just frustrating, so I will try to add it back.

jacquerie:
Gah, closest thing I have is `python setup.py test --pytest=tests/unit -a "--pep8", but this looks terrible.

jalavik:
I dunno, but it could be worth checking if  `pytest.ini` can be enhanced to play better with this?

jacquerie:
Very ugly workaround: `python setup.py test --pytest="--ignore tests/integration"` to run unit tests and vice versa for integration tests : /

jacquerie:
I give up. Maybe the only clean solution is a `run_tests.sh` script that wraps the proper invocations.

kaplun:
@jacquerie I see in invenio land that they indeed have a `run-test.sh` script:
```sh
pydocstyle invenio_foo && \                                                                                                                                                                                                                                                 
isort -rc -c -df **/*.py && \                                                                                                                                                                                                                                                  
check-manifest --ignore ".travis-*" && \                                                                                                                                                                                                                                       
sphinx-build -qnNW docs docs/_build/html && \                                                                                                                                                                                                                                  
python setup.py test && \                                                                                                                                                                                                                                                      
sphinx-build -qnNW -b doctest docs docs/_build/doctest                                                                                                                                                                                                                         
```

run-tests.sh
```diff
@@ -0,0 +1,31 @@
...
+# or submit itself to any jurisdiction.
+
+case $1 in
+    integration)
+        python setup.py test --pytest="--ignore tests/unit"
+        ;;
+    *)
+        python setup.py test --pytest="--ignore tests/integration"
+        ;;
+esac
```
jacquerie: @mihaibivol, did I do something wrong in this script?

jmartinm:
So, this ready to be merged ? As there a couple of PRs that can benefit from it

run-tests.sh
mihaibivol: I don't see anything wrong. However I think we could also benefit from a `run_docker_tests.sh` script and these commands can be easily scripted directly into `docker-compose.test.yml`. I understood after standup why we need `--ignore`.

So why doesn't this script actually run the docker test suite alltogether?

mihaibivol:
```
#!/bin/sh

cleanup () {
	docker-compose -f docker-compose.test.yml kill
	docker-compose -f docker-compose.test.yml rm -f
}
trap cleanup 0 1 2 6 

docker-compose -f docker-compose.test.yml run --rm $1
```

jacquerie: > So why doesn't this script actually run the docker test suite alltogether?

I wanted scripts that can be used in both environments, continuing the idea that we support both development environments.

jacquerie:
> So, this ready to be merged ? As there a couple of PRs that can benefit from it

If you can withstand some flakiness. Sometimes the build might fail for no good reason, but if this is blocking someone then I prefer that we merge and we fix later.

run-tests.sh
mihaibivol: Makes sesne, as long as the developer is willing to break his own db during tests.

jacquerie: Ouch, right. Didn't think of that.

jacquerie:
Now supports the non-Docker development environment. Here are the test commands:
```
$ ./run-tests.sh unit
```
and
```
$ ./run-tests.sh integration
```
for unit tests and integration tests respectively, while the Docker variants are the same as before:
```
$ docker-compose -f docker-compose.test.yml run --rm unit
```
and
```
$ docker-compose -f docker-compose.test.yml run --rm integration
```

jacquerie:
BTW, it looks like this is not so flaky after all. Please report suspicious failures, though, in case flakiness becomes an issue.

mihaibivol:
with the comment that after running the docker suite for __integration only__ you need to do

```
docker-compose -f docker-compose.test.yml kill
docker-compose -f docker-compose.test.yml rm -f
```

Otherwise you will have the records already indexed in the containers and integration tests will fail.

jacquerie:
This is probably ok to :ship:. I think that the sample test is a little silly (it's just a smoke test for the successful migration), so maybe in the future I will remove it.

kaplun:
Yay! I merge it!

PR #1039, opened 2016-04-08
WIP orcid: pushing records to orcid

Panos512:
Adds the functionality to send updated/new records to orcid.
- [x] Tweak the json converter's results to fully follow orcids schema.
- [x] Convert citations to bibtex and add them to the output json. 
- [x] Extract the needed information of the author to push on his behalf to orcid.
- [x] Check json acceptance from orcid.
- [ ] Tests.
- [ ] Code cleanup.


inspirehep/modules/orcid/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from inspirehep.modules.orcid.receivers import *
```
kaplun: Is this the right pattern to populate receivers? @jacquerie?

jacquerie: I actually have never written a receiver. I don't know! This is how it's currently done all over the codebase, though.

inspirehep/modules/orcid/receivers.py
```diff
@@ -0,0 +1,52 @@
...
+@before_record_insert.connect
+@before_record_update.connect
+def send_records_to_orcid(sender, *args, **kwargs):
+    """Sends updated records to orcid."""
+    try:
+        orcid_json = convert_to_orcid(sender)
+        api = orcid.MemberAPI(current_app.config['OAUTHCLIENT_ORCID_CREDENTIALS']['consumer_secret'],
+                              current_app.config['OAUTHCLIENT_ORCID_CREDENTIALS'][
+                                  'consumer_key'],
+                              sandbox=True
```
kaplun: Don't hardcode the sandbox flag here. Maybe that you can populate it by simply looking into `current_app.config`?

Also, not sure how heavy is the initialization of `MemberAPI` (maybe it's bare), but maybe you can move the `api` initialization in the above INSPIREOrcid one?

inspirehep/modules/orcid/receivers.py
```diff
@@ -0,0 +1,52 @@
...
+                              )
+        for author in sender.get('authors'):
+            user = RemoteAccount.query.get(1)
+            for token in user.tokens:
+                if token.token_type is '':
+                    api.add_record(user.extra_data.get(
+                        'orcid'), token, 'work', orcid_json)
```
kaplun: I assume all this is `FIXME` :dancer:

inspirehep/modules/orcid/schema.py
```diff
@@ -0,0 +1,106 @@
...
+        return{
+            'external-id-type': 'doi',
+            'external-id-value': value.get('value'),
+            'external-id-url': 'www.example.com/12344'
+        }
+    if key == 'arxiv_eprints':
+        return {
+            'external-id-type': 'arxiv',
+            'external-id-value': value.get('value'),
+            'external-id-url': 'https://arxiv.org/abs/' + value.get('value').partition(':')[2]
```
kaplun: What do you expect `value` to contain for `.partition(':')[2]` to work?
Also if `value.get('value')` is `None` it will crash.

inspirehep/modules/orcid/utils.py
```diff
@@ -0,0 +1,46 @@
...
+                final_date[0] = date[0]
+            if date[1]:
+                final_date[1] = date[1]
+            if date[2]:
+                final_date[2] = date[2]
+            publication_date = {'year': final_date[0],
+                                'month': final_date[1],
+                                'day': final_date[2]
+                                }
+            orcid_json['publication-date'] = publication_date
```
kaplun: I think it possible to move that into `dojson` as well. I have to check how but there are some tricks available.

inspirehep/modules/orcid/receivers.py
Panos512: Yeah ofc the api initialization will most probably move and the flag will be removed when this goes live.

inspirehep/modules/orcid/receivers.py
Panos512: Yes, totally 💃

inspirehep/modules/orcid/schema.py
Panos512: The value is something like `arXiv:xxxx.xxxxx` and the url needs only the second part. I already changed the whole design to a proper try-except pattern.

inspirehep/modules/orcid/utils.py
Panos512: I will also check. Indeed this need to be done in a prettier way.

inspirehep/modules/orcid/receivers.py
kaplun: Well the flag is still useful, (e.g. we can `qa.inspirehep.net` or `localhost` still using the sandbox).

kaplun:
Just to add what we discussed IRL that the proper scope to use is:
```python
'scope': '/orcid-profile/read-limited /activities/update /orcid-bio/update'
```

inspirehep/modules/orcid/schema.py
kaplun: Have you checked this is the proper way to expose external identifiers? In the old XML based method we are using on legacy there is no URL but an ID:
https://github.com/inspirehep/invenio/blob/prod/modules/miscutil/etc/templates/works.xml#L57-L64

Panos512: That's what ORCID asks on it's schema. We could of course skip it and don't provide an url if it causes trouble.

kaplun: Weird. On http://members.orcid.org/api/record-xml-structure seems to use an alternate ways of both

Panos512: Well it's weird because the schema is different depending on where you are getting it from 👅 . What I am now using is:

```
@orcid_overdo.over('work_external_identifier', 'arxiv_eprints|dois')
@dojson.utils.for_each_value
def external_id_rule(self, key, value):
    if key == 'dois':
        return{
            'external-identifier-type': 'DOI',
            'external-identifier-id': value.get('value'),
            'external-identifier-url': 'www.example.com/12344'
        }
    if key == 'arxiv_eprints':
        return {
            'external-identifier-type': 'ARXIV',
            'external-identifier-id': value.get('value'),
            'external-identifier-url': 'https://arxiv.org/abs/' + value.get('value').partition(':')[2]
        }
```

and Orcid is accepting it.

inspirehep/modules/orcid/schema.py
```diff
@@ -0,0 +1,126 @@
...
+@dojson.utils.for_each_value
+def authors_rule(self, key, value):
+    try:
+        return {
+            'credit-name': value.get('full_name'),
+            'contributor-orcid': value.get('orcid') if 'orcid' in value else '',
+            'contributor-attributes': {
+                'contributor-role': 'AUTHOR',
+                # One of 'ADDITIONAL', 'FIRST'
+                'contributor-sequence': 'ADDITIONAL'
```
Panos512: @kaplun Regarding this what should we do? `FIRST` means main authors and `ADDITIONAL` means co-authors, as far as I understood. Should we mark every author as `ADDITIONAL` or  should we not provide the information?

kaplun: authors[0] is the FIRST, and authors[1:] are all ADDITIONAL :)

Panos512: That's great! thanks

PR #1040, opened 2016-04-08
literaturesuggest: initial implementation

jmartinm:

### DIFF
--- a/inspirehep/modules/arxiv/ext.py
+++ b/inspirehep/modules/arxiv/ext.py
@@ -1,0 +1,168 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation,
+# Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Arxiv extension.
+
+Arxiv extension is initialized like this:
+
+>>> from flask import Flask
+>>> from invenio_ext.arxiv import Arxiv
+>>> app = Flask("myapp")
+>>> ext = Arxiv(app=app)
+
+Configuration Settings
+----------------------
+The details of the ArXiv URL and endpoint can be customized in the application
+settings.
+
+================= ============================================================
+`ARXIV_API_URL`   The URL of ArXiV query API.
+                  **Default:** `http://export.arxiv.org/oai2`
+`ARXIV_ENDPOINT`  The name of Flask endpoint for new application route.
+                  If the value is `False` (or `None`) the url rule is not
+                  registered.  **Default:** `_arxiv.search`
+`ARXIV_URL_RULE`  The URL for `ARXIV_ENPOINT` (i.e.
+                  `url_for(current_app.config['ARXIV_ENDPOINT'])` is equal to
+                  `current_app.config['ARXIV_ENDPOINT']`).
+                  **Default:** `/arxiv/search`
+================= ============================================================
+
+"""
+
+from __future__ import absolute_import
+
+import requests
+
+from flask import current_app, request, jsonify
+from flask_login import login_required
+from lxml.etree import fromstring
+
+from invenio_utils.xmlhelpers import etree_to_dict
+
+response_code = {'success': 200,
+                 'notfound': 404,
+                 'unsupported_versioning': 415,
+                 'malformed': 422,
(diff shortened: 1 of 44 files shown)


inspirehep/modules/literaturesuggest/templates/literaturesuggest/forms/suggest_success.html
```diff
@@ -0,0 +1,26 @@
...
+## You should have received a copy of the GNU General Public License
+## along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+#}
+{%- extends config.BASE_TEMPLATE -%}
+
+{% block body %}
+<div class="alert alert-success alert-form-success">
+  <button type="button" class="close" data-dismiss="alert" aria-hidden="true">&times;</button>
+  <strong>Thank you for your update!<br></strong><br>The INSPIRE staff will review it and your changes will be added to INSPIRE.
```
jalavik: `Thank you for your suggestion!` instead?

inspirehep/modules/arxiv/ext.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # query arxiv
+        result = self.get_json(arxiv)
+
+        resp = jsonify(result)
+        resp.status_code = response_code.get(result['status'],
+                                             result['status'])
+        return resp
+
+__all__ = ("Arxiv", )
```
kaplun: I think this whole part should be actually factor out in an dedicated repo with nice tests.

inspirehep/modules/crossref/ext.py
```diff
@@ -0,0 +1,146 @@
...
+        doi = doi or request.args.get("doi")
+
+        # query crossref
+        result = self.get_json(doi)
+
+        resp = jsonify(result)
+        resp.status_code = response_code.get(result['status'], 200)
+        return resp
+
+__all__ = ("CrossRef", )
```
kaplun: Ditto. It's kind of INSPIRE specific so it could be inspire-crossref, but at least can be easily tested and re-used.

inspirehep/modules/forms/validators/simple_fields.py
```diff
@@ -0,0 +1,186 @@
...
+                allow_redirects=True
+            )
+        except:
+            return
+        return response.headers['content-type']
+
+    message = "Please, provide an accessible direct link to a PDF document."
+
+    if field.data and get_content_type(field.data) != 'application/pdf':
+        raise StopValidation(message)
```
kaplun: Is this check asynchronous? Because then we can also download the PDF on the fly and:
```python
import requests
import PyPDF2

try:
     PyPDF2.PdfFileReader(requests.get(url), strict=False)
except PyPDF2.utils.PdfReadError as err:
    raise StopValidation(err)

File: inspirehep/modules/literaturesuggest/bundles.py [hunk 5]
```diff
@@ -0,0 +1,47 @@
...
+
+from inspirehep.modules.theme.bundles import js as _js
+
+
+js = NpmBundle(
+    "js/literaturesuggest/inspire_init.js",
+    output="gen/inspire-literature-suggest.%(version)s.js",
+    filters=RequireJSFilter(exclude=[_js]),
+    npm={
+        'buckets-js': '1.98.1',
```
kaplun: Are we always pinning minor versions?

jmartinm: Not really, usually better `~1.98.0` or `~1.98.1`, that means that new minor revisions will get installed, just pinning the major revision.

inspirehep/modules/forms/validators/simple_fields.py
jmartinm: It is async on the UI, although when somebody puts a URL in the PDF link field and moves away from the field the feedback should appear pretty quick or they will be in a different part of the form and might be a bit strange if the error appears after 20 sec.

inspirehep/modules/crossref/ext.py
jmartinm: Yes, this two extensions used to be part of `invenio-ext`. If they are not needed there we can move them to `inspirehep`, the arXiv one had some tests ready https://github.com/inveniosoftware/invenio-ext/blob/master/tests/test_ext_arxiv.py

inspirehep/modules/literaturesuggest/templates/literaturesuggest/forms/suggest_success.html
jmartinm: Right!

inspirehep/modules/forms/validators/simple_fields.py
kaplun: True that it might take some seconds, but at least it would not allow the form to submit.
This is currently the main pain point of the workflow, where users post URLs that are either invalid, or we are not allowed to fetch.

jmartinm: Yep, although now that I think about it, also when you hit submit all the validators run, so if this one takes a bit long to compute the form will appear `Submitting...` for a while which is not ideal, but not sure about a good solution.

jalavik: The validation for the PDF field could perhaps cache the result of the validation (using the url as key). Then it should feel instantaneous upon submitting. Unless there is a way to disable validation for this field onsubmit and keep it running onupdate.

kaplun:
I guess I will move all my remarks into new issues, then.

PR #1041, opened 2016-04-12
WIP: clean schemas

tomaszgy:


inspirehep/modules/records/jsonschemas/records/elements/rank.json
```diff
@@ -0,0 +1,18 @@
...
+      "SENIOR"
+      "JUNIOR"
+      "VISITOR"
+      "POSTDOC"
+      "PHD"
+      "MASTER"
+      "BACHELOR"
+      "UNDERGRADUATE"
+      "STUDENT",
+      "OTHER"
```
kaplun: Does that means that we are moving away from the current `enum` used on legacy? I am ok with that, just raising the point as this has implications e.g. in the author form etc.

kaplun:
Is inveniosoftware/invenio-jsonschemas#24 something to consider since you are including schemas inside schemas?

inspirehep/dojson/utils.py
```diff
@@ -178,3 +178,46 @@ def get_recid_from_ref(ref_obj):
...
+    else:
+        casted_value = value.upper().replace('.', '')
+        for rank_name, rank_mapping in RANKS_MAPPINGS.items():
+            if rank_name in casted_value:
+                return rank_name
+            else:
+                if rank_mapping.get('alternative_names'):
+                    for alternative in rank_mapping['alternative_names']:
+                        if alternative in casted_value:
+                            return rank_name
```
eamonnmag: Instead of having two loops here, you could just have one key, e.g. **alternative_names**, and use that since abbreviations are just an alternative name too.

tomaszgy: Would be cool but I have a slightly different logic for alternative_names and abbreviations. I check if the  the former is the part of the string whereas for the latter I check if the string is equal to it.

tomaszgy:
@kaplun, PR updated for today :smile: Have a look also at @bittirousku's work in [another PR](https://github.com/tomaszgy/inspire-next/pull/1).

kaplun:
@tomaszgy  this PR needs to be closed and reopened against `master` branch.

jacquerie:
...and please also remove https://github.com/inspirehep/inspire-next/pull/1041/files#diff-c9bea86c5aea70d00ab50d1b530aaf66R970, since it will cause weird validation errors. BTW, what about writing a test that checks that the JSONSchemas are valid?

tomaszgy:
@kaplun, it's alread done in https://github.com/inspirehep/inspire-next/pull/1069. @jacquerie I'll remove it, tests for JSONSchemas are good idea.

I suppose it's time to close this PR.

jacquerie:
That removal I mentioned was done in #1096. But please add tests, so that we don't run into it again!

PR #1046, opened 2016-04-14
records: impact graph API addition

### LINKED ISSUE #1044: Port impact graph API
Continue https://github.com/inspirehep/inspire-next/pull/885 and make it work in `invenio3`. It will be used for impact graph in detailed record page.

jmartinm:
* Adds API to retrieve information regarding references and citations
  for a given record. Needed by impact graph visualization. (closes #1044)

* Adds new record utils for record-related utilities.

* Adds get_title util.

* Adds appropriate tests for the added functionality.

### DIFF
--- a/inspirehep/modules/records/serializers/impactgraph_serializer.py
+++ b/inspirehep/modules/records/serializers/impactgraph_serializer.py
@@ -1,0 +1,125 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Impact Graph serializer for records."""
+
+from __future__ import absolute_import, print_function
+
+import json
+
+from invenio_search import current_search_client
+from invenio_search.api import Query
+
+from inspirehep.utils.record import get_title
+
+
+class ImpactGraphSerializer(object):
+
+    """Impact Graph serializer for records."""
+
+    def serialize(self, pid, record, links_factory=None):
+        """
+        Serialize a single impact graph from a record.
+
+        :param pid: Persistent identifier instance.
+        :param record: Record instance.
+        :param links_factory: Factory function for the link generation,
+                              which are added to the response.
+        """
+        out = {}
+
+        # Add information about current record
+        out['inspire_id'] = record['control_number']
+        out['title'] = get_title(record)
+        out['year'] = record['earliest_date'].split('-')[0]
+
+        # Get citations
+        citations = []
+
+        es_query = Query('refersto:' + record['control_number'])
+        es_query.body.update(
+            {
+                'size': 9999
+            }
+        )
+        record_citations = current_search_client.search(
+            index='records-hep',
+            doc_type='hep',
+            body=es_query.body,
+            _source=[
+                'control_number',
+                'citation_count',
+                'titles',
+                'earliest_date'
+            ]
+        )['hits']['hits']
+        for citation in record_citations:
+            citation = citation['_source']
+            citations.append({
+                "inspire_id": citation['control_number'],
+                "citation_count": citation.get('citation_count', 0),
+                "title": get_title(citation),
+                "year": citation['earliest_date'].split('-')[0]
+            })
+
+        out['citations'] = citations
+
+        # Get references
+        record_references = record.get('references', [])
+        references = []
+
+        reference_recids = [
+            ref['recid'] for ref in record_references if ref.get('recid')
+        ]
+
+        if reference_recids:
+            mget_body = {
+                "ids": reference_recids
+            }
+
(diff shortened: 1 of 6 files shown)


jmartinm:
Continuation of https://github.com/inspirehep/inspire-next/pull/885

Pending merging of https://github.com/inspirehep/inspire-next/pull/1036 to add an integration test.

jmartinm:
@Panos512 @kaplun in this PR there is the file `inspirehep/utils/record.py` added. I think the example is not that good as I should add the logic of what title to select when, for example, there is one from arXiv and one from a publisher (will do that).

But then, we can unify in this utility the cases when there are multiple values in a field so that we select the correct one for display and export to ORCID.

kaplun:
@jmartinm great for having started that! Wonder if we want to divide across the different types of records? But then again, the `get_title` should be generic enough to work with all types of records (also given the refactoring of schemas of @tomaszgy).

So great!

inspirehep/modules/records/serializers/impactgraph_serializer.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Add information about current record
+        out['inspire_id'] = record['control_number']
+        out['title'] = get_title(record)
+        out['year'] = record['earliest_date'].split('-')[0]
+
+        # Get citations
+        citations = []
+
+        es_query = Query('refersto:' + record['control_number'])
```
kaplun: Is this ES or `invenio_query`?

kaplun: Is this ES or `invenio_search`? I think the latter. Because otherwise in ES we could also filter out the fields without all the massive amount of metadata.

jalavik:
Regarding the `record` utility and `get_title`, we currently have `breadcrumb_title` in our data model which is basically the first title from `titles`, e.g. `titles[0]['title']`. I was musing with @tomaszgy if we should instead name it `title` or something generic, more so than `breadcrumb_title` as to not tie the name to a specific feature (breadcrumb), since it will be useful in many instances to "just get me a title" or "just get me an abstract/summary". For few, important fields this could be useful, depending on collection/data type.

kaplun:
@jalavik I only have a by-principle issue with your proposal (and indeed `breadcrumb_title` per se): such field would unnecessarily bloat the indexed JSON, even if you would never search for it, Vs. having to compute it at formatting time, which shouldn't be so expensive (given that in general there are only 1 or 2 titles to visit and only publisher should be picked up over arxiv).

WDYT?

inspirehep/modules/records/serializers/impactgraph_serializer.py
jmartinm: in this case, `Query` is from the `invenio_search` API, but then I use `current_search_client.search()` which is the ES Python library so I could maybe use [this](https://elasticsearch-py.readthedocs.org/en/master/api.html#response-filtering) in order to reduce the amount of fields returned (if that is what you meant)

jacquerie:
Almost OT: my dream data structure would be something which contains mostly raw data, which is then wrapped by the appropriate class by the controller (`views.py`) and contains all the data that the view (a template) needs to render itself (without having to use jinja filters). It might happen that the wrapping requires some querying/other expensive operation: so we cache this variable, instead of caching the views.

jmartinm:
`breadcrum_title` was a bit special case because an Invenio module required to have a clean title in some field in order to be used for the page title in the browser. So could not be computed on-the-fly.

For other cases, I would say that we can have a clean JSON representation (the one in the JSONSchema) and then for specific fields that are needed in certain views we use the serializers (like our [brief results](https://github.com/inspirehep/inspire-next/blob/invenio3/inspirehep/modules/records/serializers/json.py)) where we can enhance the returned JSON. If an operation is expensive we can cache the call.

I guess it is similar to @jacquerie's proposal (in an API manner)

jacquerie:
Ah yeah, the API way is clearly superior :+1:

inspirehep/modules/records/serializers/impactgraph_serializer.py
jmartinm: Added `_source` field now.

tests/unit/records/test_record_utils.py
```diff
@@ -0,0 +1,62 @@
...
+        "titles": []
+    }
+
+    assert get_title(empty_title) == ""
+
+    no_title_key = {
+        "not_titles": []
+    }
+
+    assert get_title(no_title_key) == ""
```
jacquerie: The actual path following the convention we have in place would be `tests/unit/utils/test_utils_record.py`. Shall we change the convention?

jalavik:
@kaplun @jmartinm @jacquerie Serializers with potentially cached fields and API is a good way. 👍 

Is it true then with v3 we then no longer need the `breadcrumb_title`?

kaplun:
> Is it true then with v3 we then no longer need the breadcrumb_title?

Just grepped in the installation and there is no single reference of `breadcrumb_title` beside our own `inspire-next` overlay.

jmartinm:
> Just grepped in the installation and there is no single reference of breadcrumb_title beside our own inspire-next overlay.

Still seems to be there on the latest [invenio-records](https://github.com/inveniosoftware/invenio-records/blob/v1.0.0a13/invenio_records/config.py#L27)

kaplun:
> Still seems to be there on the latest invenio-records

No reference beside that :+1: 
```
 <labs3> skaplun@PCSK4 ~/.virtualenvs/labs3/src➔ grep -R RECORDS_BREADCRUMB_TITLE_KEY *
invenio-records/invenio_records/config.py:RECORDS_BREADCRUMB_TITLE_KEY = 'title_statement.title'
Il file binario invenio-records/invenio_records/config.pyc corrisponde
<labs3> skaplun@PCSK4 ~/.virtualenvs/labs3/src➔ cd ../lib/
 <labs3> skaplun@PCSK4 ~/.virtualenvs/labs3/lib➔ grep -R RECORDS_BREADCRUMB_TITLE_KEY *
<labs3> skaplun@PCSK4 ~/.virtualenvs/labs3/lib➔ 
```

tests/integration/test_impact_graphs.py
```diff
@@ -0,0 +1,39 @@
...
+import json
+
+
+def test_impact_graphs_api(app):
+    """Test response of impact graph API."""
+    with app.test_client() as client:
+        result = client.get(
+            "/api/literature/613135",
+            headers={"Accept": "application/x-impact.graph+json"}
+        )
```
jacquerie: Maybe we should write tests for all our API like this one?

PR #1048, opened 2016-04-14
Error pages

eamonnmag:
* ADD Added 401, 403, 404, and 500 error pages.


jalavik:
So cool! 💃  🐧

inspirehep/config.py
```diff
@@ -889,3 +894,5 @@ def _(x):
     'stat.OT': 'Other',
     'stat.TH': 'Other'
 }
+
+ASSETS_DEBUG = True
```
jmartinm: Is this a leftover? We have been usually adding it in our instance folder local config, but maybe it makes sense to have it here?

eamonnmag: That should have been removed by me. But I think for development it makes
sense since it reduces page loading times massively :)

jmartinm: Well in theory it should slow loading times ;) as each JS creates a request to fetch the file VS production style bundles.

jmartinm:
Cannot be merged at the moment as it is lacking rebase with `invenio3`. We can :shipit: after

eamonnmag:
Will rebase when I get in :)

PR #1054, opened 2016-04-18
requirements: pin records-rest and search

jacquerie:
Hopefully fixes the build failure in https://travis-ci.org/inspirehep/inspire-next/jobs/123996582#L1004


jacquerie:
@jalavik, @jmartinm, @jmartinm: this fixes the build, but I'm not sure if I'm fixing this in the right way. Interaction between `setup.py` and `requirements.txt` is still black magic to me.

jmartinm:
LGTM. I think the fact that `invenio-records-rest` was both in `setup.py` and `requirements.txt` is an error.

PR #1067, opened 2016-04-21
global: invenio3 search compatibility

jmartinm:


jmartinm:
Things that work:

* [ ] Queries with invenio and spires syntax inside of fields, e.g `title:"Time"` and `find title Time` in all collections.

Some things to do:

* [ ] [Google style search](https://github.com/inspirehep/inspire-next/pull/1067/files#diff-6f322e9ca9e97f8b3dfb3b2933c9da1aL69) needs to be fixed. At the moment by default the query is being done on the `_all` field so not matches are found.

* [ ] Replace usages of `from invenio_search.api import Query` by the new `RecordsSearch` class.

* [ ] Figure out how to [pass the keywords](https://github.com/inveniosoftware/invenio-query-parser/blob/4bb96159cedafdda5729f4afad2eb260fbcb6415/invenio_query_parser/utils.py#L34)

* [ ] Replace [invenio query factory] by our own if we want to add enhancers and that might be the place to fix the Google style search. At the moment I am using the Invenio factory passing our own parser and walkers.

jmartinm:
@jalavik @jacquerie @kaplun @Panos512 this is ready for review and testing.

@Panos512 if you have time, you could have a look at the last point in the TO DO. I am not sure what was the difference between our DSL and the invenio one, but if we need ours, converting it to use the `Q` object from http://elasticsearch-dsl.readthedocs.org/en/latest/ should not be difficult.

And then you could also check the other point about how to pass the keywords.

@jacquerie At the moment locally the only unittests that were failing were the ones for `jinja2filters` that used to mock `es`. I made [one of them pass](https://github.com/inspirehep/inspire-next/pull/1067/files#diff-cc4ec36bfdca525994bc68c9082219e3R742) but I am not sure what would be the best way of mocking.

jmartinm:
Also how to force the latest versions of `invenio-records-rest` and `invenio-search` to be installed by Travis? https://travis-ci.org/inspirehep/inspire-next/jobs/124949166

mihaibivol:
It should try to install the latest as per the current setup.py install_requires. The new images just cache the dependencies inside them so that we don't hit pip during the build process.

mihaibivol:
And also, this matters only for pip packages. Repo installes are not cached and are pulled everytime we do a build.

requirements.txt
```diff
@@ -24,12 +24,13 @@
...
--e git+https://github.com/inspirehep/invenio-query-parser.git@invenio3#egg=invenio-query-parser==0.4.2.dev20160310
+# FIXME temporary branch for testing
+-e git+https://github.com/inspirehep/invenio-query-parser.git@invenio3-inspire#egg=invenio-query-parser==0.6.0
 -e git+https://github.com/inspirehep/invenio-records.git@invenio3#egg=invenio-records
 -e git+https://github.com/inspirehep/invenio-pidstore.git#egg=invenio-pidstore
 
 -e git+https://github.com/inspirehep/workflow.git#egg=workflow==2.0.0.dev20160223
 -e git+https://github.com/inspirehep/invenio-workflows.git@invenio3#egg=invenio-workflows==1.0.0a1.dev20160126
--e git+https://github.com/jalavik/invenio-workflows-ui.git@master#egg=invenio-workflows-ui==0.1.0.dev20160000
+-e git+https://github.com/jmartinm/invenio-workflows-ui.git@test-invenio3-search#egg=invenio-workflows-ui==0.1.0.dev20160000
```
jalavik: Remember to update this remote branch and/or push it to inspirehep/invenio-workflows-ui repo

inspirehep/modules/search/__init__.py
```diff
@@ -27,5 +27,8 @@
 from __future__ import absolute_import, print_function
 
 from .ext import INSPIRESearch
+from .query_factory import inspire_query_factory
 
-__all__ = ('INSPIRESearch', )
+IQ = inspire_query_factory()
```
jalavik: Shall we make this a LocalProxy thing? In case inspire_query_factory call will need fully loaded app?

jmartinm: I took it as inspiration from https://github.com/inveniosoftware/invenio-query-parser/blob/master/invenio_query_parser/contrib/elasticsearch/__init__.py but yeah, might be better to do a Proxy.

kaplun: I thin there is no need for a proxy because this is a thread-safe object (it's read-only so there shouldn't be any issue of concurrent access).

kaplun:
@jmartinm it's quite a huge PR :) Is there anything in particular that you wish to point out for review?

inspirehep/modules/search/__init__.py
jalavik: You are right. I was actually more worried about the loading timing, but I see now that this is just a closure anyway, so we should be good on that case too.

inspirehep/modules/search/query_factory.py
```diff
@@ -0,0 +1,93 @@
...
+from .walkers.elasticsearch import ElasticSearchDSL
+from .walkers.elasticsearch_no_keywords import ElasticSearchNoKeywordsDSL
+from .walkers.elasticsearch_no_keywords import QueryHasKeywords
+
+
+def inspire_query_factory():
+    """Create a parser returning Elastic Search DSL query instance."""
+
+    def invenio_query(pattern):
+        parser = import_string(current_app.config['SEARCH_QUERY_PARSER'])
```
jalavik: This would be called every time the function is called. Is there some performance penalty here using `import_string` every time? Same below regarding `walkers`.

inspirehep/modules/search/query_factory.py
```diff
@@ -0,0 +1,91 @@
...
+
+def inspire_query_factory():
+    """Create a parser returning Elastic Search DSL query instance."""
+
+    def invenio_query(pattern):
+        walkers = [PypegConverter(), SpiresToInvenio()]
+
+        # Enhance query first
+        # for enhancer in query_enhancers():
+        #     enhancer(self, **kwargs)
```
jacquerie: Is this still relevant?

inspirehep/utils/search.py
```diff
@@ -22,18 +22,16 @@
 
 """Functions for searching ES and returning the results."""
 
-from invenio_search import current_search_client
-from inspirehep.modules.search.query import perform_query
 
+def perform_es_search(q, index, start=0, size=10, sort=None):
+    """."""
```
jacquerie: Missing docstring.

inspirehep/modules/search/query_factory.py
jmartinm: Yes, left it as a reminder as we might still want to enhance the queries for example, excluding the `collections` a user does not have access to.

PR #1068, opened 2016-04-21
dojson: BAI support

kaplun:
* Maps 100/700__w to inspire_bai field.

* Makes inspire_bai field searchable.

* Updates demo records to have BAIs.

* Amends hep.json to support multiple editors in references.


kaplun:
cc: @jacenkow

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -454,7 +454,10 @@
                         "type": "array"
                     },
                     "editors": {
-                        "type": "string"
+                        "type": "array",
+                        "items": {
+                            "type": "string"
+                        }
```
jacquerie: Looks unrelated?

jacquerie:
The change to add `inspire_bai` looks correct (haven't tested it though), but I can't understand how are the changes to the citations data related...

kaplun:
@jacquerie they are not. But since I have refreshed the demo records, one of the record is now sporting a reference with more than one editors and I had to fix it, otherwise I wouldn't be able to load the records at all...

inspirehep/modules/records/jsonschemas/records/hep.json
kaplun: I separated it into a dedicated commit

PR #1069, opened 2016-04-21
OLD dojson: clean schemas

tomaszgy:
Supersedes #1041.


inspirehep/dojson/common/base.py
```diff
@@ -318,3 +318,30 @@ def deleted_records2marc(self, key, value):
...
+@jobs.over('urls', '^856.[10_28]')
+@journals.over('urls', '^856.[10_28]')
+def urls(self, key, value):
+    """URL to external resource."""
+    urls = []
+    value = utils.force_list(value)
+    for val in value:
+        if isinstance(val.get('y'), str):
+            description = val.get('y')
+        elif isinstance(val.get('y'), tuple):
```
jalavik: Can this also be a list? So would it be better to check `isinstance(val.get('y'), (list, tuple))`?

inspirehep/dojson/common/base.py
```diff
@@ -318,3 +318,30 @@ def deleted_records2marc(self, key, value):
...
+            description = val.get('y')[0]
+            # FIXME: raise Exception("multiple y values")
+        else:
+            description = None
+        if isinstance(val.get('u'), str):
```
jalavik: Better use `six.text_type` instead of `str` for Python 2/3 compatibility.

inspirehep/dojson/conferences/fields/bd1xx.py
```diff
@@ -50,13 +50,26 @@ def alternative_titles(self, key, value):
...
+    elif isinstance(extra_place_info, tuple):
+        for place in extra_place_info:
+            self['extra_place_info'].append(place)
+    else:
+        pass
+
+    name = value.get('p')
+    email = value.get('m')
+    return {
+        'name': name if isinstance(name, str) else None,
```
jalavik: Better use `six.text_type` here and other places.

inspirehep/dojson/jobs/fields/bd1xx.py
```diff
@@ -114,41 +82,67 @@ def institution(self, key, value):
...
-def description(self, key, value):
-    """Contact person."""
-    return value.get('a')
-
-
 @jobs.over('position', '^245..')
 def position(self, key, value):
-    """Contact person."""
+    """Breadcrumb title."""
     self.setdefault('breadcrumb_title', value.get('a'))
```
jalavik: I do not think we need the breadcrumb title anymore.

tomaszgy:
PR ready to review
@jacquerie @jalavik @kaplun @jmartinm

inspirehep/dojson/common/base.py
```diff
@@ -21,11 +21,12 @@
 # or submit itself to any jurisdiction.
```
jacquerie: Copyright dates should include 2016.

inspirehep/dojson/common/base.py
```diff
@@ -21,11 +21,12 @@
...
 
 """MARC 21 model definition."""
+import six
 
 from dojson import utils
 
 from inspirehep.dojson import utils as inspire_dojson_utils
-from inspirehep.dojson.utils import strip_empty_values
+from inspirehep.dojson.utils import strip_empty_values, classify_field
```
jacquerie: Opposite order: `classify_field, strip_empty_values`.

inspirehep/dojson/common/base.py
```diff
@@ -21,11 +21,12 @@
 # or submit itself to any jurisdiction.
 
 """MARC 21 model definition."""
+import six
```
jacquerie: Please add the `__future__` block in all the Python files you touch:
```python
from __future__ import (
    absolute_import,
    division,
    print_function,
    unicode_literals)
```

inspirehep/dojson/common/base.py
```diff
@@ -216,10 +217,13 @@ def collections(self, key, value):
 
     def get_value(value):
```
jacquerie: Not a fan of defining this function every time this rule is applied — but also not sure if there are performance issues! @mihaibivol, do you know if this hurts performance?

inspirehep/dojson/common/base.py
```diff
@@ -318,3 +322,70 @@ def deleted_records2marc(self, key, value):
...
+                        # FIXME: right now '_scheme' is what it says in the
+                        # MARCXML. Should the original term's real scheme be
+                        # detected and put here instead?
+                    if source:
+                        if 'automatically' in source:
+                            source = 'INSPIRE'
+                        field_dict['source'] = source
+                    field_categories.append(field_dict)
+                # else:
+                    # raise Exception("Field value type should be str or unicode")
```
jacquerie: Not sure what's happening here, since we are guarding against `corrected_val` being a falsy value...

inspirehep/dojson/common/base.py
jacquerie: Also: `six` import unused?

inspirehep/dojson/conferences/fields/bd1xx.py
```diff
@@ -33,40 +35,70 @@
...
-    """Contact person."""
-    if value.get('m'):
-        self['contact_email'] = value.get('m')
-    return value.get('p')
+    711__b is for indexing and will not be displayed.
+    """
+    titles = {}
+    titles['title'] = value.get('a')
+    titles['searchable_title'] = value.get('b')
+    return titles
```
jacquerie: Consistency: in the rest of the functions this is usually written as
```python
return {
    'title': value.get('a'),
    'searchable_title': value.get('b')
}
```

Is there a reason to change style here?

inspirehep/dojson/conferences/fields/bd1xx.py
```diff
@@ -33,40 +35,70 @@
...
+    """Map 270 field to contact details and extra_place_info"""
+    self.setdefault('extra_place_info', [])
+    extra_place_info = value.get('b')
+    if isinstance(extra_place_info, six.string_types):
+        self['extra_place_info'].append(extra_place_info)
+    elif isinstance(extra_place_info, tuple):
+        for place in extra_place_info:
+            self['extra_place_info'].append(place)
+    else:
+        pass
```
jacquerie: Then we don't need the `else` :D

inspirehep/dojson/geo_helpers.py
```diff
@@ -0,0 +1,420 @@
...
+           'GERMANY (DESY)'],
+    'ES': ['CANARY ISLANDS', 'MADRID'],
+    'FR': ['CORSICA'],
+    'GR': ['CRETE'],
+    'GB': ['UK', 'ENGLAND', 'ENG', 'SCOTLAND', 'WALES', 'SCOTLAND/UK',
+           'NORTHERN IRELAND', 'LONDON'],
+    'ID': ['BALI'],
+    'IL': ['JERUSALEM'],
+    'IR': ['IRAN'],
+    'IT': ['ITALIE', 'ITALIA', 'NAPOLI', 'MESSINA', 'SARDINIA', 'SICILY'],
```
jacquerie: Wait, `NAPOLI` is not an alternative spelling of `ITALIA` :D

There's other instances of cities becoming alternative spellings of countries. What's going on here?

inspirehep/dojson/geo_helpers.py
```diff
@@ -0,0 +1,420 @@
...
+    'US-WA': ['WA'],
+    'US-WV': ['WV'],
+    'US-WI': ['WI', 'WIS', 'WISC'],
+    'US-WY': ['WY'],
+}
+
+south_korean_cities = ['SEOUL', 'DAEJON', 'DAEJEON', 'MT SORAK', 'POHANG',
+                       'JEJU ISLAND', 'CHEJU ISLAND', 'GYEONGJU', 'BUSAN',
+                       'DAEGU', 'GYEONGIU', 'PUSAN', 'YONGPYONG',
+                       'PHOENIX PARK', 'CHEJU ISLAND']
```
jacquerie: In general, this file looks weird: those are not `geo_helpers`, but more `geo_data`, and I wonder if this has really to belong inside of the code, instead of being a KB file or maybe imported from somewhere. Where does this data come from, anyway?

inspirehep/dojson/hep/fields/bd01x09x.py
```diff
@@ -24,22 +24,45 @@
...
-        'medium': value.get('b')
-    }
+    """ISBN and its medium with additional comment."""
+    isbn = {}
+    if 'a' not in value:
+        return None
+    try:
+        isbn['value'] = normalize_isbn(value['a'])
+    except NotValidISBNError:
+        return None
```
jacquerie: This error checking can be written more concisely:
```python
isbn = {}

try:
    isbn['value'] = normalize_isbn(value['a'])
except (KeyError, NotValidISBNError):
    return None

if 'b' in value:
    ...
```

inspirehep/dojson/geo_helpers.py
kaplun: Uh! Indeed! This seems more like the country is missing from a raw address and the last value is the city.

inspirehep/dojson/hep/fields/bd01x09x.py
```diff
@@ -24,22 +24,45 @@
...
+            pass
+        elif 'ebook' in isbn['medium']:
+            isbn['medium'] = 'online'
+            isbn['comment'] = 'ebook'
+        elif 'hardcover' in isbn['medium']:
+            isbn['medium'] = 'print'
+            isbn['comment'] = 'hardcover'
+        else:
+            isbn['comment'] = isbn['medium']
+            isbn['medium'] = ''
```
jacquerie: I can _see_ what this is doing, but this is definitely not the most straightforward way of doing it...

inspirehep/dojson/hep/fields/bd90x99x.py
```diff
@@ -24,36 +24,47 @@
...
-            except:
-                pass
-        if 'o' in value:
+        # FIXME: There might be multiple 0, o or y values. I.e. they might be
+        # tuples, not strings.
+        isbn = ''
+        try:
+            recid = int(value.get('0', ''))
+        except (TypeError, ValueError):
+            # FIXME: log.warning("Invalid value(s) in recid")
```
jacquerie: What's the reason for keeping the logging commented? (I don't want you to change it, I want to understand what's the problem here).

inspirehep/dojson/hep/fields/bd90x99x.py
```diff
@@ -24,36 +24,47 @@
...
+        try:
+            number = int(value.get('o', ''))
+        except (TypeError, ValueError):
+            # FIXME: log.warning("Invalid value(s) in label number")
+            number = ''
+        try:
+            year = int(value.get('y', ''))
+        except (TypeError, ValueError):
+            # FIXME: log.warning("Invalid value(s) in year")
+            year = ''
```
jacquerie: A small helper in `inspire_dojson_utils` wouldn't hurt to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) this code.

inspirehep/dojson/institutions/fields/bd1xx.py
```diff
@@ -85,42 +85,48 @@ def set_value(key, val):
...
-        'address': utils.force_list(value.get('a')),
-        'city': value.get('b'),
-        'state_province': value.get('c'),
-        'country': value.get('d'),
-        'postal_code': utils.force_list(value.get('e')),
-        'country_code': value.get('g'),
-    }
+    country_code = value.get('g')
+    if isinstance(country_code, (list, tuple)):
+        country_code = list(set(country_code))
```
jacquerie: This will remove duplicates, but shuffle the list. The `list(set(...))` idiom reads nicely, but is annoying to test.

inspirehep/dojson/jobs/fields/bd1xx.py
```diff
@@ -109,46 +79,64 @@ def institution(self, key, value):
         recid = int(value.get('z'))
     return {
         'curated_relation': curated_relation,
-        'record': inspire_dojson_utils.get_record_ref(recid, 'isntitutions'),
+        'record': inspire_dojson_utils.get_record_ref(recid, 'istitutions'),
```
jacquerie: It's i**n**stitutions :D

inspirehep/dojson/geo_helpers.py
tomaszgy: Same for `Israel` and `JERUSALEM`, `Czech Republic` and `PRAGUE` etc. As @kaplun pointed out, some of conferences' raw addresses didn't contain country. If you do insist I can refactor it but as I wrote [here](https://github.com/tomaszgy/inspire-next/blob/dcf720e52ff2910ce9c211ebf5d92470d0673223/inspirehep/dojson/utils.py#L243) it is _a pretty dummy address parser (..) and should be replace with something better e.g. Google Geocoding_ :smile:

jacquerie:
Two general things: I believe my commit should go last in this series, and second to last should be @kaplun's one in #1070.

inspirehep/dojson/geo_helpers.py
tomaszgy: Agreed, it is a weird file. As I said above it should be replaced with GoogleGeocoding although AFAIK we do not use GoogleGeo yet. For now this extracts `country` and `U.S. state` from vast majority of production's records (the way it does that is silly but was quick to implement) - and those are the information needed for Neo4J. If we replace it with something smarter, I'll be more than happy! For now I'd rather have some usable data than none.

The data come more or less from [ISO_3166-1 alpha 2 standard](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) if you like we can store it the other way :smile:

jacquerie:
Another general thing: commits https://github.com/inspirehep/inspire-next/pull/1069/commits/2a261b270dba0a726af4c600fa8ae1f87acb60f4, https://github.com/inspirehep/inspire-next/pull/1069/commits/dcf720e52ff2910ce9c211ebf5d92470d0673223, https://github.com/inspirehep/inspire-next/pull/1069/commits/57d0ecab9beeed5fcdaa34ec94148db6c3b346f4, https://github.com/inspirehep/inspire-next/pull/1069/commits/0d7eb5f481f424cce695b4e085352af35279d04a
should avoid the "(#n)" in the title, since that would link to a random issue when merged, and not where they were supposed to point originally.

inspirehep/dojson/geo_helpers.py
jacquerie: Ok, I don't mind the idea of a dummy geocoder which works in the 90% of cases. I don't agree with putting those methods in `utils`, since I don't think they have other users... maybe `inspirehep_dojson_utils` is a better place?

inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -249,14 +239,23 @@ def positions(self, key, value):
     inst = {'name': value.get('a'),
             'record': inspire_dojson_utils.get_record_ref(recid,
                                                           'institutions')}
+
+    raw_rank = value.get('r')
+    if isinstance(raw_rank, tuple):
+        # FIXME: raise Exception("Multiple ranks at the same time?")
```
jacquerie: I don't think that we should ever raise an `Exception` when running `dojson`. We should be logging the issue instead to `stderr`.

inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -249,14 +239,23 @@ def positions(self, key, value):
...
+        pass
+
+    rank = inspire_dojson_utils.classify_rank(raw_rank)
+
     return {
         'institution': inst if inst['name'] else None,
-        'rank': value.get('r'),
+        'rank': rank,
+        '_rank': raw_rank if isinstance(raw_rank, str) else None,
```
jacquerie: This should use `six.string_types`.

inspirehep/modules/records/jsonschemas/records/authors.json
```diff
@@ -86,23 +86,11 @@
...
-                            "PHD",
-                            "PhD",
-                            "MAS",
-                            "UG",
-                            "Diploma",
-                            "Bachelor",
-                            "smirnov@ictp.it"
-                        ],
-                        "type": "string"
+                    "_rank": {
```
jacquerie: I like this idea of using `_field` to hold the raw value of the field. I don't think we are doing this consistently, though.

tests/unit/dojson/test_dojson_jobs.py
```diff
@@ -0,0 +1,69 @@
...
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
```
jacquerie: Please add `__future__` block.

tests/unit/dojson/test_dojson_jobs.py
```diff
@@ -0,0 +1,69 @@
...
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
```
jacquerie: Old license header. You can find the correct one in https://github.com/inspirehep/inspire-next/blob/master/setup.py.

tests/unit/dojson/test_dojson_jobs.py
```diff
@@ -0,0 +1,69 @@
...
+    """Two ranks inside one record."""
+    snippet = (
+        '<record> '
+        '<datafield tag="656" ind1=" " ind2=" ">'
+        '<subfield code="a">Senior</subfield>'
+        '</datafield>'
+        '</record>'
+    )
+
+    record = strip_empty_values(jobs.do(create_record(snippet)))
```
jacquerie: Please add a space between data and assertions, so that they are cleanly separated.

tests/unit/dojson/test_dojson_jobs.py
```diff
@@ -0,0 +1,69 @@
...
+    snippet = (
+        '<record> '
+        '<datafield tag="656" ind1=" " ind2=" ">'
+        '<subfield code="a">Senior</subfield>'
+        '<subfield code="a">Junior</subfield>'
+        '</datafield>'
+        '</record>'
+    )
+
+    record = strip_empty_values(jobs.do(create_record(snippet)))
```
jacquerie: Similarly, space between data and assertions.

tests/unit/dojson/test_dojson_jobs.py
```diff
@@ -0,0 +1,69 @@
...
+        '<datafield tag="656" ind1=" " ind2=" ">'
+        '<subfield code="a">Senior</subfield>'
+        '</datafield>'
+        '<datafield tag="656" ind1=" " ind2=" ">'
+        '<subfield code="a">Junior</subfield>'
+        '</datafield>'
+        '</record>'
+    )
+
+    record = strip_empty_values(jobs.do(create_record(snippet)))
```
jacquerie: Finally, space between data and assertions.

tests/unit/dojson/test_dojson_jobs.py
```diff
@@ -0,0 +1,69 @@
...
+    assert record['_ranks'] == ['Senior', 'Junior']
+    assert record["ranks"] == ['SENIOR', 'JUNIOR']
+
+
+def test_ranks_from_marcxml_double_656():
```
jacquerie: `def test_ranks_from_marcxml_with_double_656():`, maybe?

tests/unit/dojson/test_dojson_jobs.py
```diff
@@ -0,0 +1,69 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+from dojson.contrib.marc21.utils import create_record
+from inspirehep.dojson.jobs import jobs
+from inspirehep.dojson.utils import strip_empty_values
```
jacquerie: Missing the cases:
* No `656`.
* Single `656`, but no `a`.
* Double `656`, one with `a`, one without.

tests/unit/dojson/test_dojson_jobs.py
```diff
@@ -0,0 +1,69 @@
...
+        '<datafield tag="656" ind1=" " ind2=" ">'
+        '<subfield code="a">Senior</subfield>'
+        '<subfield code="a">Junior</subfield>'
+        '</datafield>'
+        '</record>'
+    )
+
+    record = strip_empty_values(jobs.do(create_record(snippet)))
+    assert record['_ranks'] == ['Senior', 'Junior']
+    assert record["ranks"] == ['SENIOR', 'JUNIOR']
```
jacquerie: This tests that the resulting record's `_ranks` and `ranks` keys are what we expect, while asserting equality on an excepted records asserts that ALL keys are what we expect, which is why I like the latter style more.

tests/unit/utils/test_utils_bibtex.py
```diff
@@ -1539,7 +1539,7 @@ def test_get_url_urls_without_url():
 def test_get_url_one_url_not_a_list_to_an_image():
     one_url_not_a_list_to_an_image = Record({
```
jacquerie: No need to specify `not_a_list_` in the test name and the record name, now that it can't be a list. :D

tests/unit/utils/test_utils_bibtex.py
```diff
@@ -1552,7 +1552,7 @@ def test_get_url_one_url_not_a_list_to_an_image():
 def test_get_url_one_url_not_a_list_not_to_an_image():
     one_url_not_a_list_not_to_an_image = Record({
```
jacquerie: Also here.

inspirehep/dojson/common/base.py
jacquerie: And this is what happens when I browse the entire PR instead of going commit by commit. Scratch that.

inspirehep/dojson/common/base.py
mihaibivol: I tested an example with `timeit` and it seems that this way of writing code is not that much slower. For 1M function calls the difference was two seconds.

mihaibivol: Plus, for a file this big I find it easier to read if the rules are actually self contained.

inspirehep/dojson/common/base.py
```diff
@@ -216,10 +217,13 @@ def collections(self, key, value):
 
     def get_value(value):
         primary = ''
-        if isinstance(value.get('a'), list):
+        if isinstance(value.get('a'), (list, tuple)):
             primary = value.get('a')[0]
+            if isinstance(primary, (list, tuple)):
+                primary = primary[0]
```
mihaibivol: Saw that these lines aren't covered in the coverage report. Is this case covered in the production data?

inspirehep/dojson/conferences/fields/bd1xx.py
```diff
@@ -33,40 +35,70 @@
 @utils.for_each_value
 def acronym(self, key, value):
     """Conference acronym."""
+    def append_to_self(title):
+        """Make a dictionary and append to self."""
+        titles = {}
+        titles['title'] = title
+        titles['subtitle'] = value.get('b')
```
mihaibivol: Just to be sure, a title list or tuple will always have a single subtitle? Are you sure you don't need to zip `value.get('b')` together with `value.get('a')` in the case they are both lists?

inspirehep/dojson/common/base.py
mihaibivol: Should we bring back isort :smile:

inspirehep/dojson/common/base.py
```diff
@@ -318,3 +322,70 @@ def deleted_records2marc(self, key, value):
...
+@hepnames.over('urls', '^856.[10_28]')
+@institutions.over('urls', '^856.[10_28]')
+@jobs.over('urls', '^856.[10_28]')
+@journals.over('urls', '^856.[10_28]')
+def urls(self, key, value):
+    """URL to external resource."""
+    urls = []
+    value = utils.force_list(value)
+    for val in value:
+        url_dict = {}
```
mihaibivol: You can delete this line. It doesn't do anything.

inspirehep/dojson/common/base.py
```diff
@@ -318,3 +322,70 @@ def deleted_records2marc(self, key, value):
...
+        description = utils.force_list(val.get('y'))
+        url_values = utils.force_list(val.get('u'))
+        for u in url_values:
```
mihaibivol: Again, to make sure, shouldn't the `descriptions` and `url_values` be zipped in this traversal? Seems that every url will get the same description. Is this correct?

inspirehep/dojson/institutions/fields/bd1xx.py
```diff
@@ -85,42 +85,48 @@ def set_value(key, val):
...
+        if len(country_code) == 1:
+            country_coude = country_code.pop()
```
mihaibivol: typo `country_coUde`

tests/unit/records/test_records_jsonschemas.py
```diff
@@ -0,0 +1,96 @@
...
+                'jsonschemas',
+                'records',
+                filename
+            )
+        )
+    )
+
+
+def test_acquisition_source_schema_is_valid():
+    acquisition_source_schema = fetch_schema('acquisition_source.json')
```
kaplun: Shall we automatically test every .json file under jsonschema folder? Rather than explicitly test each one and then risk to forget something?

kaplun: Also, is `Draft4Validator` able to correctly load referred schemas?

tests/unit/dojson/test_dojson_utils.py
```diff
@@ -87,3 +88,89 @@ def test_remove_duplicates_from_list_of_dicts_preserving_order():
...
+    """Test research field code mapping from arXiv to Inspire.
+
+    Case where input value is equivalent to None."""
+    raw_field_codes = [None, '']
+    expected = [None, None]
+    result = [utils.classify_field(field) for field in raw_field_codes]
+
+    assert expected == result
+
+def test_classify_field_not_a_string():
```
kaplun: Missing double spacing. @jacquerie is PEP8 check disabled?

tests/unit/dojson/test_dojson_journals.py
```diff
@@ -0,0 +1,219 @@
...
+        '</datafield> '
+        '</record>'
+    )
+
+    record = strip_empty_values(journals.do(create_record(snippet)))
+    title0 = 'PHYS REV SPECIAL TOPICS ACCELERATORS BEAMS'
+    title1 = 'PHYSICS REVIEW ST ACCEL BEAMS'
+    assert record['title_variants'][0]['title'] == title0
+    assert record['title_variants'][1]['title'] == title1
+
```
kaplun: Missing new line :stuck_out_tongue:

inspirehep/dojson/common/base.py
jacquerie: Yeah, self-contained rules read better, and a slowdown of 2 microseconds per call is very acceptable. BTW, @kaplun, this is more evidence that DoJSON's trouble is on their side...

tests/unit/dojson/test_dojson_utils.py
jacquerie: Yes, PEP8 checking is disabled in tests: https://github.com/inspirehep/inspire-next/blob/f2bf1c45920b43f52a3002406ae3351b364f762f/pytest.ini#L27-L28

tests/unit/records/test_records_jsonschemas.py
jacquerie: I'm 👍  for automatically testing all `.json` files in the folder instead of listing them.

I *think* the answer for the `Draft4Validator` is yes:
"resolver – an instance of RefResolver that will be used to resolve $ref properties (JSON references). If unprovided, one will be created." (from http://python-jsonschema.readthedocs.io/en/latest/validate/#the-validator-interface)

inspirehep/dojson/conferences/fields/bd1xx.py
tomaszgy: What about `extra_place_info` being a `list` or `int`? Is it impossible for `dojson` to output those types?

inspirehep/dojson/geo_helpers.py
tomaszgy: Sure. You mean in the directory `inspirehep.dojson.utils` (it does not exist yet)?

inspirehep/dojson/hep/fields/bd90x99x.py
tomaszgy: We're planning to configure customized machine-friendly logger for errors inside dojson. Though It's not configured yet.

inspirehep/dojson/hepnames/fields/bd1xx.py
tomaszgy: Agreed. Same as in the previous comment it should be customized logger.

inspirehep/dojson/conferences/fields/bd1xx.py
jacquerie: Writing
```python
else:
    pass
```
is equivalent to writing nothing, so those cases are still unhandled...

inspirehep/dojson/geo_helpers.py
jacquerie: Yes, the file `inspirehep/dojson/utils.py` could be a place. Or `inspirehep/dojson/geocoder.py`, or even `inspirehep/dojson/dummy_geocoder.py`.

inspirehep/dojson/common/base.py
bittirousku: Actually that was the intended behaviour.

inspirehep/dojson/conferences/fields/bd1xx.py
bittirousku: I'll update this.

inspirehep/dojson/common/base.py
bittirousku: I'll have a look at this later (didn't touch this yet).

inspirehep/dojson/hepnames/fields/bd1xx.py
bittirousku: I changed the commented exceptions to logger warnings.

inspirehep/dojson/hep/fields/bd01x09x.py
bittirousku: I try to make this more obvious.

inspirehep/dojson/conferences/fields/bd1xx.py
bittirousku: No.

inspirehep/dojson/common/base.py
tomaszgy: Yes, this case appears in the production data.

tests/unit/records/test_records_jsonschemas.py
```diff
@@ -41,56 +51,65 @@ def fetch_schema(filename):
...
+# def test_journals_schema_is_valid():
+#     journals_schema = fetch_schema('journals.json')
+#
+#     Draft4Validator.check_schema(journals_schema)
+#
+#
+# def test_json_reference_schema_is_valid():
+#     json_reference_schema = fetch_schema('json_reference.json')
+#
+#     Draft4Validator.check_schema(json_reference_schema)
```
jacquerie: NB: commenting code is usually the wrong thing to do!

tests/unit/records/test_records_jsonschemas.py
```diff
@@ -41,56 +51,65 @@ def fetch_schema(filename):
...
-
-    Draft4Validator.check_schema(conferences_schema)
-
-
-def test_experiments_schema_is_valid():
-    experiments_schema = fetch_schema('experiments.json')
-
-    Draft4Validator.check_schema(experiments_schema)
+    schemas = [os.path.split(schema)[1] for schema in glob.glob(os.path.join(
+        'inspirehep', 'modules', 'records', 'jsonschemas', 'records', '*.json'))]
```
jacquerie: Can be simplified to:
```python
schemas_dir = os.path.join(
    'inspirehep', 'modules', 'records', 'jsonschemas', 'records')
schemas  = os.listdir(schemas_dir)
```

tests/unit/records/test_records_jsonschemas.py
```diff
@@ -20,6 +20,16 @@
...
 
+from __future__ import (
+    absolute_import,
+    division,
+    print_function,
+    unicode_literals)
+
+import glob
+
+import inspirehep.dojson.utils as inspire_dojson_utils
```
jacquerie: Unused.

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -437,20 +437,20 @@ def test_succeeding_entry(marcxml_to_json, json_to_marc):
...
+    assert (marcxml_to_json['urls'][0]['value'] ==
             json_to_marc['8564'][0]['u'])
-    assert (marcxml_to_json['urls'][0]['size'] ==
-            json_to_marc['8564'][0]['s'])
-    assert (marcxml_to_json['urls'][0]['doc_string'] ==
-            json_to_marc['8564'][0]['w'])
+    #assert (marcxml_to_json['urls'][0]['size'] ==
+            #json_to_marc['8564'][0]['s'])
+    #assert (marcxml_to_json['urls'][0]['doc_string'] ==
+            #json_to_marc['8564'][0]['w'])
```
jacquerie: In general, code that is wrong should be deleted, not commented out.

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -437,20 +437,20 @@ def test_succeeding_entry(marcxml_to_json, json_to_marc):
...
-    assert (marcxml_to_json['urls'][0]['comment'] ==
-            json_to_marc['8564'][0]['z'])
-    assert (marcxml_to_json['urls'][0]['name'] ==
-            json_to_marc['8564'][0]['f'])
+    #assert (marcxml_to_json['urls'][0]['material_type'] ==
+            #json_to_marc['8564'][0]['3'])
+    #assert (marcxml_to_json['urls'][0]['comment'] ==
+            #json_to_marc['8564'][0]['z'])
+    #assert (marcxml_to_json['urls'][0]['name'] ==
+            #json_to_marc['8564'][0]['f'])
```
jacquerie: And here as well.

tests/unit/dojson/test_dojson_common_base.py
```diff
@@ -0,0 +1,159 @@
...
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+from dojson.contrib.marc21.utils import create_record
+from inspirehep.dojson.hep import hep
+from inspirehep.dojson.utils import strip_empty_values
+
+
+def test_urls_from_marcxml_856_with_single_u_single_y():
+    """Simple case."""
```
jacquerie: There's no need for a docstring when the name of the test already explains the scenario.

inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -31,7 +31,6 @@
 
 @hepnames.over('acquisition_source', '^541[10_].')
 @utils.for_each_value
-@utils.filter_values
```
jacquerie: This change looks unrelated to this commit. What's the issue here?

bittirousku: Apparently my commit message regarding this disappeared (it was in my first commit: https://github.com/tomaszgy/inspire-next/pull/1). We decided to do this because we will be filtering all None values later and this `filter_values` doesn't work with nested structures at all.

tests/unit/dojson/test_dojson_hep.py
bittirousku: This is fixed.

inspirehep/dojson/hepnames/fields/bd1xx.py
jacquerie: Ok. Keep in mind that commits should be **atomic**: they should deal with exactly one thing at a time, so this change should have been in a different commit.

Because of several other issues like this I will have to squash this PR to a single commit : (

kaplun: :+1: indeed, if some commit is about fixing a previous commit that wasn't yet merged, it's much better to rebase and fixup, i.e. to squash together the commit and the specific correction.

tests/unit/dojson/test_dojson_hep.py
jacquerie: Yes, I noticed that a later commit deletes these comments.

A better way to do it is modifying the commit itself, rather than adding a commit later to fix a previous commit: doing that creates a dependency between two commits and violates their atomicity, which makes integration harder.

tests/unit/dojson/test_dojson_common_base.py
kaplun: How to shut up linting, then?

jacquerie: We don't run linting on tests :D

jacquerie:
Closed in favor of #1170.

PR #1070, opened 2016-04-22
jsonschema: array-related fixes

kaplun:

### DIFF
--- a/inspirehep/dojson/hep/fields/bd90x99x.py
+++ b/inspirehep/dojson/hep/fields/bd90x99x.py
@@ -24,6 +24,8 @@
 
 from __future__ import absolute_import, division, print_function
 
+import re
+
 from isbnlib._exceptions import NotValidISBNError
 
 from dojson import utils
@@ -35,10 +37,24 @@ from ..model import hep, hep2marc
 from ...utils import get_recid_from_ref, get_record_ref, strip_empty_values
 
 
+RE_VALID_PUBNOTE = re.compile(".*,.*,.*(,.*)?")
+
+
 @hep.over('references', '^999C5')
 def references(self, key, value):
     """Produce list of references."""
     value = utils.force_list(value)
+
+    def get_valid_pubnotes(pubnotes):
+        valid_pubnotes = []
+        raw_references = []
+        if pubnotes:
+            for pubnote in pubnotes:
+                if RE_VALID_PUBNOTE.match(pubnote):
+                    valid_pubnotes.append(pubnote)
+                else:
+                    raw_references.append(pubnote)
+        return valid_pubnotes, raw_references
 
     def get_value(value):
         recid = None
@@ -65,6 +81,11 @@ def references(self, key, value):
         except (KeyError, NotValidISBNError):
             isbn = ''
 
+        valid_pubnotes, raw_references = get_valid_pubnotes(utils.force_list(value.get('s')))
+
+        if value.get('x'):
+            raw_references += list(utils.force_list(value.get('x')))
+
         return {
             'record': get_record_ref(recid, 'literature'),
             'texkey': value.get('1'),
@@ -80,8 +101,8 @@ def references(self, key, value):
             'report_number': utils.force_list(value.get('r')),
             'title': utils.force_list(value.get('t')),
             'urls': utils.force_list(value.get('u')),
-            'journal_pubnote': utils.force_list(value.get('s')),
-            'raw_reference': utils.force_list(value.get('x')),
+            'journal_pubnote': valid_pubnotes,
+            'raw_reference': raw_references,
             'year': year,
         }
     references = self.get('references', [])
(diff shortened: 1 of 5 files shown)


kaplun:
@tomaszgy @bittirousku here are some fixes on `hep.json` I had to implement in order to successfully use `domapping` tool.

jacquerie:
Maybe this commit can be cherry-picked in #1069 and applied before https://github.com/inspirehep/inspire-next/pull/1069/commits/c1af4ff8e545054b946ebd61b5c7032faa281864, so that we test these changes?

jacquerie:
And... what is the `domapping` tool? Does it make sense to test that it can be used?

kaplun:
@jacquerie: see: https://github.com/inveniosoftware/domapping

```
$ domapping schema_to_mapping hep.json
```

jacquerie:
Ah, it's the JSONSchema -> ES mapping thing. Well, since what it outputs is not usable as-is but needs to be refined then I'm not sure if testing that would be valuable.

jmartinm:
Is this taken into account in https://github.com/inspirehep/inspire-next/pull/1143 ?

jacquerie:
This commit is not complete, because DoJSON rules need to be amended as well, in order for the demo records to pass validation. So it's not currently part of #1143.

kaplun:
> Ah, it's the JSONSchema -> ES mapping thing. Well, since what it outputs is not usable as-is but needs to be refined then I'm not sure if testing that would be valuable.

Well a valid test is: if there is some output, then the test is passed. If error on stderr than failed :)

kaplun:
Actually we can even import `domapping` and:

```python

from domapping.cli import schema_to_mapping
schema_to_mapping(...)
```

kaplun:
@jacquerie :pray: that it works!

tests/integration/test_detailed_records.py
```diff
@@ -37,8 +37,8 @@ def test_all_records_were_loaded(app):
 
 
 def test_all_records_are_valid(app):
-    invalid = InspireProdRecords.query.filter(InspireProdRecords.valid == False)
-    recids = [el.recid for el in invalid]
+    invalid = InspireProdRecords.query.filter(InspireProdRecords.valid == False).values(InspireProdRecords.recid)
+    recids = [el[0] for el in invalid]
```
jacquerie: Uh? What do we gain from this?

inspirehep/dojson/hep/fields/bd76x78x.py
```diff
@@ -81,7 +81,7 @@ def publication_info2marc(self, key, value):
         '0': get_recid_from_ref(
             value.get('parent_record')),
         'c': value.get('page_artid'),
-        'n': value.get('journal_issue'),
+        #'n': value.get('journal_issue'),
```
jacquerie: Leftover?

inspirehep/dojson/hep/fields/bd90x99x.py
```diff
@@ -35,11 +37,25 @@
...
+    def get_valid_pubnotes(pubnotes):
+        valid_pubnotes = []
+        raw_references = []
+        if pubnotes:
+            for pubnote in pubnotes:
+                if RE_VALID_PUBNOTE.match(pubnote):
+                    valid_pubnotes.append(pubnote)
+                else:
+                    raw_references.append(pubnote)
+        return valid_pubnotes, raw_references
```
jacquerie: You could do this using `tee` from `itertools`: http://stackoverflow.com/a/949490/374865 (I'm not saying that you should, I just wanted to point out that this logic can be abstracted if it's needed in a lot of places).

jacquerie:
I'd say that this should be split in three commits, one per logical part:
1. Amend hep.json to correctly declare fields as array when necessary.
2. Amend migrate_chunk() to correctly report recid of failed record.
3. Disable porting of refextract versioning.

tests/integration/test_detailed_records.py
kaplun: Speed :smile: (it's 30 times faster). I know it's useless (1ms instead of 30ms) but since I already took the pain to write it because I didn't know that test was already there... I can remove it from the commit if you prefer.

inspirehep/dojson/hep/fields/bd76x78x.py
kaplun: Ouch! Indeed.

inspirehep/dojson/hep/fields/bd90x99x.py
kaplun: Oh Thanks! I should use more of the standard tools indeed.

kaplun:
@jalavik this PR fails because of:
```
pkg_resources.DistributionNotFound: The 'invenio_oaiharvester==1.0.0a1.dev20160406' distribution was not found and is required by inspire-crawler
```

do you know anything about this?

kaplun:
Ah yes I see #1207

jalavik:
@kaplun Yeah, on top of it. Should be resolved after #1217

inspirehep/dojson/hep/fields/bd90x99x.py
jacquerie: OT: I'm actually surprised this is not part of Python's stdlib. Ruby has `Enumerable#partition`: http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-partition.

kaplun:
Yep sorry, I meant #1217

kaplun:
@jacquerie I split the commits, however, to disentangle the refextract handling from dojson/jsonschema was too complex :)

jacquerie:
> I split the commits, however, to disentangle the refextract handling from dojson/jsonschema was too complex :)

:+1:

A nice tool to do that is `git add -p`, which uses smart heuristics to only add some changes in a file to a commit.

PR #1077, opened 2016-04-26
theme: simplify navigation header

jalavik:
* Fixes `Submit` link in the navigation bar and simplifies
  the navigation menu to be more direct.

@jmartinm


jmartinm:
Looks good. :shipit:

PR #1080, opened 2016-04-26
audit: initial Audit trail

jalavik:
* Adds back the Audit module for tracking actions in Holding Pen.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -32,7 +32,7 @@ from invenio_search import current_search_client as es
 from inspirehep.factory import create_app
 
 
-@pytest.yield_fixture(scope='session', autouse=True)
+@pytest.yield_fixture(scope='session')
 def app(request):
     """Flask application fixture."""
     app = create_app()
@@ -64,4 +64,21 @@ def app(request):
         add_citation_counts(request_timeout=20)
         es.indices.refresh('records-hep')  # Makes sure that all citation counts were added.
 
+        yield app
+
+
+@pytest.yield_fixture(scope='session')
+def db_only_app(request):
+    """Flask application fixture."""
+    app = create_app()
+
+    def teardown():
+        with app.app_context():
+            db.drop_all()
+
+    request.addfinalizer(teardown)
+
+    with app.app_context():
+        db.drop_all()
+        db.create_all()
         yield app
--- a/inspirehep/modules/workflows/models.py
+++ b/inspirehep/modules/workflows/models.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2015, 2016 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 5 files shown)


tests/unit/conftest.py
```diff
@@ -80,3 +81,33 @@ def teardown():
 
     with app.app_context():
         yield app
+
+
+@pytest.yield_fixture(scope='session', autouse=True)
+def db_app(request):
```
jalavik: @jacquerie @mihaibivol I wanted an app context that had tables created only but without the delaying factor of the ES integration. Let me know if you think this is okay, or if it should be moved to `tests/integration`.

jacquerie: This is a use case I hand't anticipated. I'd say to create this in `tests/integration/conftest.py`, and copy from the `app` fixture only the lines that create the table. What is that needs the tables to be there, but not necessarily the records?

inspirehep/modules/audit/api.py
```diff
@@ -0,0 +1,49 @@
...
+        prediction_results = prediction_results[0].get("result")
+        score = prediction_results.get("max_score")  # returns 0.222113
+        decision = prediction_results.get("decision")  # returns "Rejected"
+
+        # Map actions to align with the prediction format
+        action_map = {
+            'accept': 'Non-CORE',
+            'accept_core': 'CORE',
+            'reject': 'Rejected'
+        }
```
kaplun: Shall we have this mapping configurable, so that we can use this helper in other contextes?

inspirehep/modules/audit/models.py
```diff
@@ -0,0 +1,49 @@
...
+class Audit(db.Model):
+
+    __tablename__ = "audit_logging"
+
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+
+    # Date that the action was taken
+    created = db.Column(db.DateTime, default=datetime.now, nullable=False)
+
+    user_id = db.Column(db.Integer, default=0, nullable=False)
```
kaplun: This should be a foreign key to the `User` model.

inspirehep/modules/audit/models.py
```diff
@@ -0,0 +1,49 @@
...
+    object_id = db.Column(db.Integer, default=0, nullable=False)
```
kaplun: This should also be a foreign key to the appropriate model.

inspirehep/modules/audit/api.py
jalavik: It was not needed to be used anywhere else yet so for simplicity I kept it here, but I can put it out somewhere.

inspirehep/modules/audit/models.py
jalavik: Sure, thanks!

inspirehep/modules/audit/models.py
jalavik: On one hand this would lock it to only be used with Workflow objects, but on the other hand that has been the only use case so far and probably won't change.

jalavik:
@kaplun Aight, I changed things around to make it more concrete WRT the use case and then adding the constraints. It is part of workflows now cause thats where it is used.

tests/integration/conftest.py
```diff
@@ -32,7 +32,7 @@
 from inspirehep.factory import create_app
 
 
-@pytest.yield_fixture(scope='session', autouse=True)
+@pytest.yield_fixture(scope='session')
```
jalavik: @jacquerie @mihaibivol OK if I kill this? Then the `db_only_app` works nicely below.

jacquerie: Sure, all test I wrote inject `app` directly anyway. Might be worth checking for in-flight PRs.

PR #1082, opened 2016-04-26
converter: XSLT conversion module

jalavik:
* Adds module for converting XML documents via XSLT.

* Adds XSLT stylesheet for arXiv OAI-PMH XML to MARCXML.


tests/unit/converter/test_converter.py
```diff
@@ -0,0 +1,46 @@
...
+
+@pytest.fixture
+def oai_xml():
+    return pkg_resources.resource_string(
+        __name__,
+        os.path.join(
+            'fixtures',
+            'oai_arxiv.xml'
+        )
+    )
```
jacquerie: FYI, I really like @mihaibivol's approach: https://github.com/inspirehep/json-merger/blob/7fddfe9d1a47be0f7ef933506e82af267c4f12ec/tests/conftest.py. So eventually these fixtures will be initialized in a simpler way, which doesn't require a `__init__.py` in `tests`.

jalavik: Thanks. Yeah, I noted that this method does not play nicely locally: https://github.com/inspirehep/inspire-next/blob/master/tests/unit/dojson/test_dojson_bibfield.py#L34

it finds the first tests module it can (which varies locally).

tests/unit/converter/test_converter.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def test_xslt(oai_xml):
+    """Test conversion of XSLT from XML."""
+    xml = convert(xml=oai_xml, xslt_filename="oaiarXiv2marcxml.xsl")
+    assert xml
+    assert "10.1134/1.1533770" in xml
+    assert "oai:arXiv.org:physics/0112006" in xml
```
jacquerie: Is it wrong here to add another fixture for the expected result and assert equality between the two? Like, is it not guaranteed to be the same in all runs, stuff like that?

jacquerie:
LGTM minus the question, but remove the merge commit before merging!

PR #1084, opened 2016-04-26
setup: pytest-runner usage

jalavik:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,14 +23,11 @@
 """INSPIRE overlay repository for Invenio."""
 
 import os
-import sys
 
 from setuptools import find_packages, setup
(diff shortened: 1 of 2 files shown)


jacquerie:
I wonder why we were doing things that way.

setup.py
```diff
@@ -98,6 +95,7 @@
     'pytest-cache>=1.0',
     'pytest-cov>=1.8.0',
     'pytest-pep8>=1.0.6',
+    'pytest-runner>=2.7.0',
```
jacquerie: Ah, found the bug. `pytest-runner` should be in the `setup_require` section, not the `tests_require`.

jacquerie:
Cherry-picked this commit in #1088.

PR #1085, opened 2016-04-26
utils: dotted record key lookup

jalavik:
* Adds a utility function `get_smart_value` to allow for dotted
  record key lookup. E.g. `get_smart_value(rec, "titles.title")`.

* Migrated from legacy `invenio_utils.datastructures` package.

### DIFF
--- a/inspirehep/utils/record.py
+++ b/inspirehep/utils/record.py
@@ -22,6 +22,10 @@
 
 """Helpers for handling records."""
 
+import re
+
+SPLIT_KEY_PATTERN = re.compile('\.|\[')
+
 
 def get_title(record):
     """Get preferred title from record."""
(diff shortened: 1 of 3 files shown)


inspirehep/utils/record.py
```diff
@@ -31,3 +35,53 @@ def get_title(record):
             return title['title']
 
     return title
+
+
+def get_smart_value(record, key, default=None):
```
kaplun: I know it's silly, but to me seems this function will return me a *smart value* rather than being smart in returning values :smile: 

Wonder if it's better called: `smart_get_value()` (although it makes it a bit hidden that is a getter) or `get_value_smart()`.

jalavik: Hmm, what about `get_dotted_value`? Otherwise `get_value` could work as well. The smartness is only performed upon asking for it.

kaplun: Good point :dancer: I vote for `get_value`

PR #1089, opened 2016-04-27
workflows: antihep keyword list

jalavik:


inspirehep/kbs/corepar.kb
```diff
@@ -0,0 +1,798 @@
...
+Z0(1780)---Z0(1780)
+Z0(1865)---Z0(1865)
+Z1(1725)---Z1(1725)
+Z1(1900)---Z1(1900)
+Z1(2150)---Z1(2150)
+Z1(2500)---Z1(2500)
+Z2(4250)---Z2(4250)
+zinc---zinc
+Zino---Zino
+zirconium---zirconium
```
kaplun: Can you add a `README.rst` file to explain the meaning of these files?

kaplun:
What is antihep?

jalavik:
@kaplun This is a list of core HEP keywords that should be filtered out of the core keywords list. @fschwenn can explain better, but basically these CORE keywords should not count towards the metric number of CORE keyword, nor be displayed.

kaplun:
I love how the coverage keeps on increasing and increasing... :smile:

fschwenn:
Jan's explanation is completely right. Just the name of the issue is a bit misleading as corepar.kb should not be mixed up with https://github.com/inspirehep/inspire-next/blob/production/inspirehep/base/classifierext/taxonomies/antiHEPont.rdf which is a real 'anti'-list containing keywords like 'biology' which indicate that the content is not related to HEP.

PR #1093, opened 2016-04-28
orcid: pushing records to orcid

Panos512:
Adds the functionality to send updated/new records to orcid.
- [x] Tweak the json converter's results to fully follow orcids schema.
- [x] Convert citations to bibtex and add them to the output json. 
- [x] Extract the needed information of the author to push on his behalf to orcid.
- [x] Check json acceptance from orcid.
- [x] Tests.
- [x] Code cleanup.
- [x] Should redirect after login.

Signed-off-by: Panos Paparrigopoulos <panos.paparrigopoulos@cern

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -289,6 +289,8 @@ RECORDS_REST_ENDPOINTS = dict(
                                  ':json_v1_response'),
             'application/x-bibtex': ('inspirehep.modules.records.serializers'
                                      ':bibtex_v1_response'),
+            'application/x-orcid': ('inspirehep.modules.records.serializers'
+                                    ':orcid_response'),
             'application/x-latexeu': ('inspirehep.modules.records.serializers'
                                       ':latexeu_v1_response'),
             'application/x-latexus': ('inspirehep.modules.records.serializers'
@@ -316,6 +318,8 @@ RECORDS_REST_ENDPOINTS = dict(
             ),
             'application/x-bibtex': ('inspirehep.modules.records.serializers'
                                      ':bibtex_v1_search'),
+            'application/x-orcid': ('inspirehep.modules.records.serializers'
+                                    ':orcid_search'),
             'application/x-latexeu': ('inspirehep.modules.records.serializers'
                                       ':latexeu_v1_search'),
             'application/x-latexus': ('inspirehep.modules.records.serializers'
@@ -821,6 +825,7 @@ INDEXER_BULK_REQUEST_TIMEOUT = 120
 
 # OAuthclient
 # ===========
+orcid.REMOTE_APP['params']['request_token_params'] = {'scope': '/orcid-profile/read-limited /activities/update /orcid-bio/update'}
 OAUTHCLIENT_REMOTE_APPS = dict(
     orcid=orcid.REMOTE_APP,
 )
@@ -829,6 +834,12 @@ OAUTHCLIENT_ORCID_CREDENTIALS = dict(
     consumer_secret="CHANGE_ME",
 )
 
+OAUTHCLIENT_SETTINGS_TEMPLATE = 'inspirehep_theme/page.html'
+
+ORCID_SYNCHRONIZATION_ENABLED = False
+
+# Error Pages
+# ========
 THEME_401_TEMPLATE = "inspirehep_theme/errors/401.html"
 THEME_403_TEMPLATE = "inspirehep_theme/errors/403.html"
 THEME_404_TEMPLATE = "inspirehep_theme/errors/404.html"
--- a/inspirehep/modules/orcid/__init__.py
+++ b/inspirehep/modules/orcid/__init__.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import (
+    absolute_import,
+    division,
+    print_function)
+
+from .ext import INSPIREOrcid
+
+from .receivers import *
+
+__all__ = ('INSPIREOrcid',)
--- a/inspirehep/modules/orcid/config.py
+++ b/inspirehep/modules/orcid/config.py
@@ -1,0 +1,39 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import (
+    absolute_import,
+    division,
+    print_function)
+
+ORCID_WORK_TYPES = {
+    "book": "BOOK",
+    "conferencepaper": "CONFERENCE_PAPER",
+    "proceedings": "BOOK",
+    "preprint": "WORKING_PAPER",
+    "note": "WORKING_PAPER",
+    "published": "JOURNAL_ARTICLE",
+    "thesis": "DISSERTATION",
+    "lectures": "LECTURE_SPEECH",
+    "bookchapter": "BOOK_CHAPTER",
+    "report": "REPORT"
+}
--- a/inspirehep/modules/orcid/ext.py
+++ b/inspirehep/modules/orcid/ext.py
@@ -1,0 +1,54 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import (
+    absolute_import,
+    division,
+    print_function)
+
+import orcid
+
+
+class INSPIREOrcid(object):
+    """INSPIRE orcid extension."""
+
+    def __init__(self, app=None, **kwargs):
+        """Extension initialization."""
+        self.app = app
+        if app:
+            self.init_app(app, **kwargs)
+
+    def init_app(self, app, **kwargs):
+        """Initialize application object."""
+        app.extensions['inspire-orcid'] = self
+
+        orcid_base_url = app.config['OAUTHCLIENT_REMOTE_APPS'][
+            'orcid']['params']['base_url']
+        orcid_consumer_secret = app.config[
+            'OAUTHCLIENT_ORCID_CREDENTIALS']['consumer_secret']
+        orcid_consumer_key = app.config[
+            'OAUTHCLIENT_ORCID_CREDENTIALS']['consumer_key']
+
+        self.sandbox = (orcid_base_url == 'https://pub.sandbox.orcid.org/')
+
+        self.orcid_api = orcid.MemberAPI(
+            orcid_consumer_secret, orcid_consumer_key, sandbox=self.sandbox)
--- a/inspirehep/modules/orcid/models.py
+++ b/inspirehep/modules/orcid/models.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from invenio_db import db
+
+from sqlalchemy_utils.types import UUIDType
+
+
+class InspireOrcidRecords(db.Model):
+    __tablename__ = 'inspire_orcid_records'
+
+    orcid = db.Column(db.String(160), primary_key=True)
+    record_id = db.Column(UUIDType)
+    put_code = db.Column(db.Integer, primary_key=True)
--- a/inspirehep/modules/orcid/receivers.py
+++ b/inspirehep/modules/orcid/receivers.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from inspirehep.modules.records.signals import after_record_enhanced
+
+from invenio_records.signals import before_record_delete
+
+from .tasks import delete_from_orcid, send_to_orcid
+
+from flask import current_app
+
+
+@after_record_enhanced.connect
+def send_records_to_orcid(sender, *args, **kwargs):
+    """ Schedules a Celery task that sends every new/updated record to orcid.
+
+        :param sender: The record to be sented to orcid (in json format).
+    """
+    if current_app.config.get('ORCID_SYNCHRONIZATION_ENABLED'):
+        send_to_orcid.delay(sender=sender)
+
+
+@before_record_delete.connect
+def delete_record_from_orcid(sender, *args, **kwargs):
+    """ Schedules a Celery task that removes records from orcid.
+
+        :param sender: The record to be deleted from orcid (in json format).
+    """
+    if current_app.config.get('ORCID_SYNCHRONIZATION_ENABLED'):
+        delete_from_orcid.delay(sender=sender)
--- a/inspirehep/modules/orcid/schema.py
+++ b/inspirehep/modules/orcid/schema.py
@@ -1,0 +1,148 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from HTMLParser import HTMLParser
+
+import dojson
+
+from dojson import utils
+
+from inspirehep.utils.record import get_abstract, get_subtitle, get_title
+
+from .config import ORCID_WORK_TYPES
+
+orcid_overdo = dojson.Overdo()
+
+
+class MLStripper(HTMLParser):
+
+    def __init__(self):
+        self.reset()
+        self.fed = []
+
+    def handle_data(self, d):
+        self.fed.append(d)
+
+    def get_data(self):
+        return ''.join(self.fed)
+
+
+def strip_tags(html):
+    """ Strips html tags out of a given string.
+    """
+    if html is None:
+        html = ''
+    s = MLStripper()
+    s.feed(html)
+    return s.get_data()
+
+
+@orcid_overdo.over('title', 'titles')
+def title_rule(self, key, value):
+    title = get_title({"titles": value})
+    if title == '':
+        raise KeyError
+    subtitle = get_subtitle({"titles": value})
+    return {"title": title,
+            "subtitle": subtitle}
+
+
+@orcid_overdo.over('journal-title', 'publication_info')
+def publication_rule(self, key, value):
+    try:
+        return value[0]['journal_title']
+    except (TypeError, KeyError):
+        pass
+
+
+@orcid_overdo.over('short-description', 'abstracts')
+def abstract_rule(self, key, value):
+    return strip_tags(get_abstract({"abstracts": value}))
+
+
+@orcid_overdo.over('publication-date', 'imprints')
+def date_rule(self, key, value):
+    try:
+        date = value[0].get('date')
+        date = date.split('-')
+        final_date = ['', '', '']
+        if date[0]:
+            final_date[0] = date[0]
+        if date[1]:
+            final_date[1] = date[1]
+        if date[2]:
+            final_date[2] = date[2]
+        publication_date = {'year': int(final_date[0]),
+                            'month': int(final_date[1]),
+                            'day': int(final_date[2])
+                            }
+        return publication_date
+    except (TypeError, IndexError, AttributeError):
+        pass
+
+
+@orcid_overdo.over('work_external_identifier', 'arxiv_eprints|dois')
+@dojson.utils.for_each_value
+def external_id_rule(self, key, value):
+    if key == 'dois':
+        return{
+            'external-identifier-type': 'DOI',
+            'external-identifier-id': value.get('value')
+        }
+    if key == 'arxiv_eprints':
+        return {
+            'external-identifier-type': 'ARXIV',
+            'external-identifier-id': value.get('value')
+        }
+
+
+@orcid_overdo.over('contributors', 'authors')
+def authors_rule(self, key, value):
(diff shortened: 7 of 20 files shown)


scripts/recreate_records
```diff
@@ -64,6 +64,5 @@ inspirehep db create
 inspirehep index destroy --force --yes-i-know
 inspirehep index init
 
-inspirehep migrator populate -f inspirehep/demosite/data/demo-records.xml.gz --wait=true
-inspirehep migrator count_citations
-
+# inspirehep migrator populate -f inspirehep/demosite/data/demo-records.xml.gz --wait=true
+# inspirehep migrator count_citations
```
Panos512: ignore that

inspirehep/modules/orcid/models.py
```diff
@@ -0,0 +1,32 @@
...
+
+from invenio_db import db
+
+
+class InspireOrcidRecords(db.Model):
+    __tablename__ = 'inspire_orcid_records'
+
+    id = db.Column(db.Integer, primary_key=True, index=True)
+    orcid = db.Column(db.String(160))
+    record_id = db.Column(db.Integer)
```
kaplun: Actually I believe that the primary key in this context is given by the combination `record_id` + `orcid` which is required to retrieve the put codes. With that you then probably don't need the id.

and `record_id` should then be a foreign key to the UUID of the `Record` model. So not an integer (unfortunately...)

Mmh now I wonder what will happen when 2 records are merged :question:

inspirehep/modules/orcid/receivers.py
```diff
@@ -0,0 +1,30 @@
...
+# or submit itself to any jurisdiction.
+
+from inspirehep.modules.records.signals import after_record_enchanced
+
+from .tasks import send_to_orcid
+
+
+@after_record_enchanced.connect
+def send_records_to_orcid(sender, *args, **kwargs):
+    send_to_orcid.delay(sender, *args, **kwargs)
```
kaplun: Can you document what you expect `sender` to be? (I assume you don't use args and kwargs...)

inspirehep/modules/orcid/schema.py
```diff
@@ -0,0 +1,148 @@
...
+        "conferencepaper": "CONFERENCE_PAPER",
+        "proceedings": "BOOK",
+        "preprint": "WORKING_PAPER",
+        "note": "WORKING_PAPER",
+        "published": "JOURNAL_ARTICLE",
+        "thesis": "DISSERTATION",
+        "lectures": "LECTURE_SPEECH",
+        "bookchapter": "BOOK_CHAPTER",
+        "report": "REPORT"
+    }
```
kaplun: can you move this magic values in a `.config` file? Should ORCID/INSPIRE evolve on this...

inspirehep/modules/orcid/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+            api = current_app.extensions['inspire-orcid'].orcid_api
+            orcid_json = convert_to_orcid(sender)
+            try:
+                authors = get_orcid_valid_authors(sender)
+                for author in authors:
+                    token, author_orcid = get_authors_credentials(author)
+                    if not InspireOrcidRecords.query.filter_by(orcid=author_orcid, record_id=int(sender.get('control_number'))).all():
+                        put_code = api.add_record(
+                            author_orcid, token, 'work', orcid_json)
+                        x = InspireOrcidRecords(
```
kaplun: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

inspirehep/modules/orcid/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+            except RequestException as e:
+                print(e.response.text)
+                print sender['control_number']
+        except (KeyError, AttributeError) as e:
+            exc_type, exc_obj, exc_tb = sys.exc_info()
+            fname = os.path.split(exc_tb.tb_frame.f_code.co_filename)[1]
+            print(exc_type, fname, exc_tb.tb_lineno)
+
+
+def get_authors_credentials(author):
```
kaplun: You might want to move this function into an `.utils.py` file, since it's quite neat in terms of input/output and can in principle be tested separately.

inspirehep/modules/orcid/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+        if orcid_id['type'] == 'ORCID':
+            author_orcid = orcid_id['value']
+    raw_user = UserIdentity.query.filter_by(
+        id=author_orcid, method='orcid').first()
+    user = RemoteAccount.query.get(raw_user.id_user)
+    token = user.tokens[0].access_token
+    return (token, author_orcid)
+
+
+def get_orcid_valid_authors(sender):
```
kaplun: Can you rename `sender` to something more semantic? E.g. `record` :)

inspirehep/modules/orcid/tasks.py
jacquerie: Cat on keyboard?

inspirehep/modules/orcid/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+    }
+    authors = current_search_client.search(
+        index='records-authors',
+        doc_type='authors',
+        body=es_query
+    )['hits']['hits']
+    if authors:
+        return authors
+    else:
+        return []
```
jacquerie: Do we need this? Isn't `authors` going to be an empty list anyway? `return authors` should be enough then.

inspirehep/modules/orcid/tasks.py
```diff
@@ -0,0 +1,125 @@
...
+        id=author_orcid, method='orcid').first()
+    user = RemoteAccount.query.get(raw_user.id_user)
+    token = user.tokens[0].access_token
+    return (token, author_orcid)
+
+
+def get_orcid_valid_authors(sender):
+    """ Returns all the valid author-records from a hep-record.
+    A valid author-rerord is one that has 'curated_relation' = True
+    and contains an orcid id.
```
kaplun: Actually, you can ignore the `curated_relation` bit, because anyway you would find the ORCID if and only if it's either curated or it comes already from an external source.

inspirehep/modules/orcid/utils.py
```diff
@@ -0,0 +1,52 @@
...
+
+    orcid_json['external-identifiers'] = {
+        'work-external-identifier': orcid_json['work_external_identifier']}
+    orcid_json.pop('work_external_identifier')
+    orcid_json['citation'] = {'citation': Bibtex(
+        record).format(), 'citation-type': 'BIBTEX'}
+
+    try:
+        orcid_json['publication-date'] = orcid_json['publication_date']
+        orcid_json.pop('publication_date')
```
kaplun: Uh? Why do you need this?

inspirehep/modules/orcid/schema.py
```diff
@@ -0,0 +1,148 @@
...
+
+@orcid_overdo.over('short_description', 'abstracts')
+def abstract_rule(self, key, value):
+    try:
+        return get_abstract({"abstracts": value})
+    except (TypeError, KeyError):
+        pass
+
+
+@orcid_overdo.over('publication_date', 'imprints')
```
kaplun: Here you can use `publication-date` (rather than `publication_date`)

inspirehep/modules/orcid/schema.py
```diff
@@ -0,0 +1,148 @@
...
+
+@orcid_overdo.over('journal_title', 'publication_info')
+def publication_rule(self, key, value):
+    try:
+        return value[0].get('journal_title')
+    except (TypeError, KeyError):
+        pass
+
+
+@orcid_overdo.over('short_description', 'abstracts')
```
kaplun: See comment on below decorator

inspirehep/modules/orcid/schema.py
```diff
@@ -0,0 +1,148 @@
...
+    subtitle = ""
+    try:
+        subtitle = get_subtitle({"titles": value})
+    except KeyError:
+        pass
+    return {"title": title,
+            "subtitle": subtitle}
+
+
+@orcid_overdo.over('journal_title', 'publication_info')
```
kaplun: See comment on below decorator

inspirehep/modules/orcid/schema.py
```diff
@@ -0,0 +1,148 @@
...
+                'contributor-sequence': 'FIRST'
+            }
+        })
+        for author in value[1:]:
+            orcid_authors.append({
+                'credit-name': author.get('full_name'),
+                'contributor-orcid': author.get('orcid') if 'orcid' in author else '',
+                'contributor-attributes': {
+                    'contributor-role': 'AUTHOR',
+                    'contributor-sequence': 'ADDITIONAL'
```
kaplun: Let's see what ORCID people say about this. I remember we had some discussions in the past about the size of our giant authorlists.

inspirehep/modules/orcid/schema.py
Panos512: `-` sadly is not accepted by dojson. That's why I have to rename it afterwards.

inspirehep/modules/records/mappings/records/authors.json
```diff
@@ -113,6 +113,15 @@
                             "index": "not_analyzed"
                         }
                     }
+                },
+                "self":{
```
kaplun: What is this mapping?

inspirehep/modules/orcid/tasks.py
Panos512: I think the call just right before this returns None in case of a problem (instead of an empty list).

inspirehep/modules/records/signals.py
```diff
@@ -0,0 +1,29 @@
...
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Record module signals."""
+
+from blinker import Namespace
+
+_signals = Namespace()
+
+after_record_enchanced = _signals.signal('before-full-record-index')
```
kaplun: Can you also set the string to `after-record-enhanced`?

scripts/recreate_records
kaplun: I will

inspirehep/modules/records/mappings/records/authors.json
Panos512: In order to be able to run the query at https://github.com/Panos512/inspire-next/blob/orcid/inspirehep/modules/orcid/tasks.py#L104 refs need to be not_analyzed since `Terms` keyword needs not_analyzed fields to find exact matches.

tests/unit/orcid/test_orcid_converter.py
```diff
@@ -0,0 +1,221 @@
...
+            },
+            {
+                "title": "Title2"
+            },
+            {
+                "source": "other",
+                "title": "Title3"
+            }
+        ]
+    }
```
kaplun: All these are things that can be put in fixtures, so that you don't have to hardcode such huge blobs into python code. @mihaibivol @jacquerie can show you in case :)

inspirehep/modules/records/signals.py
Panos512: whoops. ofc

inspirehep/modules/orcid/tasks.py
kaplun: you can `return authors or []` :)

inspirehep/modules/orcid/schema.py
kaplun: image

Panos512: Yeah that was my exact reaction :D It throws an `error: bad character in group name 'Ijournal-title'`

kaplun: @jirikuncar just implemented a fix for this very case in https://github.com/inveniosoftware/dojson/pull/140. Now you can use any value for the destination and you don't need any more the above workaround.

Panos512: 👍  That's great news! I fixed the conversion.

tests/unit/orcid/test_orcid_converter.py
Panos512: I talked with @mihaibivol  and @jacquerie  and we think that maybe these should stay here. Since these two records are orcid specific in a way (they only include fields that orcid accepts) and they are not going to be used in other tests maybe we shouldn't move them to a separate file. What do you think?

jacquerie:
@Panos512, can you push a recent version here, even if it's not ready for review?

Panos512:
@jacquerie Just did! Sorry for the delay. There are still some issues I am fixing right now though.

inspirehep/config.py
```diff
@@ -760,7 +764,8 @@ def _(x):
             view="invenio_oauthclient.handlers:signup_handler",
         ),
         params=dict(
-            request_token_params={'scope': '/authenticate'},
+            request_token_params={
+                'scope': '/orcid-profile/read-limited /activities/update /orcid-bio/update'},
```
jacquerie: Not an expert of this, but why is `/authenticate` gone?

kaplun: Because it's a subset of the other scopes... See bottom of https://members.orcid.org/api/orcid-scopes

inspirehep/modules/orcid/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from .ext import INSPIREOrcid
```
jacquerie: Please add `__future__` block:
```python
from __future__ import (
    absolute_import,
    division,
    print_function,
    unicode_literals)
```

inspirehep/modules/orcid/config.py
```diff
@@ -0,0 +1,34 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+work_types = {
```
jacquerie: This guy should probably be called `ORCID_WORK_TYPES`.

inspirehep/modules/orcid/config.py
```diff
@@ -0,0 +1,34 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
```
jacquerie: `__future__` block.

inspirehep/modules/orcid/ext.py
```diff
@@ -0,0 +1,44 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+import orcid
```
jacquerie: __future__ block.

inspirehep/modules/orcid/__init__.py
jirikuncar: I would recommend not to use `unicode_literals`.

inspirehep/modules/orcid/ext.py
```diff
@@ -0,0 +1,44 @@
...
+        self.app = app
+        if app:
+            self.init_app(app, **kwargs)
+
+    def init_app(self, app, **kwargs):
+        """Initialize application object."""
+        app.extensions['inspire-orcid'] = self
+        self.sandbox = False
+        if app.config['OAUTHCLIENT_REMOTE_APPS']['orcid']['params']['base_url'] == 'https://pub.sandbox.orcid.org/':
+            self.sandbox = True
```
jacquerie: These lines can become:
```python
orcid_base_url = app.config['OAUTHCLIENT_REMOTE_APPS']['orcid']['params']['base_url']
self.sandbox = (orcid_base_url == 'https://pub.sandbox.orcid.org/')
```

inspirehep/modules/orcid/ext.py
```diff
@@ -0,0 +1,44 @@
...
+        self.orcid_api = orcid.MemberAPI(
+            app.config['OAUTHCLIENT_ORCID_CREDENTIALS']['consumer_secret'],
+            app.config['OAUTHCLIENT_ORCID_CREDENTIALS']['consumer_key'],
+            sandbox=self.sandbox)
```
jacquerie: Or we can go full [extract variable](https://sourcemaking.com/refactoring/extract-variable) and rewrite this method as:
```python
def init_app(self, app, **kwargs):
    """Initialize application object."""
    app.extensions['inspire-orcid'] = self

    orcid_base_url = app.config['OAUTHCLIENT_REMOTE_APPS']['orcid']['params']['base_url']
    orcid_consumer_secret = app.config['OAUTHCLIENT_ORCID_CREDENTIALS']['consumer_secret']
    orcid_consumer_key = app.config['OAUTHCLIENT_ORCID_CREDENTIALS']['consumer_key']

    self.sandbox = (orcid_base_url == 'https://pub.sandbox.orcid.org/')

    self.orcid_api = orcid.MemberAPI(
         orcid_consumer_secret, orcid_consumer_key, sandbox=self.sandbox)
```

inspirehep/modules/orcid/__init__.py
jacquerie: I saw that this in fact debated (http://python-future.org/unicode_literals.html#unicode-literals), but decided to try to add it everywhere and see what breaks. Was that a stupid idea?

inspirehep/modules/orcid/models.py
```diff
@@ -0,0 +1,33 @@
...
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from invenio_db import db
+from sqlalchemy_utils.types import UUIDType
+
+class InspireOrcidRecords(db.Model):
+    __tablename__ = 'inspire_orcid_records'
```
jacquerie: PEP8 / general formatting quite broken in this file.

tests/unit/utils/test_utils_record.py
```diff
@@ -52,28 +54,145 @@ def single_title():
...
+
+
+@pytest.fixture
+def empty_abstract():
+    return {
+        "abstracts": []
+    }
+
+
+def test_get_title(double_title, single_title, empty_title, double_title_with_multiple_sources, title_only_from_arxiv):
```
jacquerie: Some time ago I rewrote these tests in a style I like more, but I never pushed it anywhere: https://github.com/jacquerie/inspire-next/commit/ecb6826a4fa169efea634dc3672cc59edc4c4d09

inspirehep/modules/orcid/tasks.py
```diff
@@ -0,0 +1,131 @@
...
+@shared_task(ignore_result=True)
+def send_to_orcid(sender):
+    """Sends records to orcid."""
+    index, doc_type = schema_to_index(sender['$schema'])
+    # FIXME: next two lines are there for debugging reasons.
+    if 'authors' in sender:
+        sender['authors'].append({u'affiliations': [{u'value': u'St. Petersburg, INP'}],  u'curated_relation': True,  u'full_name': u'Paparrigopoulos, Panos',  u'profile': {
+                                 u'__url__': u'http://inspirehep.net/record/1064588'},  u'record': {u'$ref': u'http://localhost:5000/api/authors/10'}})
+    if doc_type == 'hep':
+        logger.info("Sending " + sender.get('control_number') + " to orcid.")
```
jacquerie: Can you show me an example of what the logging looks like when sending stuff to ORCID?

inspirehep/modules/orcid/tasks.py
```diff
@@ -0,0 +1,131 @@
...
+        try:
+            api = current_app.extensions['inspire-orcid'].orcid_api
```
jacquerie: For testability `api` should be injected in the method. Wouldn't hurt to call it `orcid_api`, since it's what it is :D

inspirehep/modules/orcid/config.py
Panos512: even if I put it on `__init__.py`?

tests/unit/utils/test_utils_record.py
Panos512: great! I will cherrypick it and rewrite my tests on top 😄

inspirehep/modules/orcid/__init__.py
jirikuncar: It causes a lot of problems with DB, templates, etc. Try to avoid it.

jacquerie: Thanks! :+1:

Well, I guess this weekend I will submit the big import cleanup PR.

inspirehep/modules/orcid/tasks.py
Panos512:
```
17:05:08 worker.1    | [2016-05-20 17:05:08,805: INFO/Worker-4] inspirehep.modules.orcid.tasks.send_to_orcid[effa1dba-709d-4f13-8c75-bc1cd7f60e50]: Sending 712925 to orcid.
```

👯

kaplun:
Uh @jacquerie what is this last commit?

Panos512:
@kaplun It is the test skeleton I have cherry-picked from Jacopo to write the tests on top.

kaplun:
Oh. Because it look empty.

jacquerie:
Yeah, it looks empty because all its changes were already part of @Panos512's commit. I think it can be removed.

kaplun:
@Panos512, as mentioned this morning please introduce a global config flag into root `config.py` such as `ORCID_SYNCHRONIZATION_ENABLED = False` which basically protect the registration of the receiver.

Panos512:
@kaplun @jacquerie The commit is gone and in it's place we have the flag 💃

inspirehep/modules/orcid/tasks.py
```diff
@@ -0,0 +1,168 @@
...
+                    author_orcid = author[2]
+                    record_id = author[3]
+                    if not put_code:
+                        put_code = api.add_record(  # try-continue
+                            author_orcid, token, 'work', orcid_json)
+                        orcid_log_record = InspireOrcidRecords(
+                            orcid=author_orcid,
+                            record_id=record_id,
+                            put_code=put_code,
+                            token=token)
```
kaplun: Why here you store the token? Isn't this de-normalizing the token information which is already in other tables?

inspirehep/modules/orcid/utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+from inspirehep.utils.bibtex import Bibtex
+
+from invenio_oauthclient.models import RemoteAccount, UserIdentity
+
+from .schema import orcid_overdo
+
+
+def convert_to_orcid(record):
+    """ Converts a given record to a json containging the information that
```
kaplun: Typo: `containging`

inspirehep/modules/orcid/utils.py
```diff
@@ -0,0 +1,56 @@
...
+    orcid_json['citation'] = {'citation': Bibtex(
+        record).format(), 'citation-type': 'BIBTEX'}
+
+    return orcid_json
+
+
+def get_authors_credentials(author):
+    """Returns the orcid-id and the orcid-token for a specific author (if available)."""
+    author_orcid = ''
+    author = author['_source']
```
kaplun: Here it is not declared that the author is actually coming from elastic search. And for simplicity reason we could infact make the `get_authors_credentials` function to also work with an author coming from the DB.

Can you extract the `_source` field in the caller, rather than here.

inspirehep/modules/records/signals.py
```diff
@@ -0,0 +1,29 @@
...
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Record module signals."""
+
+from blinker import Namespace
+
+_signals = Namespace()
+
+after_record_enchanced = _signals.signal('after-record-enchanced')
```
jmartinm: `after_record_enchanced` -> `after_record_enhanced` ?

kaplun:
@Panos512 indeed 0 is not equal to 1: https://travis-ci.org/inspirehep/inspire-next/jobs/138915148#L2168 :dancer:

inspirehep/modules/records/receivers.py
```diff
@@ -32,6 +30,28 @@
...
+
+from invenio_indexer.signals import before_record_index
+
+import six
+
+from .signals import after_record_enhanced
+
+
+@before_record_index.connect
+def enhance_record(recid, json, *args, **kwargs):
```
jirikuncar: Please note that first argument should be `sender` which points to the connected Flask application.

jirikuncar:
What about making the module generic enough so any instance can use it? See an example with Invenio-GitHub module. If you agree I can prepare template for Invenio-ORCID module.

inspirehep/modules/theme/templates/inspirehep_theme/accounts/login.html
```diff
@@ -33,7 +33,7 @@
 <div class="col-xs-12 col-sm-10 col-md-8 col-lg-6 col-centered">
   <div class="well">
     <br/>
-    {% set next = request.url if request.endpoint != 'security.logout' and request.endpoint != 'security.login' else config.SERVER_NAME %}
+    {% set next = request.url if request.endpoint != 'security.logout' and request.endpoint != 'security.login' else request.args.get('next', '') %}
```
jirikuncar: what about fixing a default macro in https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/templates/invenio_oauthclient/_macros.html#L54 ?

kaplun:
@jirikuncar we have thought about that but we fear that are maybe too many points that would be specific to an instance, and that would render making such module generic. In particular we would need to make configurable:
* [ ] What records would need to be sent to ORCID (and would therefore trigger an update/insert (in our case any record belonging to `Literature` and `Data`.
* [ ] An algorithm to retrieve which author with an ORCID are involved for the affected record (this is the tricky part: in our case we are retrieving all the signature from a paper, then we consider all the claimed ones, then we are following up into the corresponding authority record and retrieve the ORCID from there.
* [ ] A way to retrieve from the ORCID the given token (or maybe we can just specify here the name of the oauth service used, in our case `orcid`.
* [ ] A mapping from the inner schema to the ORCID specific one (we have developed our internal `dojson` mapping, for that.

As you can see if you move all this into a `contrib`, what is left is just the model needed to store ORCID's `putcode` for the association between a profile and a paper.

On the other hand it's true that would make such integration more elegant and reusable (assuming other installation are willing to reimplement all the above points).

kaplun:
One complication that I also see, if we were going generic and storing INSPIRE specific bits into `contrib` folder, those bits would depend sometime on the INSPIRE overlay. If we were not then shipping those bits, a complete example of usage of such module would then miss...

jirikuncar:
> What records would need to be sent to ORCID (and would therefore trigger an update/insert (in our case any record belonging to Literature and Data.

It can be defined as a query to which a record has to match.

> An algorithm to retrieve which author with an ORCID are involved for the affected record (this is the tricky part: in our case we are retrieving all the signature from a paper, then we consider all the claimed ones, then we are following up into the corresponding authority record and retrieve the ORCID from there.

Installations can provide `extract_authors` function. For records coming from deposit, it can be pretty straight forward.

> A way to retrieve from the ORCID the given token (or maybe we can just specify here the name of the oauth service used, in our case orcid.

It can be done through following relation: Author -> User -> Linked OAuth account (ORCID). You can see an example in Invenio-GitHub.

> A mapping from the inner schema to the ORCID specific one (we have developed our internal dojson mapping, for that.

One can configure a serializer as it is done in Invenio-Records-REST.

You specific implementation can be in INSPIRE overlay the same way as Zenodo is doing.

kaplun:
OK, @jirikuncar thanks for the suggestions and indications.
Since this particular project is basically ready to be merged we are still going to proceed with this particular PR, since it's in our milestones.
Nevertheless we are going to refactor it into an Invenio module, as it might be useful to others.

I am going to create it in our  Github project, because we still need to ensure it first meets our service needs (also in time of releases and overall control), but we are going to base it on all the patterns of a generic Invenio module.

jirikuncar:
@kaplun thank you for your answer. Please note, that I have added Invenio-ORCID repo under @inveniosoftware organization so everybody can find this integration module. I can try to quickly filter history and code from this PR to the new repo so it's easy to start there in the future.

kaplun:
OK @jirikuncar thanks for the history bits.

As long as @inspirehep/core have also admin rights on the repo I am fine with that.

kaplun:
@Panos512 Can you just rebase it? I will then open an issue to migrate to official inveniosoftware/invenio-orcid.

kaplun:
@Panos512 test is currently failing. D'oh!

tests/unit/orcid/test_orcid_converter.py
```diff
@@ -0,0 +1,497 @@
...
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Tests for json convertion to orcid form."""
```
kaplun: Note that it's `conversion`.

tests/unit/orcid/test_orcid_converter.py
```diff
@@ -0,0 +1,497 @@
...
+
+from inspirehep.modules.orcid.utils import convert_to_orcid
+
+import pytest
+
+
+def test_succesfull_convertion():
```
kaplun: Ditto :)

tests/unit/orcid/test_orcid_converter.py
```diff
@@ -0,0 +1,497 @@
...
+                                                                       'external-identifier-type': 'ARXIV'}]},
+                'journal-title': 'Nuovo Cim.',
+                'short-description': 'Abstract',
+                'title': {'subtitle': 'Subtitle', 'title': 'Title3'},
+                'type': 'REPORT'}
+    result = convert_to_orcid(record)
+    assert expected == result
+
+
+def test_arxiv_missing_convertion():
```
kaplun: ditto

PR #1096, opened 2016-05-02
schemas: fix hep collections bug

jacenkow:
* Removes string type of items in collections field.


jacquerie:
Thanks! Will merge immediately after green build.

PR #1103, opened 2016-05-09
docker: add ignore and update lib

jalavik:
Thanks to @mihaibivol and @jacquerie


jalavik:
Can be merged without CI as it is not part of our CI.

PR #1104, opened 2016-05-09
docs: brings back docs

jalavik:
Brings back docs from v2 branches.

To build do: `sphinx-build -qnNW docs docs/_build/html`.


INSTALL.rst
```diff
@@ -91,8 +91,3 @@ Access the record (web/rest)
 
    firefox http://localhost:5000/records/1
    curl -i -H "Accept: application/json" http://localhost:5000/api/records/1
-
-
-======
-ENJOY!
-======
```
kaplun: No!? We shall be sad?

docs/grobid.rst
```diff
@@ -0,0 +1,74 @@
...
+The models available after cloning are not using the new available training
+data. To generate the new ones we need to go inside of the root folder and
+call:
+
+.. code-block:: bash
+
+    $ cd grobid
+    $ java -Xmx1024m -jar grobid-trainer/target/grobid-trainer-0.3.4-SNAPSHOT.one-jar.jar 0 $MODEL -gH grobid-home
+
+where ``$MODEL`` is the model we want to train. Note that there's new data
```
kaplun: Is there more information on what is a `$MODEL`? Can you add some online pointer?

jacquerie: That would be http://grobid.readthedocs.io/en/latest/Training-the-models-of-Grobid/, but in fact the only models that makes sense to train are the ones listed below (`segmentation` and `header`).

kaplun: I miss something: "should I put really the word `segmentation` there? And where the input is provided?

jacquerie: Yes, you should put the word `segmentation` there. Training data (the input) MUST be in folder `grobid-trainer/resources/dataset/$MODEL/corpus`.

jalavik: So shall I update something here or are we okay to 🚢  for now?

INSTALL.rst
jalavik: Hehe, I brought it back, but without the headline causing issues with table of contents.

docs/grobid.rst
jacquerie: For me, :shipit:, but let's hear what @kaplun says.

kaplun: Ah no please go on! We can always fix it later.

kaplun: Is there just a reference that can be put (E.g. to official grobid documentation) to exaplain how the corpus should be structured then?

PR #1106, opened 2016-05-10
WIP authors: split blueprints

jacenkow:
* Split author module view into holdingpen and publications views.

* Adds basic test to prevent regression.

Co-authored-by: Jacopo Notarstefano <jacopo.notarstefano@cern.ch>

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -43,7 +43,7 @@ def app(request):
 
             sleep(10)  # Makes sure that ES is up.
             _es = app.extensions['invenio-search']
-            list(_es.delete(ignore=[400]))
+            list(_es.delete(ignore=[404]))
 
     request.addfinalizer(teardown)
 
@@ -56,6 +56,7 @@ def app(request):
 
         sleep(10)  # Makes sure that ES is up.
         _es = app.extensions['invenio-search']
+        list(_es.delete(ignore=[404]))
         list(_es.create(ignore=[400]))
 
         migrate('./inspirehep/demosite/data/demo-records.xml.gz', wait_for_results=True)
--- a/tests/integration/test_author_views.py
+++ b/tests/integration/test_author_views.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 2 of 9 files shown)


jacenkow:
CC: @jacquerie

tests/integration/test_author_views.py
```diff
@@ -0,0 +1,46 @@
...
+        assert client.get(
+            '/author/publications?recid=984519').status_code == 200
+
+
+def test_author_publications_returns_the_right_data(app):
+    with app.test_client() as client:
+        response = json.loads(
+            client.get('/author/publications?recid=984519').data)
+
+        # FIXME: Assert should be more specific.
```
jacquerie: Nit: `# FIXME: asserts should be more specific.`

jacquerie:
### TODO:

- [ ] We fail due to a bundling error (https://travis-ci.org/inspirehep/inspire-next/jobs/129155669#L1264). In fact, we are still declaring that our templates and static files are at the same level as the blueprint, which is no longer the case. This should be an easy fix for tomorrow.
- [ ] Add a commit to fix the conference view issue we have been seeing for a while (@eamonnmag).
- [ ] Add a commit to delete ES data in the integration fixture, so that we don't get bogus failures.

When this is done, I think we can 🚢 this PR.

tests/integration/conftest.py
```diff
@@ -56,6 +56,7 @@ def teardown():
 
         sleep(10)  # Makes sure that ES is up.
         _es = app.extensions['invenio-search']
+        list(_es.delete(ignore=[404]))
```
jacquerie: Move these changes to the previous commit, because they belong there.

tests/integration/conftest.py
```diff
@@ -56,6 +56,7 @@ def teardown():
 
         sleep(10)  # Makes sure that ES is up.
         _es = app.extensions['invenio-search']
+        list(_es.delete(ignore=[400]))
```
jacquerie: The problem with doing this is that, if you `git bisect`, you get a broken build on this commit. Ideally we should be passing the suite at every commit!

tests/integration/conftest.py
jacenkow: Done.

jacquerie: Thanks!

jacquerie:
Commit 2479f6f should be replaced with https://github.com/jacquerie/inspire-next/commit/f56e5810cdf0044d745581d11cb652ef699a8e52 due to an error found by @jmartinm. Then, after a rebase, we :shipit:.

jacquerie:
Superseded by #1112.

PR #1083, opened 2016-04-26
format: Job detailed record

nikpap:
capture d ecran 2016-05-09 a 11 24 28

Signed-off-by: Nick Papoutsis <admin@nils.gr>

PR #1115, opened 2016-05-13
tests: integration tests for exporters

jacquerie:


tests/integration/test_latex_exporting.py
```diff
@@ -0,0 +1,64 @@
...
+
+@pytest.mark.xfail(reason='wrong output')
+def test_format_latex_eu(app):
+    article = get_db_record('literature', 4328)
+    today = date.today().strftime('%d %b %Y')
+
+    expected = u'''%\cite{Glashow:1961tr}
+\\bibitem{Glashow:1961tr}
+  S.~L.~Glashow,
+  %``Partial Symmetries of Weak Interactions,''
```
jacquerie: @kaplun Here is where we are missing the ending quotes in the actual output.

jacquerie:
I will rebase after merging #1083, but other than that this looks good to go.

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -218,26 +218,6 @@ def remove_duplicates_from_dict(value):
...
-def cv_latex(record):
-    return Cv_latex(record).format()
-
-
-@blueprint.app_template_filter()
-def cv_latex_html_text(record, format_type, separator):
-    return Cv_latex_html_text(record, format_type, separator).format()
-
-
-@blueprint.app_template_filter()
```
kaplun: Can you explain better in the commit log why you are removing them? What is there instead?

tests/integration/test_latex_exporting.py
```diff
@@ -0,0 +1,64 @@
...
+from inspirehep.utils.latex import Latex
+from inspirehep.utils.record_getter import get_db_record
+
+import pytest
+
+
+@pytest.mark.xfail(reason='wrong output')
+def test_format_latex_eu(app):
+    article = get_db_record('literature', 4328)
+    today = date.today().strftime('%d %b %Y')
```
kaplun: There's a slight chance that this fails if the test is run after midnight but the records where loaded before midnight... I wonder if there is a `py.test` pattern for mocking the system time.

jacquerie: I'd say that if we are running builds around midnight we probably have bigger problems than a flaky test :D

jacquerie: Mocking system time is actually a little trickier than it looks because most data structures that hold the time are immutable. It can be done, like I did for unit tests, but I'd rather not put too much mocking/logic inside of an integration test. I'm already dissatisfied for having to do `+ today`...

kaplun:
I think it's good to go then

inspirehep/modules/theme/jinja2filters.py
jacquerie: Sure! The gist: we no longer use that code, just the serializers.

jacquerie:
Then :ship:

PR #1116, opened 2016-05-17
WIP global: invenio-collections initial usage

jmartinm:

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -23,7 +23,15 @@ Next, install and build assets:
    (inspirehep)$ inspirehep assets build   # NOTE: Might give an error about missing `minjs`, pip install it to fix.
 
 
-Next, create the database and database tables if you haven't already done so:
+Next, create the ElasticSearch Indices and Aliases (this should be done before database init for collection percolators to work):
+
+.. code-block:: console
+
+   (inspirehep)$ inspirehep index init
+
+**NOTE:** you have to install the analysis-icu plugin for this command to work.
+
+Next, create the database and database tables:
 
 .. code-block:: console
 
@@ -55,16 +63,6 @@ Start honcho
 And the site is now available on `http://localhost:5000`.
 
 **PS: Note the new port!**
-
-
-Create ElasticSearch Indices and Aliases
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-.. code-block:: console
-
-   (inspirehep)$ inspirehep index init
-
-Note, that you have to install the analysis-icu plugin for this command to work.
 
 
 Add demo records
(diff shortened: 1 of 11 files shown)


jmartinm:
Can be merged after https://github.com/inveniosoftware/invenio-collections/pull/53

jmartinm:
I think we can now name our main collection `Literature` instead of `HEP`, that will make it easier for display (and for functions like [this](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/theme/jinja2filters.py#L398) and [this](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/theme/jinja2filters.py#L685)

INSTALL.rst
```diff
@@ -23,7 +23,15 @@ Next, install and build assets:
...
-Next, create the database and database tables if you haven't already done so:
+Next, create the ElasticSearch Indices and Aliases (this should be done before database init for collection percolators to work):
+
+.. code-block:: console
+
+   (inspirehep)$ inspirehep index init
+
+**NOTE:** you have to install the analysis-icu plugin for this command to work.
+
+Next, create the database and database tables:
```
jacquerie: This doc change should be in another commit, IMHO.

jacquerie: Oh, I see, this commit introduces a different ordering of ES creation/DB creation. Then this doc change should stay in this commit.

inspirehep/demosite/data/collection.json
```diff
@@ -0,0 +1,82 @@
+[
```
kaplun: This file is not really demo, but should actually be considered real configuration. Is there a better place for it?

requirements.txt
```diff
@@ -26,6 +26,8 @@
 -e git+https://github.com/inspirehep/invenio-query-parser.git@invenio3-inspire#egg=invenio-query-parser==0.6.0
 -e git+https://github.com/inspirehep/invenio-pidstore.git#egg=invenio-pidstore
 
+-e git+https://github.com/inveniosoftware/invenio-collections.git@master#egg=invenio-collections==1.0.0a3.dev20160204
+
```
kaplun: Why not the official package?

jmartinm: Because the there has not been a release since the last commit (the one that makes `invenio-collections` work with Invenio 3)

inspirehep/demosite/data/collection.json
jmartinm: Not sure, it is a fixture that will have to be loaded once into the DB (and also needs to be loaded into the demo site for developers). Dunno what would be a good place.

kaplun: What about creating a collection directory or a `fixtures` one, in the root. A bit like we currently have `kbs` and `taxonomies`. The thing is that in principle we could remove the `demosite` directory and everything should run smoothly. But this particular file is really needed in order to implement the production configuration.

jmartinm:
Superseded by https://github.com/inspirehep/inspire-next/pull/1589

PR #1098, opened 2016-05-02
WIP search: google like search fields

kaplun:
* Searches on different fields based on the index we are searching on.

Signed-off-by: Panos Paparrigopoulos <panos.paparrigopoulos@cern.ch>
Co-authored-by: Samuele Kaplun <samuele.kaplun@cern.ch>

PR #1121, opened 2016-05-17
search: google like search fields

Panos512:
* Searches on different fields based on the index we are searching on.

Co-authored-by: Samuele Kaplun <samuele.kaplun@cern.ch>

### DIFF
--- a/inspirehep/modules/search/query_factory.py
+++ b/inspirehep/modules/search/query_factory.py
@@ -27,28 +27,61 @@
 from __future__ import absolute_import, print_function
 
 import json
-import pypeg2
-
-from flask import current_app
 
 from elasticsearch_dsl import Q
 
+from flask import current_app
+
 from invenio_query_parser.ast import MalformedQuery
+
+import pypeg2
 
 from .parser import Main
 
-from .walkers.pypeg_to_ast import PypegConverter
-from .walkers.spires_to_invenio import SpiresToInvenio
 from .walkers.elasticsearch import ElasticSearchDSL
 from .walkers.elasticsearch_no_keywords import ElasticSearchNoKeywordsDSL
 from .walkers.elasticsearch_no_keywords import QueryHasKeywords
+from .walkers.pypeg_to_ast import PypegConverter
+from .walkers.spires_to_invenio import SpiresToInvenio
+
+
+walkers = [PypegConverter(), SpiresToInvenio()]
+
+
+def get_fields_by_index(index):
+    if index == "records-hep":
+        return [
+            "title^3",
+            "title.raw^10",
+            "abstract^2",
+            "abstract.raw^4",
+            "author^10",
+            "author.raw^15",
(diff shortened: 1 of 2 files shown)


inspirehep/modules/search/query_factory.py
```diff
@@ -44,10 +45,40 @@
...
+    else:
+        return ['global_fulltext']
+
+
 def inspire_query_factory():
     """Create a parser returning Elastic Search DSL query instance."""
 
-    def invenio_query(pattern):
+    def invenio_query(pattern, index='records-hep'):
         walkers = [PypegConverter(), SpiresToInvenio()]
```
Panos512: @kaplun @jmartinm should I move this on top?

jmartinm:
By the way, in case it is useful, this is the config of current production regarding `global index`

screen shot 2016-05-18 at 1 27 35 pm

So the work there would be to check which fields correspond to the MARC tags indicated in the screenshot (looking at our [dojson config for each collection](https://github.com/inspirehep/inspire-next/tree/master/inspirehep/dojson))

kaplun:
If I am not mistaken, in global we index everything but fulltext files. There is a white list there just not to index erroneous fields.

inspirehep/modules/search/query_factory.py
jmartinm: Sure, just double check that everything works as expected after the move ;)

annetteholtkamp:
Is it really everything? Then it's different behaviour to prod. Is that intended?

- Annette


On May 18, 2016, at 4:38 AM, Samuele Kaplun <notifications@github.com<mailto:notifications@github.com>> wrote:


If I am not mistaken, in global we index everything but fulltext files. There is a white list there just not to index erroneous fields.

kaplun:
@annetteholtkamp what do you mean? Is what different from what? I just say that on prod it seems we basically index everything. But indeed being a white list we might miss some important subfield...

kaplun:
But yest, in Labs we want a different behaviour to prod. In particular we decided in the past to try going for google like results (i.e. default terms in OR) and boost certain fields more.

annetteholtkamp:
E.g. the recid is not in the global index. Neither is the collection in 980.

- Annette

On May 21, 2016, at 11:13 AM, Samuele Kaplun <notifications@github.com<mailto:notifications@github.com>> wrote:


@annetteholtkamp<https://github.com/annetteholtkamp> what do you mean? Is what different from what? I just say that on prod it seems we basically index everything. But indeed being a white list we might miss some important subfield...

kaplun:
@Panos512 can you rebase one last time? I think we can merge this even if not perfect. At least we have a base to iterate on top for specifying fields.

kaplun:
BTW this superseded #1098, correct?

Panos512:
@kaplun just rebased. Yes it superseeds it.

Issue #1124, opened 2016-05-18
workflows UI 'New author' detailed page crash


jmartinm:
After creating a new author (with `/author/create`) and then navigating to the detailed page in `/holdingpen` I am getting the following error:

```
BuildError: Could not build url for endpoint 'invenio_workflows_ui.resolve_action'. Did you mean 'invenio_workflows_ui.static' instead?
```

jmartinm:
@jalavik

PR #1126, opened 2016-05-18
authors: blueprint fix

jmartinm:
* Reorganizes blueprints to make user facing forms work again.


jacquerie:
This test will fail: https://github.com/inspirehep/inspire-next/blob/master/tests/unit/authors/test_authors_views.py

jmartinm:
@jalavik this is an attempt to revive the author update/new forms.

I don't remember when the reorganization was first done but just realised that they were not working any more under `/author/new` and `/author/update`

I have left the view related to the holding pen in its own file and created a new file with the views related to forms.

Can you have a look to see if it makes sense?

Due to https://github.com/inspirehep/inspire-next/issues/1124 I could not check if the whole circle of Create new-review-accept review was working.

jmartinm:
> This test will fail: https://github.com/inspirehep/inspire-next/blob/master/tests/unit/authors/test_authors_views.py

Right. Updated.

jacquerie:
This needs to be done by AB, right? Otherwise it would be a cool thing to test these views (it's 300+ lines of untested code!)

jalavik:
Ah, yes, so @jacenkow was reorganizing the blueprints after the workflows PR got merged as author submission URL for others was aligned with `submit/literature` was moved to `submit/author/new` instead of `author/new` I believe there was some clash with the author profiles page that may also live under `author` prefix?

Anyway, if we wanna keep `author/new` instead of `submit/author/new`, perhaps the holding pen blueprint prefix `submit/author` is not suitable for the one endpoint there, but thats a minor thing.

jmartinm:
I see, yes I like the consistency of `submit/literature` and `submit/author`. I will have another look and update this PR.

jmartinm:
Continued here https://github.com/inspirehep/inspire-next/pull/1131

There were just some wrong blueprint names. I leave the URLs under `/submit/author`

PR #1130, opened 2016-05-19
theme: footer style fix in error pages

jacquerie:
The CSS rule for `.error-pages h4` was leaking in the footer. I fixed it by making the footer CSS more specific and specifying the default values.

Also contains a commit to fix two HTML validity issues I found while investigating this tiny bug.

WDYT about tests that check the validity of HTML pages we generate? Spoiler alert: we fail.


jmartinm:
👍 :shipit:

jacquerie:
:ship:

PR #1135, opened 2016-05-20
Fixes from QA

jalavik:
Assortement of minor fixes that was needed to properly run QA.

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -89,6 +89,11 @@ ACCOUNTS_SESSION_REDIS_URL = "redis://localhost:6379/2"
 # ====
 REST_ENABLE_CORS = True
 
+# Logging
+# =======
+# To enable file logging set it to e.g. "{sys_prefix}/var/log/inspirehep.log"
+LOGGING_FS_LOGFILE = None
+
 # Accounts
 # ========
 RECAPTCHA_PUBLIC_KEY = "CHANGE_ME"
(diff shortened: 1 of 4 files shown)


inspirehep/modules/search/views.py
```diff
@@ -69,6 +69,12 @@ def search():
                            **ctx)
 
 
+@blueprint.route("/ping")
+def ping():
+    """Ping for load balancer."""
+    return "OK", 200
```
jacquerie: Uh? Who is looking for this guy under `/search/ping`? It should be pointing to `/ping`...

jalavik: The url_prefix for the blueprint is `/` so it will be `/ping`.

jacquerie: Whoops, you're right about the `url_prefix`. On the other hand, Isn't this guy already defined? https://github.com/inspirehep/inspire-next/blob/644530717f503b4913f58336d7b5246088fbac24/inspirehep/modules/theme/views.py#L172-L174

Maybe we just need to add an explicit 200 to that?

jalavik: Ah yeah, you are right!

jacquerie:
LGTM, rebase and merge?

requirements.txt
```diff
@@ -22,6 +22,9 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
+# Raven backported fix for new Flask Login on old Sentry
```
lnielsen: FYI https://github.com/zenodo/zenodo/blob/master/zenodo/config.py#L197-L198

lnielsen: + https://github.com/zenodo/zenodo/blob/master/setup.py#L132

jalavik: Thanks @lnielsen!

jacquerie:
:ship:

PR #1141, opened 2016-05-24
theme: detailed record references fix

jmartinm:
* Fixes references endpoint to correctly retrieve references via search call
  instead of mget.

* Changes bundle addition in detailed page template to make sure datatables
  js gets loaded before it is needed.

### DIFF
--- a/inspirehep/utils/references.py
+++ b/inspirehep/utils/references.py
@@ -17,9 +17,11 @@
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 #
 
-from invenio_records.api import Record
-from invenio_search import current_search_client as es
 from jinja2 import render_template_to_string
+
+from invenio_records.api import Record
+
+from inspirehep.utils.search import perform_es_search
 
 
 class Reference(object):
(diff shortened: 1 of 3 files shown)


inspirehep/utils/references.py
```diff
@@ -42,24 +44,21 @@ def references(self):
             refs_to_get_from_es = [
                 ref['recid'] for ref in references if ref.get('recid')
```
mihaibivol: Are you sure this can't be solved by using `replace_refs` ?

mihaibivol: it sohuld be something like
```
references = replace_refs(references, 'es')
# Now we have something like this:
# [{'record': REF_RECORD_FROM_ES,
#     'original_fields_in_references'}]

for reference in references:
    ref_record = reference.get('record')
    if ref_record:
        ref_record = Record(ref_record)  # For compatibility
        recid = ref_record['control_number']
        row.append(.....)
   else:
        row.append(....)

jmartinm: I think it would work. The only concern was, how many requests to ES are done when calling `replace_refs` if there are 100 references? In this case there would be only one request doing the [search](https://github.com/inspirehep/inspire-next/pull/1141/files/83881fb1b490bd3b4dafa6a7afdc4007584e579e#diff-bc37d9f037fdf5e12a93242bc0ee6b91R49)

Also I think I am going to modify this code and call directly `es.search()` so that I can pass the `_source` parameter and only get back the fields I really need (like `title` and `authors`)

jmartinm:
Continued here https://github.com/inspirehep/inspire-next/pull/1142

PR #1146, opened 2016-05-24
WIP dojson: update dojson to 1.2.1

jacquerie:
Preparatory work for #1093, which requires a recent DoJSON version, but is hitting a performance regression in the `authors` rule in `inspirehep/dojson/hep/fields/bd1xx.py`.

This PR fixes that, but discovers three new (unrelated) failures caused by updating DoJSON.

htismpc

### DIFF
--- a/inspirehep/dojson/utils.py
+++ b/inspirehep/dojson/utils.py
@@ -19,13 +19,15 @@
 
 """DoJSON related utilities."""
 
-import six
 import re
+import six
 
 try:
     from flask import current_app
 except ImportError:
     current_app = None
+
+from inspirehep.utils.dedupers import dedupe_list, dedupe_list_of_dicts
 
 
 def legacy_export_as_marc(json, tabsize=4):
@@ -113,35 +115,11 @@ def strip_empty_values(obj):
 
 
 def remove_duplicates_from_list(l):
-    """Remove duplicates from a list preserving the order.
-
-    We might be tempted to use the list(set(l)) idiom,
-    but it doesn't preserve the order, which hinders
-    testability."""
-    result = []
-
-    for el in l:
-        if el not in result:
-            result.append(el)
-
-    return result
+    return dedupe_list(l)
 
 
 def remove_duplicates_from_list_of_dicts(ld):
-    """Remove duplicates from a list of dictionaries preserving the order.
-
-    We can't use the generic list helper because a dictionary isn't
-    hashable. Taken from http://stackoverflow.com/a/9427216/374865."""
-    result = []
-    seen = set()
-
-    for d in ld:
-        t = tuple(d.items())
-        if t not in seen:
-            result.append(d)
-            seen.add(t)
-
-    return result
+    return dedupe_list_of_dicts(ld)
 
 
 def get_record_ref(recid, record_type='record'):
--- a/inspirehep/dojson/hep/fields/bd1xx.py
+++ b/inspirehep/dojson/hep/fields/bd1xx.py
@@ -88,11 +88,10 @@ def authors(self, key, value):
     else:
         for single_value in value:
             authors.append(get_value(single_value))
-    filtered_authors = []
-    for element in authors:
-        if element not in filtered_authors:
-            filtered_authors.append(element)
-    return filtered_authors
+
+    # XXX: duplicates are not stripped here, because this is an expensive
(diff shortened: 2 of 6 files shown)


inspirehep/dojson/utils.py
```diff
@@ -131,15 +131,27 @@ def remove_duplicates_from_list_of_dicts(ld):
...
+    def _freeze(o):
+        """Recursively freezes a dict in an hashable object.
+
+        Adapted from http://stackoverflow.com/a/21614155/374865."""
+        if isinstance(o, dict):
+            return frozenset({k: _freeze(v) for k, v in o.items()}.items())
+        elif isinstance(o, (list, tuple)):
+            return tuple([_freeze(v) for v in o])
+        else:
+            return o
```
jacquerie: @mihaibivol, this is what I use to serialize dicts in order to find duplicates. Did you find a better solution?

kaplun: BTW, I think we should move this as `before_record_insert/update` receiver. This is because we'll need this also when we cut the rope from legacy.

kaplun: Same thing for the stripping of empty fields we are doing in migrator.

mihaibivol: @jacquerie I found this library https://pypi.python.org/pypi/pyrsistent/

It has two methods:
* freeze
* thaw

It does more than this and you can actually use their wrapped types but I think this is a huge overhead :)

inspirehep/dojson/utils.py
```diff
@@ -131,15 +131,27 @@ def remove_duplicates_from_list_of_dicts(ld):
...
     We can't use the generic list helper because a dictionary isn't
-    hashable. Taken from http://stackoverflow.com/a/9427216/374865."""
+    hashable. Adapted from http://stackoverflow.com/a/9427216/374865."""
+
+    def _freeze(o):
+        """Recursively freezes a dict in an hashable object.
+
+        Adapted from http://stackoverflow.com/a/21614155/374865."""
+        if isinstance(o, dict):
+            return frozenset({k: _freeze(v) for k, v in o.items()}.items())
```
mihaibivol: .items() generates a new list why not `frozenset((k, _freeze(v)) for k, v in six.iteritems(o))`

inspirehep/dojson/utils.py
```diff
@@ -131,15 +131,27 @@ def remove_duplicates_from_list_of_dicts(ld):
     """Remove duplicates from a list of dictionaries preserving the order.
 
     We can't use the generic list helper because a dictionary isn't
-    hashable. Taken from http://stackoverflow.com/a/9427216/374865."""
+    hashable. Adapted from http://stackoverflow.com/a/9427216/374865."""
+
+    def _freeze(o):
+        """Recursively freezes a dict in an hashable object.
```
mihaibivol: nitpick: into a hashable object.

jacquerie:
BTW, this version is still too slow for Docker/Travis: https://travis-ci.org/inspirehep/inspire-next/jobs/132689070#L1844

kaplun:
> BTW, this version is still too slow for Docker/Travis: https://travis-ci.org/inspirehep/inspire-next/jobs/132689070#L1844

It complains about no output in the last 10 minutes. What are we doing in those 10 minutes? Loading the records? Shall we print at least this fact?

jacquerie:
Yes, it's stuck loading the records. I don't think we should work around that by outputting stuff...

inspirehep/dojson/utils.py
kaplun: BTW Just confirming that validation is happening after signals `before_record_insert/update` are sent. So we can call your `remove_duplicates_from_list_of_dicts()` there, and remove all duplicate stripping from dojson.

Additionally I would move there also the calls to `strip_empty_values()` in https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/migrator/tasks.py#L368-L382

jacquerie:
@kaplun green. Now you can build your commit on top, and I will cherry-pick and merge.

@Panos512 this means that in your PR, if this is merged before, you can drop the requirement that DoJSON is our forked version, and can just be 1.2.1

Panos512:
Great news! I set it to 1.2.1

kaplun:
Patch coming...

PR #1147, opened 2016-05-25
workflows: beard API integration

jalavik:
CC @eamonnmag @jacenkow


jacenkow:
@jalavik should we first merge `beard-server`?

jalavik:
@jacenkow Yeah, that would be ideal!

PR #1148, opened 2016-05-25
WIP workflows: Magpie API integration

jalavik:
* Adds Magpie workflow tasks for guessing of categories, keywords and
  experiments.

* Unifies code across Beard and Magpie tasks.

@jstypka @eamonnmag


inspirehep/modules/workflows/tasks/magpie.py
```diff
@@ -0,0 +1,123 @@
...
+    )
+    return payload
+
+
+def filter_magpie_response(labels, limit):
+    """Filter response from Magpie API, keeping most relevant labels."""
+    filtered_labels = []
+    for word, score in labels:
+        if score >= limit:
+            filtered_labels.append((word, score))
```
jstypka: how about `filtered_labels = [(word, score) for word, score in labels if score >= limit]

jstypka:
👍

inspirehep/modules/workflows/tasks/magpie.py
jalavik: Yes, we can make this a one line comprehension as it should be readable enough. Thanks!

inspirehep/modules/workflows/tasks/magpie.py
```diff
@@ -0,0 +1,123 @@
...
+    """Workflow task to ask Magpie API for a subject area assessment."""
+    magpie_url = get_magpie_url()
+    if not magpie_url:
+        # Skip task if no API URL set
+        return
+    payload = prepare_magpie_payload(obj.data, corpus="experiments")
+    results = json_api_request(magpie_url, payload)
+    if results:
+        labels = results.get('labels', [])
+        experiments = filter_magpie_response(labels, limit=0.5)
```
jalavik: @eamonnmag @jstypka I took this number out of the air so may have to be adjusted.

jstypka: yeah, I wanted to ask about that. I think it's fair, the confidence values for experiments are usually very high, because they are often explicitly mentioned in the abstract. You might want to extract it as a configuration constant or something, but not sure if it's necessary.

jstypka:
do you predict experiments for all abstracts that come in or only for the ones that belong to the `experiment-HEP` subject category?

eamonnmag:
@jstypka only the ones in the experiments-validation file you provided :)

inspirehep/modules/workflows/workflows/hep_ingestion.py
```diff
@@ -125,6 +130,11 @@ class mysource(HEPIngestion):
             with_author_keywords=True,
         ),
         filter_core_keywords,
+        guess_categories,
+        IF(is_experimental_paper, [
```
jalavik: @jstypka You have a great point. I added a check for this in the workflow.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -111,3 +113,10 @@ def is_record_relevant(obj, *args, **kwargs):
                 len(core_keywords) == 0:
             return False
     return True
+
+
+def is_experimental_paper(obj, eng):
+    """Is the record an experimental paper."""
+    categories = get_value(obj.data, "arxiv_eprints.categories", []) + \
+        get_value(obj.data, "subject_terms.term", [])
+    return "hep-ex" in categories or "Experiment-HEP" in categories
```
kaplun: Is this the correct criteria? 
E.g. 
https://inspirehep.net/search?ln=en&ln=en&p=collaboration%3Aatlas+-subject%3Ahep-ex+&of=hb&action_search=Search&sf=earliestdate&so=d&rm=&rg=25&sc=0

shows that there 3K papers from ATLAS that are not `hep-ex`.

@annetteholtkamp what do you think? For which INSPIRE/arXiv categories shall we guess the experiment? Shall we be consverative and just opt for the safe bet of `hep-ex`/`Experiment-HEP`?

inspirehep/modules/workflows/utils.py
```diff
@@ -22,12 +22,34 @@
...
+    current_app.logger.debug("POST {0} with \n{1}".format(
+        url, json.dumps(data, indent=4)
+    ))
+    response = requests.post(
+        url=url,
+        headers=final_headers,
+        data=json.dumps(data)
+    )
+    if response.status_code == 200:
+        return response.json()
```
kaplun: Can you make a configurable timeout? In this way we can fine tune, in case `beard` etc. are slower than we thought.

inspirehep/modules/workflows/tasks/beard.py
```diff
@@ -69,27 +53,28 @@ def guess_coreness(obj, eng):
...
         scores = results.get('scores') or []
         max_score = max(scores)
         decision = results.get('decision')
-        relevance_score = max_score
         scores = {
             "CORE": scores[0],
             "Non-CORE": scores[1],
             "Rejected": scores[2],
         }
+        # Generate a normalized relevance_score useful for sorting
```
kaplun: Can you further explain how this computation work?

inspirehep/modules/workflows/workflows/hep_ingestion.py
```diff
@@ -125,6 +130,11 @@ class mysource(HEPIngestion):
...
+            guess_experiments,
+        ]),
+        guess_keywords,
```
kaplun: @jalavik I see (as it should indeed be) that arXiv harvesting and hep ingestion workflows are quite similar. Is there a way to factor out the common part, so that enhancement such as the ones in this PR can be written only once?

inspirehep/modules/workflows/tasks/actions.py
jalavik: Yeah, indeed there are probably better criterias. We could look in the collaboration field as well as your inspire search indicate - however do we need to guess the experiment at all if collaboration found to be ATLAS? To rephrase, could we rely on assigning experiments via the collaboration field when available as well?

inspirehep/modules/workflows/utils.py
jalavik: Sure, I will add some timeout handling here.

inspirehep/modules/workflows/tasks/actions.py
kaplun: Good point! The 2 fields are definitively correlated. I wonder which one comes from arXiv?

inspirehep/modules/workflows/workflows/hep_ingestion.py
jalavik: The arXiv harvesting is currently inheriting from the hep ingestion workflow to reuse some common parts. Now the `before_halt_check` part is indeed becoming longer and parts of it is equal between them. We could split `before_halt_check` up to more parts to just add the arXiv parts before/after.

But looking at it again now I am honestly not sure that is the right direction to share the workflow across the two.

In terms of readability and helping understanding what is going on in the different workflows, you see it's already today quite complex to understand what is happening and one must go back and forth between those two files to see the whole picture. The best way today is actually is to open a shell and import the workflow and echo the workflow attribute.

So, I am wondering now if it would just be better to just be more direct in the workflow definitions and not inherit parts at all. Similar to how literature vs. arxiv workflow is today. Or, that we have one workflow with multiple branches (if arxiv etc.). WDYT?

kaplun: I really like your latter proposal. IMHO we should have one workflow per document type. The fact that something comes from arXiv Vs. publisher Vs. user is of smaller importance given all the enhancements, checks, and involvements of cataloguers, the end product being always the same, i.e. an article.

So somehow I see these workflows as part of input branches of a tree going then to one output trunk.

On the other hand this might force the workflow to be unnecessary even more complex. Would be nice to once pict the overall steps involved...

jalavik: Yeah. You make some good points. I will look into this shortly on organizing these workflows altogether. For this PR however, we can leave it I think.

inspirehep/modules/workflows/tasks/actions.py
annetteholtkamp: Subject categories where experiment guessing makes sense are:
Experiment-HEP
Experiment-Nucl
Astrophysics
Instrumentation
And probably the primary arXiv category should be one of these.

Collaboration is a good indicator. But since arXiv does not have a collaboration field it is somehow extracted from the author field and doesn't come in standardised form. If the experiment guesser works as good as we believe it's maybe an unnecessary complication to look at the collaboration?
Btw, Kirsten runs a job (nut sure whether it's in bibcheck) to standardise the collaboration field a bit. Maybe that should we integrated into the ingestion workflow?

- Annette


On May 25, 2016, at 2:35 PM, Samuele Kaplun <notifications@github.com<mailto:notifications@github.com>> wrote:


In inspirehep/modules/workflows/tasks/actions.py<https://github.com/inspirehep/inspire-next/pull/1148#discussion_r64564001>:

> @@ -111,3 +113,10 @@ def is_record_relevant(obj, *args, **kwargs):
>                  len(core_keywords) == 0:
>              return False
>      return True
> +
> +
> +def is_experimental_paper(obj, eng):
> +    """Is the record an experimental paper."""
> +    categories = get_value(obj.data, "arxiv_eprints.categories", []) + \
> +        get_value(obj.data, "subject_terms.term", [])
> +    return "hep-ex" in categories or "Experiment-HEP" in categories


Is this the correct criteria?
E.g.
https://inspirehep.net/search?ln=en&ln=en&p=collaboration%3Aatlas+-subject%3Ahep-ex+&of=hb&action_search=Search&sf=earliestdate&so=d&rm=&rg=25&sc=0

shows that there 3K papers from ATLAS that are not hep-ex.

@annetteholtkamp<https://github.com/annetteholtkamp> what do you think? For which INSPIRE/arXiv categories shall we guess the experiment? Shall we be consverative and just opt for the safe bet of hep-ex/Experiment-HEP?

PR #1151, opened 2016-05-26
WIP: Holding pen UI changes

eamonnmag:
- [ ] Dashboard page
  - [x] Block view
  - [ ] Timeline
- [ ] List page
  - [x] Authors
  - [x] Literature
- [ ] Detail page
  - [x] Authors
  - [x] Literature
- [x] Action buttons
- [ ] Tests
- [ ] Fully working search


inspirehep/modules/theme/jinja2filters.py
```diff
@@ -684,21 +684,19 @@ def format_date(datetext):
 @blueprint.app_template_filter()
 def find_collection_from_url(url):
     """Returns the collection based on the URL."""
-    if 'conferences' in url:
+    if 'cc=conferences' in url or '/conferences' in url:
```
jacquerie: Aww, we have to say goodbye to the coolest bug/feature ever.

inspirehep/modules/theme/static/scss/holding-pen/holding-pen.scss
```diff
@@ -0,0 +1,460 @@
...
+      }
+    }
+
+    .authorupdate {
+      color: $belize-hole !important;
+      &.active {
+        background-color: $belize-hole;
+        color: white !important;
+      }
+    }
```
jacquerie: Maybe we can refactor these with a mixin:
```sass
@mixin facet-colors($color) {
  color: $color !important;
  &.active {
    background-color: $color;
    color: white !important;
  }
}

.arxiv {
  @include facet-colors($alizarin);
}

[...]

.authorupdate {
  @include facet-colors($belize-hole);
}
```

eamonnmag: Awesome!

inspirehep/modules/theme/jinja2filters.py
kaplun: Would be great if this was actually done by properly parsing the URL query and path. In this way we would avoid unexpected behaviors.

jacquerie: I think that a better solution uses something like this: http://stackoverflow.com/a/25013141/374865, rather than parsing the URL.

eamonnmag: This current code will cover all use cases from now until we add a new collection. It’s not the most elegant, but sometimes it doesn’t have to be.
All we’d be doing is coming up with a more convoluted way of writing what we have here.

kaplun: It has though some troubles:

If I just have 'jobs' in the URL it will think it's about Jobs. If I have 'cc=conferencesfoo'  it will think it's about conferences.

inspirehep/utils/stats.py
```diff
@@ -0,0 +1,68 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+import operator
+
+
+class Stats(object):
```
jmartinm: Any advantage of a class vs just functions?

inspirehep/utils/stats.py
```diff
@@ -0,0 +1,68 @@
...
+        returns the hindex.
+        :param args:
+        :param kwargs:
+        :return: h index for a given list of citations.
+        """
+        _record_citations = kwargs.pop('citations')
+        _sorted = Stats.get_sorted_citations(_record_citations)
+
+        h_index = 0
+        print _sorted
```
jmartinm: leftover print

inspirehep/utils/stats.py
```diff
@@ -0,0 +1,68 @@
...
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+import operator
+
+
+class Stats(object):
+
+    """Stats is a class to calculate common metrics"""
+
+    @staticmethod
+    def calculate_hindex(*args, **kwargs):
```
jmartinm: Does this function need much future flexibility? (to have `*args, **kwargs`) or we just specify the parameters in the signature.

inspirehep/utils/stats.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        Given a citation dictionary, in the format {recid: citation_count}
+        returns the hindex.
+        :param args:
+        :param kwargs:
+        :return: h index for a given list of citations.
+        """
+        _record_citations = kwargs.pop('citations')
```
jmartinm: Note that if the parameter `citations` is not mandatory this line will give `KeyError`. So you can either add a default value to the `pop` function or make `citations` mandatory by adding it to the `calculate_hindex` parameters.

eamonnmag:
cc @kaplun holding-pen-status

jacquerie:
Holy Batman atomic commits!

kaplun:
@jacquerie  I guess @eamonnmag will squash all of them into a nice:
```
holdingpen: AngularJS porting

Signed-off-by....
```
:smile_cat:

eamonnmag:
Something messed up in Pycharm. Or my brain. Or both.

eamonnmag:
cc @jalavik - pretty cool! holding-pen-updates

jalavik:
Starting to look really good @eamonnmag 👏 

I would favor squash in fact to not have all the nitty gritty steps of the development, unless the commits have a working state for each step that is not refactored in the next commit.

Another minor thing is that the commit log message subject could be set to `workflows:` instead of `holding_pen` as the changes occur primarily under the `workflows` module.

Finally, there seem to be some merge issues here as commits already in master branch is added here. Perhaps just missing rebase.

inspirehep/config.py
```diff
@@ -827,8 +828,10 @@ def _(x):
                              ':json_serializer'),
     },
     search_serializers={
-        'application/json': ('invenio_workflows_ui.serializers'
+        'application/vnd+inspire.brief+json': ('invenio_workflows_ui.serializers'
```
jalavik: Eventually I guess we will directly use the invenio-search angular module instead of the inspire-search and this is not needed?

inspirehep/modules/theme/static/js/build.js
```diff
@@ -19,7 +19,7 @@
 
 ({
   preserveLicenseComments: false,
-  optimize: 'uglify2',
+  optimize: 'none',
```
jalavik: debug leftover?

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -598,8 +598,7 @@ def publication_info(record):
             if 'conference_record' in pub_info:
                 conference_rec = replace_refs(pub_info['conference_record'],
                                               'es')
-                if conference_rec and conference_rec.get('control_number'):
-                    conference_recid = conference_rec['control_number']
+                conference_recid = conference_rec['control_number']
```
jmartinm: watch out, seems like the line got deleted by merging. This is causing QA search page to break.

jalavik:
Seperate branch containing this work is now merged. See #1214

PR #1152, opened 2016-05-26
utils: search helper improvement

jmartinm:
* Allows filtering of fields in ElasticSearch response.


kaplun:
Can you rebase it? I have fixed all issues that were breaking master.

PR #1153, opened 2016-05-26
dojson: collaboration fix

jmartinm:
* Fixes data format for collaboration field to match ES mapping.

### DIFF
--- a/inspirehep/dojson/hep/fields/bd70x75x.py
+++ b/inspirehep/dojson/hep/fields/bd70x75x.py
@@ -87,5 +87,5 @@ def collaboration(self, key, value):
 def collaboration2marc(self, key, value):
     """Added Entry-Corporate Name."""
     return {
-        'g': value,
+        'g': value.get('value'),
     }
--- a/inspirehep/modules/deposit/dojson/fields/literature.py
+++ b/inspirehep/modules/deposit/dojson/fields/literature.py
@@ -102,7 +102,7 @@ def categories(self, key, value):
 
 @literature.over('collaboration', '^collaboration$')
 def collaboration(self, key, value):
-    return value
+    return [{"value": value}]
 
 
 @literature.over('hidden_notes', '^hidden_note$')


jalavik:
Looks good. I assume from the form there is only ever one collaboration.

jmartinm:
> I assume from the form there is only ever one collaboration.

Yes, thats correct.

Thanks.

annetteholtkamp:
Are you talking about the current 710__g field? That is a multiple field.

- Annette


On May 26, 2016, at 2:54 PM, Javier Martin Montull <notifications@github.com<mailto:notifications@github.com>> wrote:


I assume from the form there is only ever one collaboration.

Yes, thats correct.

Thanks.

jmartinm:
I think `710__g` being multiple when importing records from marcxml and when exporting them is handled correctly. 

If I understood @jalavik correctly, he meant that from the form we only get one collaboration, which is the case.

annetteholtkamp:
So why don’t we allow multiple collaborations also in the form? It’s a rare case. But what are people supposed to do in case of several coll’s?

- Annette

jmartinm:
> So why don’t we allow multiple collaborations also in the form? It’s a rare case. But what are people supposed to do in case of several coll’s?

Like in some other cases, it is just the simplifying the form VS supporting the 1% use cases.

If there was a repeated case when somebody wanted to add multiple collaborations and they could not, we should add it.

eamonnmag:
There are quite a few older ones e.g. from the CERN-Soviet and French-Soviet collaborations (https://hepdata.net/search/?collaboration=French-Soviet&page=1), but probably < 100 in total. Many of the new papers have one collaboration, except for one paper in HEPData which has CMS and TOTEM experiments https://hepdata.net/record/ins1294140.

Some more:
1294140
1282028
1377206
513337
247579

kaplun:
Actually: there are 4K records on production with more than one collaboration, but that's out of 1.3M so 0.3% :)

kaplun:
Since being able to handle all these small use cases, shall we have a free note field (maybe it's already there), that would trigger the attention of the cataloguer? E.g. *"You wanted to provide more information but there wasn't a field for it, just tell us here:"*.

jmartinm:
> Since being able to handle all these small use cases, shall we have a free note field (maybe it's already there), that would trigger the attention of the cataloguer? E.g. "You wanted to provide more information but there wasn't a field for it, just tell us here:".

yes, there is at the moment a field for extra comments and we have gotten in the past some feedback when a field was missing or the user could not fill correctly a field for some reason. Don't think there was so far a comment about multiple collaborations though.

PR #1160, opened 2016-05-30
migrator: bulk indexing and inspire_prod_records usage

jmartinm:
* Catches validation errors when creating records and inserts them in
  the table inspire_prod_records with valid=False.

* Uses invenio-indexer bulk indexing functionality.


jmartinm:
Regression tests should pass when https://github.com/inveniosoftware/invenio-indexer/pull/25 is merged and the timeout raised.

inspirehep/modules/migrator/tasks.py
```diff
@@ -122,7 +127,7 @@ def migrate_broken_records(broken_output=None, dry_run=False):
 @shared_task(ignore_result=True)
 def migrate(source, broken_output=None, dry_run=False, wait_for_results=False):
     """Main migration function."""
-    from celery.task.sets import TaskSet
+    from celery import group
```
jacquerie: Does this import have to happen inside of the task?

inspirehep/modules/migrator/tasks.py
```diff
@@ -199,96 +206,48 @@ def continuous_migration():
...
-#         db.session.commit()
-#         logger.info("Sending chunk to elasticsearch")
-#         es_bulk(es, records_to_index, request_timeout=60)
-#     finally:
-#         # models_committed.connect(record_modification)
-#         # after_record_insert.connect(catch_citations_insert)
-#         # before_record_index.connect(add_citation_count_on_insert_or_update)
-#         # before_record_index.connect(catch_citations_update)
-#         db.session.close()
-
```
jacquerie: :+1:

inspirehep/modules/migrator/tasks.py
```diff
@@ -378,8 +337,11 @@ def create_record(record, force=True, dry_run=False):
     else:
         json = hep.do(record)
 
-    if dry_run:
-        return errors, json
+    if '$schema' in json:
+        json['$schema'] = url_for(
+            'invenio_jsonschemas.get_schema',
+            schema_path="records/{0}".format(json['$schema'])
+        )
```
jacquerie: @mihaibivol, where's the other place where you were doing something similar? Is this solution better?

inspirehep/modules/migrator/tasks.py
jmartinm: Not really, can be moved out.

inspirehep/modules/migrator/tasks.py
jmartinm: This code was actually just moved from https://github.com/inspirehep/inspire-next/pull/1160/files#diff-2b6424df440e3e9cba58ea4a4cab141aL272 as I thought it was better to have here (so that `create_record` does create the full record)

jmartinm:
@mihaibivol 😉 

I guess this will have to do. It reduces integration test a bit more than 30sec so it's something.

PR #1161, opened 2016-05-30
authors: add phonetic blocks generator

jacenkow:
* Adds phonetic block to each signature on
  before_record_insert and before_record_update signals.

### DIFF
--- a/inspirehep/modules/authors/receivers.py
+++ b/inspirehep/modules/authors/receivers.py
@@ -22,14 +22,78 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
+from __future__ import absolute_import, division, print_function
+
+import json
 import uuid
+
+from flask import current_app
+import requests
 
 from invenio_indexer.signals import before_record_index
 from invenio_records.signals import (
     before_record_insert,
-    before_record_update)
+    before_record_update,
+)
 
 from .utils import author_tokenize
+
+
+def _query_beard_api(full_names):
+    """Query Beard API.
+
+    This method allows for computation of phonetic blocks
+    from a given list of strings.
+    """
+    headers = {
+        "Content-Type": "application/json",
+        "Accept": "application/json"
+    }
+
+    text_endpoint = "{base_url}/text/phonetic_blocks".format(
+        base_url=current_app.config.get('BEARD_API_URL')
+    )
+
+    response = requests.post(
+        url=text_endpoint,
+        headers=headers,
+        data=json.dumps({'full_names': full_names})
+    )
+
+    return response.json()['phonetic_blocks']
+
+
+@before_record_insert.connect
+@before_record_update.connect
+def assign_phonetic_block(sender, *args, **kwargs):
+    """Assign phonetic block to each signature.
+
+    The method extends the given signature with a phonetic
+    notation of the author's full name, based on
+    nysiis algorithm. The phonetic block is assigned before
+    the signature is indexed by an Elasticsearch instance.
+    """
+    if current_app.config.get('BEARD_API_URL'):
+        authors = sender.get('authors', [])
+        authors_map = {}
+
+        for index, author in enumerate(authors):
+            if 'full_name' in author:
+                authors_map[author['full_name']] = index
+
+        # Call Beard API to generate phonetic blocks.
+        signatures_blocks = _query_beard_api(authors_map.keys())
+
+        # Add signature block to an author.
+        for full_name, signature_block in signatures_blocks.iteritems():
+            authors[authors_map[full_name]].update(
+                {"signature_block": signature_block})
+
+        # # For missing phonetic blocks (not valid full names) add None.
+        for full_name in list(
+                set(authors_map.keys()) - set(signatures_blocks.keys())):
+            authors[authors_map[full_name]].update(
+                {"signature_block": None})
 
 
 @before_record_insert.connect
--- a/tests/unit/authors/test_authors_receivers.py
+++ b/tests/unit/authors/test_authors_receivers.py
@@ -22,6 +22,11 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
+from __future__ import absolute_import, division, print_function
+
+import mock
+import httpretty
+
 from inspirehep.modules.authors import receivers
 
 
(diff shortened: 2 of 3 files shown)


jacenkow:
WIP since no tests and to be honest there is no good error handling.

inspirehep/modules/authors/receivers.py
```diff
@@ -32,6 +36,60 @@
...
+
+    The method extends the given signature with a phonetic
+    notation of the author's full name, based on
+    nysiis algorithm. The phonetic block is assigned before
+    the signature is indexed by an Elasticsearch instance.
+    """
+    if current_app.config.get('BEARD_API_URL'):
+        authors = sender.get('authors', [])
+        authors_map = {}
+        full_names = []
```
jacquerie: This will just contain the `.keys()` of `authors_map`, so it's not really needed.

jacenkow: 👍

inspirehep/modules/authors/receivers.py
```diff
@@ -32,6 +36,60 @@
...
+                full_names.append(full_name)
+                authors_map[full_name] = index
+
+        # Add signature block to an author.
+        signatures_blocks = _query_beard_api(full_names)
+
+        for full_name, signature_block in signatures_blocks.iteritems():
+            authors[authors_map[full_name]].update(
+                {"signature_block": (
+                    lambda x: x if x else None)(signature_block)})
```
jacquerie: Python supports ternary expressions without needing lambdas: `signature_block if signature_block else None`.

jacquerie:
Sigh. Add a commit to increase [this timeout](https://github.com/inspirehep/inspire-next/blob/ac5e520d27a2212173b35657c859f92bfff240df/tests/integration/conftest.py#L65) to 80 : (

jacenkow:
@jacquerie `Elapsed time 27 min 37 sec` 😃

jacquerie:
Refresh the page! https://travis-ci.org/inspirehep/inspire-next/builds/134148655 shows "Elapsed time 17 min 50 sec"

inspirehep/modules/authors/receivers.py
```diff
@@ -32,6 +36,63 @@
...
+@before_record_update.connect
+def assign_phonetic_block(sender, *args, **kwargs):
+    """Assign phonetic block to each signature.
+
+    The method extends the given signature with a phonetic
+    notation of the author's full name, based on
+    nysiis algorithm. The phonetic block is assigned before
+    the signature is indexed by an Elasticsearch instance.
+    """
+    if current_app.config.get('BEARD_API_URL'):
```
kaplun: Since this is None by default, am I correct in assuming this part is not executed ATM on Travis?

inspirehep/modules/authors/receivers.py
```diff
@@ -32,6 +36,63 @@
...
+                authors_map[author['full_name']] = index
+
+        # Call Beard API to generate phonetic blocks.
+        signatures_blocks = _query_beard_api(authors_map.keys())
+
+        # Add signature block to an author.
+        for full_name, signature_block in signatures_blocks.iteritems():
+            authors[authors_map[full_name]].update(
+                {"signature_block": signature_block})
+
```
kaplun: I enjoyed trying to understand this mind-twisting implementation.

inspirehep/modules/authors/receivers.py
jacquerie: You can inject the proper configuration at test-time, so this should not be an issue.

inspirehep/modules/authors/receivers.py
jacquerie: "Everyone knows that debugging is twice as hard as writing a program in the first place. So if you're as clever as you can be when you write it, how will you ever debug it?" — Brian Kernighan

tests/unit/authors/test_authors_receivers.py
```diff
@@ -63,6 +77,37 @@ def test_name_variations():
...
+        }, {
+            "full_name": "**BROKEN**",
+        }, {
+            "full_name": "John Richard Ellis"
+        }]
+    }
+
+    # In order to run the method, BEARD_API_URL value must be set.
+    temp_config = current_app.config.get('BEARD_API_URL')
+    current_app.config['BEARD_API_URL'] = True
```
jacenkow: @kaplun this is how I injected 🎉

jacenkow:
@jacquerie can you check my tests?

tests/unit/authors/test_authors_receivers.py
jalavik: `True` is technically not the right value as it should really be an URL string, but for your case it works I guess.

Btw, this is a neat trick to avoid reverting manually: https://github.com/inspirehep/inspire-next/blob/master/tests/integration/workflows/test_arxiv_workflow.py#L252-L259

jacenkow: @jalavik looks much better 👍

tests/unit/authors/test_authors_receivers.py
```diff
@@ -63,6 +73,93 @@ def test_name_variations():
             'Richard Ellis'])
 
 
+@httpretty.activate
+def test_phonetic_block_generation_ascii():
+    from flask import current_app
```
jacquerie: Uh? What's wrong with the `app` fixture?

jacenkow: Seems like the `app` argument was provided by `mock` decorator. After removing `mock`, the method was not receiving any argument.

jacenkow:
@jacquerie in tests I left commented `# unicode_literals` import on purpose, importing it fixes my problem.

tests/unit/authors/test_authors_receivers.py
jacquerie: Yeah, you should tell the test to use the `app` fixture in `tests/unit/conftest.py` by adding `app` as a parameter to every test.

jacenkow:
@jacquerie :shipit:

kaplun:
@jacenkow please rebase it first, seems outdated...

jacenkow:
@kaplun I would like to apologise for being sloppy 😞

tests/unit/conftest.py
```diff
@@ -242,3 +243,10 @@ def dummy_empty_response():
             "hits": []
         }
     }
+
+
+@pytest.yield_fixture
+def httppretty_mock():
+    httpretty.enable()
+    yield
+    httpretty.disable()
```
jacquerie: The only nitpick I could really find is that this fixture could have used a comment explaining that the decorator doesn't work with pytest, but who cares.

jacquerie:
LGTM :ship:

PR #1162, opened 2016-05-30
migrator: add citations with progress bar

kaplun:


kaplun:
@jmartinm next time we know how much it takes :)

eamonnmag:
How can it know how long it takes? Is this going to be like installing
windows where it says it will take 1 hour but takes 2? :)

PR #1157, opened 2016-05-27
authors: add uuid generator

jacenkow:
* Adds uuid to each signature on before_record_insert signal.

* Adds test to check if uuid field is not empty.

Signed-off-by: Grzegorz Jacenków <grzegorz.jacenkow@cern.ch>

PR #1155, opened 2016-05-27
workflows: custom search factory

jalavik:
This is the custom searcher for the new angular based Holding Pen list view filtering out some of the JSON for performance reasons.

PR #1164, opened 2016-05-31
tests: increase again timeout of bulk ES query

jacquerie:
PR #1157 increased the work we do on ES record update, and, as a consequence, we sometimes hit the timeout during the ES bulk query that updates citations: https://travis-ci.org/inspirehep/inspire-next/jobs/133985931#L4780

This PR increases the timeout, but if this becomes a problem again we will have to rethink that bulk query.


jacquerie:
Already included in #1155.

PR #1165, opened 2016-05-31
WIP Mathoid

kaplun:

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -85,6 +85,12 @@ CACHE_REDIS_URL = os.environ.get(
 CACHE_TYPE = "redis"
 ACCOUNTS_SESSION_REDIS_URL = "redis://localhost:6379/2"
 
+# Mathoid
+# =======
+MATHOID_SERVER = os.environ.get(
+    "MATHOID_SERVER",
+    "http://localhost:10044")
+
 # REST
 # ====
 REST_ENABLE_CORS = True
--- a/inspirehep/utils/formulas.py
+++ b/inspirehep/utils/formulas.py
@@ -1,0 +1,187 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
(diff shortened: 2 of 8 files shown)


inspirehep/utils/formulas.py
```diff
@@ -0,0 +1,187 @@
...
+        elif c == ')':
+            count -= 1
+        if count < 0:
+            return False
+    return count == 0
+
+
+def tex2mml(formula):
+    """Converts a LaTeX formula to MathML."""
+    mathjax_node_server = current_app.config['MATHOID_SERVER']
```
jalavik: Here we should check if it is not `None` perhaps.

inspirehep/config.py
```diff
@@ -85,6 +85,12 @@ def _(x):
 CACHE_TYPE = "redis"
 ACCOUNTS_SESSION_REDIS_URL = "redis://localhost:6379/2"
 
+# Mathoid
+# =======
+MATHOID_SERVER = os.environ.get(
```
jalavik: Shall we set this to `None` as default and have a check in the receiver for people who are not running mathoid locally?

kaplun: :+1: That very precise change was sitting in my laptop, not yet pushed to :-)

In particular since I am not going to fiddle with npm and honcho I know many people will not have mathoid running...

Additionally I think I will separate mathoid testing from the integration ones.

jacquerie:
Abandoned? Also doing an extra synchronous HTTP request before indexing slows the migration way too much. I don't think Mathoid is a good solution for this problem.

kaplun:
Nope not abandoned. Simply postponed due to different priority. 

By parsing formulas we can make the difference in delivering our service. I firmly believe in this. And as far as I know Mathoid is out of the box the best tool in town. What we can do to improve the performance is by caching. Afterall the number of involved formulas is enumerable in our domain.

kaplun:
But yes, we could extract formulas in background and not at migration time.

jmartinm:
ping @kaplun , shall we close this PR?

kaplun:
But then I would loose it :) Can we have a postpone feature?

jacquerie:

kaplun:
But what about enjoying weekends instead of closing others PRs? :stuck_out_tongue_closed_eyes:

jacquerie:
That's how you enjoy weekends!

PR #1166, opened 2016-05-31
authors: ORCID duplicate validation

jmartinm:
* Checks for ORCID duplicates in order to not allow new author information
  to be submitted if an ORCID already exists in inspirehep.net.

### DIFF
--- a/inspirehep/modules/authors/templates/authors/forms/update_form.html
+++ b/inspirehep/modules/authors/templates/authors/forms/update_form.html
@@ -18,6 +18,7 @@
 #}
 {%- extends "forms/form.html" -%}
 {% set title = "New author information - INSPIRE Labs" if new else "Update author information - INSPIRE Labs" %}
+{% set is_update = False if new else True %}
 
 {% block global_bundles %}
   {{ super() }}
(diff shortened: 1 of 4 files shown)


jmartinm:
This needs to be ported to `master` branch.

inspirehep/modules/authors/templates/authors/forms/update_form.html
```diff
@@ -52,9 +53,9 @@
 
     var config = {
       form: {
-        save_url: '{{ url_for("inspire_authors.validate") }}',
-        save_all_url: '{{ url_for("inspire_authors.validate", all='1') }}',
-        complete_url: '{{ url_for("inspire_authors.validate", submit='1') }}',
+        save_url: '{{ url_for("inspire_authors.validate", is_update=is_update) }}',
+        save_all_url: '{{ url_for("inspire_authors.validate", all='1', is_update=is_update) }}',
+        complete_url: '{{ url_for("inspire_authors.validate", submit='1', is_update=is_update) }}',
```
kaplun: Does that means that the `is_update` flag can in principle be tweaked client side? Can't you always know if it's an update just by looking server-side?

inspirehep/modules/authors/views.py
```diff
@@ -146,9 +147,11 @@ def validate():
     """Validate form and return validation errors."""
     if request.method != 'POST':
         abort(400)
+
+    is_update = True if request.args.get('is_update') == 'True' else False
```
kaplun: IMHO, this should be read from the workflow, and not received by the client.

jmartinm: It's a bit tricky because until the form is submitted no workflow exists.

The only thing that you can do by modifying client side is to skip the validation. I think if I add validation on the server side on form submission then it would not be a big deal?

kaplun: Ah OK, understood. :+1:

PR #1168, opened 2016-05-31
fixtures: initial module

jalavik:
* Adds new fixtures module used when setting up the INSPIRE instance
  to pre-load database tables with content and setting up folders etc.

* NOTE There is a new CLI command you need to run to setup the site properly:
  `inspirehep fixtures init`

E.g. we could potentially later hook up the collection loading here also from #1116 

@kaplun @jmartinm

### DIFF
--- a/inspirehep/modules/fixtures/files.py
+++ b/inspirehep/modules/fixtures/files.py
@@ -1,0 +1,77 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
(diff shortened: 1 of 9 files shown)


scripts/recreate_records
```diff
@@ -63,7 +63,7 @@ inspirehep db drop --yes-i-know
 inspirehep db create
 inspirehep index destroy --force --yes-i-know
 inspirehep index init
+inspirehep fixtures init
```
jalavik: @jacquerie We would need to run this extra CLI command during CI testing as well. Would importing and calling it in conftests be the way?

inspirehep/modules/fixtures/files.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def init_workflows_storage_path(default=False):
+    """Init workflows file store location."""
+    try:
+        uri = current_app.config['WORKFLOWS_FILE_LOCATION']
+        if uri.startswith('/') and not os.path.exists(uri):
+            os.makedirs(uri)
+        loc = Location(
+            name=current_app.config["WORKFLOWS_DEFAULT_FILE_LOCATION"],
```
jmartinm: Is `WORKFLOWS_DEFAULT_FILE_LOCATION ` a location or a name?

jmartinm:
👍 Sounds good.

scripts/recreate_records
jacquerie: Well, not the actual CLI command, but the method it then invokes.

kaplun:
How will we go when the fixture changes, in order to apply this to production? E.g. say we want to create a new hidden collection for a new set of notes.

inspirehep/modules/fixtures/files.py
jalavik: Its both kinda, hehe, but I will rename it to `WORKFLOWS_DEFAULT_FILE_LOCATION_NAME`

jalavik:
So we can have an update command perhaps eventually that can resync files or be given a file to sync in case of collection information. Or for collections we could have an admin interface or cli to add/remove collections and dump configuration to a file.

PR #1174, opened 2016-06-01
docker: race condition on *.pyc removal fix

mihaibivol:


mihaibivol:
cc @jmartinm

PR #1176, opened 2016-06-02
tests: ensure that all records are valid

jacquerie:
As suggested during standup.


jacquerie:
Not very useful now, but might be useful for the DoJSON stuff.

PR #1177, opened 2016-06-02
travis: fix the build

jacquerie:


jacquerie:
With @mihaibivol we determined that our trouble is on Docker's end, so there's no use for this PR.

PR #1179, opened 2016-06-03
docs: install guide update

jalavik:


INSTALL.rst
```diff
@@ -75,12 +98,17 @@ Add demo records
...
+    Remember to keep `honcho` running in a separate window.
+
+
 Adding a single record with CLI
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. code-block:: console
 
-   (inspirehep)$ demouuid = $(dojson do -l marcxml -i inspirehep/demosite/data/sample.xml hep | inspirehep records create)
+   (inspirehep)$ demouuid = $(dojson -l marcxml -i inspirehep/demosite/data/sample.xml do hep | inspirehep records create)
```
jmartinm: Does this index the record? we can also run the migrator command pointing to a marcxml with one record.

INSTALL.rst
```diff
@@ -2,95 +2,166 @@
...
+    (inspirehep)$ npm install
+    (inspirehep)$ inspirehep collect -v
+    (inspirehep)$ inspirehep assets build
+
 
+.. note::
 
-Usage
-=====
+    Might give an error about missing `minjs`, pip install it to fix.
```
jacquerie: This is no longer true, I believe? Now `minjs` is part of `setup.py`.

PR #1180, opened 2016-06-03
dojson: remove bibfield code and tests

jacquerie:
@jalavik told me IRL that this code is no longer needed.

Since this code was well-tested I expect coverage to go down.


jalavik:
:shipit:

jacquerie:
:ship:

PR #1145, opened 2016-05-24
format: institution detailed record

nikpap:
capture d ecran 2016-05-27 a 15 06 57

Signed-off-by: Nick Papoutsis <admin@nils.gr>

PR #1181, opened 2016-06-06
Institute detailed page

jmartinm:


jmartinm:
Takes over from https://github.com/inspirehep/inspire-next/pull/1145

inspire-inst

eamonnmag:
Awesome!

kaplun:
This is indeed great! As we mused IRL (and partially with Salvatore), would be great if:
* [ ] List of papers is moved above other stats
* [ ] Sorted by number of papers
* [ ] Clearly understand if we consider papers from people *currently* at the given institutions (so Institute -> People -> papers) Vs. Papers with authors affiliated at the given institution
* [ ] Boxes should be memoized 
* [ ] While we are basing this on ES rather than Neo4J, we should then make use of nested documents (in particular to search for all the signatures which have been made at a given institute and then be able to group by that author)

jmartinm:
> Clearly understand if we consider papers from people currently at the given institutions (so Institute -> People -> papers) Vs. Papers with authors affiliated at the given institution

So after nesting authors in `Literature` what I am doing is getting all authors that wrote a paper while being affiliated at the institution and counting the number of papers they wrote (at the institution).

So it is historical data (not current) which has the benefit of only involving querying `Literature` and also our `current` status is usually not very up-to-date.

PR #991, opened 2016-03-14
WIP search: suggestion messages

Panos512:
* Adds suggestion messages on search.

* Closes inspirehep/inspire-next#939.

Signed-off-by: Panos Paparrigopoulos <panos.paparrigopoulos@cern.ch>

PR #1182, opened 2016-06-07
WIP records/search/theme: suggestion messages

Panos512:
- Adds suggestion messages on search.
- Succeeds #939 and #991.

Preview Screenshots:

prev1

prev2

prev3

Currently we have 3 types of messages (as shown on the above screenshot):
- Malformed queries (queries with unclosed parentheses etc).
- Unsupported keywords (queries containing currently unsupported keywords that require nested searches such as refersto).
- Extra keyword queries (queries such as `author: foo bar` in which only `foo` is searched in authors while `bar` is searched in the `full_text`).

TODO:
- [ ] Pass tests.
- [x] Write tests.
- [ ] Decide on final form of messages (and, why not, add more messages).

### DIFF
--- a/tests/integration/test_api_literature.py
+++ b/tests/integration/test_api_literature.py
@@ -1,0 +1,82 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+import json
+
+
+def test_api_literature_search_no_warnings_when_query_is_well_formed(app):
+    with app.test_client() as client:
+        response = client.get(
+            '/api/literature/?cc=literature&q=author:+Panos',
+            headers = {
+                'Accept': 'application/vnd+inspire.brief+json',
+            },
+        )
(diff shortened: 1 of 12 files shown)


jmartinm:
A small design comment, I would add a bit of opacity to the black background of the text (see https://github.com/inveniosoftware/invenio/search?utf8=%E2%9C%93&q=ast+%29).

For the messages:

* First one seems fine.
* Second one. The only problem I see is that you tell me `Malformed query` without explaining what I did wrong so I cannot improve next time (and will feel insulted ;) )
* The last one is hard to understand. Is it possible to have more concrete info in the message? Like `You can use quotes author:"Foo Bar" to search for multiple words.`

Panos512:
@jmartinm Thanks for the suggestions! For the `malformed query` the problem is that we cannot identify the exact problem. It basically just wraps the `SyntaxError` exception e.g. for the query `author: panos (` it wraps 

```
SyntaxError(u"expecting one of [<class 'invenio_query_parser.parser.AndQuery'>, <class 'invenio_query_parser.parser.OrQuery'>, <class 'invenio_query_parser.parser.ImplicitAndQuery'>]",)
```

We could of course change the message in something more polite 😃  .

For the last one I will check if it's possible to produce the 'quoted-query'  so that we can present it.

tests/integration/test_api_literature.py
```diff
@@ -0,0 +1,84 @@
...
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+import json
+
+import requests
```
jmartinm: `requests` is not used I think?

tests/integration/test_api_literature.py
```diff
@@ -0,0 +1,84 @@
+# -*- coding: utf-8 -*-
```
jmartinm: Name of the file could be better `test_literature_api.py`

tests/integration/test_api_literature.py
Panos512: ooops leftover

tests/integration/test_api_literature.py
Panos512: sure if @jacquerie is ok with this. I believe he proposed it in order to have `test_api_authors` etc and keep the the files in order that way.

jmartinm: Ah ok, was saying because our tests are named at the moment:

`test_<whatever>_views.py`

So thought might be better for API tests to be named:

``test_<whatever>_api.py`

kaplun:
@jacquerie proposes some radical change to the architecture of invenio_query_parser that we can overall attack once we work on the search milestone.

jacquerie:

PR #1187, opened 2016-06-08
setup: gunicorn for development

jalavik:
* Moves away from the standard development server to use
  gunicorn in order to address the resource unavailable issue
  which often occured during local development.

* NOTE You can edit the gunicorn settings locally in `gunicorn.cfg`

Co-authored-by: Javier Martin Montull <javier.martin.montull@cern.ch>
Co-authored-by: Jacopo Notarstefano <jacopo.notarstefano@cern.ch>

CC @Panos512

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
-web: inspirehep --debug run
+web: gunicorn inspirehep.wsgi -c gunicorn.cfg
 cache: redis-server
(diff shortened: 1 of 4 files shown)


kaplun:
Since this is used for development in Honcho, shall we force it to have a single process (so that not 5 processes are being allocated). Or is this already the default?

jalavik:
@kaplun Only one is actually the default http://gunicorn-docs.readthedocs.io/en/latest/settings.html#workers

PR #1195, opened 2016-06-09
theme: permanent redirect for /record/N

### LINKED ISSUE #1076: /record/NNNN should redirect to the appropriate /literature/NNN or similar

jacquerie:
Uses HTTP 301 code, because redirection is permanent.

Closes #1076

### DIFF
--- a/inspirehep/modules/theme/views.py
+++ b/inspirehep/modules/theme/views.py
@@ -3,30 +3,31 @@
 # This file is part of INSPIRE.
 # Copyright (C) 2016 CERN.
 #
-# INSPIRE is free software; you can redistribute it
-# and/or modify it under the terms of the GNU General Public License as
-# published by the Free Software Foundation; either version 2 of the
-# License, or (at your option) any later version.
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
 #
-# INSPIRE is distributed in the hope that it will be
-# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-# General Public License for more details.
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


jmartinm:
Looks good.

jacquerie:
Let me just add a test that covers the case in which there was no collection.

jacquerie:
> Let me just add a test that covers the case in which there was no collection.

Which, in fact, contained a bug 👍

jacquerie:
If you want to take another look, @jmartinm, and merge if you want.

inspirehep/modules/theme/views.py
```diff
@@ -549,3 +550,18 @@ def menu_fixup():
...
+#
+
+@blueprint.route('/record/<control_number>')
+def record(control_number):
+    try:
+        pid = PersistentIdentifier.query.filter_by(
+            pid_value=control_number).one()
+        return redirect('/{collection}/{control_number}'.format(
+            collection=pid.pid_type, control_number=control_number)), 301
+    except Exception:
```
jmartinm: Is there a concrete exception we can capture here?

jmartinm: Apart from this, if you can rebase the branch we can merge it.

jacquerie: Since we are falling back to SQLAlchemy I guess we can catch `SQLAlchemyError`. On the other hand I wasn't entirely sure if the second line could fail in weirder ways...

jacquerie:
This can probably be resurrected and merged.

inspirehep/modules/theme/views.py
jalavik: I'd say catch the exception and do the abort when no result is found or multiple rows is found, see exception types here http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.one

Then the return statement should be outside of the try..except.

jalavik:
👍  , just minor comment on the exception handling.

inspirehep/modules/theme/views.py
jacquerie: Oh, nice idea. Should we fail more loudly than `404` when `MultipleResultsFound`? That's a fairly big error for us...

jalavik: Right. Good point. Maybe that kind of exception can be allowed to just raise all the way.

jacquerie:
:ship:

PR #1196, opened 2016-06-09
dojson: fix acquisition_source

jalavik:
I would like this to be a single item field. Then we can more easily filter/display this field for new Holding Pen, and keep our sanity. 

Thoughts?

### DIFF
--- a/tests/unit/dojson/test_dojson_hep.py
+++ b/tests/unit/dojson/test_dojson_hep.py
@@ -30,7 +30,7 @@ import pytest
 from dojson.contrib.marc21.utils import create_record
 
 from inspirehep.dojson.hep import hep, hep2marc
-from inspirehep.dojson.utils import get_recid_from_ref
+from inspirehep.dojson.utils import get_recid_from_ref, strip_empty_values
 
 
 @pytest.fixture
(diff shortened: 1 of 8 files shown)


jalavik:
cc @annetteholtkamp @kaplun @jacquerie @jmartinm

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -312,16 +312,16 @@ def test_licence(marcxml_to_json, json_to_marc):
...
+    assert (marcxml_to_json['acquisition_source']['source'] ==
+            json_to_marc['541']['a'])
+    assert (marcxml_to_json['acquisition_source']['email'] ==
+            json_to_marc['541']['b'])
+    assert (marcxml_to_json['acquisition_source']['method'] ==
+            json_to_marc['541']['c'])
+    assert (marcxml_to_json['acquisition_source']['date'] ==
+            json_to_marc['541']['d'])
+    assert (marcxml_to_json['acquisition_source']['submission_number'] ==
+            json_to_marc['541']['e'])
```
jacquerie: If you have time it would be cool if you could convert these tests to the new style, like in https://github.com/inspirehep/inspire-next/blob/master/tests/unit/dojson/test_dojson_common_base.py.

jalavik: Sure. Done,

kaplun:
I have so far not looked at this field because IMHO this should be treated not in the metadata but as part of an audit log.

The acquisition source should be a piece of information attached to a specific JSON that is then merged into the current one.

This requires a bit more investigation on how to best implement it though.

jalavik:
@kaplun I agree we should have an additional store for this. Meanwhile I think we can merge this for now as field `541`/`acquisition_source`  is being exported/imported to PROD metadata so keeping some reference of this in the metadata I guess is necessary for the short term, or at least for simplicity right now.

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -529,3 +515,50 @@ def test_book_link(marcxml_to_json_book):
...
             1409249)
+
+
+def test_acquisition_source_field():
+    """Test acquisition_source."""
+    snippet = (
+        '<record>'
+        '   <datafield tag="541" ind1=" " ind2=" ">'
+        '       <subfield code="a">inspire:uid:50000</subfield>'
+        '       <subfield code="b">montull@gmail.com</subfield>'
```
jmartinm: Will I get spammed having my email here? ;)

PR #1170, opened 2016-05-31
schemas: clean schemas

jacquerie:
Continues #1069, #1070, and #1143. Here's the plan:

- [x] Fix imports from `inspirehep.dojson.utils`
- [x] Implement `remove_duplicates` from `inspirehep.dojson.utils` in terms of `inspirehep.utils.dedupers`
- [x] Implement `remove_duplicates` from `inspirehep.modules.theme` in terms of `inspirehep.utils.dedupers`
- [x] Move `inspirehep/dojson/utils.py` in `inspirehep/dojson/utils/__init__.py`
- [x] Add @tomaszgy's non-geo helpers to `inspirehep/dojson/utils/__init__.py`, with tests
- [x] Add @bittirousku's helpers to `inspirehep/dojson/utils/__init__.py`, with tests
- [x] Add `inspirehep/dojson/utils/geo.py` from @tomaszgy, with tests
- [x] Add element schemas in `inspirehep/modules/records/jsonschemas/records/elements`, add test to check their validity
- [x] Amend schemas in `inspirehep/modules/records/jsonschemas` to use elements, collection by collection (with DoJSON rules and tests)

    - [x] Address
    - [x] Contact
    - [x] Fields
    - [x] ISBN & ISSN
    - [x] Ranks
    - [x] Titles
    - [x] URLs

- [x] Final cleanup

Extra:
- Fix DoJSON docstrings
- Test current methods in `inspirehep/dojson/utils.py`
- See if `inspirehep/dojson/utils/geo.py` can be better implemented in terms of `pycountry`
- Refactor DoJSON tests
- Integrate #1070

PR #1194, opened 2016-06-09
workflows, crawler and dojson: consolidation of fixes

jalavik:
Includes:

* fixes to workflows tasks
* integration of inspire-crawler
* minor dojson/schema updates

PR #1197, opened 2016-06-09
theme: amend templates to pass tests

### LINKED ISSUE #1178: profiles: missing email / gravatar bug
If an author does not have an e-mail address, rendering of an author profile will fail due to this line
```html
<img class="profile-image img-circle" src="{{ record['positions'][0]['email'] ...>
```

jacquerie:
* First commit is unrelated, but while investigating this issue I wrote a few tests for the corresponding DoJSON rules to see if the bug is there.
* Second commit fixes the bugs currently present in templates after #1170.
* Third commit adds `DEBUG = True` to the integration `app` fixture, so that we can more easily discover the bugs fixed in the previous commit.

Closes #1178

### DIFF
--- a/inspirehep/modules/theme/templates/inspirehep_theme/format/record/authors/Author_HTML_detailed.html
+++ b/inspirehep/modules/theme/templates/inspirehep_theme/format/record/authors/Author_HTML_detailed.html
@@ -1,20 +1,23 @@
 {#
-## This file is part of INSPIRE.
-## Copyright (C) 2014 CERN.
-##
-## INSPIRE is free software; you can redistribute it and/or
-## modify it under the terms of the GNU General Public License as
-## published by the Free Software Foundation; either version 2 of the
-## License, or (at your option) any later version.
-##
-## INSPIRE is distributed in the hope that it will be useful, but
-## WITHOUT ANY WARRANTY; without even the implied warranty of
-## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-## General Public License for more details.
-##
-## You should have received a copy of the GNU General Public License
-## along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
-## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+# This file is part of INSPIRE.
+# Copyright (C) 2014, 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
 #}
 
 {% from "inspirehep_theme/format/record/Inspire_Default_HTML_general_macros.tpl" import mathjax %}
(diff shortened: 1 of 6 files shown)


inspirehep/modules/theme/templates/inspirehep_theme/format/record/authors/Author_HTML_detailed.html
```diff
@@ -34,7 +34,7 @@
       <div class="row">
         <div class="col-md-12 summary" id="detailed-header-top">
           <div class="col-md-2">
-            <img class="profile-image img-circle" src="{{ record['positions'][0]['email'] | gravatar(size=90, default='identicon') }}" data-toggle='tooltip' data-placement='left' title='Your profile picture can be set at gravatar.com' width="100%">
+            <img class="profile-image img-circle" src="{{ record['positions'][0].get('email', 'gravatar@inspirehep.net') | gravatar(size=90, default='identicon') }}" data-toggle='tooltip' data-placement='left' title='Your profile picture can be set at gravatar.com' width="100%">
```
jmartinm: Might be that some author does not have any items in `positions` by the way, so this line could continue failing.

Also with @jacenkow we were looking into making the default image dynamic based on the initials of the author, a la google https://raw.githubusercontent.com/hellsan631/material-avatar/master/sample.png so I guess the trick of registering our email in gravatar would not work in that case.

jacquerie: Right. I guess I can amend this commit to exclude this modification (and also remove the commit that adds `DEBUG = True` so that the build does not fail).

jmartinm: Or to make the build pass you could for the moment leave it with the `default='identicon'` and just use `get()` to avoid crash in case of no email.

jacquerie: Like this?
```python
record.get('positions'], [])[0].get('email', '') | gravatar(size=90, default='identicon')
```

inspirehep/modules/theme/templates/inspirehep_theme/format/record/authors/Author_HTML_detailed.html
```diff
@@ -34,7 +34,7 @@
       <div class="row">
         <div class="col-md-12 summary" id="detailed-header-top">
           <div class="col-md-2">
-            <img class="profile-image img-circle" src="{{ record['positions'][0]['email'] | gravatar(size=90, default='identicon') }}" data-toggle='tooltip' data-placement='left' title='Your profile picture can be set at gravatar.com' width="100%">
+            <img class="profile-image img-circle" src="{{ record.get('positions', [])[0].get('email', '') | gravatar(size=90, default='identicon') }}" data-toggle='tooltip' data-placement='left' title='Your profile picture can be set at gravatar.com' width="100%">
```
jmartinm: I guess if there are no `positions` this line will still fail as it will try to access index `0` of empty list.

We might need to add a bit extra code:

```jinja2
{% set email = '' %}
{% if  record.get('positions', []) %}
    {% if  len(record.get('positions')) > 0 %}
        {% set email = record['positions'][0].get('email', '') %}
    {% endif %}
{% endif %}
<img class="profile-image img-circle" src="{{ email | gravatar(size=90, default='identicon') }}" data-toggle='tooltip' data-placement='left' title='Your profile picture can be set at gravatar.com' width="100%">
```

jacquerie: Right. I'm dumb.

inspirehep/modules/theme/templates/inspirehep_theme/format/record/authors/Author_HTML_detailed.html
```diff
@@ -34,7 +37,11 @@
       <div class="row">
         <div class="col-md-12 summary" id="detailed-header-top">
           <div class="col-md-2">
-            <img class="profile-image img-circle" src="{{ record['positions'][0]['email'] | gravatar(size=90, default='identicon') }}" data-toggle='tooltip' data-placement='left' title='Your profile picture can be set at gravatar.com' width="100%">
+            {% set email = '' %}
+            {% if record.get('positions') and (record['positions'])|length > 0 %}
```
jalavik: Generally, in jinja you strictly do not need to `.get` I believe. `record.positions | length > 0 ` may just work even if positions is not filled.

jacquerie: Ah, nice to know :+1:

Will amend.

jacquerie:
Nothing had to be changed after #1194, I just incorporated some changes suggested by @jalavik in code review. :shipit:?

PR #1198, opened 2016-06-09
Conferences detailed

jmartinm:


kaplun:
Nice! I guess you can skip the year in this case since everything will belong to the same conference. I have some comments on the first block (e.g. how series is presented) but it's easier in real life. Anyway nothing major :dancer:

jmartinm:
> Nice! I guess you can skip the year in this case since everything will belong to the same conference. I have some comments on the first block (e.g. how series is presented) but it's easier in real life. Anyway nothing major 

Yes, the first block is not very clean yet and the series and contributions don't have all the numbers in place. Feedback welcome next week.

annetteholtkamp:
We should also distinguish between proceedings and contributions. And we should discuss which info to show for the contributions. Makes a difference whether they have been published or not (yet).

- Annette

PR #1200, opened 2016-06-10
tests: unit tests for DoJSON institutions

### LINKED ISSUE #1202: schemas: historical_data should be a list of strings
After fixing [this bug](https://github.com/inspirehep/inspire-next/pull/1200/files#diff-f9c3d63484a960776a18587a59d6c1c5L147) in the DoJSON rule for `historical_data` we get [this validation failure](https://travis-ci.org/inspirehep/inspire-next/jobs/136654556#L2046), which is caused by the fact that [its record](https://inspirehep.net/record/902725/export/xm) has multiple `a` fields in `6781_`.

(This issue might become obsolete with the schema redesign, but I wanted to write it down.)

jacquerie:
~~This PR adds more tests to the DoJSON conversions for the "minor" collections: Institutions, Conferences, Experiments, Jobs, and Journals.~~

~~The main purpose is to see what we actually have in the data and discover its quirks. Some will have to be solved in Legacy (like https://github.com/inspirehep/inspire/issues/170), others will have to be taken in account when building the DoJSON rules.~~

This was growing too much in scope, so in this PR I will only deal with Institutions. Other minor collections are tracked in #1240.

Only the assertions will have to be changed when we finalise the schemas, the test cases will stay the same (which is why this work will not be wasted).

- [x] Institutions
    - [x] Test `location`
    - [x] Fix population of `name_variants`
    - [x] Test `name_variants`
    - [x] Fix population of `hidden_notes`
    - [x] Test `hidden_notes`
    - [x] Refactor `location`

Closes #1202

### DIFF
--- a/inspirehep/dojson/institutions/fields/bd1xx.py
+++ b/inspirehep/dojson/institutions/fields/bd1xx.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of INSPIRE.
-# Copyright (C) 2014, 2015 CERN.
+# Copyright (C) 2014, 2015, 2016 CERN.
 #
 # INSPIRE is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -21,6 +21,8 @@
 # or submit itself to any jurisdiction.
 
 """MARC 21 model definition."""
+
+from __future__ import absolute_import, division, print_function
 
 from dojson import utils
 
@@ -144,8 +146,10 @@ def public_notes(self, key, value):
     return value.get('i')
 
 
-@institutions.over('historical_data', '^6781..')
-@utils.for_each_value
+@institutions.over('historical_data', '^6781.')
 def historical_data(self, key, value):
     """Historical data."""
-    return value.get('a')
+    values = self.get('historical_data', [])
+    values.extend(el for el in value.get('a'))
+
+    return values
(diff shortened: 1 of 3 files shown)


jacquerie:
This is becoming way too complicated for a single PR, and might get split in 5 (one per collection) + 1 for cleaning up.

jacquerie:
Code for this PR is ready, but this PR depends on https://github.com/inspirehep/inspire/issues/170 to be solved on legacy.

jacquerie:
Mh, I'm still going to rebase and merge this, so that #1239 can proceed, but https://github.com/inspirehep/inspire/issues/170 needs to be fixed before deploying.

inspirehep/dojson/institutions/fields/bd1xx.py
```diff
@@ -32,20 +36,15 @@
...
-        except:
-            pass
+            return float(value[c])
+        except (KeyError, ValueError):
+            return ''
+
     return {
-        'longitude': longitude,
-        'latitude': latitude
+        'longitude': _get_float(value, 'd'),
```
jmartinm: Will the code using these fields rely on them being `floats`?

In which case the above:

```python
except (KeyError, ValueError):
             return ''
```

might give some trouble.

jacquerie: Then they will be skipped, as documented by the tests: https://github.com/jacquerie/inspire-next/blob/ff5b8b2be01c6386973f17ae2c587a9664ccac54/tests/unit/dojson/test_dojson_institutions.py#L80-L90

PR #1201, opened 2016-06-10
jinja2filters: pubinfo sanity check

jmartinm:
* Avoids crash when parent recid does not exist in ES.

* Adds testcase.


jacquerie:
Nice stuff (pinging @mihaibivol so that he knows this is fixed).

LGTM!

PR #1204, opened 2016-06-13
tests: unit tests for authors DoJSON

jacquerie:
Pretty straightforward tests. Still missing a few, but this was a quick coverage win and will help refactoring after #1170.

A similar PR will come for `inspirehep/modules/literaturesuggest/dojson`.

### DIFF
--- a/tests/unit/authors/test_authors_dojson.py
+++ b/tests/unit/authors/test_authors_dojson.py
@@ -1,0 +1,479 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


jacquerie:
These tests could use a `GroupableOrderedDict` for forms to test the `append` behaviour currently not covered by tests.

tests/unit/authors/test_authors_dojson.py
```diff
@@ -0,0 +1,479 @@
...
+
+
+def test_urls_from_websites_webpage():
+    form = {
+        'websites': [
+            {
+                'webpage': 'foo',
+            },
+        ],
+    }
```
jacquerie: This doesn't use a `GroupableOrderedDict` because I couldn't figure out the syntax to have a value of type `list`.

jacquerie:
:shipit: ?

PR #1205, opened 2016-06-13
theme: fix google maps markers

jmartinm:
* Use png instead of svg for markers and set a scaled size.


jmartinm:
this should help with issues like https://github.com/inspirehep/inspire-next/pull/1083#discussion_r62815939

jacquerie:
LGTM, tested working locally.

Issue #1206, opened 2016-06-13
Validation errors on production records


jmartinm:
After the last test migrating an INSPIRE dump into qa.inspirehep.net, the following validation errors appeared.

https://www.dropbox.com/s/0xnwtmxunvfdopg/validation-errors.txt?dl=1

jacquerie:
# Validation Errors #

## Wrong Enums ##

```
============
1442216
============
ValidationError: Record None: 'magpie' is not one of [u'INSPIRE', u'submitter', u'curator', u'publisher']

Failed validating u'enum' in schema[u'properties'][u'field_categories'][u'items'][u'properties'][u'source']:
    {u'enum': [u'INSPIRE', u'submitter', u'curator', u'publisher'],
     u'type': u'string'}

On instance[u'field_categories'][0][u'source']:
    'magpie'
```
```
============
1448256
============
ValidationError: Record None: 'online' is not one of [u'electronic', u'print']

Failed validating u'enum' in schema[u'properties'][u'issn'][u'items'][u'properties'][u'medium']:
    {u'description': u'Medium (type) of the ISSN.',
     u'enum': [u'electronic', u'print'],
     u'type': u'string'}

On instance[u'issn'][0][u'medium']:
    'online'
```

## Wrong Types ##

```
============
1442215
============
ValidationError: Record None: 'Skobelltsyn,D.V.' is not of type u'array'

Failed validating u'type' in schema[u'properties'][u'references'][u'items'][u'properties'][u'editors']:
    {u'items': {u'type': u'string'}, u'type': u'array'}

On instance[u'references'][15][u'editors']:
    'Skobelltsyn,D.V.'
```
```
============
1448393
============
ValidationError: Record None: ("Sel'tser, Sh.Z.", u'\u0421\u0435\u043b\u044c\u0446\u0435\u0440, \u0428.\u0417.') is not of type u'string'

Failed validating u'type' in schema[u'properties'][u'authors'][u'items'][u'properties'][u'alternative_name']:
    {u'description': u'FIXME: needed for import only',
     u'format': u'.+, .+',
     u'title': u'Alternative author name',
     u'type': u'string'}

On instance[u'authors'][2][u'alternative_name']:
    ("Sel'tser, Sh.Z.",
     u'\u0421\u0435\u043b\u044c\u0446\u0435\u0440, \u0428.\u0417.')
```

kaplun:
[...]

PR #1212, opened 2016-06-15
schemas: make hidden_notes values an array

jacquerie:
First commit addresses #1206 (first bullet point).

Second and third commit add tests to the DoJSON rule in question and then refactor it, ensuring that the same behaviour is preserved.


jacquerie:
Coverage decreased because number of lines decreased, so it's fine.

:shipit: ?

PR #1218, opened 2016-06-15
utils: stats class for h and i10 indexes

jalavik:
By request I separated this one from #1214 

CC @eamonnmag

### DIFF
--- a/inspirehep/utils/stats.py
+++ b/inspirehep/utils/stats.py
@@ -20,7 +20,6 @@ import operator
 
 
 class Stats(object):
-
     """Stats is a class to calculate common metrics"""
 
     @staticmethod
@@ -33,13 +32,16 @@ class Stats(object):
         :return: h index for a given list of citations.
         """
         _record_citations = kwargs.pop('citations')
+        if not Stats.check_type_is_dict(_record_citations):
+            raise TypeError("citations should be a dict")
+
         _sorted = Stats.get_sorted_citations(_record_citations)
 
         h_index = 0
         print _sorted
         for _index, _val in enumerate(_sorted):
             if _val[1] > _index:
-                h_index = _index+1
+                h_index = _index + 1
             else:
                 break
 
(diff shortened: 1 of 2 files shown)


inspirehep/utils/stats.py
```diff
@@ -0,0 +1,78 @@
...
+        :return: h index for a given list of citations.
+        """
+        _record_citations = kwargs.pop('citations')
+        if not Stats.check_type_is_dict(_record_citations):
+            raise TypeError("citations should be a dict")
+
+        _sorted = Stats.get_sorted_citations(_record_citations)
+
+        h_index = 0
+        print _sorted
```
jmartinm: leftover print

inspirehep/utils/stats.py
```diff
@@ -0,0 +1,78 @@
...
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+import operator
```
jmartinm:
```
```

inspirehep/utils/stats.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class Stats(object):
```
jmartinm: If a class is not needed we can consider just having functions that receive the needed parameters instead of `*args, **kwargs`

eamonnmag: I agree. I was being 'pythonic' whose only known synonym is 'convoluted'.

jmartinm:
So where are these utility functions used? I can take care of clean up and merging them when they are needed.

inspirehep/utils/stats.py
```diff
@@ -0,0 +1,78 @@
...
+        Given a citation dictionary, in the format {recid: citation_count}
+        returns the i10 index, which is simply the number of records with
+        10 or more citations.
+        :param args:
+        :param kwargs:
+        :return: i10 index for a given list of citations.
+        """
+
+        _record_citations = kwargs.pop('citations')
+        if not Stats.check_type_is_dict(_record_citations):
```
jmartinm: Another nice way of doing this in Python is:

```python
assert _record_citations is dict, "Citations should be a dict"
```

kaplun: The only theoretical issue of using `assert` in Python is that if you execute Python in optimized mode all the assert statements will be stripped. Now I don't believe this is happening ever.

kaplun:
> So where are these utility functions used? I can take care of clean up and merging them when they are needed

They are not yet used, but should be used in the Author profile pages, IMHO.

PR #1224, opened 2016-06-17
theme: use count instead of hits total

jacquerie:
Doesn't yet do what the title says, but the first two commits refactor the somewhat messy Jinja2 filters and add a few missing tests. There will be a third commit implementing what @eamonnmag suggested.

(I would also like to move ES logic into the controller from the template, but this PR will not yet do that.)

### DIFF
--- a/inspirehep/modules/theme/jinja2filters.py
+++ b/inspirehep/modules/theme/jinja2filters.py
@@ -34,6 +34,7 @@ from operator import itemgetter
 from flask import current_app
 from jinja2.filters import do_join, evalcontextfilter
 
+from invenio_search import current_search_client as es
 from invenio_search.api import RecordsSearch
 
 from inspirehep.modules.records.json_ref_loader import replace_refs
@@ -364,7 +365,9 @@ def collection_select_current(collection_name, current_collection):
 def number_of_records(collection_name):
     """Returns number of records for the collection."""
     index = collection_to_index(collection_name)
-    return perform_es_search('', index).hits.total
+    result = es.count(index=index)
+
+    return result['count']
 
 
 @blueprint.app_template_filter()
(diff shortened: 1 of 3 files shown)


jacquerie:
Now does what the title says.

jacquerie:
Ready for review.

jacquerie:
I have the suspicion that the problem of bogus counts being shown is more the result of caching by Flask/Jinja2 rather than using the incorrect method of ES.

eamonnmag:
You are right. After further testing on my end, similar behaviour can be found on hepdata.net if you try long enough.

jmartinm:
That could be, doing a quick search seems to be quite common 'problem' http://stackoverflow.com/a/27891981/890185

I guess because our input to the filter is constant it gets cached:

```
{{ "HEP" | number_of_records }}
```

If instead it was:

```
{{ RANDOM() | number_of_records }}
```

It wouldn't (according to the link above).

jacquerie:
That doesn't look so bad. We could cache the result of the `count` query for 30/60 minutes, so that we don't constantly query ES, but constantly call the `number_of_records` filter using the `RANDOM` trick above. WDYT?

jmartinm:
Yes, that could work. If jinja2 caches the whole template (not only the filter) then with this we will lose that caching so landing pages will always be recompiled on every request.

I am not sure if that would have a visible impact.

jacquerie:
Ah, I was assuming the caching invalidation was per-filter, not per-page.

jacquerie:
This can probably be merged, since I'm working on another PR to evict ES queries from filters that builds on top of this.

kaplun:
@jacquerie can you rebase it? So that then we merge it?

jacquerie:
Just sent a rebase your way. :ship:

PR #1225, opened 2016-06-19
tests: unit tests for literaturesuggest DoJSON

jacquerie:
Another quick coverage win.

### DIFF
--- a/inspirehep/modules/literaturesuggest/dojson/fields/literature.py
+++ b/inspirehep/modules/literaturesuggest/dojson/fields/literature.py
@@ -182,7 +182,7 @@ def issue(self, key, value):
     if 'publication_info' in self:
         self['publication_info'][0].update(dict(journal_issue=value))
     else:
-        self['publication_info'] = [dict(issue=value)]
+        self['publication_info'] = [dict(journal_issue=value)]
     raise IgnoreKey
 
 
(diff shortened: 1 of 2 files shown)


jacquerie:
First commit adds unit tests for the `literaturesuggest` DoJSON rules. Second commit fixes a bug found while writing these tests.

jacquerie:
Ready for review.

kaplun:
I am understanding this tests (although long) are very simple, but still allowed you to spot a bug. Well done. I wonder if they are making it though more rigid our ability to change the data model (in the sense that a small change will now imply fixing several tests as well), but I guess that's a price to pay for stability.

jacquerie:
The biggest benefit of unit tests is evident when refactoring: if you want to ensure that behaviour stays the same, being constrained by what the tests say the code used to do is very beneficial.

If you're reimplementing from scratch, instead, you can probably throw all the unit tests away. Hopefully then you are constrained by your integration tests.

jacquerie:
BTW: this test is similar in spirit to https://github.com/inspirehep/inspire-next/blob/6cf8c2d110833e69b608f7383d305e37a81e202a/tests/test_hep_submission.py, so we can say that this PR addresses #918.

kaplun:
LGBT

eamonnmag:
Are you wearing your pink cardigan again? 💃

kaplun:
:two_women_holding_hands:

PR #1227, opened 2016-06-21
Legacy user and remote account migration

jmartinm:


lnielsen:
It would be nice if you could submit this against Invenio-Migrator. I need it as well on Zenodo.

inspirehep/modules/migrator/legacy/remoteaccount.py
```diff
@@ -0,0 +1,75 @@
...
+
+def dump(u, from_date, **kwargs):
+    """Dump the remoteaccounts as a list of dictionaries."""
+    from invenio_oauthclient.models import RemoteAccount
+    from invenio_accounts.models import UserEXT
+
+    # First fetch all the remote account information
+    remoteaccount = {}
+    try:
+        result = RemoteAccount.query.filter_by(user_id=u.id).one()
```
lnielsen: I think you also need to dump ``RemoteToken``?

inspirehep/modules/migrator/legacy/users.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def get(*args, **kwargs):
+    """Get users."""
+    from invenio_accounts.models import User
+    q = User.query
+    return q.count(), q.all()
+
+
+def dump(u, from_date, with_json=True, latest_only=False, **kwargs):
```
lnielsen: I think this one already exists in Invenio-Migrator.

inspirehep/modules/migrator/legacy/remoteaccount.py
jmartinm: Thing is that we are updating the scope of the ORCID authentication so we would like users to relogin and get a new token once we deploy Invenio 3.

inspirehep/modules/migrator/legacy/users.py
jmartinm: Yes, had to copy it over because `invenio-migrator` is based on Invenio 2.0 while our installation went ahead and is more bleeding edge so all the imports for us are:

```
```

Instead of

```
from invenio.modules.accounts.models import User
```

lnielsen: Since they are basically equivalent, it's fine to just add:
```python
try:
    from invenio.modules.accounts.models import User
except ImportError:
```
to https://github.com/inveniosoftware/invenio-migrator/blob/master/invenio_migrator/legacy/users.py#L34

jmartinm: Sure, that is fine in that case. Created PR https://github.com/inveniosoftware/invenio-migrator/pull/34

Can remove from here once merged.

jmartinm:
> It would be nice if you could submit this against Invenio-Migrator. I need it as well on Zenodo.

I can create a PR there although as before, we rely on a bit more modern version than Invenio 2.0

I will do the same `try:...except...` for the imports and you could give it a try as I don't have any Invenio 2.0 installation to try it. I think it should work any way.

inspirehep/modules/migrator/legacy/users.py
lnielsen: Thanks a lot. I've merged it.

lnielsen:
@jmartinm Would be great. For the dump, if you don't need the remote token you can just leave it out in the PR. I can add it later.

jmartinm:
@lnielsen PR Created now https://github.com/inveniosoftware/invenio-migrator/pull/36

jmartinm:
The PR might not be needed after we merge https://github.com/inveniosoftware/invenio-migrator/pull/36

jmartinm:
Not needed anymore since the functionality was merged into `invenio-migrator` - https://github.com/inveniosoftware/invenio-migrator/commit/87b3f766a31d16337cc38c1804e885f80468929d

PR #1236, opened 2016-06-22
jsonschema: page_artid -> page_start/end + artid

kaplun:
* Amends JSONSchema, dojson, serializer, search, Jinja templates in
  order to correctly capture the concepts of page_start, page_end,
  and artid.

### DIFF
--- a/inspirehep/modules/literaturesuggest/dojson/fields/literature.py
+++ b/inspirehep/modules/literaturesuggest/dojson/fields/literature.py
@@ -34,6 +34,8 @@ from dojson import utils
 from dojson.errors import IgnoreKey
 from idutils import is_arxiv_post_2007
 
+from inspirehep.dojson.utils import split_page_artid
+
 from ..model import literature
 
 
@@ -215,10 +217,11 @@ def year(self, key, value):
 
 @literature.over('_page_range_article_id', '^page_range_article_id$')
 def page_range_article_id(self, key, value):
-    if 'publication_info' in self:
-        self['publication_info'][0].update(dict(page_artid=value))
-    else:
-        self['publication_info'] = [dict(page_artid=value)]
+    page_start, page_end, artid = split_page_artid(value)
+    self.setdefault('publication_info', [{}])[0].update(dict(
+        page_start=page_start,
+        page_end=page_end,
+        artid=artid))
     raise IgnoreKey
 
 
--- a/inspirehep/modules/literaturesuggest/tasks.py
+++ b/inspirehep/modules/literaturesuggest/tasks.py
@@ -83,7 +83,8 @@ def formdata_to_model(obj, formdata):
                 }]
     if "publication_info" in data:
         if all([key in data['publication_info'][0].keys() for key in
-               ('year', 'journal_issue', 'journal_volume', 'page_artid')]):
+               ('year', 'journal_issue', 'journal_volume', 'page_start',
+                'page_end', 'artid')]):
             # NOTE: Only peer reviewed journals should have this collection
             # we are adding it here but ideally should be manually added
             # by a curator.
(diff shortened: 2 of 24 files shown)


inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -147,16 +148,16 @@ def formdata_to_model(obj, formdata):
...
     if 'page_nr' not in data:
-        if data.get("publication_info", [{}])[0].get("page_artid"):
-            pages = data['publication_info'][0]['page_artid'].split('-')
-            if len(pages) == 2:
-                try:
-                    data['page_nr'] = int(pages[1]) - int(pages[0]) + 1
-                except ValueError:
-                    pass
+        if (data.get("publication_info", [{}])[0].get("page_start") and
+                data.get("publication_info", [{}])[0].get("page_end")):
```
jacquerie: Explainer variables would improve this:
```python
first_publication_info = data.get('publication_info', [{}])[0]
page_start = first_publication_info.get('page_start')
page_end = first_publication_info.get('page_end')

if page_start and page_end:
    try:
        data['page_nr'] = page_end - page_start + 1
    except ValueError:
        pass
```
which suggests that this code never throws a `ValueError` after all.

inspirehep/utils/cv_latex_html_text.py
```diff
@@ -203,19 +203,21 @@ def _get_publi_info(self):
...
                     if isinstance(field['journal_title'], list):
-                        if not ('journal_volume' in field or
-                                'journal_issue' in field or
-                                'page_artid' in field or
-                                'doi' in self.record):
+                        if (('journal_volume' not in field or
+                                'journal_issue' not in field or
+                                ('page_start' not in field and
+                                 'artid' not in field)) and
+                                'doi' not in self.record):
```
jacquerie: Logic here has changed, because a `not` of `or`s is different from an `or` of `not`s.

kaplun: Yes, I have corrected the wrong logic there :)

kaplun: This is because this complex query is trying to understand if something has been published or not. something is published if you have either a DOI or journal_title + volue + issue + page_start/artid

kaplun: Maybe calling for a utility function

jacquerie: Definitely use a utility function + an explainer variable. I actually understood that for @ioannistsanaktsidis a paper is published if it has _any one_ of volume, issue, page_start/artid, which is what the tests are asserting.

inspirehep/modules/literaturesuggest/tasks.py
kaplun: That code is really convoluted.

inspirehep/modules/literaturesuggest/dojson/fields/literature.py
```diff
@@ -215,10 +217,11 @@ def year(self, key, value):
 
 @literature.over('_page_range_article_id', '^page_range_article_id$')
 def page_range_article_id(self, key, value):
-    if 'publication_info' in self:
-        self['publication_info'][0].update(dict(page_artid=value))
-    else:
-        self['publication_info'] = [dict(page_artid=value)]
+    page_start, page_end, artid = split_page_artid(value)
+    self.setdefault('publication_info', [{}]).update(dict(
```
jacquerie: Uh? `setdefault` should return the second argument, which then doesn't have an `update` method defined on it because it is a list (and not a dict).

kaplun: Well spotted. Should be:
```python
self.setdefault('publication_info', [{}])[0].update(dict(
```

kaplun:
One more bug to fix and it's good.

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -147,16 +148,17 @@ def formdata_to_model(obj, formdata):
...
-                    data['page_nr'] = int(pages[1]) - int(pages[0]) + 1
-                except ValueError:
-                    pass
+        first_publication_info = data.get('publication_info', [{}])[0]
+
+        page_start = first_publication_info.get('page_start')
+        page_end = first_publication_info.get('page_end')
+
+        if page_start and page_end:
+            data['page_nr'] = page_end - page_start + 1
```
jacquerie: Ouch, I was wrong here. I was assuming `page_start` and `page_end` were integers from the schema, but they are not, so the difference will throw a `TypeError` here. Which means in fact this should be:
```python
try:
    data['page_nr'] = int(page_end) - int(page_start) + 1
except (TypeError, ValueError):
    pass
```

kaplun: Ha! Right! Additionally there might be weird prefixes such as 'pp.' in front `page_start`.

Will send a fix with a test tomorrow.

Issue #861, opened 2016-02-09
Support more document types


jmartinm:
# Book chapter
- Remove Journal information box
- New box "Book information"
# Book
- Basic information like Book chapter "Book information" box

## Book information box contains

- a field for ISBN, with an optional [medium](https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/hep.json#L553) dropdown, and a possibility to add more ISBNs. Ideally, the ISBN should be normalized and validated. Schema: `isbns`
- a field for publisher. Schema: `publisher` in `imprints`
- a field for publication date. Schema: `date` in `imprints`
- a field for publication place. Schema: `place` in `imprints`
- a field for edition. Schema: `editions`
- a field for book series title. Schema: `title` in `book_series`
- a field for book series volume. Schema: `volume` in `book_series`

# Proceedings
- Conference lookup like in the Article Submission.

jmartinm:
A follow up of this task will be https://github.com/inspirehep/inspire-next/issues/2118

PR #1237, opened 2016-06-22
forms: remove dead code

jacquerie:
Instead of writing tests for code that is never called I propose to delete it. Dead code was found using [vulture](https://pypi.python.org/pypi/vulture), imports where amended using [pylint](https://pypi.python.org/pypi/pylint).

### DIFF
--- a/inspirehep/modules/forms/validation_utils.py
+++ b/inspirehep/modules/forms/validation_utils.py
@@ -1,172 +1,39 @@
 # -*- coding: utf-8 -*-
 #
-# This file is part of Invenio.
-# Copyright (C) 2013, 2014, 2015 CERN.
+# This file is part of INSPIRE.
+# Copyright (C) 2013, 2014, 2015, 2016 CERN.
 #
-# Invenio is free software; you can redistribute it and/or
-# modify it under the terms of the GNU General Public License as
-# published by the Free Software Foundation; either version 2 of the
-# License, or (at your option) any later version.
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
 #
-# Invenio is distributed in the hope that it will be useful, but
-# WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-# General Public License for more details.
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
 #
 # You should have received a copy of the GNU General Public License
-# along with Invenio; if not, write to the Free Software Foundation, Inc.,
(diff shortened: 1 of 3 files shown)


jmartinm:
Having a very quick look, this seems like code that indeed we don't need

inspirehep/modules/forms/validation_utils.py
```diff
@@ -1,172 +1,39 @@
...
-
-def doi_syntax_validator(form, field):
-    """DOI syntax validator. Deprecated.
-
-    :param field: validated field.
-    :param form: validated form.
-    """
-    import warnings
-    warnings.warn("Please use DOISyntaxValidator instead.", DeprecationWarning)
-    return DOISyntaxValidator()(form, field)
```
kaplun: We don't validate DOI?

inspirehep/modules/forms/validators/simple_fields.py
```diff
@@ -42,16 +39,6 @@ def arxiv_syntax_validation(form, field):
         raise StopValidation(message)
 
 
-def isbn_syntax_validation(form, field):
```
kaplun: What about ISBNs? @jmartinm how do we validate them? Or it's simply because users can't suggest books?

inspirehep/modules/forms/validation_utils.py
jacquerie: `doi_syntax_validator` has no uses:
```shell
% ag doi_syntax_validator
inspirehep/modules/forms/validation_utils.py
161:def doi_syntax_validator(form, field):
```
while `DOISyntaxValidator` has several:
```shell
% ag DOISyntaxValidator
inspirehep/modules/forms/fields/doi.py
26:from ..validation_utils import DOISyntaxValidator
51:                DOISyntaxValidator(),

inspirehep/modules/forms/validation_utils.py
168:    warnings.warn("Please use DOISyntaxValidator instead.", DeprecationWarning)
169:    return DOISyntaxValidator()(form, field)
172:class DOISyntaxValidator(object):

inspirehep/modules/literaturesuggest/forms.py
44:from inspirehep.modules.forms.validation_utils import DOISyntaxValidator
234:        validators=[DOISyntaxValidator("The provided DOI is invalid - it should look similar to '10.1086/305772'."),
```
since it was already deprecated, why don't we remove it?

inspirehep/modules/forms/validators/simple_fields.py
jacquerie: Again, `isbn_syntax_validation` has no uses:
```shell
% ag isbn_syntax_validation
inspirehep/modules/forms/validators/simple_fields.py
45:def isbn_syntax_validation(form, field):
```

inspirehep/modules/forms/validation_utils.py
jmartinm: Yes, `DOISyntaxValidator` is the one we use. The other one came along when we moved to the overlay the code from the old `invenio-deposit` and can be removed.

inspirehep/modules/forms/validators/simple_fields.py
jmartinm: This one we might want to leave. As we don't currently have books in the suggestion form but is something that has been asked several times so once we add the `isbn` field we would make use of it. There is an old issue https://github.com/inspirehep/inspire-next/issues/861

jacquerie: Ok, this has been added back.

PR #1245, opened 2016-06-23
search: BAI search support

### LINKED ISSUE #1242: author search for BAI
As a follow up of https://github.com/inspirehep/inspire-next/pull/1222 I have seen that search for a given BAI currently returns no results. There is a related [x-failed test](https://github.com/inspirehep/inspire-next/blob/master/tests/unit/search/test_search_query.py#L645).

An exact query like author:"Ellis" produces the following ES query:

```
{
"bool": {
    "must": [
             {
                 "match": {
                     "authors.name_variations": "Ellis"
                 }
             }
         ],
         "should": [
             {
                 "match": {
                     "authors.full_name": "Ellis"
                 }
             }
         ]
     }
 }
```

I guess we need to add in the `must` query that either there is an exact match in `"authors.name_variations"` or there is an exact match in the [inspire_bai](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/mappings/records/hep.json#L319) field.

jmartinm:
* Adds support for matching author search by BAI. (closes #1242)

### DIFF
--- a/inspirehep/modules/search/walkers/elasticsearch.py
+++ b/inspirehep/modules/search/walkers/elasticsearch.py
@@ -118,12 +118,11 @@ class ElasticSearchDSL(object):
 
             fields = self.get_fields_for_keyword(keyword, mode='a')
             if fields == ['authors.full_name', 'authors.alternative_name']:
-                return Q({"bool":
-                          {"should": [
-                              {"match": {
-                                  "authors.name_variations": str(node.value)}},
-                              {"match": {"authors.full_name": str(node.value)}}]
-                           }})
+                return (
+                    Q("match", authors__name_variations=str(node.value)) |
+                    Q("match", authors__full_name=str(node.value)) |
+                    Q("match", authors__inspire_bai=str(node.value))
+                )
             return Q({
                 'multi_match': {
                     'query': node.value,
@@ -145,12 +144,12 @@ class ElasticSearchDSL(object):
         def query(keyword):
             fields = self.get_fields_for_keyword(keyword, mode='e')
             if fields == ['authors.full_name', 'authors.alternative_name']:
-                return Q({"bool":
-                          {"must": [
-                              {"match": {"authors.name_variations": str(node.value)}}],
-                           "should": [
-                              {"match": {"authors.full_name": str(node.value)}}]
-                           }})
+                return (
+                    Q("term", authors__name_variations=str(node.value)) |
+                    Q("term", authors__inspire_bai=str(node.value)) |
+                    Q("match", authors__full_name=str(node.value))
+                )
+
             if (len(fields) > 1):
                 return Q({"bool":
                           {"should": [{"term": {k: str(node.value)}}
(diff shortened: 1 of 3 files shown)


jmartinm:
Missing looking into `xfailed` tests and updating those.

jmartinm:
Ready for review. I will then rebase and merge.

tests/unit/search/test_search_query.py
```diff
@@ -115,6 +115,41 @@ def test_author_bai():
...
 
+def test_author_bai():
+    query = IQ('find a r.j.hill.1')
+
+    expected = {
+        "bool": {
+            "should": [
+              {
+                "bool": {
+                  "should": [
```
jacquerie: I wonder if this should be a `must` instead.

tests/unit/search/test_search_query.py
```diff
@@ -115,6 +115,41 @@ def test_author_bai():
...
+                }
+              },
+              {
+                "match": {
+                  "authors.full_name": "r.j.hill.1"
+                }
+              }
+            ]
+        }
+    }
```
Panos512: @jmartinm If this is what we were discussing the other day I think it should be like: 

```
    expected = {
        "bool": {
            "must": [
              {
                "bool": {
                  "should": [
                    {
                      "term": {
                        "authors.name_variations": "r.j.hill.1"
                      }
                    },
                    {
                      "term": {
                        "authors.inspire_bai": "r.j.hill.1"
                      }
                    }
                  ]
                }
              }],
              "should": [
              {
                "match": {
                  "authors.full_name": "r.j.hill.1"
                }
              }]
            
        }
    }
```
Since what we want is that inspire_bai `must` be in `either (should)` "authors.name_variations" or "authors.inspire_bai" and it `would also be good (should)` for it to be matching the "authors.full_name" .

tests/unit/search/test_search_query.py
```diff
@@ -186,10 +221,29 @@ def test_author_colon():
...
+                        "authors.full_name": "vagenas"
+                      }
+                    }
+                  ]
+                }
+              },
+              {
+                "match": {
+                  "authors.inspire_bai": "vagenas"
+                }
```
jacquerie: I also wonder why the ordering/nesting is different here: is the query produced non-deterministic?

inspirehep/modules/search/walkers/elasticsearch.py
```diff
@@ -145,12 +144,12 @@ def visit(self, node):
...
-                              {"match": {"authors.name_variations": str(node.value)}}],
-                           "should": [
-                              {"match": {"authors.full_name": str(node.value)}}]
-                           }})
+                return (
+                    Q("term", authors__name_variations=str(node.value)) |
+                    Q("term", authors__inspire_bai=str(node.value)) |
+                    Q("match", authors__full_name=str(node.value))
+                )
+
```
Panos512: I think this should be:

```
                    ( Q("term", authors__name_variations=str(node.value)) |
                    Q("term", authors__inspire_bai=str(node.value)) )&
                    Q("match", authors__full_name=str(node.value))
```

if `&` is the `and` symbol.

PR #1246, opened 2016-06-23
schemas: populate related_institutes

### LINKED ISSUE #1239: Institutions DoJSON - related institutes missing
Current DoJSON for Institutes doesn't use field `510` which contains information about related institution's `recid` (subfield `0`), `name` (`a`) and type of relation (`w`) (see all the fields explained [here](https://twiki.cern.ch/twiki/bin/view/Inspire/DevelopmentRecordMarkupInstitutions)).

Twiki documents four possible types of relation: `parent` (value `t`), `predecessor` (`a`), successor (`b`) and `otherwise related` (`r`).

Field `related_institutes` is already defined in [JSON Schema for Institutions](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/jsonschemas/records/institutions.json#L151) and seems to be good with two  exceptions:
 - it contains subfield `curated_relation` which seems to not have its equivalent in MarcXML
 - subfield `relation_type`, being an enum, should probably contain also value `otherwise_related`.

Last thing to be discussed: is it possible to get rid of the `successor` value in records XMLs? We could instead add (pred_recid, `predecessor`) pair inside successor's XML where `pred_recid` is its predecessor's recid. That needs though to edit legacy's XMLs.

@jacquerie @kaplun

jacquerie:
~~It's missing a few tests.~~

Closes #1239

### DIFF
--- a/inspirehep/dojson/institutions/fields/bd1xx.py
+++ b/inspirehep/dojson/institutions/fields/bd1xx.py
@@ -29,6 +29,7 @@ from dojson import utils
 from inspirehep.utils.dedupers import dedupe_list, dedupe_list_of_dicts
 
 from ..model import institutions
+from ...utils import get_record_ref
 from ...utils.geo import parse_institution_address
 
 
@@ -154,3 +155,27 @@ def historical_data(self, key, value):
     values.extend(el for el in value.get('a'))
 
     return values
+
+
+@institutions.over('related_institutes', '^510..')
+@utils.for_each_value
+def related_institutes(self, key, value):
+    """Related institutes."""
+    def _classify_relation_type(c):
+        if c == 'a':
+            return 'predecessor'
+        elif c == 'b':
+            return 'successor'
+        elif c == 't':
+            return 'parent'
+        elif c == 'r':
+            return 'other'
+        else:
+            return ''
+
+    return {
+        'curated_relation': bool(value.get('0')),
+        'name': value.get('a'),
+        'relation_type': _classify_relation_type(value.get('w')),
+        'record': get_record_ref(value.get('0'), record_type='institutions'),
+    }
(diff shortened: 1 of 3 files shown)


jacquerie:
First commit is unrelated, and fixes an xfailed test for a DoJSON rule for Institutions.
Second commit is the one closing #1239 (CC: @tomaszgy).

I'll wait for green and then :shipit:.

tomaszgy:
@jacquerie - thanks for fixing it! :smile:
I've one tiny doubt though concerning what @kaplun has said in https://github.com/inspirehep/inspire-next/issues/1239:

Imagine you have **Old Institution** with `old_institution_recid` being replaced by **New Institution** with `new_institution_recid`.
Now, we find inside the **Old Institution's record** reference to the **New Institution** being the `successor` of the **Old Institution**.

If I understood @kaplun correctly (please, confirm), on the legacy we're going to populate **New Institution record** with information that **Old Institution** is its `predecessor`.
Then, once it's done for all our record on legacy, in DoJSON we should ignore the fields with a value `successor` and only save the ones with `predecessor`. Thus, predecessors records wouldn't contain field `related_institutes`.

Is it correct?

jacquerie:
The DoJSON rules that I wrote are just doing a per-record dumb translation of the old data to the new data model. IMO it's a bad idea to put more logic (eg, code that modifies other records) than that in a DoJSON rule. That's what migrations are for.

kaplun:
What I am thinking is to deprecate the successor field, and silver the issue on legacy before migrating it.

jacquerie:
Yeah, that would be better. BTW, have you noticed that we could write migrations using DoJSON? Reversible migrations, even, by writing a forward rule and a backwards rule...

kaplun:
You mean there is official support for migrations? I.e. a place where to drop dojson rules that would be automatically called for migrating?

jacquerie:
Nope, not official. Just a random (a bit crazy) thought.

kaplun:
ah sure. Originally, when we thought about moving away from MARCXML one of the scenarios we had in mind was precisely upgreadability of the data model. (this was even before dojson). That's why the JSON Schema should in principle be versioned.

PR #1251, opened 2016-06-25
workflows: holding pen ui updates

eamonnmag:
* ADD Better error handling
* ADD Autocomplete on some fields
* ADD Support for new schemas
* ADD Better definition of keywords and field categories


inspirehep/config.py
```diff
@@ -1188,3 +1188,9 @@ def _(x):
     "stat.OT": "Other",
     "stat.TH": "Other"
 }
+
+# Import local config file if it is present.
+try:
+    from inspirehep.config_local import *
```
jacquerie: Not a fan of this... we already have `inspirehep.cfg` in the `static` folder, why the duplicate? It will just add one step to debugging.

eamonnmag: It's what I do in HEPData too. I think it's more obvious than having inspirehep.cfg in the var/instance dir since people normally forget about it.

jacquerie: Ah, yeah, you're right, I saw people forget about that file. 👍

jmartinm: And where is the `config_local` file? In Puppet for example we are filling the `inspirehep.cfg` file that goes into `var/instance`. When would we use the `config_local` file?

eamonnmag: When you're developing locally. E.g.use of assets debug. Setting the magpie
or beard endpoints etc.

jmartinm: Ok. I see it is added above `https://github.com/inspirehep/inspire-next/pull/1251/files/ae94900a7ca8c08f1ddb50f57244c5415da95919#diff-a084b794bc0759e7a6b77810e01874f2R6` in `.gitignore` I personally don't see the need for it and the least lines the better but if other people find it useful its ok.

inspirehep/config.py
```diff
@@ -846,7 +846,7 @@ def _(x):
 
 # Web services and APIs
 # =====================
-BEARD_API_URL = None   # e.g. "http://beard.inspirehep.net/api"
+BEARD_API_URL = "http://beard.inspirehep.net/api"
```
jmartinm: This supposed to be `None` by default as far as I know. In order for developers, Travis and so on to avoid querying the beard server every time and slowing down loading records.

inspirehep/modules/workflows/static/js/inspire_workflows_ui/templates/results.html
```diff
@@ -147,6 +159,20 @@
               </span>
           </div>
 
+          <br/>
+
+          <div if="!record._extra_data">
```
jmartinm: should it be `ng-if` ?

jmartinm:
Just had a look and looks good to me, except the minor comments should be ready

inspirehep/modules/workflows/static/js/inspire_workflows_ui/templates/results.html
eamonnmag: Yes! Indeed. Thanks!

eamonnmag:
@jmartinm @jacquerie - I'll leave it to someone else to merge. I am now looking at the submission timeline visualization for the dashboard.

jacquerie:
Ok, I will take care of merging this.

PR #1241, opened 2016-06-23
literaturesuggest: graceful page_nr

kaplun:
Signed-off-by: Samuele Kaplun <samuele.kaplun@cern.ch>

PR #1252, opened 2016-06-25
literaturesuggest: graceful page_nr

jacquerie:
* First commit adds unit tests for most of `inspirehep/modules/literaturesuggest/tasks.py`. This also uncovered a few bugs, documented by xfailed tests.
* Second commit is taken (and slightly modified) from #1241, which fixes a mistake introduced in #1236.

The style of testing used in the first commit can probably be applied to testing the other tasks in the codebase.

Closes #1241

### DIFF
--- a/tests/unit/literaturesuggest/test_literaturesuggest_tasks.py
+++ b/tests/unit/literaturesuggest/test_literaturesuggest_tasks.py
@@ -1,0 +1,695 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


jacquerie:
Ready for review.

tests/unit/literaturesuggest/test_literaturesuggest_tasks.py
```diff
@@ -0,0 +1,693 @@
...
+            'bar',
+        ],
+    }
+
+    obj = StubObj(data, extra_data)
+
+    expected = 'http://localhost:5000/schemas/records/hep.json'
+    result = formdata_to_model(obj, formdata)
+
+    assert expected == result['$schema']
```
jmartinm: So the `title` and `type_of_doc` that are in `formdata` do not produce anything in `result`?

jmartinm: Ok, I guess here you are only testing the `$schema` part

jmartinm:
Looks ok to me 👍

tests/unit/literaturesuggest/test_literaturesuggest_tasks.py
jacquerie: They do produce something in `result`, but I'm not testing for that. They have to be there, otherwise the method under test throws a `KeyError` or something like that.

Issue #1240, opened 2016-06-22
dojson: debug current DoJSON rules for conferences, experiments, jobs, and journals


jacquerie:
Spun from #1200, because it was growing too much in scope. #1200 will only be about institutions.
The work that was there has been preserved here: https://github.com/jacquerie/inspire-next/tree/test-dojson.

- [x] Institutions (#1200)
- [x] Jobs (#1253)
- [x] Journals (#1254)
- [x] Conferences (#1320)
- [x] Experiments (#1338)

PR #1253, opened 2016-06-26
tests: unit tests for DoJSON jobs

jacquerie:
Continues to address #1240.

- [x] Jobs
    - [x] Test `date_closed`
    - [x] Test `contact_details`
    - [x] Test `continent`
    - [x] Test `experiments`
    - [x] Test `institution`
    - [x] Test `description`
    - [x] Test `position`
    - [x] Test `ranks`
    - [x] ~~Test `experiment`~~ (obsolete rule)

### DIFF
--- a/inspirehep/dojson/jobs/fields/bd1xx.py
+++ b/inspirehep/dojson/jobs/fields/bd1xx.py
@@ -130,9 +130,3 @@ def ranks(self, key, value):
         for _rank in _ranks:
             self['_ranks'].append(_rank)
             self['ranks'].append(classify_rank(_rank))
-
-
-@jobs.over('experiment', '^693..')
-@utils.for_each_value
-def experiment(self, key, value):
-    return value.get('e')
--- a/inspirehep/dojson/jobs/fields/bd1xx.py
+++ b/inspirehep/dojson/jobs/fields/bd1xx.py
@@ -29,36 +29,41 @@ import six
 from dojson import utils
 
 from ..model import jobs
-from ...utils import classify_rank, get_record_ref
+from ...utils import classify_rank, force_single_element, get_record_ref
 
 
 @jobs.over('date_closed', '^046..')
 def date_closed(self, key, value):
     """Date the job was closed."""
-    value = utils.force_list(value)
-    closed_date = None
-    deadline_date = None
-    for val in value:
-        if val.get('i'):
-            deadline_date = val.get('i')
-        if val.get('l'):
-            if "@" in val.get('l'):
-                # Its an email
-                if "reference_email" in self:
-                    self["reference_email"].append(val.get('l'))
-                else:
-                    self["reference_email"] = [val.get('l')]
-            elif "www" in val.get('l') or "http" in val.get('l'):
-                # Its a URL
-                if "urls" in self:
-                    self["urls"].append({'value': val.get('l')})
-                else:
-                    self["urls"] = [{'value': val.get('l')}]
-            else:
-                closed_date = val.get('l')
+    def _contains_email(val):
+        return '@' in val
+
+    def _contains_url(val):
+        return 'www' in val or 'http' in val
+
+    el = force_single_element(value)
+
+    deadline_date = force_single_element(el.get('i'))
     if deadline_date:
         self['deadline_date'] = deadline_date
-    return closed_date
+
+    closing_date = None
+    closing_info = force_single_element(el.get('l'))
+    if closing_info:
+        if _contains_email(closing_info):
+            if 'reference_email' in self:
+                self['reference_email'].append(closing_info)
+            else:
+                self['reference_email'] = [closing_info]
+        elif _contains_url(closing_info):
+            if 'urls' in self:
+                self['urls'].append({'value': closing_info})
+            else:
+                self['urls'] = [{'value': closing_info}]
+        else:
+            closing_date = closing_info
+
+    return closing_date
 
 
 @jobs.over('contact_details', '^270..')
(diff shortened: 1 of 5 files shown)


inspirehep/dojson/jobs/fields/bd1xx.py
```diff
@@ -130,9 +130,3 @@ def ranks(self, key, value):
         for _rank in _ranks:
             self['_ranks'].append(_rank)
             self['ranks'].append(classify_rank(_rank))
-
-
-@jobs.over('experiment', '^693..')
-@utils.for_each_value
-def experiment(self, key, value):
-    return value.get('e')
```
kaplun: Why it disappears?

jacquerie: Because `experiment` is not a key in the schema: https://github.com/inspirehep/inspire-next/blob/303444d0e344d3a4c750d82c45c6094a6f224e05/inspirehep/modules/records/jsonschemas/records/jobs.json, and `693__e` is already consumed by rule `experiments`, which populates the right key `experiments`.

inspirehep/dojson/jobs/fields/bd1xx.py
```diff
@@ -51,9 +51,9 @@ def date_closed(self, key, value):
             elif "www" in val.get('l') or "http" in val.get('l'):
```
mihaibivol: Can you commit this snippet first in utils?
```
def force_single_element(obj):
    """Force an object to a list and return the first element."""
    lst = force_list(obj)
    if lst:
        return lst[0]
    return None
```

Just to make sure that the field 1 is never repeated

inspirehep/modules/records/jsonschemas/records/jobs.json
```diff
@@ -22,21 +22,17 @@
...
+            "type": "string",
+            "enum": [
+                "North America",
+                "Europe",
+                "Asia",
+                "South America",
+                "Australasia",
+                "Middle East",
+                "Africa",
+                "Europe / South America"
```
mihaibivol: what?

tests/unit/dojson/test_dojson_jobs.py
```diff
@@ -22,12 +22,82 @@
...
 
+def test_date_closed_from_046__i():
+    snippet = (
+        '<datafield tag="046" ind1=" " ind2=" ">'
+        '  <subfield code="i">2015-12-15</subfield>'
+        '</datafield>'
+    )  # record/1310294
+
+    expected = '2015-12-15'
+    result = strip_empty_values(jobs.do(create_record(snippet)))
```
mihaibivol: use `clean_record` instead of `strip_empty_values`. It should be a bit cleaner

mihaibivol: .. for everything else also :)

inspirehep/modules/records/jsonschemas/records/jobs.json
jacquerie: Oh, right, I cut/pasted the values from [here](https://github.com/inspirehep/inspire-next/blob/fec36228b9b06a1d2da18c2df825558f323fbc0a/inspirehep/dojson/current_marcxml_usage/jobs.txt#L45-L65), but I guess it's better to have a correct `enum` here and fix the data on Legacy.

mihaibivol: Would it make sense to do a translation in dojson for the continents. It's a handful of them anyway.

jacquerie: That's actually my argument for doing it manually — it's just a handful of them anyway.

inspirehep/dojson/jobs/fields/bd1xx.py
```diff
@@ -29,36 +29,41 @@
...
-                closed_date = val.get('l')
+    def _contains_email(val):
+        return '@' in val
+
+    def _contains_url(val):
+        return 'www' in val or 'http' in val
+
+    el = force_single_element(value)
+
+    deadline_date = el.get('i')
```
mihaibivol: I would also filter a single element out of `el['i']` and `el['1']`. I didn't check against repeatable fields but this should help a lot

inspirehep/dojson/utils/__init__.py
```diff
@@ -270,3 +270,11 @@ def split_page_artid(page_artid):
...
 
     return page_start, page_end, artid
+
+
+def force_single_element(obj):
+    """Force an object to a list and return the first element."""
+    lst = force_list(obj)
+    if lst:
+        return lst[0]
+    return None
```
mihaibivol: Can you randomly return a list though. The behavior of this function is too deterministic.

tests/unit/dojson/test_dojson_utils.py
```diff
@@ -127,6 +128,17 @@ def test_create_profile_url_returns_empty_string_when_not_given_an_int():
...
 
 
+def test_force_single_element_returns_first_element_on_a_list():
+    expected = 'foo'
+    result = force_single_element(['foo', 'bar', 'baz'])
+
+    assert expected == result
+
+
+def test_force_single_element_returns_none_on_empty_list():
```
mihaibivol: If you're writing this test, you can also test that forcing a single element out of a thing that is already a single element returns that element (e.g. `force_single_element('foo') == 'foo'`)

inspirehep/dojson/utils/__init__.py
jacquerie: `from random import random; return [] if random() > 0.5 else None`?

mihaibivol:
LGTM :+1: :ship: :bike: :bike: :bike:

PR #1254, opened 2016-06-26
tests: unit tests for DoJSON journals

jacquerie:
Continues to address #1240.

- [x] Journals
    - [x] Test `issn`
    - [x] Test `coden`
    - [x] Test `publisher`
    - [x] Test `titles`
    - [x] Test `short_titles`
    - [x] Test `title_variants`
    - [x] ~~Test `name_variants`~~ (obsolete rule)

### DIFF
--- a/inspirehep/dojson/journals/fields/bd1xx.py
+++ b/inspirehep/dojson/journals/fields/bd1xx.py
@@ -108,10 +108,3 @@ def title_variants(self, key, value):
     return {
         'title': value.get('a'),
     }
-
-
-@journals.over('name_variants', '^730..')
-@utils.for_each_value
-def name_variants(self, key, value):
-    """Variants of the name."""
-    return value.get('a')
--- a/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Journal_HTML_detailed.tpl
+++ b/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Journal_HTML_detailed.tpl
@@ -1,20 +1,23 @@
 {#
-## This file is part of INSPIRE.
-## Copyright (C) 2015 CERN.
-##
-## INSPIRE is free software; you can redistribute it and/or
-## modify it under the terms of the GNU General Public License as
-## published by the Free Software Foundation; either version 2 of the
-## License, or (at your option) any later version.
-##
-## INSPIRE is distributed in the hope that it will be useful, but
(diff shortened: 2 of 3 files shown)


jacquerie:
[This block](https://github.com/inspirehep/inspire-next/blob/1d3d2320ea11a7a947cda88a6283cead966f41ae/inspirehep/dojson/journals/fields/bd1xx.py#L57-L59) is not tested because it never happens on INSPIRE: https://inspirehep.net/search?ln=en&cc=Journals&p=022__b%3A%27hardcover%27&action_search=Search.

On the other hand, it sounded reasonable...

jacquerie:
Ready for review.

inspirehep/dojson/journals/fields/bd1xx.py
```diff
@@ -108,10 +108,3 @@ def title_variants(self, key, value):
     return {
         'title': value.get('a'),
     }
-
-
-@journals.over('name_variants', '^730..')
-@utils.for_each_value
-def name_variants(self, key, value):
-    """Variants of the name."""
-    return value.get('a')
```
kaplun: Is it normal that this disappeared?

kaplun:
Why are you removing name variants? They are the core functionality of the Journal DB.

inspirehep/dojson/journals/fields/bd1xx.py
jacquerie: `730__a` is already consumed by rule `title_variants`, which populates the field `title_variants`, which is in the schema, while `name_variants` isn't there: https://github.com/inspirehep/inspire-next/blob/303444d0e344d3a4c750d82c45c6094a6f224e05/inspirehep/modules/records/jsonschemas/records/journals.json

jacquerie: In any case, I'll add a bullet point to the commit message to say this.

inspirehep/modules/theme/templates/inspirehep_theme/format/record/Journal_HTML_detailed.tpl
```diff
@@ -112,7 +112,7 @@
       </div>
       <div class="modal-body">
         <div style="text-align:left;">
-        {{ record['name_variants']|join(', ')}}
+        {{ record['title_variants']|join(', ')}}
```
jacquerie: In fact, their leftover use had to be amended.

PR #1263, opened 2016-06-28
search: elasticsearch dsl fix for author queries

jmartinm:
* AMENDS 98a634797f8d996ee1540f71df4dc02fcce5543a

### DIFF
--- a/inspirehep/modules/search/walkers/elasticsearch.py
+++ b/inspirehep/modules/search/walkers/elasticsearch.py
@@ -118,11 +118,11 @@ class ElasticSearchDSL(object):
 
             fields = self.get_fields_for_keyword(keyword, mode='a')
             if fields == ['authors.full_name', 'authors.alternative_name']:
-                return (
-                    Q("match", authors__name_variations=str(node.value)) |
-                    Q("match", authors__full_name=str(node.value)) |
+                return Q('bool', should=[
+                    Q("match", authors__name_variations=str(node.value)),
+                    Q("match", authors__full_name=str(node.value)),
                     Q("match", authors__inspire_bai=str(node.value))
-                )
+                ])
             return Q({
                 'multi_match': {
                     'query': node.value,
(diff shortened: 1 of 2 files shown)


jmartinm:
@Panos512 @jacquerie I think this looks like what we want. Can you have a look before I merge?

jacquerie:
Yes, I like this version of the queries better.

BTW, do you think these tests add value, or is the overhead of mantaining them too high?

Panos512:
Yes, regarding the queries it looks great.

jmartinm:
> BTW, do you think these tests add value, or is the overhead of mantaining them too high?

So far they have not been very helpful. The only thing is that they make you stop and look at the queries so you might realize if something is wrong.

Apart from that what I have been doing to fix them is produce the query dict `query.to_dict()` and copy-pasting into the test.

I guess one ideal thing to have would be integration tests where we can have a query and the expected results back. So that when we check code we can see if we are returning less/more results. At the moment I cannot check that.

Panos512:
@jmartinm Regarding this test I am working on something similar now. I am trying to get the response of a given query to see if suggestion_messages are in it. So given a query we could check the length of the hits that are on the response.

jmartinm:
> @jmartinm Regarding this test I am working on something similar now. I am trying to get the response of a given query to see if suggestion_messages are in it. So given a query we could check the length of the hits that are on the response.

Yes, I wonder how difficult it will be to come up with the 'correct' number of search results that should be returned. For exact matches should be ok, for google-style might be a bit more difficult due to the fuzzy queries.

jacquerie:
> Yes, I wonder how difficult it will be to come up with the 'correct' number of search results that should be returned. For exact matches should be ok, for google-style might be a bit more difficult due to the fuzzy queries.

Exactly the reason I wanted to specify the queries produced, rather than the results, since those are harder to specify.

But if they don't add value I don't see the point of mantainining them.

jmartinm:
> Exactly the reason I wanted to specify the queries produced, rather than the results, since those are harder to specify.
 But if they don't add value then I don't see the point of mantainining them. 

I still think they have some value. It's good to see how a user query gets transformed into the final ES DSL. Maybe at some point we can reduce the tests as it seemed to me that some of them were repetitive so changing the code had to change quite a lot of them. This should not happen often though, the author queries are the most tested so had to amend lots of tests but usually it is not the case.

jacquerie:
Heh, on the other hand if you are pasting `query.body` then you are testing the implementation against itself, plus you're adding manual work on top : )

As long as we check what `query.body` is, and that it is a good query, then these tests provide some value. If we find that they are a schlep to mantain then my recommendation is to kill them.

PR #1265, opened 2016-06-28
fixtures: add users and permissions

jmartinm:


jmartinm:
With this, as part of `./scripts/recreate_records` we will have a demo user with a role and permission to access the admin interface.

inspirehep/modules/fixtures/users.py
```diff
@@ -0,0 +1,50 @@
...
+from invenio_db import db
+
+
+def init_users_and_permissions():
+    ds = current_app.extensions['invenio-accounts'].datastore
+    with db.session.begin_nested():
+        superuser_role = ds.create_role(name='superuser')
+        ds.create_user(
+            email='admin@inspirehep.net',
+            password=encrypt_password("123456"),
```
kaplun: Please, amend it to the more secure `dadada`.

PR #1267, opened 2016-06-28
schema: better thesis field for HEP

### LINKED ISSUE #1260: Cleaning up thesis field
`thesis` should not be repeatable: only the `thesis.university` field should be. (additionally the `thesis.record` field should be aligned with the `thesis.university`). And probably `thesis.university` should be called `thesis.institution`.

mihaibivol:
* Makes `thesis` field a single field instead of a list;
* Adds support for multiple institiutions and records for a given
  thesis.
* Closes #1260

### DIFF
--- a/inspirehep/dojson/hep/fields/bd5xx.py
+++ b/inspirehep/dojson/hep/fields/bd5xx.py
@@ -85,32 +85,37 @@ def hidden_note2marc(self, key, value):
 
 
 @hep.over('thesis', '^502..')
-@utils.for_each_value
 @utils.filter_values
 def thesis(self, key, value):
     """Get Thesis Information."""
-    curated_relation = False
-    if value.get('z'):
-        curated_relation = True
-    return {
+    res = {
         'defense_date': value.get('a'),
         'degree_type': value.get('b'),
-        'university': value.get('c'),
         'date': value.get('d'),
-        'record': get_record_ref(value.get('z'), 'institutions'),
-        'curated_relation': curated_relation,
     }
+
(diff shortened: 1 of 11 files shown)


mihaibivol:
Closes #1267

mihaibivol:
https://imgur.com/w4PS3FX

@kaplun. Still have to do hepcrawl

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -246,45 +246,48 @@
...
+            "properties": {
+                "date": {
+                    "type": "string",
+                    "format": "date"
+                },
+                "defense_date": {
+                    "type": "string",
+                    "format": "date"
+                },
+                "curated_relation": {
```
kaplun: Actually this flag should be part of institutions properties

jacquerie:
LGTM. Rebasing it for you.

inspirehep/dojson/hep/fields/bd5xx.py
```diff
@@ -85,32 +85,37 @@ def hidden_note2marc(self, key, value):
 
 
 @hep.over('thesis', '^502..')
-@utils.for_each_value
```
kaplun: Luckily there are no records on legacy with more than on 502 field, because otherwise you would have had a `list` of `dict`s rather than a `dict` in `value`, by removing this decorator.

mihaibivol: I checked against the field usage output :)

PR #1268, opened 2016-06-29
migrator: improvements

jmartinm:
* Truncates all record related tables and allows cleaning up ES.

### DIFF
--- a/inspirehep/modules/migrator/cli.py
+++ b/inspirehep/modules/migrator/cli.py
@@ -28,7 +28,7 @@ import os
 import sys
 
 from flask import current_app
-from flask.ext.script import prompt_bool
+from flask_cli import with_appcontext
 
 from invenio_db import db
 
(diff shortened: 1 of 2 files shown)


jmartinm:
This should facilitate recreating the records in QA.

inspirehep/modules/migrator/cli.py
```diff
@@ -126,39 +126,54 @@ def remove_legacy_tables():
...
-            db.engine.execute("TRUNCATE TABLE {0}".format(table))
-            print(">>> Truncated {0}".format(table))
 
-        db.engine.execute("DELETE FROM pidSTORE WHERE pid_type='recid'")
-        print(">>> Truncated pidSTORE WHERE pid_type='recid'")
+        click.secho('Truncating tables...', fg='red', bold=True,
+                    file=sys.stderr)
+        with click.progressbar(tables_to_truncate) as tables:
+            for table in tables:
+                db.engine.execute("TRUNCATE TABLE {0} CASCADE".format(table))
```
kaplun: You should add `RESTART IDENTITY` as per https://www.postgresql.org/docs/current/static/sql-truncate.html

Otherwise the sequences are not going to be reinitialized.

inspirehep/modules/migrator/cli.py
```diff
@@ -126,39 +126,54 @@ def remove_legacy_tables():
...
+                click.secho("\tTruncated {0}".format(table))
 
-        db.engine.execute("SET FOREIGN_KEY_CHECKS=1;")
         db.session.commit()
+
+        if click.confirm("Do you want to remove records from ElasticSearch?"):
```
kaplun: I don't think there is ever going to be a situation where you truncate DB but not ES, so probably this second confirmation step can be avoided :)

Issue #1257, opened 2016-06-28
Simplifying titles


kaplun:
Titles are not used for searching (i.e. nobody searches for an exact title). What they are useful for is to convey concepts that might be searched for. 

For this reason it is proposed to:
* [x] kill old titles (they are used only 6 times!!! and 1 out of 6 is a DOI instead of a title)
* [x] kill title variations (which contains only slight variation, e.g.:
  * a LaTeX formula expressed in unicode
  * an acronym
  * and -> & and other useless substitutions
* [ ] LaTeX to unicode can be automatized
* [ ] Acronyms can be computed directly by stripping stop words and considering upper-case words (anyway they will be there in the abstract and can be boosted, i.e. receive more weight when searched for it).
* [ ] we could provide a synonym table to expand known acronym and symbols.

kaplun:
cc: @inspirehep/inspire-dir

ksachs:
I'm not sure what you are up to here.
We need a title 'history', i.e. different versions from arXiv, journal, what have you.
A synonym might have different meaning in math, astro or hep. going only with a table you will create false positives or miss some variants. we might decide another field is not worth the effort but we should discuss it.

kaplun:
> I'm not sure what you are up to here.
> We need a title 'history', i.e. different versions from arXiv, journal, what have you.

@ksachs, why do you need a title history? Since we don't display them, I assume it's only for searching. Now:
* real users do not search really for titles. However they might search a la Google. In that case, in the modern world (so not with old Invenio exact-only search-engine), the search engine is able to search in approximate way, using synonyms etc. and displaying at the top documents that are more likely been searched by the user.
* Note that: we still plan to store both arXiv and publisher version explicitly. However not to store all the past versions from arXiv, or all the past versions from a publisher

> A synonym might have different meaning in math, astro or hep. going only with a table you will create false positives or miss some variants.

* Re synonyms, I am not proposing to go only with a table, but to additionally expand with synonyms: we have a modern search engine now and we can boost exact matches, and also allow for less boosted synonym matches. The final goal when using phrase fields such as title, abstract, fulltext, is to present on top the record the user was searching for, not to match all and only the records the user might have been looking (for that we would still use the normal fields such as category, subject, experiments, identifiers, etc.), i.e. the false positive will be there no matter what, but relegated to the bottom of the search engine.

> we might decide another field is not worth the effort but we should discuss it.

That's the very purpose of this ticket :+1: 
[...]

PR #1291, opened 2016-07-05
jsonschemas: schema clean up

jmartinm:
* Removes fields only used on ElasticSearch and not part of the record
  data model.

### DIFF
--- a/inspirehep/dojson/common/base.py
+++ b/inspirehep/dojson/common/base.py
@@ -25,6 +25,7 @@
 from __future__ import absolute_import, division, print_function
 
 from dojson import utils
+from dojson.errors import IgnoreKey
 
 from ..conferences.model import conferences
 from ..experiments.model import experiments
@@ -143,38 +144,37 @@ def creation_modification_date2marc(self, key, value):
 @utils.ignore_value
 def spires_sysnos(self, key, value):
     """Old SPIRES number and new_recid from 970."""
+    external_system_numbers = self.get('external_system_numbers', [])
     value = force_force_list(value)
-    sysnos = []
     new_recid = None
     for val in value:
         if 'a' in val:
-            # Only append if there is something
-            sysnos.append(val.get('a'))
+            external_system_numbers.append(
+                {
+                    "institute": "SPIRES",
+                    "value": val.get('a'),
+                    "obsolete": True
+                }
+            )
         elif 'd' in val:
             new_recid = val.get('d')
     if new_recid is not None:
-        # FIXME we are currently using the default /record API. Which might
-        # resolve to a 404 response.
         self['new_record'] = get_record_ref(new_recid)
-    return sysnos or None
+
+    self['external_system_numbers'] = external_system_numbers
 
 
-@hep2marc.over('970', '(spires_sysnos|new_record)')
-@hepnames2marc.over('970', '(spires_sysnos|new_record)')
+@hep2marc.over('970', 'new_record')
+@hepnames2marc.over('970', 'new_record')
 def spires_sysnos2marc(self, key, value):
     """970 SPIRES number and new recid."""
     value = force_force_list(value)
     existing_values = self.get('970', [])
 
-    if key == 'spires_sysnos':
-        existing_values.extend(
-            [{'a': val} for val in value if val]
-        )
-    elif key == 'new_record':
-        val_recids = [get_recid_from_ref(val) for val in value]
-        existing_values.extend(
-            [{'d': val} for val in val_recids if val]
-        )
+    val_recids = [get_recid_from_ref(val) for val in value]
+    existing_values.extend(
+        [{'d': val} for val in val_recids if val]
+    )
     return existing_values
 
 
--- a/inspirehep/modules/records/jsonschemas/records/hep.json
+++ b/inspirehep/modules/records/jsonschemas/records/hep.json
@@ -31,10 +31,6 @@
                     "experiment": {
                         "type": "string"
                     },
-                    "facets_experiment": {
-                        "description": "Experiment name to be used on facets.",
-                        "type": "string"
-                    },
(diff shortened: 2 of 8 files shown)


jmartinm:
I have some questions for other fields. I am thinking here from the perspective of the new [record editor](https://github.com/inspirehep/record-editor) were we would use the schema as a template to fill the record so having too many separate fields seems a bit confusing.

* [ ] `spyres_sysnos` - [see schema](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/jsonschemas/records/hep.json#L106-L113) does not seem to be [much used](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/dojson/current_marcxml_usage/hep.txt#L8782-L8818). Shall we instead use [persistent identifiers](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/jsonschemas/records/hep.json#L972) or [external_system_numbers](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/jsonschemas/records/hep.json#L949) ?

* [ ] [titles_old](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/jsonschemas/records/hep.json#L149). Is this something we want to have as separate field? Do we want, for example, the editor to show that field as something to fill in? Or it would be better as part of titles (with a flag or comment saying that it is old)

* [ ] [title_variations](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/jsonschemas/records/hep.json#L5250). Same as above, is there a good use on keeping it separate? Wouldn't it be more natural to edit everything in the `titles` field ?

* [ ] [hidden_note CDS](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/jsonschemas/records/hep.json#L249) - [see usage](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/dojson/current_marcxml_usage/hep.txt#L4917) , shall we move the content the `source` subfield writing `CERN` or `CDS`

jmartinm:
@inspirehep/inspire-content

jacquerie:
Do `titles_old` and `title_variations` fall under #1257, which wasn't warmly welcomed by @ksachs and @fschwenn?

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -654,12 +641,6 @@
                         "description": "Just in case...",
                         "title": "Other IDs"
                     },
-                    "uuid": {
-                        "title": "UUID",
-                        "description": "Universally unique identifier of the author.",
-                        "type": "string",
-                        "pattern": "[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}"
-                    },
```
kaplun: Actually this should be in the schema. It should not be touched by catalogers though. This is added on `before_record_update`/`_inser`, so it goes directly into the DB.

kaplun:
Re: #1257, as long as we preserve all titles, they are fine (e.g. we came IRL to the proposal of storing all titles as a list, where anything that is not the first title is either an alternative or older title.) The use case of old_titles is for retrievability so having them in the list without them being displayed is ok. Plus @mihaibivol says he can devise a merging algorithm that puts olders value at the end of a list.

kaplun:
>  spyres_sysnos - see schema does not seem to be much used. Shall we instead use persistent identifiers or external_system_numbers ?

I think we can kill this. Does it has any use at all nowadays?

> hidden_note CDS - see usage , shall we move the content the source subfield writing CERN or CDS

:+1:

inspirehep/modules/records/jsonschemas/records/hep.json
jacquerie: Ok, then I'm readding a WIP tag and we can wait for @jmartinm to be back. If something else is blocked by this (I don't see how) I'll resurrect and integrate.

jmartinm:
I will take care of this PR today for the fields without doubts and create a different issue with the pending ones.

inspirehep/modules/records/jsonschemas/records/hep.json
jmartinm: 👍

jmartinm:
Ready for review. Have not touched titles for the moment.

inspirehep/dojson/common/base.py
```diff
@@ -143,38 +144,41 @@ def creation_modification_date2marc(self, key, value):
...
+                    "institute": "SPIRES",
+                    "value": val.get('a'),
+                    "obsolete": True
+                }
+            )
         elif 'd' in val:
             new_recid = val.get('d')
     if new_recid is not None:
         # FIXME we are currently using the default /record API. Which might
         # resolve to a 404 response.
```
jacquerie: I believe this is no longer true?

inspirehep/dojson/common/base.py
```diff
@@ -143,38 +144,41 @@ def creation_modification_date2marc(self, key, value):
...
         self['new_record'] = get_record_ref(new_recid)
-    return sysnos or None
 
+    self['external_system_numbers'] = external_system_numbers
 
-@hep2marc.over('970', '(spires_sysnos|new_record)')
-@hepnames2marc.over('970', '(spires_sysnos|new_record)')
+    raise IgnoreKey
```
jacquerie: Doesn't `@utils.ignore_value` at the top already achieve this?

jacquerie:
LGTM (but I haven't tested). Left a couple of questions.

inspirehep/dojson/common/base.py
jmartinm: Looking at the code of that utility it seems to be the same indeed.

jmartinm:
Comments addressed and rebased.

PR #1299, opened 2016-07-07
docs:Update the install docs

david-caro:
* Add some missing steps
* Add centos examples
* Add user admin access grant

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -2,6 +2,46 @@
 Installation
 ============
 
+System prerequisites
+=====================
+
+This guide expects you to have installed in your system the following tools:
+
+* pip
+* git
+* virtualenv
+* virtualenvwrapper
+* npm > 3.0
+* postgresql + devel headers
+* libxml2 + devel headers
+* libxslt + devel headers
+* ImageMagick
+* redis
+* elasticsearch
+
+CentOS
+-------
+.. code-block:: console
+
+    $ sudo yum install python-pip python-virtualenv python-virtualenvwrapper \
+        npm postgresql postgresql-devel libxml2-devel ImageMagick redis git \
+        libxslt-devel
+    $ sudo npm -g install npm
+
+For elasticsearch you can find the installation instructions on the
+`elasticsearch install page`_, and, to run the development environment, you
+will need also add the following workarounds:
+
+.. code-block:: console
+
+    $ sudo usermod -a -G $USER elasticsearch
+    $ newgrp elasticsearch  # or log out and in again
+    $ sudo ln -s /etc/elasticsearch /usr/share/elasticsearch/config
+
+.. TODO::
+
+    Add also Ubuntu and MacOS
+
 Create a virtual environment
 ============================
 
@@ -13,18 +53,57 @@ Create a virtual environment and clone the INSPIRE source code using `git`:
     $ workon inspirehep
     (inspirehep)$ cdvirtualenv
     (inspirehep)$ mkdir src
-    (inspirehep)$ git clone https://github.com/inspirehep/inspire-next.git inspirehep
+    (inspirehep)$ git clone https://github.com/inspirehep/inspire-next.git src/inspirehep
 
 
 Install requirements
 ====================
 
-Use `pip` to install all requirements:
+Use `pip` to install all requirements, it's recommended to upgrade pip and
+setuptools to latest too:
 
 .. code-block:: console
 
+    (inspirehep)$ pip install --upgrade pip setuptools
     (inspirehep)$ cdvirtualenv src/inspirehep
     (inspirehep)$ pip install -r requirements.txt --pre --exists-action i
+
+And for development:
+
+.. code-block:: console
+
+    (inspirehep)$ pip install -e .[development]
+
+
+Custom configuration and debug mode
+====================================
+
+If you want to change the database url, or enable the debug mode for
+troubleshooting, you can do so in the `inspirehep.cfg` file under
+`var/inspirehep-instance`, you might need to create it:
+
+.. code-block:: console
+
+    (inspirehep)$ cdvirtualenv var/inspirehep-instance
+    (inspirehep)$ vim inspirehep.cfg
+
+
+There you can change the value of any of the variables that are set under the
+file `src/inspirehep/inspirehep/config.py`, for example:
+
+
+.. code-block:: python
+
+    DEBUG = True
+    SQLALCHEMY_DATABASE_URI = "postgresql+psycopg2://someuser:somepass@my.postgres.server:5432/inspirehep"
+
+
+.. note::
+
+    Make sure that the configuration keys you override here have the same exact
+    name as the ones in the config.py file, as it will not complain if you put
+    a key that did not exist.
+
 
 
 Build assets
@@ -34,7 +113,16 @@ We build assets using `npm`. Make sure you have installed it system wide.
 
 .. code-block:: console
 
-    (inspirehep)$ npm update && npm install -g node-sass clean-css requirejs uglify-js
+    (inspirehep)$ sudo npm update
+    (inspirehep)$ sudo npm install -g node-sass clean-css requirejs uglify-js
+
+
+.. note::
+
+    If you don't want to use sudo to install the npm packages globally, you can
+    still setup a per-user npm modules installation that will allow you to
+    install/remove modules as normal user. You can find more info `in the npm
+    docs here`_.
 
 
 Then we build the INSPIRE assets:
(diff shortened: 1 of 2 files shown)


INSTALL.rst
```diff
@@ -13,28 +48,42 @@ Create a virtual environment and clone the INSPIRE source code using `git`:
...
 Build assets
 ============
 
 We build assets using `npm`. Make sure you have installed it system wide.
 
+.. note::
+
+    You will have to install a specific version of clean-css until `this issue`_
+    is fixed
+
```
eamonnmag: Moving the CSS in to the bundle rather than using as an import solves this, so this problem should now be fixed after the changes made yesterday evening.

david-caro: Forgot to remove it :see_no_evil:

INSTALL.rst
```diff
@@ -2,6 +2,41 @@
 Installation
 ============
 
+System prerequisites
+=====================
+
+This guide expects you to have installed is your system the following tools:
```
jmartinm: in your system*

INSTALL.rst
```diff
@@ -2,6 +2,41 @@
...
+
+    $ sudo yum install python-pip python-virtualenv python-virtualenvwrapper \
+        npm postgresql postgresql-devel lxml ImageMagick redis 
+    $ sudo npm -g install npm
+
+For elasticsearch you can find the installation instructions on the
+`elasticsearch install page`_, and, to run the development environment, you
+will need also add the following workarounds:
+
+.. code-block:: console
```
jmartinm: Note that this is only in CentOS ? At least on MacOS it was not needed, might be good making it clear.

INSTALL.rst
```diff
@@ -175,4 +239,33 @@ entry points and then do the following (e.g. for the `hep` rule):
...
 .. code-block:: console
 
-    dojson -l marcxml -i inspirehep/demosite/data/sample.xml do hep
+    (inspirehep)$ dojson -l marcxml -i inspirehep/demosite/data/sample.xml do hep
+
+
+Giving your user admin access
+==============================
+
+If you want access to the admin console, you'll have to execute:
```
jmartinm: Did you create a user after installation?

Note that there is a [fixture](https://github.com/inspirehep/inspire-next/blob/cd50b1cbcd028ba6a00d87d676ed0ed187a77382/inspirehep/modules/fixtures/users.py#L35) that creates an initial user with `superuser` role that should be able to access everything.

So either we mention that the user exists, or we teach how to create a new user (and then add the permissions you mention here)?

INSTALL.rst
```diff
@@ -175,4 +239,33 @@ entry points and then do the following (e.g. for the `hep` rule):
...
+
+    This feature is still unstable, and the name of the authorization access
+    might change, if you are having problems with it, you can check all the
+    access actions with:
+
+    .. code-block:: console
+
+        (inspirehep)$ inspirehep access list
+
+.. _this issue: https://github.com/inspirehep/inspire-next/issues/1296
```
jmartinm: What are these links? They don't seem to render in the final docs?

INSTALL.rst
david-caro: that's generic, you'll need those tools in any system where you plan to run the dev environment, were they installed by default or not. My (CentOS) system specifics are below.

INSTALL.rst
david-caro: See that at line 21 it states it's an example for CentOS, should I put it in a note block instead?

INSTALL.rst
david-caro: In the guide it tells you to create that user, I just followed the guide, line 73, has it's own section.

I was unaware of that fixture, is it run by default on install? I can replace the user creation section with this.

INSTALL.rst
david-caro: They do, just only on pythonhosted (or readthedocs). It's sphinx (restructured text) syntax, and github does not like it, I can try to investigate if it's possible to please both systems in any way

INSTALL.rst
david-caro: Was that merged? I'm not seeing it in the code, I still see the include in inspirehep/modules/theme/static/scss/holding-pen/holding-pen.scss and It's missing also from the bundles.py file (and it fails to build assets).

eamonnmag: We did it on your machine. So it can be your first code contribution 💃

david-caro: sent :smile: #1302

david-caro:
I'm retesting with a bare centos docker container, will update if any other package requirements come up

david-caro:
Now it's ready imo

david-caro:
Dammit... no it's not... the todo extension is not added by default to sphinx :/, I'll add it

david-caro:
Done, let me know if you don't want todos there, I can just remove it and leave a comment or something.

INSTALL.rst
```diff
@@ -13,19 +53,58 @@ Create a virtual environment and clone the INSPIRE source code using `git`:
...
+
+    (inspirehep)$ pip install -e .[development]
+
+
+Custom configuration and debug mode
+====================================
+
+If you want to change the database url, or enable the debug mode for
+troubleshooting, you can do so in the `inspirehep.cfg` file under
+`var/inspirehep`, you might need to create it:
```
jmartinm: Folder is `var/inspirehep-instance/`

INSTALL.rst
```diff
@@ -13,19 +53,58 @@ Create a virtual environment and clone the INSPIRE source code using `git`:
...
+
+.. code-block:: console
+
+    (inspirehep)$ cdvirtualenv var/inspirehep
+    (inspirehep)$ vim ispirehep.cfg
```
jmartinm: :s/ispirehep/inspirehep/g

inspirehep/config.py
```diff
@@ -61,7 +61,7 @@ def _(x):
 # ========
 SQLALCHEMY_DATABASE_URI = os.environ.get(
     "SQLALCHEMY_DATABASE_URI",
-    "postgresql+psycopg2://inspirehep:dbpass123@localhost:5432/inspirehep")
+    "postgresql+psycopg2://inspirehep:dbpass123@172.17.0.2:5432/inspirehep")
```
jmartinm: This should not be commited?

INSTALL.rst
```diff
@@ -34,7 +113,8 @@ We build assets using `npm`. Make sure you have installed it system wide.
 
 .. code-block:: console
 
-    (inspirehep)$ npm update && npm install -g node-sass clean-css requirejs uglify-js
+    (inspirehep)$ sudo npm update
+    (inspirehep)$ sudo npm install -g node-sass clean-css requirejs uglify-js
```
jacquerie: FYI `npm install -g` does not require `sudo` using the following configuration: https://docs.npmjs.com/getting-started/fixing-npm-permissions

david-caro: That does not look really like 'fixing', it's changing the ownership of a system directory... though the second option there (per-user "global" npm directory is a nice option).

In any case, this is meant to work on a fresh install, with the default permissions. I can add a note pointing to that page if you want. Will have to check if the solution is generic or macos bound tohugh. Btw. there's already a todo message to add ubuntu and macos examples too, might be a subtask of that one

jacquerie: Ah, yes, the solution I wanted to point to was the second one, not the ugly `chown`. I _believe_ the official NPM docs discourage using `sudo` in that situation, so I don't think we should recommend that in our installation instructions.

tiborsimko: FWIW wouldn't it be easier to lead `npm` installation related discussions in upstream Invenio, and then simply tell the INSPIRE developers (or CDS, or Zenodo...) in the overlay instructions to run the appropriate `provision-web.sh --with-foo --with-bar`, or `provision-elasticsearch.sh` etc scripts, in order to have all the pre-requisites installed? Might be easier to maintain instructions up-to-date in this way.

david-caro: I was not aware that those scripts exist, can you point me to them?

david-caro: Found them :smile: (thanks @jacquerie), though I don't think it should be added to this pull request right now (it's not just executing a command, but you have to pull in the invenio scripts and setup env vars and such). I think that should go on it's own pr, once we (aka me) have a clear idea of the setup process.

david-caro: @jacquerie: I've added a note and a link to that npm docs page, let me know if that's enough (I don't think I should copy the whole stuff, and well, I think they deserve a visit for their info ;) )

INSTALL.rst
```diff
@@ -13,19 +53,58 @@ Create a virtual environment and clone the INSPIRE source code using `git`:
...
 
-Use `pip` to install all requirements:
+Use `pip` to install all requirements, it's recommended to upgrade pip and
+setuptools to latest too:
 
 .. code-block:: console
 
+    (inspirehep)$ pip install --upgrade pip setuptools
     (inspirehep)$ cdvirtualenv src/inspirehep
     (inspirehep)$ pip install -r requirements.txt --pre --exists-action i
```
jacquerie: We can probably get rid of the `--exists-action i` now... it's an historical artifact of the Invenio/INSPIRE overlay concept.

INSTALL.rst
```diff
@@ -13,19 +53,58 @@ Create a virtual environment and clone the INSPIRE source code using `git`:
...
 
+And for development:
+
+.. code-block:: console
+
+    (inspirehep)$ pip install -e .[development]
+
+
+Custom configuration and debug mode
+====================================
```
jacquerie: Note also that commit 8dacc07d7 introduced another way of overriding configuration via a file called `inspirehep/config_local.py`. I'm not sure it's worth mentioning in the docs, I believe only @eamonnmag uses this variant.

jacquerie:
LGTM. :ship:

PR #1301, opened 2016-07-07
HAL validator

jmartinm:


jmartinm:
A number of steps are needed before deploying this code.

* [ ] Create french cataloger role and attach `usedeposit` action.
* [ ] Assign all catalogers to this role.

PR #1302, opened 2016-07-07
holdingpen: cleancss include error

### LINKED ISSUE #1296: Pin down clean-css version
Investigating with @eamonnmag and @jmartinm, we found out that the newer clean-css (3.4.18) breaks the assets build step, so the version of clean-css should be pinned down (last known working version being 3.4.9).

Probably remove from puppet installation and move to the build step if possible.

david-caro:
Fix the current issue when using a new cleancss module that fails to
import the css file. Moved it to a bundle instead. (fixes #1296)

### DIFF
--- a/inspirehep/modules/theme/bundles.py
+++ b/inspirehep/modules/theme/bundles.py
@@ -100,6 +100,7 @@ detailedjs = NpmBundle(
 )
 
 holding_pen_css = NpmBundle(
+    "node_modules/angular-xeditable/dist/css/xeditable.css",
     "scss/holding-pen/holding-pen.scss",
     filters="scss, cleancss",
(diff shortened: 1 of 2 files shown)


inspirehep/modules/theme/bundles.py
```diff
@@ -100,8 +100,9 @@
 )
 
 holding_pen_css = NpmBundle(
+    "node_modules/angular-xeditable/dist/css/xeditable.css",
     "scss/holding-pen/holding-pen.scss",
     filters="scss, cleancss",
     output="gen/inspirehep.holding.%(version)s.css",
-    depends="scss/**/*.scss"
+    depends="scss/*/.scss"
```
david-caro: This does work, but I'm not sure if the glob is correct

PR #1303, opened 2016-07-07
dojson: faster dedupes

mihaibivol:
* Currently `dedupe_lists` is called on the full list, every time a record is appended. (Dedupe is n^2 or some heavy O(n)) Multiply this with the length of the list and we get to unnecessary n^3 or n^2. Would be cool to have the `freeze` + `set` logic but that would be some massive hack which sets function attributes.


mihaibivol:
@jmartinm This doesn't make the tests faster, but for migrating the legacy records + their history it makes a __huge__ difference

PR #1305, opened 2016-07-08
WIP setup: add flask-shell-ipython dependency

jmartinm:
* Allows `$ inspirehep shell` to use Ipython instead of default
  Python shell.


jacquerie:
This merge conflict really makes me think that we need a commit that applyes some rules for the order of stuff in `setup.py`.

setup.py
```diff
@@ -31,7 +31,13 @@
 
 install_requires = [
     'amqp>=1.4.9,<2.0',
+    'Flask>=0.11.1',
+    'Flask-Admin>=1.4.2',
+    'Flask-Breadcrumbs>=0.3.0',
+    'Flask-Script>=2.0.5',
+    'Flask-CLI>=0.4.0',
```
jirikuncar: do not use Flask-CLI with Flask>=0.11.1

jirikuncar: do not use Flask-CLI with Flask>=0.11.1

jacquerie: It's required by a lot of other packages anyway, `invenio-search` included: https://github.com/inveniosoftware/invenio-search/blob/0fc8f760fa3f5abb47e02eaa70735778cdc9fc3f/setup.py#L71

jirikuncar: > It's required by a lot of other packages anyway

Yes, but we are going to remove this dependency and your overlay should not depend on it directly. The same applies to most of Flask* dependencies in your setup.py. Pin only top-level dependencies and use `pip-tools` to compile your full production requirements.

jacquerie: Yes, in fact I resubmitted this patch without a dependency on Flask-CLI: https://github.com/inspirehep/inspire-next/pull/1319#issue-165122304. Our `setup.py` is without doubt full of scar tissue.

PR #1306, opened 2016-07-08
dojson: dedupe and clean once per record

mihaibivol:
* Adds `dedupe_all_lists` function
* Calls `dedupe_all_lists` for any `dojson` output

Signed-off-by: Mihai Bivol <mm.bivol@gmail.com>

### DIFF
--- a/inspirehep/dojson/utils/__init__.py
+++ b/inspirehep/dojson/utils/__init__.py
@@ -39,6 +39,8 @@ from inspirehep.config import (
     INSPIRE_RANK_TYPES,
 )
 
+from inspirehep.utils.dedupers import dedupe_list, dedupe_list_of_dicts
+
 
 _RE_2_CHARS = re.compile(r"[a-z].*[a-z]", re.I)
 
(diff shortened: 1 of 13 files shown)


mihaibivol:
Let the build be fast

mihaibivol:
@jacquerie @kaplun I left the `processors.py` file there since it's a good centralized point in which one can find generic dojson conversion functions.

inspirehep/dojson/utils/__init__.py
```diff
@@ -196,21 +198,50 @@ def encode_for_marcxml(value):
...
     """Recursively strips empty values."""
     if isinstance(obj, dict):
-        for key, value in obj.items():
-            value = strip_empty_values(value)
-            if value or value is False or value == 0:
-                obj[key] = value
+        for key, val in obj.items():
+            new_val = strip_empty_values(val)
+            # There are already lots of leaks in the callers of this function,
+            # there's no need to make it idempotent and double memory.
```
kaplun: What do you mean?

jacquerie: Same question as @kaplun. AFAICT, `value or value is False or value == 0` is `True` if it is a truthy value or a falsy value different from `[]`, `{}`, and `None`. I believe you are changing behaviour here.

mihaibivol: It would be more "correct" to create a new dict rather than reusing the old one. But this might leave memory used by the old one and also be a bit slower.

inspirehep/dojson/utils/__init__.py
```diff
@@ -196,21 +198,50 @@ def encode_for_marcxml(value):
...
+        return obj or None
+    elif isinstance(obj, (list, tuple, set)):
+        new_obj = []
+        for val in obj:
+            new_val = strip_empty_values(val)
+            if new_val is not None:
+                new_obj.append(new_val)
+        return type(obj)(new_obj) or None
+    elif obj or obj is False or obj == 0:
+        return obj
```
jacquerie: Ooh, I see. Scrap that earlier comment.

inspirehep/dojson/utils/__init__.py
mihaibivol: @jacquerie that's why I return `obj or None` and `type(obj)(new_obj) or None`.  So it's the same behavior of returning `None` for everything that is either a truthy value or `0` or `False`

mihaibivol: Also `0` is already equal to `False` but it reads better this way :)

jacquerie:
LGTM, just a tiny rebase away from :ship:

PR #1312, opened 2016-07-10
setup: add pin on celery

jacquerie:
Build failed with: https://travis-ci.org/inspirehep/inspire-next/jobs/143751233#L1147. Commit 8f19cb9b added a pin on `aqmp` without an explanation (but it was probably fixing a build). This PR tries to fix the current build failure by removing this pin. If this fails I'll instead add a pin on `kombu`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ readme = open('README.rst').read()
 
 install_requires = [
     'amqp>=1.4.9,<2.0',
+    'celery<4.0',
     'Flask-Gravatar>=0.4.2',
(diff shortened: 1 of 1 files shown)


jacquerie:
A pin on `celery` probably makes more sense. This will have to be merged first tomorrow morning, so that no build fails.

david-caro:
Looks ok, but the subject is wrong no? It's actually adding the pin?

jacquerie:
> Looks ok, but the subject is wrong no? It's actually adding the pin?

:dancer:

PR #1313, opened 2016-07-11
dojson: ValueError on split page_artid fix

mihaibivol:

### DIFF
--- a/inspirehep/dojson/utils/__init__.py
+++ b/inspirehep/dojson/utils/__init__.py
@@ -252,7 +252,11 @@ def split_page_artid(page_artid):
         if page_artid:
             if '-' in page_artid:
                 # if it has a dash it's a page range
-                page_start, page_end = page_artid.split('-')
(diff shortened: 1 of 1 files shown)


mihaibivol:
When migrating a production dump with dojson, half of the exceptions were caused by this.

inspirehep/dojson/utils/__init__.py
```diff
@@ -252,7 +252,11 @@ def split_page_artid(page_artid):
         if page_artid:
             if '-' in page_artid:
                 # if it has a dash it's a page range
-                page_start, page_end = page_artid.split('-')
+                page_range = page_artid.split('-')
+                if len(page_range) == 2:
+                    page_start, page_end = page_range
+                else:
+                    artid = page_artid
```
kaplun: I wonder what actual values there are on production. E.g. If it's 123- it's probably just a `start_page`. (-123 instead if of dubious meaning).

PR #1316, opened 2016-07-12
ci: pip use latest deps

david-caro:
* Upgrade any deps if available too, so we make sure to test with the
  latest version of the deps (and not just use whatever version was
  installed)
* This allows to discover version incompatibilities early in the
  process, before it hits production

### DIFF
--- a/docker-compose.deps.yml
+++ b/docker-compose.deps.yml
@@ -27,7 +27,7 @@ services:
     extends:
       file: services.yml
       service: base
-    command: pip-accel install -r requirements.txt --pre -e .[tests]
+    command: pip-accel install -r requirements.txt --upgrade --pre -e .[tests]
     volumes_from:
       - static
   assets:


jacquerie:
People have complained about build stability in the past, this will not make them very happy : )

kaplun:
Shall we rather copy Invenio's technique:
https://github.com/inveniosoftware/invenio-records/blob/master/.travis.yml#L63-L65

david-caro:
looks ok, but multiplies the number of tests by 3 (not sure though what's the diff between the 'pypi' and 'dev' deps lists though, we might only need one of them).

Are we willing to do so? (might mean more test run time and more probability of random failure, like net or something)

david-caro:
So according to the docs, 'min' is the minimal version that satisfies the deps, 'pypi' the  maximum ones, and 'dev' the ones from --pre, so technically, we are using 'dev' all the time, not sure if that will work for us, let me test it

david-caro:
dammit... the min thingie forces you to have a minimal version specifier on all the deps (we don't), the 'pypi' works well, but does not really change anything (same deps list as we have now) and the 'dev' one requires us to have a requirements.txt or similar file

kaplun:
What about setting the min version to today *status quo*, then from today onwards we should be good.

david-caro:
Will move the instability to the docker image build process, doing there the upgrade to latest and failing there if the tests don't pass so the image used for the tests has always a working set of pip packages installed :smiling_imp:

Issue #1287, opened 2016-07-01
Make workflow files compatible with invenio-records-files>=1.0.0a5


jmartinm:
Updating the following packages past their current version:

```
invenio-files-rest==1.0.0a3,
invenio-records-files==1.0.0a5
```

Makes `test_harvesting_arxiv_workflow_rejected` break. See [Travis job](https://travis-ci.org/inspirehep/inspire-next/jobs/141625472#L2094)

```python
  File "/code/inspirehep/modules/workflows/tasks/arxiv.py", line 62, in arxiv_fulltext_download
    arxiv_id=arxiv_id
  File "/code/inspirehep/utils/helpers.py", line 49, in download_file_to_record
    record.files[name] = req.raw
  File "/virtualenv/lib/python2.7/site-packages/invenio_records_files/api.py", line 94, in wrapper
    return method(self, *args, **kwargs)
  File "/virtualenv/lib/python2.7/site-packages/invenio_records_files/api.py", line 159, in __setitem__
    self.filesmap[key] = self.file_cls(obj, {}).dumps()
AttributeError: 'WorkflowFilesIterator' object has no attribute 'file_cls'
```

david-caro:
The change that introduces the issue on invenio-records-files is inveniosoftware/invenio-records-files@c46676d91ed66444981fb44d16849d51204d347f

looking

david-caro:
mered on #1317

PR #1317, opened 2016-07-12
global: invenio files packages upgrade

david-caro:
* Upgrade to latest invenio-records-files and invenio-files-rest
  packages

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,8 +62,8 @@ install_requires = [
     'invenio-search>=1.0.0a7',
     'invenio-records-rest>=1.0.0a15',
     'invenio-records-ui>=1.0.0a6',
-    'invenio-files-rest==1.0.0a3',
-    'invenio-records-files==1.0.0a5',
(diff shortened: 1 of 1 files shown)


jacquerie:
Uh? I don't understand how this solves #1287, just judging by the description of the bug...

BTW if you add a "Closes #1287" in the text of the PR or in one of the commit messages GitHub automatically closes the associated issue. More variations: https://help.github.com/articles/closing-issues-via-commit-messages/

david-caro:
It does not, not unless we merge #1316, it just does not fail if you use the lastest allowed version for all the deps, I'm trying to figure out exactly the packages that are needed to be upgraded for it to work and will bump all the minimal versions for all of them at the same time

david-caro:
Okok, so with the latest docker image this should be working well, and with the pip upgrade process moved to the docker image setup, this should be good for the future too, I'll rebase to make sure it still works ok and imo it can be merged

Issue #1282, opened 2016-07-01
Phonetic blocks from ES


kaplun:
`Beard` depends on the `fuzzy` python library for phonetic blocks computation. However this library seems abandoned and is not being ported to Python 3.

ES natively support the phonetic blocks such as `nysiis` used by INSPIRE: https://github.com/elastic/elasticsearch-analysis-phonetic

PR #1318, opened 2016-07-12
orcid: profile chooser

### LINKED ISSUE #1281: Identifying correct author signature upon sign up.
When an author signs up using orcid we should associate him with a signature if there is one presented on our records.

The steps we could follow to do that are:

1) When the author logins we query ORCiD for the records that are associated with him.
2) We map these records to the records that we have in our database (actually elasticsearch).
3) We query these records for signatures given the full name of the author.
          a) If there is one signature matching then that's his.
          b) If there are more than one signatures we select a number of records for each of these signatures (maybe 3) and we present them to the author, suggesting him to click his records. Given his answers we can identify the correct signature for him.
          c) If there are no signatures matching we repeat step 3 but for all our records instead of only the ones that correspond (given the informations ORCiD provides us) to the logged in author.

Panos512:
- Enhances sign up to identify possible profiles of signed-up author.
- Closes #1281.

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -28,6 +28,7 @@ import os
 import sys
 
 from invenio_oauthclient.contrib import orcid
+
 from invenio_records_rest.facets import terms_filter
 
 
@@ -1018,7 +1019,8 @@ INDEXER_BULK_REQUEST_TIMEOUT = float(120)
 
 # OAuthclient
 # ===========
-orcid.REMOTE_APP['params']['request_token_params'] = {'scope': '/orcid-profile/read-limited /activities/update /orcid-bio/update'}
+orcid.REMOTE_APP['params']['request_token_params'] = {
+    'scope': '/orcid-profile/read-limited /activities/update /orcid-bio/update'}
 OAUTHCLIENT_REMOTE_APPS = dict(
     orcid=orcid.REMOTE_APP,
 )
@@ -1030,6 +1032,8 @@ OAUTHCLIENT_ORCID_CREDENTIALS = dict(
 OAUTHCLIENT_SETTINGS_TEMPLATE = 'inspirehep_theme/page.html'
 
 ORCID_SYNCHRONIZATION_ENABLED = False
+
+ORCID_PROFILE_MATCHING_ENABLED = False
 
 # Error Pages
 # ========
--- a/inspirehep/modules/authors/receivers.py
+++ b/inspirehep/modules/authors/receivers.py
@@ -25,16 +25,19 @@
 from __future__ import absolute_import, division, print_function
 
 import json
+import requests
 import uuid
 
 from flask import current_app
-import requests
 
 from invenio_indexer.signals import before_record_index
+
 from invenio_records.signals import (
     before_record_insert,
     before_record_update,
 )
+
+from invenio_search.utils import schema_to_index
 
 from .utils import author_tokenize
 
@@ -45,22 +48,27 @@ def _query_beard_api(full_names):
     This method allows for computation of phonetic blocks
     from a given list of strings.
     """
-    headers = {
-        "Content-Type": "application/json",
-        "Accept": "application/json"
-    }
+    base_url = current_app.config.get('BEARD_API_URL')
 
-    text_endpoint = "{base_url}/text/phonetic_blocks".format(
-        base_url=current_app.config.get('BEARD_API_URL')
-    )
+    if base_url:
+        headers = {
+            "Content-Type": "application/json",
+            "Accept": "application/json"
+        }
 
-    response = requests.post(
-        url=text_endpoint,
-        headers=headers,
-        data=json.dumps({'full_names': full_names})
-    )
+        text_endpoint = "{base_url}/text/phonetic_blocks".format(
+            base_url=base_url
+        )
 
-    return response.json()['phonetic_blocks']
+        response = requests.post(
+            url=text_endpoint,
+            headers=headers,
+            data=json.dumps({'full_names': full_names})
+        )
+
+        return response.json()['phonetic_blocks']
+    else:
+        return {full_names[i]: '' for i in range(len(full_names))}
 
 
 @before_record_insert.connect
@@ -136,3 +144,9 @@ def generate_name_variations(recid, json, *args, **kwargs):
                     "output": name,
                     "payload": {"bai": bai[0] if bai else None}
                 }})
+    index, doc_type = schema_to_index(json['$schema'])
+    if doc_type == 'authors':
+        authors_name = json.get('name')
+        if authors_name:
+            name_variations = author_tokenize(authors_name['value'])
+            authors_name['name_variations'] = name_variations
--- a/inspirehep/modules/orcid/static/js/inspire_profile_matcher/templates/match_profile.html
+++ b/inspirehep/modules/orcid/static/js/inspire_profile_matcher/templates/match_profile.html
@@ -1,0 +1,84 @@
+<form  ng-controller="inspireProfileMatcherCtrl" ng-submit="postMatches()">
+
+  <div id="collection-landing-page-text">
+    <div class="row">
+      <div class="col-xs-12">
+          <h3 class="text-center">
+            Please confirm your authorship of {{vm.invenioSearchResults.hits.hits.length}} papers.
+          </h3>
+          <br>
+          <p class="text-muted text-center">
+             Please deselect any papers not authored by you.<br>
+             This will help our algorithm get better at spotting your papers in the future.
+          </p>
+      </div>
+    </div>
+  </div>
+  <article class="panel panel-default profile-matching header record-brief-literature">
+   <div class="panel-body">
+      <div class="row">
+        <div class="col-md-9 text-muted">
+    <h2 class="record-brief-title">Publication Details</h2>
+    </div>
+    <div class='col-md-3'>
+        <h2 class="record-brief-title text-muted text-center">Are you the author?</h2>
+    </div>
+  </div>
+  </div>
+</article>
+  <article class="panel panel-default profile-matching record-brief-literature" ng-repeat="record in vm.invenioSearchResults.hits.hits track by $index">
+    <hr>
+    <div class="panel-body">
+      <div class="row">
+        <div class="col-md-9">
+          <h2 class="record-brief-title">
+            <a ng-href="/literature/{{record.id}}" ng-bind-html='record.metadata.titles[0] | title'>
+            </a>
+          </h2>
+          <div class="record-brief-subtitle">
+            <span ng-bind-html='record.metadata.collaboration | collaboration : record.metadata.corporate_author'></span>
+              <span ng-bind-html='record.metadata.authors | corporateAuthor : record.metadata.corporate_author'></span>
+              <authors-affiliation authors-template="/static/node_modules/inspirehep-search-js/dist/templates/authors/authors.html" authors="{{record.metadata.authors}}" collaboration="{{record.metadata.collaboration}}" number-of-authors={{record.display.number_of_authors}} control-number="{{record.metadata.control_number}}"></authors-affiliation>
+              <span class="record-brief-muted" ng-bind-html="record.display.date | date : record.metadata.authors : record.metadata.collaboration : record.metadata.corporate_author"></span>
+          </div>
+          <div class="record-brief-details">
+              <div ng-if="record.display.publication_info_line.pub_info && record.metadata.dois" >
+                <div ng-if="record.display.publication_info_line.pub_info.length == 1 && record.metadata.dois.length == 1">
+                  <div ng-bind-html='record.metadata.dois | publicationInfoDoi : record.display.publication_info_line'></div>
+                </div>
+                <div ng-if="record.display.publication_info_line.pub_info.length !== 1 || record.metadata.dois.length !== 1">
+                  <div ng-bind-html='record.display.publication_info_line | publicationInfo' ></div>
+                  <div ng-bind-html='record.metadata.dois | doi' ></div>
+                </div>
+              </div>
+              <div ng-if="record.display.publication_info_line && !(record.display.publication_info_line.pub_info && record.metadata.dois)">
+                <div ng-bind-html='record.display.publication_info_line | publicationInfo' ></div>
+              </div>
+              <div ng-if="record.metadata.dois && !(record.display.publication_info_line && record.display.publication_info_line.pub_info)">
+                <div ng-bind-html='record.metadata.dois | doi'></div>
+              </div>
+              <div ng-bind-html='record.metadata.arxiv_eprints | arxivInfo'></div>
+          </div>
+        </div>
+        <div class="col-md-3">
+          <div class="onoffswitch col-centered">
+            <input type="checkbox" name="onoffswitch" class="onoffswitch-checkbox" id="myonoffswitch{{$index}}"
+                  ng-model="record.matched"
+                  ng-change="changeState(record)"
+                  checked>
+            <label class="onoffswitch-label" for="myonoffswitch{{$index}}">
+              <span class="onoffswitch-inner"></span>
+              <span class="onoffswitch-switch"></span>
+            </label>
+          </div>
+        </div>
+      </div>
+  </article>
+  </div>
+  <div class="profile-matching submition">
+    <h4 class = "profile-matching">
+      You have selected {{Matches.length}} of {{vm.invenioSearchResults.hits.hits.length}} papers as yours.
+    </h4>
+    <button type="submit" class="btn btn-primary profile-matching">Submit</button>
+  </div>
+</form>
--- a/inspirehep/modules/orcid/utils.py
+++ b/inspirehep/modules/orcid/utils.py
@@ -20,9 +20,21 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-from inspirehep.utils.bibtex import Bibtex
+from __future__ import absolute_import, division, print_function
+
+
+from flask import current_app, session
 
 from .schema import orcid_overdo
+
+from inspirehep.modules.authors.receivers import _query_beard_api
+from inspirehep.modules.records.json_ref_loader import replace_refs
+from inspirehep.modules.search import AuthorsSearch, LiteratureSearch
+from inspirehep.utils.bibtex import Bibtex
+
+from invenio_orcid.proxies import current_orcid
+from invenio_records.api import Record
+from invenio_search import current_search_client
 
 
 def convert_to_orcid(record):
@@ -30,13 +42,11 @@ def convert_to_orcid(record):
     orcid needs.
     """
     orcid_json = orcid_overdo.do(record)
+    orcid_json['external-ids'] = {'external-id': orcid_json['external-id']}
+    orcid_json.pop('external-id')
 
-    orcid_json['external-identifiers'] = {
-        'work-external-identifier': orcid_json['work_external_identifier']}
-    orcid_json.pop('work_external_identifier')
     orcid_json['citation'] = {'citation': Bibtex(
         record).format(), 'citation-type': 'BIBTEX'}
-
     return orcid_json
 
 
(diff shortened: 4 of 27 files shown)


kaplun:
`Enchances` -> `Enhances` :dancer:

eamonnmag:
@Panos512 perhaps we can have a look at this from the UI point of view? Do you have anything for this yet?

inspirehep/config.py
```diff
@@ -877,7 +878,7 @@ def _(x):
 
 # Web services and APIs
 # =====================
-BEARD_API_URL = None  # e.g. "http://beard.inspirehep.net/api"
+BEARD_API_URL = "http://beard.inspirehep.net/api"
```
kaplun: This and the following, would be better kept disabled. But I guess it's needed for testing, until #1282 is implemented.

Panos512:
@kaplun  I do this every time... 😞 

@eamonnmag That would be great. I've just made a blueprint and had the final sign-up button redirect to my view, but that's all.

inspirehep/modules/orcid/utils.py
```diff
@@ -53,3 +54,341 @@ def get_authors_credentials(author):
     user = RemoteAccount.query.filter_by(user_id=raw_user.id_user).first()
     token = user.tokens[0].access_token
     return (token, author_orcid)
+
+class DiscoverPossibleProfiles:
```
kaplun: You don't need to make this as a class...

Panos512:
Current WIP ui preview:

screen shot 2016-07-27 at 14 08 25

inspirehep/config.py
```diff
@@ -83,9 +84,7 @@ def _(x):
     "CACHE_REDIS_URL",
     "redis://localhost:6379/0")
 CACHE_TYPE = "redis"
-ACCOUNTS_SESSION_REDIS_URL = os.environ.get(
-    "ACCOUNTS_SESSION_REDIS_URL",
-    "redis://localhost:6379/2")
+ACCOUNTS_SESSION_REDIS_URL = "redis://localhost:6379/2"
```
jmartinm: I am not sure if we were using the environment variable, but why the removal?

inspirehep/config.py
```diff
@@ -886,7 +886,7 @@ def _(x):
 
 # Web services and APIs
 # =====================
-BEARD_API_URL = None  # e.g. "http://beard.inspirehep.net/api"
+BEARD_API_URL = "http://beard.inspirehep.net/api"  # For testing purposes.
```
jmartinm: Was this supposed to be commited? Or we want to keep it as None.

inspirehep/modules/authors/receivers.py
```diff
@@ -28,14 +28,16 @@
...
 
 from invenio_indexer.signals import before_record_index
 from invenio_records.signals import (
     before_record_insert,
     before_record_update,
 )
 
+from invenio_search.utils import schema_to_index
+
+import requests
```
jmartinm: Better to keep the library imports `import *` all together on top.

inspirehep/config.py
Panos512: Nope, that's a leftover. It was uncommented to get `phonetic_blocks` on the records, but I think there is another way now.

inspirehep/modules/orcid/static/js/inspire_profile_matcher/templates/match_profile.html
```diff
@@ -0,0 +1,84 @@
...
+  <div id="collection-landing-page-text">
+    <div class="row">
+      <div class="col-xs-12">
+          <p class="h3 text-center">
+            Please confirm your authorship of {{vm.invenioSearchResults.hits.hits.length}} papers.
+          </p>
+          <br>
+          <p class="text-muted text-center">
+             Please deselect any papers not authored by you.<br>
+             This will help our algorithm get better at spotting your papers in future.
```
jmartinm: `in the future.` instead of `in future.`

inspirehep/modules/orcid/static/js/inspire_profile_matcher/templates/match_profile.html
```diff
@@ -0,0 +1,84 @@
...
+      </div>
+    </div>
+  </div>
+  <article class="panel panel-default profile-matching header record-brief-literature">
+   <div class="panel-body">
+      <div class="row">
+        <div class="col-md-9 text-muted">
+    <h2 class="record-brief-title">Publication Details</h2>
+    </div>
+    <div class='col-md-3' style='text-align: center'>
```
jmartinm: Avoid inline styling, you can use `class: text-center` instead.

inspirehep/modules/orcid/static/js/inspire_profile_matcher/templates/match_profile.html
```diff
@@ -0,0 +1,84 @@
+<form  ng-controller="inspireProfileMatcherCtrl" ng-submit="postMatches()">
+
+  <div id="collection-landing-page-text">
+    <div class="row">
+      <div class="col-xs-12">
+          <p class="h3 text-center">
```
jmartinm: Where is the `h3` class defined? It does not seem a good name to style a `p` element.

inspirehep/modules/orcid/static/js/inspire_profile_matcher/templates/match_profile.html
```diff
@@ -0,0 +1,84 @@
...
+              <div ng-if="record.display.publication_info_line && !(record.display.publication_info_line.pub_info && record.metadata.dois)">
+                <div ng-bind-html='record.display.publication_info_line | publicationInfo' ></div>
+              </div>
+              <div ng-if="record.metadata.dois && !(record.display.publication_info_line && record.display.publication_info_line.pub_info)">
+                <div ng-bind-html='record.metadata.dois | doi'></div>
+              </div>
+              <div ng-bind-html='record.metadata.arxiv_eprints | arxivInfo'></div>
+          </div>
+        </div>
+        <div class="col-md-3 profile-matching" align="center">
```
jmartinm: http://www.w3schools.com/tags/att_div_align.asp

`The <div> align attribute is not supported in HTML5. Use CSS instead.`

inspirehep/modules/orcid/static/templates/match_profile.html
```diff
@@ -0,0 +1,21 @@
+
```
jmartinm: This is a jinja2 template so should be in `inspirehep/modules/orcid/templates/orcid`

`static` is reserved for static files (JS, Angular templates and so on)

inspirehep/modules/orcid/utils.py
```diff
@@ -24,21 +24,27 @@
...
+
+from invenio_search import current_search_client
+
 
 def convert_to_orcid(record):
     """ Converts a given record to a json containing the information that
     orcid needs.
     """
     orcid_json = orcid_overdo.do(record)
+    print(orcid_json)
```
jmartinm: leftover

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+
+
+def match_profiles(token, response):
+    author_name = response['name']
+    author_token = token.token()[0]
+    author_orcid = response['orcid']
+
+    # Query ORCiD for the records that are associated with the logged in
+    # author.
+    orcid_records_dois, orcid_records_arxivs = get_identifiers_from_orcid_records(
```
jmartinm: `orcid_records_arxiv_ids`

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+
+
+def get_identifiers_from_orcid_records(orcid, token):
+
+    api = current_app.extensions['inspire-orcid'].orcid_api
+
+    orcid_records = api.read_record_public(
+        orcid, 'activities', token)
+    orcid_records_dois = []
+    orcid_records_arxivs = []
```
jmartinm: Like above, I think `orcid_records_arxiv_ids` is more accurate.

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+    records = current_search_client.search(
+        index='records-hep',
+        doc_type='hep',
+        body=es_query,
+        size=10000
+    )['hits']['hits']
+    return records
+
+
+def query_for_authors_references_using_phonetic_block(phonetic_block, dois, arxivs):
```
jmartinm: arxiv_ids

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+            for author in record['inner_hits']['authors']['hits']['hits']:
+                possible_profiles.add(author['_source']['record']['$ref'])
+        except KeyError:
+            continue
+    return possible_profiles
+
+
+def solve_references_of_authors(refs):
+    possible_profiles = []
+    current_app.logger.info(refs)
```
jmartinm: Do we need to log this?

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+        except KeyError:
+            continue
+    return possible_profiles
+
+
+def pick_records_from_author_profiles(profiles):
+    number_of_suggested_records = len(profiles)
+    number_of_records_per_profile = 1
+    picked_records = {}
+    picked_recids = []
```
jmartinm: To keep track of the ids a `set()` is better than a list.

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+    if len(profiles) is 2:
+        number_of_suggested_records = 4
```
jmartinm: I am not sure I understand this variable. It seems that you use it [here](https://github.com/inspirehep/inspire-next/pull/1318/files#diff-167ecde30d58be9b213c70305d70ee33R400) for the maximum number of profiles to consider.

But then in this `if` if there are 2 profiles you will consider 4 profiles ?

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+        number_of_records_per_profile = 2
```
jmartinm: Do you need this variable at all? When you are [querying ES](https://github.com/inspirehep/inspire-next/pull/1318/files#diff-167ecde30d58be9b213c70305d70ee33R157) you are already setting the `size` to 2 so you will always get two records.

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+    if len(profiles) > 10:
+        number_of_suggested_records = 10
+    for i, profile in enumerate(profiles):
+        picked_records[profile] = []
+        if i is number_of_suggested_records:
+            break
+        records = query_elasticsearch_for_literature_records_using_a_profile_reference(
+            profile)
+        for j, record in enumerate(records):
+            if j is number_of_records_per_profile:
```
jmartinm: See comment above, you probably don't need this since size of records will always be 2 maximum.

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+                break
+            if record['_source']['control_number'] not in picked_recids:
+                record['_source']['_profile'] = profile
```
jmartinm: I am a bit confused by this, why are you altering the `record` object retrieved from ElasticSearch? Where do you use this `_profile` key?

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+                picked_records[profile].append(record)
+                picked_recids.append(record['_source']['control_number'])
+    if len(picked_recids) is 1:
+        assign_profile(profile)
+    return picked_records
+
+
+def find_profiles(possible_profiles):
```
jmartinm: All these functions need the parameters documented as there are all kinds of data structures traveling around, and for example here it is not clear what `possible_profiles` might be.

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+    if len(possible_profiles) is 1:
+        profile = possible_profiles.pop()
+        assign_profile(profile)
+        return [profile]
+    if len(possible_profiles) is 0:
+        return None
+    chosen_records = pick_records_from_author_profiles(possible_profiles)
+    return chosen_records
```
jmartinm: If I am correct this function can return a list, None or a dictionary (different on each `return`) this needs to be unified/documented.

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+
+def assign_profile(profile, orcid=None):
+    if not orcid:
+        orcid = session['users_orcid']
+    author = query_for_author_profiles_using_ref(profile)[0]['_source']
+    author['ids'].append({
+        "type": u"ORCID",
+        "value": orcid
+    })
+    current_search_client.index(
```
jmartinm: So the change you make just goes to ES.

Is this intended? Or it should also go to the DB?

inspirehep/modules/orcid/views.py
```diff
@@ -0,0 +1,89 @@
...
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+from collections import Counter
+
+from flask import Blueprint, render_template, session, request
+
+from inspirehep.modules.records.serializers.json import process_es_hit
+
+import json
```
jmartinm: Keep this imports on top.

inspirehep/modules/orcid/views.py
```diff
@@ -0,0 +1,89 @@
...
+
+# from inspirehep.modules.pidstore.fetchers import inspire_recid_fetcher as pid_fetcher
+# from inspirehep.modules.records.serializers import JSONBriefSerializer
+# from inspirehep.modules.records.serializers.schemas.json import RecordSchemaJSONBRIEFV1
+blueprint = Blueprint(
+    'inspire_oauthclient',
+    __name__,
+    url_prefix="/oauth",
+    static_folder="static",
+    template_folder="static/templates/",
```
jmartinm: Should be `template_folder='templates'`

inspirehep/modules/orcid/views.py
```diff
@@ -0,0 +1,89 @@
...
+            "hits": {
+                "hits": [],
+                "total": 0
+            },
+            "links": {
+                "self": ""
+            }
+        }
+
+        return json.dumps(response)
```
jmartinm: You can keep only one `return json.dumps(response)`

inspirehep/modules/orcid/views.py
```diff
@@ -0,0 +1,89 @@
...
+    try:
+        data = session['oauth_possible_profiles']
+
+        records = [process_es_hit(x) for y, z in data.items() for x in z]
+
+        response = dict(
+            hits=dict(
+                hits=[{'metadata': process_es_hit(
+                    hit['_source']),
+                    'id': hit['_source']['control_number']}for hit in records]
```
jmartinm: space missing after }

inspirehep/modules/orcid/views.py
```diff
@@ -0,0 +1,89 @@
...
+def match_data():
+    """Extra signup step."""
+    try:
+        data = session['oauth_possible_profiles']
+
+        records = [process_es_hit(x) for y, z in data.items() for x in z]
+
+        response = dict(
+            hits=dict(
+                hits=[{'metadata': process_es_hit(
```
jmartinm: Do you need to call `process_es_hit` twice?

inspirehep/modules/records/jsonschemas/records/authors.json
```diff
@@ -247,9 +247,27 @@
...
                 },
                 "properties": {
-                    "name": {
-                        "type": "string"
-                    }
+                  "name": {
+                    "type": "string"
+                  }
+                },
+                "name_variations": {
```
jmartinm: This should not be part of the schema afaik, as it is something automatically computed that is only in ES. Not something that, for example, would be edited by a cataloger.

inspirehep/modules/theme/static/js/settings.js
```diff
@@ -55,6 +55,8 @@ require.config({
     'inspirehep-clipboard': 'js/inspire_clipboard',
     'impact-graphs': 'node_modules/impact-graphs/impact-graph',
     'inspirehep-search': 'node_modules/inspirehep-search-js/dist/inspirehep-search',
+    'inspire-profile-matcher-module': 'js/inspire_profile_matcher/profile_matcher/profilematcher.module',
+    'inspire-profile-matcher-controllers': 'js/inspire_profile_matcher/profile_matcher/profilematcher.controller',
```
jmartinm: I think you don't need `inspire-profile-matcher-controllers` in `settings.js`

inspirehep/modules/theme/static/scss/base/core.scss
```diff
@@ -18,13 +18,107 @@
...
-*/
+// TODO: These will be moved to a new file
+article.header {
+  height: 28px;
+}
 
-/* ===========================================================
-    Document Setup (body, common classes, structure etc)
-============================================================ */
+h4.profile-matchizng{
```
jmartinm: Typo?

inspirehep/modules/theme/static/scss/base/core.scss
```diff
@@ -18,13 +18,107 @@
...
 
-/*
-Generic core styles that are included in all INSPIRE pages.
-*/
+// TODO: These will be moved to a new file
+article.header {
+  height: 28px;
+}
 
-/* ===========================================================
```
jmartinm: The comment mentions what this file is for. So it should be kept and the styles related only to matching page moved to their own file.

tests/unit/orcid/test_orcid_converter.py
```diff
@@ -135,8 +135,8 @@ def test_successfull_conversion():
                                                                              'contributor-sequence': 'ADDITIONAL'},
```
jmartinm: You are adding quite a bit of untested code in this PR, so once it is more stable would be good to add them.

inspirehep/modules/orcid/utils.py
Panos512: Yeah it's for accuracy. We didn't want to propose only 2 records. So if the possible profiles are two then we propose 2 from each profile. If there are more we propose one from every profile (with a max of 10 records).

inspirehep/modules/orcid/utils.py
Panos512: It's more like a flag, that decides how many of the two records `es` returns we keep. If we have two possible profiles we need both the records. But if we have more than two profiles we need only one record per profile.

inspirehep/modules/orcid/utils.py
Panos512: I get that from angular to know which one of the possible profiles is associated with this record.

inspirehep/modules/orcid/utils.py
Panos512: If there is a table that we store ids then it should also go to the db.

jmartinm: Well it should be just part of the record https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/jsonschemas/records/authors.json#L71 and commited to DB otherwise fetching from DB or from Elasticsearch would give different values which should not be the case.

inspirehep/modules/orcid/utils.py
jmartinm: I see, I think it will work although it was a bit difficult to read. Maybe adding a test will improve that ;)

eamonnmag:
Looks good. The submit button should be more aligned to the right hand side
though.

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+        number_of_records_per_profile = 2
+    if len(profiles) > 10:
+        number_of_suggested_records = 10
+    for i, profile in enumerate(profiles):
+        picked_records[profile] = []
+        if i is number_of_suggested_records:
+            break
+        records = query_elasticsearch_for_literature_records_using_a_profile_reference(
+            profile)
+        for j, record in enumerate(records):
```
jmartinm: instead of `enumerate` and then `break` you could do:

```python
for record in records[:number_of_records_per_profile]:
```

inspirehep/modules/orcid/utils.py
```diff
@@ -54,3 +60,378 @@ def get_authors_credentials(author):
...
+                if 'orcid' in identifier:
+                    continue
+            possible_profiles.append(ref)
+        except KeyError:
+            continue
+    return possible_profiles
+
+
+def pick_records_from_author_profiles(profiles):
+    number_of_suggested_records = len(profiles)
```
jmartinm: As we spoked before, you might not need this variable, keeping track of `number_of_records_per_profile` should be enough.

inspirehep/modules/theme/static/js/settings.js
Panos512: It doesn't load the controller without it. Can I declare the dependency somewhere else?

inspirehep/modules/orcid/utils.py
Panos512: Is adding:
```
db_record = Record(author)
db_record.update()
```
the correct way to do that?

Panos512:
@jmartinm @eamonnmag 
There is a small problem with the mobile-view of the interface:
Since the switch-button moves to the center so does the `Are you the author text`

screen shot 2016-08-08 at 13 48 26


any suggestions?

Panos512:
Just updated this adding: 

* Runs queries using search-classes.
* Redirects after matching.
* The profile linking now correctly reflects to the database.

inspirehep/modules/orcid/utils.py
kaplun: Yep, the change should totally go to the DB (and as a consequence it will be reflected into ES).
@jacquerie has just amended the schema WRT IDs.
https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/jsonschemas/records/authors.json#L98

Panos512:
Regarding the delay issues we were talking about it seems that most of the time is spent retrieving the information from the `orcid-api` which I assume will be much faster on production since we will not be using the sandbox-app.

profiling

Panos512:
@kaplun @jmartinm this is ready if you want to take a look

jacquerie:

PR #1322, opened 2016-07-13
dojson: clear forgotten dedupes.

mihaibivol:

### DIFF
--- a/tests/unit/dojson/test_dojson_institutions.py
+++ b/tests/unit/dojson/test_dojson_institutions.py
@@ -538,24 +538,6 @@ def test_hidden_notes_from_double_595__a():
     assert expected == result['hidden_notes']
 
 
-def test_hidden_notes_from_double_595__a_removes_duplicates():
-    snippet = (
-        '<record>'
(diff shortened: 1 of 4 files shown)


tests/unit/dojson/test_dojson_institutions.py
```diff
@@ -538,24 +538,6 @@ def test_hidden_notes_from_double_595__a():
     assert expected == result['hidden_notes']
 
 
-def test_hidden_notes_from_double_595__a_removes_duplicates():
```
jacquerie: Why was this removed?

mihaibivol: Because we don't remove duplicates in the rules and just in the end.

The test would have looked like:
```
result = clear_duplicates(dojson(smth))
assert no duplicates
```

It doesn't make much more sense to test deduping per specific field since the rule itself is not deduping.

jacquerie: Oh, you're absolutely right.

jacquerie:
LGTM minus a terribly small nit: you have a dot at the end of the commit title : (

PR #1324, opened 2016-07-13
Tei module

liamkirsh:
I haven't thoroughly tested this since the most recent changes and still have some debugging code included.


inspirehep/modules/converttohal/tei.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/python
```
jacquerie: The [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) is only useful if you expect users to run this script directly, i.e. `$ ./tei.py`. If you expect users to import it and use the functions inside, it doesn't matter.

inspirehep/modules/converttohal/tei.py
```diff
@@ -0,0 +1,115 @@
...
+
+from jinja2 import Environment, PackageLoader
+import json
+
+TEMPLATE = 'default.xml'
+#from inspirehep.utils.record_getter import get_es_record
+from inspirehep.modules.records.json_ref_loader import replace_refs
+#from inspirehep.modules.authors.utils import scan_author_string_for_phrases
+from nameparser import HumanName
+import sys
```
jacquerie: PEP8 contains some suggestions on how to order the imports: https://www.python.org/dev/peps/pep-0008/#imports

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.
>
> Imports should be grouped in the following order:
> 
> 1. standard library imports
> 2. related third party imports
> 3. local application/library specific imports
>
> You should put a blank line between each group of imports.

Also, committing commented code should be avoided as much as possible (I'm almost sure you know this, but better safe than sorry).

inspirehep/modules/converttohal/tei.py
```diff
@@ -0,0 +1,115 @@
...
+        publication = None
+
+    my_affiliations = []
+    seen = []
+    structures = []
+    for author in data.get('authors', []):
+        for affiliation in author.get('affiliations', []):
+            if 'recid' in affiliation and affiliation['recid'] not in seen:
+                my_affiliations.append(affiliation)
+                seen.append(affiliation['recid'])
```
jacquerie: Have a look at these helpers: https://github.com/inspirehep/inspire-next/blob/1f1774933e711982f884ccd7a45cb01daeccb851/inspirehep/utils/dedupers.py

inspirehep/modules/converttohal/tei.py
```diff
@@ -0,0 +1,115 @@
...
+        ref = replace_refs(affiliation, 'db')
+
+        if 'record' in ref and 'collections' in ref['record']:
+            structures.append(
+                _structure_data(ref['record'], affiliation['recid'])
+            )
+
+    print template.render(titles=titles, doi=doi, authors=authors,
+                          publication=publication, structures=structures)
+
```
jacquerie: PEP8 requires two spaces between top-level functions: https://www.python.org/dev/peps/pep-0008/#blank-lines

> Surround top-level function and class definitions with two blank lines.

jacquerie:
Are you familiar with the concept of squashing in Git? If not, I can show you tomorrow how to avoid sending two commits like:

1. Do the thing.
2. Whoops! Fix all the things I messed up in the previous commit.

inspirehep/modules/converttohal/tei.py
```diff
@@ -0,0 +1,115 @@
...
+TEMPLATE = 'default.xml'
+#from inspirehep.utils.record_getter import get_es_record
+from inspirehep.modules.records.json_ref_loader import replace_refs
+#from inspirehep.modules.authors.utils import scan_author_string_for_phrases
+from nameparser import HumanName
+import sys
+
+
+def tei_response(record):
+
```
jacquerie: Top-level functions should have a docstring, a particular comment which becomes the `__doc__` attribute of the function and acts as an inline help when using an interactive environment (like IPython). Docstrings formatting has its own PEP: https://www.python.org/dev/peps/pep-0257/

inspirehep/modules/converttohal/tei.py
```diff
@@ -0,0 +1,115 @@
...
+    for author in authors:
+        if 'full_name' in author and author['full_name']:
+            parsed = HumanName(author['full_name'])
+            author['parsed_name'] = parsed
+
+    titles = data.get('titles', [])
+
+    doi = data['dois'][0]['value'] if 'dois' in data else ""
+
+    if 'publication_info' in data:
```
jacquerie: Have a look at EAFP (Easier to ask forgiveness than permission): https://docs.python.org/2/glossary.html#term-eafp

PR #1327, opened 2016-07-14
utils: add h-index and i10-index utils

jacquerie:
I simplified the API to be just functions rather than static methods in a class.

Closes #1218

### DIFF
--- a/inspirehep/utils/stats.py
+++ b/inspirehep/utils/stats.py
@@ -1,0 +1,61 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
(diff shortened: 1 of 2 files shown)


jacquerie:
Ready for review.

inspirehep/utils/stats.py
```diff
@@ -0,0 +1,73 @@
...
+    An author has i10-index X if she has X papers with at least 10 citations
+    each. See: https://en.wikipedia.org/wiki/H-index#i10-index
+
+    :param citations: a dictionary in the format {recid: citation_count}
+    :return: i10-index of the dictionary of citations.
+    """
+    sorted_citations = _sort_citations(citations)
+
+    i10_index = 0
+    for citation in sorted_citations:
```
mihaibivol: don't need to sort here. Just return `len(x for r, c in citations.items() if c >= 10)` or write it in the verbose mode without early exit.

eamonnmag: Indeed. I was just using the sort since I already had the function from the h index. You approach would be quicker of course.

inspirehep/utils/stats.py
```diff
@@ -0,0 +1,73 @@
...
+        if citation[1] >= 10:
+            i10_index += 1
+        else:
+            break
+
+    return i10_index
+
+
+def _sort_citations(citations):
```
mihaibivol: :nit: you actually need only values

inspirehep/utils/stats.py
jacquerie: My intent here was to write the two algorithms as similar as possible so that one could tell at a glance the difference. But yeah, sorting here is useless (but N log N and N are not that distant when N ~ 1000...)

mihaibivol: Can't help nitpicking algo things

eamonnmag: So initially, I had the idea of passing one sort list through both N methods and gathering all stats in one go. This is why the i10 index function also had sorting on the citations. But I forgot to remove it and create the overall method to calculate all stats on one sorted collection.

mihaibivol:
LGTM in either way :ship:

jacquerie:
Now with shiny linear implementations.

PR #1329, opened 2016-07-14
adding IndexError catch to get_value

liamkirsh:
Modified utils.record.get_value to handle invalid indices that throw an IndexError. Previously it would raise the error; now it should return the default value.

### DIFF
--- a/tests/unit/utils/test_utils_record.py
+++ b/tests/unit/utils/test_utils_record.py
@@ -300,3 +300,17 @@ def test_get_value_returns_empty_dic_when_there_are_no_titles():
     result = get_value(empty_titles, "foo")
 
     assert expected == result
+
+def test_get_value_returns_none_on_index_error():
+    single_title = Record({
+        "titles": [
+            {
+                "title": "Importance of a consistent choice of alpha(s) in the matching of AlpGen and Pythia"
+            }
+        ],
+    })
+
+    expected = None
+    result = get_value(single_title, "titles.title[1]")
+    
+    assert expected is result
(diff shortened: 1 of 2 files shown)


tests/unit/utils/test_utils_record.py
```diff
@@ -300,3 +300,17 @@ def test_get_value_returns_empty_dic_when_there_are_no_titles():
...
+            {
+                "title": "Importance of a consistent choice of alpha(s) in the matching of AlpGen and Pythia"
+            }
+        ],
+    })
+
+    expected = None
+    result = get_value(single_title, "titles.title[1]")
+    
+    assert expected is result
```
jacquerie: In this case I prefer to write `assert get_value(single_title, 'titles.title[1]') is None`, because this looks a little silly : )

tests/unit/utils/test_utils_record.py
```diff
@@ -300,3 +300,17 @@ def test_get_value_returns_empty_dic_when_there_are_no_titles():
     result = get_value(empty_titles, "foo")
 
     assert expected == result
+
```
jacquerie: Two spaces between functions (that includes tests).

jacquerie:
LGTM, except for things in comments.

Now it's time to open the can of worms of Git commit messages : )

Invenio has lots of rules for those: http://invenio.readthedocs.io/en/latest/technology/git.html#r2-remarks-on-commit-log-messages. In INSPIRE we adopt (cargo-cult?) a small subset, of which you can find numerous examples by doing `git log`. In particular look at `git commit -s` to soft-sign your commit, and the typical commit structure.

jacquerie:
Oh, and have a look at `git rebase` and things called "triangular workflow" and "topic branches". If you have questions I am here!

jacquerie:
Ok, this can be closed in favor of #1335.

PR #1279, opened 2016-06-30
jsonschema: pumped up references

mihaibivol:
## Provide a `hep`-like reference field inside `hep` references
Closes #430

* [x] Add reference new schema (similar to `hep`)
* [x] Fix mapping - it was fixed already
* [x] Amend `dojson` to work with this new schema
* ~~Amend `hepcrawl` to work with this new schema~~ Moved to next bullet
* ~~Amend holdingpen to save this new schema~~ Refs not use in holdingpen
* [x] Amend HEP Detail View to work with this new schema

* ~~Create follow-up issue for merging good ideas into the `hep` schema~~

### Follow-up + extra findings

* Parse author names better #1344 
* Figure out fate of `dois` `isbns` fields in `hep` schema TODO
* Better rendering of reference datatables (we now curate external ids so we can link to them) TODO
* Fix error in double harvesting #1341 
* Fix some other errors in harvesting (won't make an issue)
* Bring references back TODO

PR #1330, opened 2016-07-14
references: first commits

mihaibivol:
* Adds `inspirehep.references` module
* Moves some references utils here
* Moves utils from dojson to `inspirehep.utils`
  * So we can have 'dojson imports modules which import inspirehep utils' and not the other way around since we're going to use this module both in workflows and dojson.
* Adds a non-pythonic non-dict-inheriting non-regex-rule-implementing class that builds jsons by calling methods.
  * Did this so we can transform anything into this json with simple calls rather than fancy transformers


mihaibivol:
We'll use this in #1279 instead of the hardcoded dojson thing and also for moving holdingpen records into hep records without breaking the whole holdingpen :)

eamonnmag:
I love this :)

mihaibivol:
@jacquerie @kaplun

inspirehep/modules/references/processors.py
```diff
@@ -0,0 +1,176 @@
...
+import re
+
+import idutils
+from isbn import ISBNError
+
+from inspirehep.utils.pubnote import split_pubnote
+
+
+def _is_arxiv(obj):
+    arxiv_test = obj.split()
```
kaplun: What is obj in this context? What does this line do?

mihaibivol: I'll add a comment. `idutils` does not parse `arXiv:2905.0001 [hep-th]` but only the first part. At least in references we have lots of strings including the category (`[hep-th]`) so I am stripping that out so I can move forward with #1279

inspirehep/modules/references/processors.py
```diff
@@ -0,0 +1,176 @@
...
+            self.obj['arxiv_eprints'].append(_normalize_arxiv(repno))
+        else:
+            self._ensure_field('publication_info', {})
+            self.obj['publication_info']['reportnumber'] = repno
+
+    def add_uid(self, uid):
+        """Add unique identifier in correct field."""
+        # We might add None values from wherever. Kill them here.
+        uid = uid or ''
+        if _is_arxiv(uid):
```
kaplun: actually idutils has `detect_identifier_schemes` that would reduce a bit of boiler plate.

inspirehep/modules/references/processors.py
mihaibivol: Most likely I still have some errors / corner cases in this logic that I will find when indexing a production dump.

inspirehep/utils/pubnote.py
```diff
@@ -0,0 +1,79 @@
...
+        if page_artid:
+            if '-' in page_artid:
+                # if it has a dash it's a page range
+                page_range = page_artid.split('-')
+                if len(page_range) == 2:
+                    page_start, page_end = page_range
+                else:
+                    artid = page_artid
+            elif _RE_2_CHARS.search(page_artid):
+                # if it it has 2 ore more letters it's an article ID
```
kaplun: typo 'it it'

tests/unit/references/test_processors.py
```diff
@@ -0,0 +1,120 @@
...
+    rb.add_uid('hdl:10443/1646')
+
+    expected = {
+        'arxiv_eprints': ['arXiv:hep-th/0603001', 'arXiv:0705.0017'],
+        'publication_info': {
+            'isbn': '978-1-4493-4485-6',
+            'cnum': 'C87-11-11',
+            'doi': '10.3972/water973.0145.db'
+        },
+        'persistent_identifiers': ['10443/1646']
```
kaplun: This is not good. We are loosing the 'hdl' information.

inspirehep/modules/references/processors.py
```diff
@@ -0,0 +1,176 @@
...
+        self._ensure_field('urls', [])
+        self.obj['urls'].append({'value': url})
+
+    def add_author(self, full_name, role=None):
+        self._ensure_field('authors', [])
+        author = {} if not role else {'role': role}
+        author['full_name'] = full_name
+        self.obj['authors'].append(author)
+
+    def set_pubnote(self, pubnote):
```
kaplun: Mmh. Here you are forcing to build a pubnote even if I had the separate information (pubnote is an invention of INSPIRE, and it's only refextract which is really outputting it).

kaplun: Would be great if you can instead make it possible to independently set a journal title etc.

Additionally, what about allowing the class to be initialized with an optional journal title normalizer?

inspirehep/modules/references/processors.py
mihaibivol: I actually work around some idutils limitations here (like the arXiv thing) which might take a while to integrate and release.

inspirehep/utils/pubnote.py
mihaibivol: fixing it :laughing:

inspirehep/modules/references/processors.py
kaplun: The arXiv specific one is actually INSPIRE-only issue, but the other might be good if at least you open bug reports to idutils, so that others are aware.

inspirehep/modules/references/processors.py
mihaibivol: For the moment we won't use the other information. I'd rather add `set_journal_title` and friends when we will actually use them.

tests/unit/references/test_processors.py
mihaibivol: hmm. I'd just add prepend `hdl:` in front of them when normalizing. maybe we'll change them together with the schema in #1279

inspirehep/modules/references/processors.py
kaplun: Uh? But what about hepcrawl? What I mean is the input not necesseraily has a pubnote, but rather already splitted components.

inspirehep/utils/pubnote.py
```diff
@@ -0,0 +1,79 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Helpers for handling pubnote parsing."""
```
david-caro: Sorry for the 'basic' questions, but what is a pubnote?

inspirehep/utils/pubnote.py
david-caro: s/ore/or

inspirehep/utils/pubnote.py
kaplun: A pubnote is the `,`-concatenation of: `short journal title`, `volume`, `page(s)`.

kaplun: E.g.: https://inspirehep.net/search?p=find+j+phys.rev.%2CD50%2C1140+or+j+jhep%2C0903%2C112

inspirehep/modules/references/processors.py
mihaibivol: I think that for the moment we might create pubnotes so we parse them back. But, the PRs for references will grow a lot if we affect the `hepcrawl -> holdingpen -> inspirehep` stack since there's a need for:
* unifying hepcrawl output
* changing holdingpen `publication` schema
* fixing holdingpen views
* fixing workflows to save the correct thing in our db

That can be a full and different PR

inspirehep/modules/references/processors.py
mihaibivol: inveniosoftware/idutils#14

PR #1328, opened 2016-07-14
ci: use pip to install packages

david-caro:
* Instead of using pip-accel external package, use the new pip
  features directly.

Signed-off-by: David Caro <dcaroest@redhat.com>

PR #1340, opened 2016-07-18
dojson: remove OAI-PMH handling

### LINKED ISSUE #1339: Schema and record inconsistencies
- [ ] `oai_pmh` field doesn't have `items.properties.set`, although some records have it. It is a `string array` and can be seen in `/literature/1404685`
- [ ]  `arxiv_eprints.items.type` doesn't exists but it should be `object`.

kaplun:
* Removes OAI-PMH handling which should be better handled
  natively via invenio-oaiserver. (closes #1339)

### DIFF
--- a/inspirehep/modules/records/jsonschemas/records/hep.json
+++ b/inspirehep/modules/records/jsonschemas/records/hep.json
@@ -518,7 +518,8 @@
                         },
                         "type": "array"
                     }
-                }
+                },
+                "type": "object"
             },
             "type": "array"
         },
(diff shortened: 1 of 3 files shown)


inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -518,7 +518,8 @@
                         },
                         "type": "array"
                     }
-                }
+                },
+                "type": "object"
```
jacquerie: ...Atomicity...

jacquerie:
LGTM minus the atomicity violation that partially addresses #1339

jacquerie:
> - Removes OAI-PMH handling which should be better handled natively via invenio-oaiserver. (closes #1339)

The fact is, this doesn't close #1339.

kaplun:
You mean? Well it addresses both points. Although I have not mentioned in the commit log about the arXiv bit.

jacquerie:
Heh, I was wrong. This still needs a rebase because #1328 was more urgent, though.

PR #1192, opened 2016-06-08
WIP tests: check all licences

jacquerie:
Since you're probably tired of receiving my emails pestering you about the state of your licences, I present to you a test that complains about licences for me.

This is expected to fail because, well, licences are pretty broken ATM.

PR #1351, opened 2016-07-19
global: make JSON pretty again!

kaplun:
* Introduces new scripts/prettyjson that recursively amend all JSON
  sources in the repo in order to prettify them, i.e. indenting them
  by 4 chars, and sorting keys in objects.

* Enforces pretty JSONs by testing for them actually being pretty.

* Amends all current JSONs to be pretty.

* Note: from now own, if you touch a schema or a mapping just rerun
  scripts/prettyjson before committing. In this way we can ensure
  patches are minimal and referred only to the actual code piece of
  JSON being really amended.


kaplun:
cc: @jacquerie

jacquerie:
> * Note: from now own, if you touch a schema or a mapping just rerun
>   scripts/prettyjson before committing. In this way we can ensure
>   patches are minimal and referred only to the actual code piece of
>   JSON being really amended.

This screams "commit hook"...

kaplun:
> This screams "commit hook"...

Indeed. I'll look into that.

kaplun:
Mmh. But actually a commit hook is executed only locally. So each developer should install it. Any suggestion?

eamonnmag:
Actually commit hooks were something to be looked in to anyway to run tests
(selenium) before a push is made.

kaplun:
@eamonnmag since tests are typically slow, wouldn't that be slowing down developers? We are running tests anyway on Travis.

jacquerie:
It's much slower to have people submit code on Travis on which they didn't run the unit tests, like it often happens, because they get their failure after 5+ minutes rather than the 30 seconds it would take locally.

tests/unit/test_json.py
```diff
@@ -0,0 +1,39 @@
...
+
+def test_pretty_json():
+    for root_dir in JSON_PATHS:
+        for json_dir, _, json_files in os.walk(root_dir):
+            for json_file in json_files:
+                if not json_file.endswith(".json"):
+                    continue
+                json_path = os.path.join(json_dir, json_file)
+                json_content = open(json_path).read()
+                assert json_content == json.dumps(json.loads(json_content), indent=4, sort_keys=True), "{file} is not pretty. Please run ./scripts/prettyjson before committing"
```
jacquerie: Hear, hear! I'm actually against this test: since this is still running the beautification anyway to check that the JSON looks good, why don't we just beautify the JSON in the commit hook?

eamonnmag: Exactly, makes no sense waiting 20 mins to find out something you'd see in
a few seconds. If the tests were taking minutes to run then obviously this
would be a no go. But still, overall it's all relative. I would argue that
development would speed up in this way and travis would merely be there as
a sanity check for completely fresh installations.

kaplun:
I agree for unittests. But I believe Selenium is going to be as slow as regression tests, so not quickly runnable locally.

eamonnmag:
It's even slower on travis

kaplun:
I know. But when I push to Travis I can checkout to another branch and start working on something else while Travis gives me the response in parallel. If we use local commit hooks, then I'd have the repo blocked until tests have executed.

kaplun:
Note also that coverage and running test is part of our integration anyway. A developer would need to amend his branch in case of decreased coverage/broken tests. So we do indeed follow common industry practices :)

kaplun:
@jacquerie I have added stripping of trailing spaces and adding of new lines :) I think it is good to :ship:

eamonnmag:
In theory. In practice, with coveralls it's impossible to know if your coverage actually went down or it just messed up. It's more noise than utility on occasions. 

>  If we use local commit hooks, then I'd have the repo blocked until tests have executed.

Again, tests take much less time running locally than on travis. I'd prefer if a commit was validated as I was working on it than having to context switch again to another task 20 mins after starting something else. I think your opinion of how things work in reality is skewed by the fact you don't normally work on inspire-next and you're not dealing with the pains of travis' slow speeds.

kaplun:
> In theory. In practice, with coveralls it's impossible to know if your coverage actually went down or it just messed up. It's more noise than utility on occasions.

Why you say that? I agree coverage is fucked up when tests fail. But when they succeed it's pretty clear if it increased or not and where. 

> Again, tests take much less time running locally than on travis. I'd prefer if a commit was validated as I was working on it than having to context switch again to another task 20 mins after starting something else. 

OK. Fair enough: I think everybody is free to enable a git hook (being them local). Anyway there is no way to enforce them. I am not expert on how to set them up. I guess it's possible to simply symlink `pre-commit` to `./run-test.sh`

> I think your opinion of how things work in reality is skewed by the fact you don't normally work on inspire-next and you're not dealing with the pains of travis' slow speeds.

Might very well be, but it's also a matter of different development process. Travis in background albeit slow, is still not blocking me. And I am happy with that. But then again: everybody is free to run tests locally, and hook them up in pre-commit.

eamonnmag:
It drops even when tests pass on many occasions. The problem is that all of these things waste time. Waiting on travis to report back on things that you would have spotted by running tests locally (which should be done anyway, but is often forgotten) is a complete waste of time. 

People can choose indeed. But the point of a development process is that everyone should follow it. 

Anyway, by reducing the number of erroneous executions on travis, there will be fewer builds in the end because more pass first time. So to me, it's a win-win.

https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks

kaplun:
BTW I confirm one can simply symlink run-tests.sh to pre-commit to have test run locally.

scripts/prettyjson
```diff
@@ -0,0 +1,32 @@
...
+# or submit itself to any jurisdiction.
+
+# Reformat all JSONs in the source tree
+
+for basedir in inspirehep/modules/records/jsonschemas/ \
+        inspirehep/modules/workflows/mappings/ \
+        inspirehep/modules/records/mappings/; do
+    for name in `find $basedir -name *.json`; do
+        echo Cleaning $name
+        python -c "import json; input = json.load(open('${name}')); output = json.dumps(input, indent=4, sort_keys=True); open('${name}', 'w').write(output.replace(', \n', ',\n') + '\n')"
```
jacquerie: This can be just `python -m json.tool`:
```shell
% echo '{"foo": "foo", "bar": "baz"}' | python -m json.tool
{
    "bar": "baz",
    "foo": "foo"
}
```

kaplun: Ha! I haven't understood it was part official library. It's good it's adding newline at the end:
https://hg.python.org/cpython/file/3.5/Lib/json/tool.py#l46

However it's not stripping trailing spaces, so test would fail as soon as one is opening a .json file with an editor that automatically strips trailing spaces.

tests/unit/records/test_records_jsonschemas.py
```diff
@@ -39,5 +39,6 @@ def test_schemas_are_valid():
     for schemas_dir, _, schemas in os.walk(root_dir):
         schemas_path = os.path.sep.join(schemas_dir.split(os.path.sep)[1:])
         for schema in schemas:
-            schema_path = os.path.join(schemas_path, schema)
-            Draft4Validator.check_schema(fetch_schema(schema_path))
+            if schema.endswith(".json"):
```
jacquerie: What is this protecting against?

kaplun: `*.json~` left overs from editors :-) (I was particularly hit by them)

jacquerie:
IMO the schema changes are OK, but I don't like the test: it's not like the application fails if the JSON files are not formatted nicely. It's why I retracted the test that was checking the licenses (#1192).

tests/unit/records/test_records_jsonschemas.py
jacquerie: OK. But I will split this in another commit : )

scripts/prettyjson
jacquerie: TIL that `python -m json.tool` in Python 2 doesn't handle in-place modification of files. We really have to do something like that.

eamonnmag: Also, you should use simplejson, not json. json is much slower than simplejson.

jacquerie: I believe that `simplejson` is just the name of `json` before it was added to the stdlib... but who knows, Stackoverflow is dow.

eamonnmag: It was the same. But they have diverged a little. Simplejson now has lots of nice optimisations, including c bindings. If you're loading lots of JSON files, or dumping them, it'll make a substantial difference.

tests/unit/records/test_records_jsonschemas.py
```diff
@@ -39,5 +39,6 @@ def test_schemas_are_valid():
...
+                schema_path = os.path.join(schemas_path, schema)
+                Draft4Validator.check_schema(fetch_schema(schema_path))
```
jacquerie: This could be simplified and prettified to:
```shell
find inspirehep -name "*.json" | while read -r file; do
  python -c "import json; \
    input = json.load(open('${file}', 'r')); \
    output = json.dumps(input, indent=4, sort_keys=True); \
    open('${file}', 'w').write(output.replace(', \n', ',\n') + '\n')"
done
```
since we don't really have other JSON files outside of these folders that don't have to be touched.

scripts/prettyjson
jacquerie: I get the opposite results:
```
% time scripts/prettify_simplejson
scripts/prettify_json  0.62s user 0.34s system 94% cpu 1.019 total
```
and
```
% time scripts/prettify_json
scripts/prettify_json  0.32s user 0.18s system 93% cpu 0.538 total
```

scripts/prettyjson
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
jacquerie: I also propose to call this `prettify_json`, since all others scripts so far use verbs...

scripts/prettyjson
eamonnmag: I don't think you're using the C bindings, this is where the speed up comes from.

kaplun:
> IMO the schema changes are OK, but I don't like the test: it's not like the application fails if the JSON files are not formatted nicely. It's why I retracted the test that was checking the licenses (#1192).

But... but... That test would have enforced just to run a simple script (beside spit out by the failed test itself), it wasn't adding any further burden on the developer...

Now you will end up, from time to time, with commit re-prettifying JSONs (and if developer do this together with a change of JSON, then it will hinder the very purpose of this patch, i.e. to make it easier to review JSON-changes.

Issue #301, opened 2015-08-17
INSPIRE beard client


kaplun:
Toghether with the beard server (#300) we shall develop the client part that would:
* [x] collect all modified records WRT their signatures (if Invenio allows that), and the claims/unclaims that have happened since last time.
* [x] retrieve all the affected blocks (given a signature in a paper it will belong to a very defined and stable block, e.g. based on a transformation of a surname, see: https://github.com/inveniosoftware/beard/blob/master/beard/clustering/blocking_funcs.py#L174) 
* [x] send to beard all the records and all the signatures and all the claims and all the cluster linked with the affected blocks for beard to give back the new cluster
* [x] rematch the new clusters against the known ones (a.k.a. profiles), and in case create/delete some of them. This must be done respecting explicitly claims and unclaims.
* [x] port to Invenio 3 (@jlavik)

kaplun:
@MSusik, @glouppe please feel free to amend this task as you deem appropriate

glouppe:
Just a comment regarding the design this is taking. Beard has been developed as a _library_ providing all the tools needed to implement author disambiguation. In particular, the `examples` folder is only meant to  illustrate how put all these pieces together to solve disambiguation.  They are not meant to be used as an application themselves. 

In other words, I would rather see a single `inspire-disambiguation` module implementing tasks for author disambiguation (this can be copy pasted from the examples in beard), tasks for matching current and predicted clusters, tasks for getting signatures from records, etc rather than a collection of many decoupled modules (`invenio-beard`, `beard-server`, the glue for putting that into the overlay, etc). 

Or I would even first start by having the machinery working in the inspire overlay only, and only then think about generalizing it to other invenio instances.

Just my 2 cents.

kaplun:
Sure `invenio-beard` could be part of the Overlay, but what would be the benefit?
The fact that we package all these tasks in the form of an Invenio module doesn't imply that is generic. On the other hand it will allow us to lower the barrier for making it generic in the future. 

Beside `beard-server` should really leave as a separate module, since it is a wrapper of beard library that would live on separate machine. It wouldn't make sense to have `beard-server` in INSPIRE overlay.

IMHO, the current decoupling represent on which piece of hardware each task is going to be executed.

kaplun:
> (invenio-beard, beard-server, beardy_json, the glue for putting that into the overlay, etc).

[...]

PR #1353, opened 2016-07-20
disambiguation: alpha release

jacenkow:
* Adds a method to communicate with Beard Celery instances
  in order to delegate a clustering job.

* Adds methods to create a new author profile and update
  signatures to point to new author profiles.

* Adds methods to keep track of new/updated records.

### DIFF
--- a/inspirehep/modules/disambiguation/__init__.py
+++ b/inspirehep/modules/disambiguation/__init__.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Inspire wrapper for Beard author disambiguation service."""
+
+from __future__ import (
+    absolute_import,
+    division,
+    print_function,
+)
+
+from .ext import InspireDisambiguation
+from .tasks import disambiguation_daemon
+
+__all__ = ('InspireDisambiguation', 'disambiguation_daemon')
--- a/inspirehep/modules/disambiguation/beard.py
+++ b/inspirehep/modules/disambiguation/beard.py
@@ -1,0 +1,106 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Communication handler with Beard Celery service."""
+
+from __future__ import (
+    absolute_import,
+    division,
+    print_function,
+)
+
+import celery
+from flask import current_app
+
+
+def make_beard_clusters(records, signatures):
+    """Dispatch a clustering task to Beard Celery instance.
+
+    The method receives a list of records and a list of signatures representing
+    records that will be clustered by Beard algorithm.
+
+    The argument 'records' is a list of dictionaries containing values like
+    authors of a particular record, earliest date recorded of the publication
+    and finally publication UUID.
+
+    The argument 'signatures' represents the phonetic block,
+    that is being currently computed by one of 'disambiguation' workers.
+
+    In order to check what is being used in order to cluster signatures,
+    check 'create_beard_record' and 'create_beard_signatures' methods in
+    'search.py' as the reference.
+
+    The method dispatches a Celery task to Beard server.
+    Beard algorithm distinguishes different authors for
+    *the same* signature block. The clustered authors are then returned
+    in the format of dictionaries.
+
+    :param records:
+        A list of the records, where at least one author will be
+        clustered (by having the same phonetic block).
+
+        Example:
+            records = [{'authors': [u'Hohm, Olaf', u'Wang, Yi-Nan'],
+                       'publication_id': u'13c3cca8-b0bf-42f5-90d4-...',
+                       'year': u'2015'}]
+
+    :param signatures:
+        A list of signatures belonging to the same signature block,
+        which is currently being clustered.
+
+        Example:
+            signatures =  [{'author_affiliation': u'MIT, Cambridge, CTP',
+                            'author_name': u'Wang, Yi-Nan',
+                            'publication_id': u'13c3cca8-b0bf-42f5-90d4-...',
+                            'signature_id': u'a4156520-4248-a57f-949c361e0dd0',
+                            'author_recid': u'10123',
+                            'author_claimed': False}]
+
+    :return:
+        A tuple containing clusters matched with existing author profiles and
+        clusters for which new profile must be created.
+
+        The first 'bucket' is a list of dictionaries, where each key
+        represents recid of an existing author profile. The second 'bucket'
+        is a list of dictionaries as well, however with enumerated keys,
+        which are meaningless.
+
+        Except clustering, Beard Celery instance is also responsible for
+        matching output of a clustering job with a current state of Inspire
+        system, ie. current links between signatures and their author profiles.
+        The matching process is done using simplex algorithm for maximising
+        overlap between new clusters (Beard output) with signatures clustered
+        by belonging to the same profile.
+
+        To see what is the workflow behind Beard Celery instance, check
+        https://github.com/inspirehep/beard-server
+
+        Example:
+            [{u'10123': [u'a4156520-4248-a57f-949c361e0dd0']}, {}]
+    """
+    if records and signatures:
+        clusters = celery.current_app.send_task(
+            'beard_server.tasks.make_clusters',
+            (records, signatures),
+            queue=current_app.config.get('DISAMBIGUATION_QUEUE'))
+
+        return clusters
--- a/inspirehep/modules/disambiguation/config.py
+++ b/inspirehep/modules/disambiguation/config.py
@@ -1,0 +1,39 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Configuration of Disambiguation module."""
+
+from __future__ import (
+    absolute_import,
+    division,
+    print_function,
+)
+
+# Celery.
+DISAMBIGUATION_QUEUE = "celery"
+
+# Elasticsearch.
+DISAMBIGUATION_RECORD_INDEX = "records-hep"
+DISAMBIGUATION_RECORD_TYPE = "hep"
+
+# Invenio-Records.
+DISAMBIGUATION_AUTHORS_SCHEMA_PATH = "records/authors.json"
--- a/inspirehep/modules/disambiguation/logic.py
+++ b/inspirehep/modules/disambiguation/logic.py
@@ -1,0 +1,433 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Logic of Disambiguation module."""
+
+from __future__ import (
+    absolute_import,
+    division,
+    print_function,
+)
+
+from operator import add
+
+import celery
+from celery.utils.log import get_task_logger
+from flask import current_app
+
+from inspirehep.modules.disambiguation.records import create_author
+from inspirehep.modules.disambiguation.search import (
+    create_beard_record,
+    get_signature,
+)
+
+logger = get_task_logger(__name__)
+
+
+def _check_if_claimed(signature):
+    """Check if the signature is claimed.
+
+    In order to proceed with processing the clusters, the signatures
+    must be checked if they are claimed by authors.
+    However in order to spot a collision (at least two claimed
+    signatures from different authors) the method returns a tuple
+    representing a claim together with an author profile, to which this
+    signature has been assigned.
+
+    :param signature:
+        A dictionary representing a signature.
+
+        Example:
+            signature = {'author_affiliation': u'MIT, Cambridge, CTP',
+                         'author_name': u'Wang, Yi-Nan',
+                         'publication_id': u'13c3cca8-b0bf-42f5-90d4-...',
+                         'signature_id': u'a4156520-560d-4248-a57f',
+                         'author_recid': u'10123',
+                         'author_claimed: False}
+
+    :return:
+        A tuple representing a claim and an author profile recid.
+
+        Example:
+            (False, '1337489')
+    """
+    return (bool(signature.get('author_claimed', False)),
+            str(signature.get('author_recid', None)))
+
+
+def _create_distance_signature(signatures_map, uuid):
+    """Create a signature for a similarity measurement.
+
+    Given a UUID as the parameter, the method
+    creates a signature, which will be used to calculate
+    the similarity against claimed signatures.
+
+    :param uuid:
+        A string representing UUID of a given signature.
+
+        Example:
+            uuid = 'd63537a8-1df4-4436-b5ed-224da5b5028c'
+
+    :return:
+        Example:
+            {u'affiliations': u'Yerevan Phys. Inst.',
+             u'publication_id': u'13c3cca8-b0bf-42f5-90d4-e3dfcced0511',
+             u'full_name': u'Chatrchyan, Serguei',
+             u'uuid': u'd63537a8-1df4-4436-b5ed-224da5b5028c'}
+
+    """
+    record = create_beard_record(
+        signatures_map[uuid].get('publication_id'))
+
+    beard_signature = signatures_map[uuid].copy()
+    beard_signature['publication'] = record
+
+    return beard_signature
+
+
+def _create_set_from_signature(signature):
+    """Create a set based on a given signature.
+
+    The method receives a dictionary representing a signature.
+    From a given dictionary it builds a set that will be
+    later used in select_profile_base method.
+
+    :param signature:
+        A dictionary representing a signature.
+
+        Example:
+            signature = {'author_affiliation': u'MIT, Cambridge, CTP',
+                         'author_name': u'Wang, Yi-Nan',
+                         'publication_id': u'13c3cca8-b0bf-42f5-90d4-...',
+                         'signature_id': u'a4156520-560d-4248-a57f',
+                         'author_recid': u'10123',
+                         'author_claimed: False}
+
+    :return:
+        A set of dictionary's values.
+
+        Example:
+            {u'MIT, Cambridge, CTP', u'Wang, Yi-Nan'}
+    """
+    result = set([signature['author_name']])
+
+    if 'author_affiliation' in signature:
+        result.update([signature['author_affiliation']])
+
+    return result
+
+
+def _create_uuid_key_dictionary(signatures):
+    """Create a dictionary with UUIDs being used as keys.
+
+    This method creates a dictionary, which keys are exactly
+    the same as UUID values of each signatures within the list.
+    The value under each key is the signature.
+    """
+    signatures_dictionary = {}
+
+    for signature in signatures:
+        signatures_dictionary[signature['signature_id']] = signature
+
+    return signatures_dictionary
+
+
+def _select_profile_base(signatures_map, uuids):
+    """Select the most rich-in-information signature.
+
+    The method selects a UUID, which signature contains the biggest
+    amount of information from the given list of UUIDs, of
+    signatures selected to represent the same author.
+    Given signatures are converted to sets, then compared against
+    each other to calculate the amount of the common values.
+    The results are stored in a symmetric matrix, where afterwards
+    the shared elements are summed up.
+    The signature with the biggest sum is select to be the most
+    "representative". In case of the same score, the longest set is
+    selected (for example, a signature can have an additional
+    affiliation).
+
+    :param signatures_map:
+        A dictionary representing signatures with keys
+        being exactly the same as the values of UUID of each
+        signature.
+
+        Example:
+            signatures_map = {u'a4156520-4248-a57f':
+                                {'author_affiliation': u'MIT, Cambridge, CTP',
+                                 'author_name': u'Wang, Yi-Nan',
+                                 'publication_id': u'13c3cca8-b0bf-42f5-90d4-',
+                                 'signature_id': u'a4156520-4248-a57f',
+                                 'author_recid': u'10123',
+                                 'author_claimed': False}
+                             }
+
+    :param uuids:
+        A list of UUIDs, representing the same author.
+
+        Example:
+            uuids = ['6e6e8b99-ee08-43a2-a7c6-136a531a47cc',
+                     '4a964f85-762f-4bff-8472-2f47e5f3bb26']
+
+    :return:
+        A UUID of a signature selected to be the most representative.
+
+        Example:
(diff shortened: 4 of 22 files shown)


jacenkow:
This PR suppresses `inspire-disambiguation` repo.

WIP status since missing
- [ ] Cleaning up `config.py` 
- [ ] Tests

jacenkow:
Now you can see this funny error, which I am always getting while migrating records.
Obviously that's my fault, however I cannot spot what is exactly the problem.
CC: @kaplun, @jacquerie

jacquerie:
Can you paste here a longer stacktrace than https://travis-ci.org/inspirehep/inspire-next/jobs/146047716#L1571?

inspirehep/modules/disambiguation/config.py
```diff
@@ -0,0 +1,46 @@
...
+    'run-beard-every-24h': {
+        'task': 'inspirehep.modules.disambiguation.tasks.disambiguation_demon',
+        'schedule': timedelta(hours=24)
+    },
+}
+DISAMBIGUATION_QUEUE = "beard"
+
+# Elasticsearch.
+HEP_ELASTICSEARCH_INDEX = "records-hep"
+HEP_ELASTICSEARCH_TYPE = "hep"
```
jacenkow: @kaplun should I use `INDEXER_DEFAULT_INDEX` and `INDEXER_DEFAULT_DOC_TYPE` which have exactly the same values but a little bit different meaning?

kaplun: I would rather have a variable:

DISAMBIGUATION_RECORD_TYPES = ["hep"]
(and derive the index from that), so that we can plug more than one type of record (e.g. soon "Data").

inspirehep/modules/disambiguation/receivers.py
```diff
@@ -99,7 +99,8 @@ def append_new_record_to_queue(sender, *args, **kwargs):
     appends the given record UUID to the queue of records to be
     processed by Beard.
     """
-    if 'HEP' in [i['primary'] for i in sender['collections']]:
+    # FIXME: Use a dedicated method when #1355 will be resolved.
```
kaplun: :+1:

jacenkow:
@kaplun I have added unit tests, which only cover a few methods from `logic.py` and one from `receivers.py`. Unfortunately the rest must be tested as a part of integration tests, since they require the whole system to be on. As @jacquerie suggested, I will create `fixture`s which will populate few records as well as a small number of profiles to check my use cases (mocking the whole output of demo records processed by `Beard` is not the point, right?). So from now on, I will focus on integration tests.

kaplun:
If this takes too much time, what about creating a new element in the Travis matrix where we load again the records?

inspirehep/modules/disambiguation/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Inspire wrapper for Beard author disambiguation service."""
+
+from __future__ import absolute_import, division, print_function, \
+    unicode_literals
```
jacquerie: `from __future__ import absolute_import, division, print_function` (no `unicode_literals`).

inspirehep/modules/disambiguation/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+#
+# Inspire is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Inspire; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
```
jacquerie: Low priority, but this is GPLv2, and we try to use GPLv3.

inspirehep/modules/disambiguation/logic.py
```diff
@@ -0,0 +1,428 @@
...
+                         'signature_id': u'a4156520-560d-4248-a57f',
+                         'author_recid': u'10123',
+                         'author_claimed: False}
+
+    :return:
+        A tuple representing claim and author profile recid.
+
+        Example:
+            (False, '1337489')
+    """
```
jacquerie: IMO these detailed docstrings for internal helpers are not needed...

inspirehep/modules/disambiguation/logic.py
```diff
@@ -0,0 +1,428 @@
...
+    if 'author_affiliation' in signature:
+        result.update([signature['author_affiliation']])
+
+    return result
+
+
+def _create_uuid_key_dictionary(signatures):
+    """Create a dictionary with UUIDs being used as keys.
+
+    This method creates a dictionary, which keys are exactly
```
jacquerie: whose?

tests/unit/disambiguation/test_receivers.py
```diff
@@ -0,0 +1,89 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Inspire; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Test *helpers* (_method_name(foo, bar):) from receivers.py"""
```
jacquerie: Leftover?

inspirehep/modules/disambiguation/search.py
```diff
@@ -0,0 +1,366 @@
...
+from dateutil.parser import parse
+
+from elasticsearch.helpers import scan
+from flask import current_app
+
+from invenio_search import current_search_client as es
+
+
+def _get_record(record_id, source):
+    """Make a call to the Elasticsearch instace with known record_id.
```
jacquerie: `s/instance/instance/`

inspirehep/modules/disambiguation/search.py
```diff
@@ -0,0 +1,366 @@
...
+            source = ["collaboration", "author.full_name"]
+    """
+    elasticsearch_index = current_app.config.get('DISAMBIGUATION_RECORD_INDEX')
+    elasticsearch_type = current_app.config.get('DISAMBIGUATION_RECORD_TYPE')
+
+    return es.get_source(
+        index=elasticsearch_index,
+        id=str(record_id),
+        doc_type=elasticsearch_type,
+        _source=source)
```
jacquerie: What about adding this feature to https://github.com/inspirehep/inspire-next/blob/59fb321dbb71e64684617aa7e2c63e5268ff1505/inspirehep/utils/record_getter.py#L66-L74 ? Or, at least, wrapping that method with your additional behaviour.

inspirehep/modules/disambiguation/beard.py
```diff
@@ -0,0 +1,103 @@
...
+
+def make_beard_clusters(records, signatures):
+    """Dispatch a clustering task to Beard Celery instance.
+
+    The method receives a list of records and a list of signatures representing
+    records that will be clustered by Beard algorithm.
+
+    The argument 'records' is a list of dictionaries containing values like
+    authors of a particular record, earliest date recorded of the publication
+    and finally publication UUID.
```
jacquerie: So this means that an argument called `records` is not a list of records?!

jacenkow:
@jacquerie, question. I have added a new fixture called `small_app`, which is exactly the same code like `app` with a small difference during migration (check the last commit). I created a new `xml` file containing only two HEP records. It's convenient for testing on a local machine, since you don't have to wait 10 minutes for the records to migrate. However, do we really care about testing time on Travis, since we still have to wait for other integration test to finish (and they run in parallel)?

jacquerie:
> However, do we really care about testing time on Travis, since we still have to wait for other integration test to finish (and they run in parallel)?

Yes. My hope is that we can split the builds in such a way that the longest test suite takes ~10 minutes (4 minutes for Docker, 4 minutes for demo records, 2 minutes for testing) and that they are more or less balanced. Since our slow suite is 15 minutes this means that we are currently running 7 minutes worth of tests, which might mean we could do this in 4 parallel builds that take ~10 minutes.

inspirehep/modules/disambiguation/records.py
```diff
@@ -0,0 +1,256 @@
...
+
+    :param profile_recid:
+        A string representing author profile recid, that
+        updated signature should point to.
+
+        Example:
+            profile_recid = '1'
+    """
+
+    trial_count = 0
```
jmartinm: Missing increment inside of the `while` loop?

inspirehep/modules/disambiguation/search.py
```diff
@@ -0,0 +1,365 @@
...
+    ]
+
+    response = _get_record(record_id, _source)
+
+    # Get the year of the record publication.
+    try:
+        earliest_date = response.get("earliest_date")
+        year = parse(earliest_date).strftime("%Y")
+    except:
+        year = False
```
jmartinm: Making `year` a boolean will not cause problems below when you do `int(year)` ?

jmartinm: Replying to myself:


```python
In [1]: int(False)
Out[1]: 0
```

inspirehep/modules/disambiguation/records.py
jacenkow: 😉

tests/integration/disambiguation/test_daemon.py
```diff
@@ -0,0 +1,87 @@
...
+        assert send_to_clustering.call_count == 2
+
+
+def test_check_if_daemon_has_cleaned_the_queue(small_app):
+    """Check if the daemon has cleaned up the queue."""
+    from inspirehep.modules.disambiguation.models import \
+        DisambiguationRecord
+    from inspirehep.modules.disambiguation.tasks import \
+        disambiguation_daemon
+
```
jmartinm: You can check here `DisambiguationRecord.query.count() != 0`

inspirehep/modules/disambiguation/search.py
jacquerie: (Nit) eh, this is actually guaranteed in Python 3, but not necessarily true in Python 2:
```
In [1]: False = 5

In [2]: int(False)
Out[2]: 5
```

inspirehep/modules/disambiguation/records.py
```diff
@@ -0,0 +1,259 @@
...
+        if author['uuid'] == uuid:
+            author['recid'] = str(profile_recid)
+            update_flag = True
+
+    if update_flag:
+        # Disconnect the signal on insert of a new record.
+        before_record_index.disconnect(append_updated_record_to_queue)
+
+        # Update the record in the database.
+        record.commit()
```
kaplun: Protect this with a:
```python
try:
....
```

inspirehep/modules/disambiguation/records.py
```diff
@@ -0,0 +1,259 @@
...
+        db.session.commit()
+
+        # Update the record in Elasticsearch.
+        indexer = RecordIndexer()
+        indexer.index_by_id(record.id)
+
+        # Reconnect the disconnected signal.
```
kaplun:
```python
finally:
   before...
```

inspirehep/modules/disambiguation/config.py
```diff
@@ -0,0 +1,37 @@
...
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Configuration of Disambiguation module."""
+
+from __future__ import absolute_import, division, print_function
+
+# Celery.
+DISAMBIGUATION_QUEUE = "beard"
```
kaplun: By default should be empty and on production we should set it to `beard`.

inspirehep/modules/disambiguation/logic.py
```diff
@@ -0,0 +1,438 @@
...
+        # Select the most rich-in-information signature.
+        base_profile = get_signature(_select_profile_base(signatures_map,
+                                                          uuids))
+        # Create a new profile.
+        recid = create_author(base_profile)
+
+        # Update all signatures with the new profile (recid).
+        for uuid in uuids:
+            record = signatures_map[uuid].get('publication_id')
+            celery.current_app.send_task(
```
kaplun: Here you are in inspirehep context, so you can happily import `inspirehep.modules.disambiguation.tasks.update_authors_recid` 
and call:
```python
update_authors_recid.delay(record, uuid, recid)
```

inspirehep/modules/disambiguation/logic.py
```diff
@@ -0,0 +1,438 @@
...
+
+    # If there are not claimed signatures, but there is a match.
+    elif len(claims) == 0 and recid_key:
+        # Update all signatures with the profile of the 'old' cluster.
+        for uuid in uuids:
+            record = signatures_map[uuid].get('publication_id')
+            celery.current_app.send_task(
+                'inspirehep.modules.disambiguation.tasks.update_authors_recid',
+                (record, uuid, recid_key)
+            )
```
kaplun: Mmh... This smells like some repeated code ;-)

inspirehep/modules/disambiguation/logic.py
```diff
@@ -0,0 +1,438 @@
...
+        # claims format: (False, u'1234')
+        recid = claims.pop()[1]
+
+        # Update all signatures with a profile of the claimed signature.
+        for uuid in uuids:
+            record = signatures_map[uuid].get('publication_id')
+            celery.current_app.send_task(
+                'inspirehep.modules.disambiguation.tasks.update_authors_recid',
+                (record, uuid, recid)
+            )
```
kaplun: Ditto

inspirehep/modules/disambiguation/logic.py
```diff
@@ -0,0 +1,438 @@
...
+        # For each claimed signature, assign recid of it to not claimed ones.
+        for claimed_uuid, uuids in matched_signatures.items():
+            recid = signatures_map[claimed_uuid].get('author_recid')
+
+            for uuid in uuids:
+                record = signatures_map[uuid].get('publication_id')
+                celery.current_app.send_task(
+                    'inspirehep.modules.disambiguation.tasks.update_authors_recid',
+                    (record, uuid, recid)
+                )
```
kaplun: Ditto

inspirehep/modules/disambiguation/logic.py
jacenkow: I have some problems with imports when I do
```python
from inspirehep.modules.disambiguation.tasks import update_authors_recid
```
in `logic.py`, my `Celery` worker is crashing.

inspirehep/modules/disambiguation/receivers.py
```diff
@@ -0,0 +1,134 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+from invenio_indexer.signals import before_record_index
+from invenio_records.signals import after_record_insert
+
+from inspirehep.modules.disambiguation.models import DisambiguationRecord
+
+
+def _check_if_record_eligible(authors_before, authors_after):
```
kaplun: Can you rename it to what it does? `needs_beard_reprocessing` or similar.

inspirehep/modules/disambiguation/logic.py
kaplun: You might need to import it locally to the function

tests/unit/disambiguation/test_receivers.py
```diff
@@ -82,8 +82,8 @@ def test_check_if_record_eligible_method():
         "full_name": "Grzegorz Jacenków"
     }]
 
-    assert _check_if_record_eligible(base_author, base_author) is not True
-    assert _check_if_record_eligible(base_author, author_new_affiliation) is \
+    assert _needs_beard_reprocessing(base_author, base_author) is not True
+    assert _needs_beard_reprocessing(base_author, author_new_affiliation) is \
```
jacquerie: There's no need for `is True`: just `assert stuff_that_is_True`. Analogously, just `assert not stuff_that_is_False`.

jacenkow:
@kaplun for HOWTO check #1421.

inspirehep/modules/disambiguation/beard.py
```diff
@@ -0,0 +1,101 @@
...
+        overlap between new clusters (Beard output) with signatures clustered
+        by belonging to the same profile.
+
+        To see what is the workflow behind Beard Celery instance, check
+        https://github.com/inspirehep/beard-server
+
+        Example:
+            [{u'10123': [u'a4156520-4248-a57f-949c361e0dd0']}, {}]
+    """
+    # Do not run if any of the two lists are empty.
```
jacquerie: Obvious comment is obvious.

inspirehep/modules/disambiguation/beard.py
```diff
@@ -0,0 +1,101 @@
...
+    if records and signatures:
+        clusters = celery.current_app.send_task(
+            'beard_server.tasks.make_clusters',
+            (records, signatures),
+            queue=current_app.config.get('DISAMBIGUATION_QUEUE'))
```
jacquerie: Why can't this invocation be direct?

jacenkow: This how you dispatch a task to another machine with no function definition on the machine that you dispatch the task from.

inspirehep/modules/disambiguation/logic.py
```diff
@@ -0,0 +1,428 @@
...
+
+from operator import add
+
+import celery
+from celery.utils.log import get_task_logger
+from flask import current_app
+
+from inspirehep.modules.disambiguation.records import create_author
+from inspirehep.modules.disambiguation.search import create_record, \
+    get_signature
```
jacquerie: One of these snippets is a syntax error:
```python
import foo, \
    bar
```
and
```python
import foo, \ 
    bar
```
Which one?

jacenkow: Sorry, what?

jacquerie: The second snippet has an extra space after the `\`, which makes it invalid. Which is why I don't like this style of imports...

kaplun: Oooh, the recommended be:
```python
import (foo,
     bar)
```
right?

jacenkow: Thanks, anyway I would appreciate straightforward comments.

Imports using `\` are following `PEP-8`.
I mean I cannot blame myself for not putting whitespace after `\` and thus having imports working correctly.

kaplun: The interesting gotcha reported by @jacquerie underlines indeed that we should use parenthesis to protect multiline statements, even if PEP8 allows for the trailing backslash.

We should indeed use this technique everywhere.

jacquerie: > Oooh, the recommended be:
> 
> ```python
> import (foo,
>     bar)
> ```
> right?

```python
import (
    foo,
    bar,
)
```
is better, because then you can add/remove imports and have cleaner Git histories.

> Imports using \ are following PEP-8.

Nope. From https://www.python.org/dev/peps/pep-0008/#maximum-line-length:

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

There's also https://www.python.org/dev/peps/pep-0328/#rationale-for-parentheses, and PEP8 explicitly mentions this line continuation invisible gotcha.

> Thanks, anyway I would appreciate straightforward comments.
Heh, I wasn't getting any answer before...

inspirehep/modules/disambiguation/beard.py
kaplun: @jacquerie I think `record` in this context is to be considered as a `beard_record`. @jacenkow indeed what are the keys that are expected? Is there a formal description to which you can refer in this description?

E.g.  to say: "see docstring of function `foo` for a description of the structure of a record as expected by beard."

inspirehep/modules/disambiguation/logic.py
jacenkow: I wasn't giving any answers since I would like to focus on merging this PR rather than solving puzzles...

kaplun:
Please, add to commit log that this this closes #301.

tests/unit/disambiguation/test_receivers.py
jacenkow: This is the maximum that I can test a part of unit tests. The other methods require the whole system to be set up. This is way I only check *helpers*.

inspirehep/modules/disambiguation/logic.py
jacenkow: I believe in English, we refer whose to human beings only.

tests/unit/disambiguation/test_receivers.py
jacquerie: Oh, no, here I meant the weird `(_method_name(foo, bar):)` which looked auto-generated. But then I understood what you meant.

inspirehep/modules/disambiguation/beard.py
jacenkow: Done.

jacenkow:
@kaplun thanks for merging #1426.

So that's a green light from my side, all my problems are now resolved. I have updated the branch.
However, I still haven't touched `get_es_record` since now it became a class. Need time to investigate.

CC: @jacquerie

jacquerie:
I will take care of fixing some style issues myself. Just ping me when you (and @kaplun) think it's ready to go, and I will take care of merging.

inspirehep/modules/disambiguation/records.py
```diff
@@ -0,0 +1,158 @@
...
+    return url_for(
+        "invenio_jsonschemas.get_schema",
+        schema_path=current_app.config.get(
+            'DISAMBIGUATION_AUTHORS_SCHEMA_PATH'
+        )
+    )
+
+
+def create_author(profile):
+    """Create a new profile in authors index.
```
kaplun: Better say: "create a new a author based on the given signature" (not profile).

inspirehep/modules/disambiguation/records.py
```diff
@@ -0,0 +1,158 @@
...
+
+    The method receives a dictionary representing an author.
+    Based on the values, it creates a dictionary in the invenio_records format.
+    After all the fields are processed, the method calls create_record
+    from invenio_records.api to put the new record.
+
+    :param profile:
+        A signature representing an author's to be created as a profile
+        with additional values for an experiment and a field_categories
+        taken from the hep paper.
```
kaplun: Can you show in the example below where the experiment and field_category goes?

inspirehep/modules/disambiguation/records.py
```diff
@@ -0,0 +1,158 @@
...
+                       u'profile': u'',
+                       u'recid': None,
+                       u'role': None,
+                       u'uuid': u'd63537a8-1df4-4436-b5ed-224da5b5028c'}
+
+    :return:
+        A recid, where the new profile can be accessed.
+
+        Example:
+            "fb8cecf2-ed73-4240-ac70-6bdf8e834819"
```
kaplun: It seems to me that this function actually returns a record ID not a UUID

inspirehep/modules/disambiguation/search.py
```diff
@@ -0,0 +1,367 @@
...
+    for response in scan(
+            es, query=query,
+            index=elasticsearch_index,
+            doc_type=elasticsearch_type):
+        result.append(response)
+
+    return result
+
+
+def create_record(record_id):
```
kaplun: Better call it then: `create_beard_record`.

inspirehep/modules/disambiguation/tasks.py
```diff
@@ -0,0 +1,191 @@
...
+    dispatch each of the phonetic blocks found within
+    these records to Celery workers and finally dumps the records.
+    """
+    phonetic_blocks = set()
+
+    # Get the records classified for clustering.
+    records_to_cluster = DisambiguationRecord.query.all()
+
+    # For each record get the phonetic blocks.
+    for record in records_to_cluster:
```
kaplun: Here you might have more times the same record. So you are going to call `get_blocks_from_record()` in the loop too many times. It's better if you can actually extract directly the list of record IDs involved and transform them into a set.

inspirehep/modules/disambiguation/tasks.py
```diff
@@ -0,0 +1,191 @@
...
+        if clusters_created:
+            for beard_uuids in list(clusters_created.values()):
+                process_clusters(beard_uuids, signatures)
+
+        db.session.commit()
+    finally:
+        db.session.close()
+
+
+@shared_task(ignore_result=True)
```
kaplun: Why `ignore_result=True` here?

jacenkow:
@kaplun  🍍

inspirehep/modules/disambiguation/records.py
kaplun: @jacenkow this point is still missing.

inspirehep/modules/disambiguation/search.py
kaplun: @jacenkow I believe here you should also add a FIXME, in case we are going to run disambiguation over multiple indexes (which is the case, i.e. Literature and Data)

kaplun:
LGTM.

inspirehep/modules/disambiguation/records.py
jacquerie: He can't, because https://github.com/inspirehep/inspire-next/pull/1353/files#diff-a9904eafdef51b485bb1dcb7e59343e8R129.

In fact, this method is assuming things that are not true in the schema, therefore reading and creating invalid author records.

inspirehep/modules/disambiguation/records.py
```diff
@@ -0,0 +1,162 @@
...
+                       u'role': None,
+                       u'uuid': u'd63537a8-1df4-4436-b5ed-224da5b5028c'}
+
+    :return:
+        A recid, where the new profile can be accessed.
+
+        Example:
+            "1234"
+    """
+    # Get the author's name.
```
jacquerie: This style of commenting is only appropriate when writing a tutorial for people that have never programmed before. It's completely obvious that `name = profile.get('full_name')` gets the author's name; if it didn't, something would be very very wrong. Comments should never say _what_ are you doing, but _why_ are you doing it.

inspirehep/modules/disambiguation/records.py
```diff
@@ -0,0 +1,162 @@
...
+    # Apply the changes.
+    record.commit()
+    db.session.commit()
+
+    # Add the record to Elasticsearch.
+    indexer = RecordIndexer()
+    indexer.index_by_id(record_pid.object_uuid)
+
+    # Reconnect the disconnected signal.
+    after_record_insert.connect(append_new_record_to_queue)
```
jacquerie: All this logic that deals with low level implementation details (pids, receivers, ES...) of the creation of an author record should be abstracted in an helper (if it wasn't already).

inspirehep/modules/disambiguation/search.py
```diff
@@ -0,0 +1,367 @@
...
+        Example:
+            record_id = "a5afb151-8f75-4e91-8dc1-05e7e8e8c0b8"
+
+    :param source:
+        A list of fields to be extracted from the Elasticsearch response.
+
+        Example:
+            source = ["collaboration", "author.full_name"]
+    """
+    elasticsearch_index = current_app.config.get('DISAMBIGUATION_RECORD_INDEX')
```
jacquerie: As I said today during standup, record search MUST be centralized, otherwise we would have to change the index name everywhere it's used when we deploy on production.

inspirehep/modules/disambiguation/records.py
jacenkow: I understand the point, however I don't see the reason of creating a helper method, which wouldn't be reusable.

inspirehep/modules/disambiguation/search.py
jacenkow: I would like to leave it for the next iteration.

jacenkow:
ping @kaplun

jacenkow:
❤️

inspirehep/modules/disambiguation/receivers.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    if len(authors_before) == len(authors_after):
+        for index, author_before in enumerate(authors_before):
+            # Not every author has an affiliation.
+            before_affiliations = author_before.get(
+                'affiliations', [])
+
+            # We don't iterate over authors_after, we take the index.
+            after_affiliations = authors_after[index].get(
+                'affiliations', [])
```
jacquerie: FYI, this should have been a [`zip`](https://docs.python.org/2/library/functions.html#zip), which is the pythonic way of iterating on two lists in parallel.

inspirehep/modules/disambiguation/receivers.py
```diff
@@ -0,0 +1,136 @@
...
+
+            before = (author_before['full_name'], before_affiliations)
+            after = (authors_after[index]['full_name'], after_affiliations)
+
+            if before != after:
+                return True
```
jacquerie: ...and this kind of logic is an [`any`](https://docs.python.org/2/library/functions.html#any).

kaplun: Never used `any`. Thanks for the tip

PR #1354, opened 2016-07-20
references: support for old refextract authors strings

### LINKED ISSUE #1344: references: try authors string split
The `999C5` field contains only the full authors string. Since it's hard to change this in legacy it's better to dump another piece of logic into `modules/references/processors.py`

mihaibivol:
* Adds method to parse out individual authors from legacy refextract
  author strings containing multiple authors.
* Uses added method in dojson.

Closes #1344 

### DIFF
--- a/inspirehep/modules/references/processors.py
+++ b/inspirehep/modules/references/processors.py
@@ -23,11 +23,64 @@
 """References Processors."""
 
 import re
+import six
 
 import idutils
 from isbn import ISBNError
 
 from inspirehep.utils.pubnote import split_pubnote
+
+
+# Matches any separators for author enumerations.
+RE_SPLIT_AUTH = re.compile(r',?\s+and\s|,?\s*&|,|et al\.?|\(?eds?\.\)?',
+                           re.I | re.U)
+# Matches any stream of initials (A. B C D. -E F).
+RE_INITIALS_ONLY = re.compile(r'^\s*-?[A-Z]((\.|\s)\s*-?[A-Z])*\.?\s*$',
+                              re.U)
+
+
+def _split_refextract_authors_str(authors_str):
+    """Extract author names out of refextract authors output."""
+    author_seq = (x.strip() for x in RE_SPLIT_AUTH.split(authors_str) if x)
+    res = []
+
+    current = ''
+    for author in author_seq:
+        if not isinstance(author, six.text_type):
+            author = six.text_type(author.decode('utf8', 'ignore'))
+
+        # First clean the token.
+        author = re.sub(r'\(|\)', '', author, re.U)
+        # Names usually start with characters.
+        author = re.sub(r'^[\W\d]+', '', author, re.U)
+        # Names should end with characters or dot.
+        author = re.sub(r'[^.\w]+$', '', author, re.U)
+
+        # If we have initials join them with the previous token.
+        if RE_INITIALS_ONLY.match(author):
+            current += ', ' + author.strip()
+        else:
+            if current:
+                res.append(current)
+            current = author
+
+    # Add last element.
+    if current:
+        res.append(current)
+
+    # Manual filterings that we don't want to add in regular expressions since
+    # it would make them more complex.
+    #  * ed might sneak in
+    #  * many legacy refs look like 'X. and Somebody E.'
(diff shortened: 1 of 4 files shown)


mihaibivol:
Hope that integration tests are not a lot slower :)

inspirehep/modules/references/processors.py
```diff
@@ -23,13 +23,66 @@
...
+    for author in author_seq:
+        if not isinstance(author, six.text_type):
+            author = six.text_type(author.decode('utf8', 'ignore'))
+
+        # First clean the token.
+        author = re.sub(r'\(|\)', '', author, re.U)
+        # Names usually start with characters.
+        author = re.sub(r'^[\W\d]+', '', author, re.U)
+        # Names should end with characters or dot.
+        author = re.sub(r'[^.\w]+$', '', author, re.U)
```
mihaibivol: @jacquerie whoops http://stackstatus.net/post/147710624694/outage-postmortem-july-20-2016

mihaibivol: In any case, refextract won't produce a string bad for this regex.

tests/unit/references/test_processors.py
```diff
@@ -120,3 +123,31 @@ def test_reference_builder_add_uid():
...
+        'Gómez R, Reilly P, Winicour J and Isaacson R'
+    ]
+
+    expected = [
+        ['Butler, D.', 'Demarque, P.', 'Smith, H. A.'],
+        ['Cenko, S. B.', 'Kasliwal, M. M.', 'Perley, D. A.'],
+        ['J. Kätlne'],
+        ['J. Kätlne'],
+        ['Hoaglin D. C.', 'Mostellar F.', 'Tukey J. W.'],
+        ['V.M. Zhuravlev', 'S.V. Chervon', 'V.K. Shchigolev'],
```
kaplun: Mmh, does it mean you actually don't try to put twist these name into "surname, firstnames"?
Would be great if you can do this as well, since this is the way they are then encoded in the system.

mihaibivol: Hm, This might make the function a bit heavy and also I have no clue how it will look for our firends `New, A.`, `Study, A.` :P

kaplun: Just take the last word and put it as first before the comma. We will then fix it in another step ;)

mihaibivol: I tried `scan_author_string_for_phrases` on these examples and it looks pretty bad. Do we have anything else for this type of normalization?

kaplun: What about the library @liamkirsh was playing with? https://pypi.python.org/pypi/nameparser

mihaibivol: Still, kind of the same. I wouldn't proceed with anything else for the moment.

* In matching I always used raw tokens without first name / last name info mostly because of these things being wrong
* In display it's the same thing for me :)

inspirehep/modules/references/processors.py
```diff
@@ -23,13 +23,66 @@
...
 import re
+import six
 
 import idutils
 from isbn import ISBNError
 
 from inspirehep.utils.pubnote import split_pubnote
 
 
+# Matches any separators for author enumeratoins.
```
jacquerie: `s/enumeratoins/enumerations`

inspirehep/modules/references/processors.py
```diff
@@ -23,13 +23,66 @@
...
+RE_SPLIT_AUTH = re.compile(r',? and |,?\s*&|,|et al\.?|\(?eds?\.\)?',
```
jacquerie: Isn't this missing some cases? Like, arbitrary spaces after the comma but `and` instead of `&`?

jacquerie:
LGTM, minus one nit and a comment.

inspirehep/modules/references/processors.py
mihaibivol: I'd rather strip the strings than integrate arbitrary whitespaces here.

mihaibivol: ah, that thing. It seemed alright in the tests, there were no leftovers for this thing

PR #1358, opened 2016-07-20
global: make JSON pretty again

jacquerie:
Closes #1351


jacquerie:
@eamonnmag I didn't follow your lead on `simplejson` + C extensions because at most it would have netted us 1 second every time we reformat the JSON files, which is not very frequently...

PR #1359, opened 2016-07-21
dojson: hidden publication_info

kaplun:
- Introduces support for hidden publication_info, i.e. 7731_ in
  MARCXML.

### DIFF
--- a/inspirehep/dojson/hep/fields/bd76x78x.py
+++ b/inspirehep/dojson/hep/fields/bd76x78x.py
@@ -78,6 +78,7 @@ def publication_info(self, key, value):
         'year': year,
         'isbn': force_single_element(value.get('z')),
         'notes': dedupe_list(force_force_list(value.get('m'))),
+        'hidden': key[3:4] == '1' or None,  # ind1="1" means hidden
     }
 
     return res
@@ -94,9 +95,8 @@ def publication_info2marc(self, key, value):
         page_artid.append('{page_start}'.format(**value))
     if value.get('artid'):
         page_artid.append('{artid}'.format(**value))
-    return {
-        '0': get_recid_from_ref(
-            value.get('parent_record')),
+    ret = {
+        '0': get_recid_from_ref(value.get('parent_record')),
         'c': page_artid,
         'n': value.get('journal_issue'),
         'o': value.get('conf_acronym'),
@@ -110,6 +110,10 @@ def publication_info2marc(self, key, value):
         'z': value.get('isbn'),
         'm': value.get('notes')
     }
+    if value.get('hidden'):
+        self.setdefault('7731_', []).append(ret)
+        return {}
+    return ret
 
 
 @hep.over('succeeding_entry', '^785..')
--- a/inspirehep/modules/records/jsonschemas/records/hep.json
+++ b/inspirehep/modules/records/jsonschemas/records/hep.json
@@ -611,6 +611,11 @@
                     "curated_relation": {
                         "type": "boolean"
                     },
+                    "hidden": {
+                        "default": false,
+                        "description": "Whether these publication info are hidden because (e.g. because of legacy reasons)",
+                        "type": "boolean"
+                    },
                     "isbn": {
                         "type": "string"
                     },
--- a/inspirehep/modules/theme/jinja2filters.py
+++ b/inspirehep/modules/theme/jinja2filters.py
@@ -488,6 +488,8 @@ def publication_info(record):
     pub_infos = []
     if 'publication_info' in record:
         for pub_info in record.publication_information:
+            if pub_info.get('hidden'):
+                continue
             pub_info_html = render_macro_from_template(
                 name="pub_info",
                 template="inspirehep_theme/format/record/Publication_info.tpl",
--- a/tests/unit/dojson/fixtures/test_hep_record.xml
+++ b/tests/unit/dojson/fixtures/test_hep_record.xml
@@ -225,6 +225,21 @@
     <subfield code="o">ConfAcr</subfield>
     <subfield code="0">1</subfield>
   </datafield>
+  <datafield tag="773" ind1="1" ind2=" ">
+    <subfield code="c">R123</subfield>
+    <subfield code="n">2</subfield>
+    <subfield code="p">Phys.Rev.Lett.</subfield>
+    <subfield code="v">105</subfield>
+    <subfield code="y">2010</subfield>
+    <subfield code="m">note</subfield>
+    <subfield code="z">isbn</subfield>
+    <subfield code="x">pubinfo_freetext</subfield>
+    <subfield code="w">cnum</subfield>
+    <subfield code="t">confpaper_info</subfield>
+    <subfield code="r">reportNumber</subfield>
+    <subfield code="o">ConfAcr</subfield>
+    <subfield code="0">1</subfield>
+  </datafield>
   <datafield tag="785" ind1="0" ind2="2">
     <subfield code="r">CERN-PH-EP-2014-262</subfield>
     <subfield code="w">recid</subfield>
--- a/tests/unit/dojson/test_dojson_hep.py
+++ b/tests/unit/dojson/test_dojson_hep.py
@@ -1649,8 +1649,11 @@ def test_publication_info(marcxml_to_json, json_to_marc):
     assert marcxml_to_json['publication_info'][0]['artid'] == '026802'
     assert marcxml_to_json['publication_info'][0]['page_start'] == '123'
     assert marcxml_to_json['publication_info'][0]['page_end'] == '456'
+    assert marcxml_to_json['publication_info'][1]['artid'] == 'R123'
+    assert marcxml_to_json['publication_info'][1]['hidden']
     assert '026802' in json_to_marc['773'][0]['c']
     assert '123-456' in json_to_marc['773'][0]['c']
+    assert 'R123' in json_to_marc['7731_'][0]['c']
     assert (marcxml_to_json['publication_info'][0]['journal_issue'] ==
             json_to_marc['773'][0]['n'])
     assert (marcxml_to_json['publication_info'][0]['journal_title'] ==


kaplun:
cc @tsgit here's the corresponding implementation for Invenio 3.

inspirehep/dojson/hep/fields/bd76x78x.py
```diff
@@ -77,6 +77,7 @@ def get_int_value(val):
         'year': year,
         'isbn': value.get('z'),
         'note': value.get('m'),
+        'hidden': key[3] == '1' or None,  # ind1="1" means hidden
```
jacquerie: This will raise `KeyError` if there's no `3`, though.

inspirehep/dojson/hep/fields/bd76x78x.py
```diff
@@ -94,22 +95,23 @@ def publication_info2marc(self, key, value):
...
+           'o': value.get('conf_acronym'),
+           'p': value.get('journal_title'),
+           'r': value.get('reportnumber'),
+           't': value.get('confpaper_info'),
+           'v': value.get('journal_volume'),
+           'w': value.get('cnum'),
+           'x': value.get('pubinfo_freetext'),
+           'y': value.get('year'),
+           'z': value.get('isbn'),
+           'm': value.get('note')}
```
jacquerie: IMO formatting was better before.

inspirehep/dojson/hep/fields/bd76x78x.py
```diff
@@ -94,22 +95,23 @@ def publication_info2marc(self, key, value):
...
+    if value.get('hidden'):
+        self.setdefault('7731_', []).append(ret)
```
jacquerie: Uh, let me understand the use case here, because this is hidden, but it's still part of the metadata, just under a different key...

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -486,6 +486,8 @@ def publication_info(record):
         journal_title, journal_volume, year, journal_issue, pages = \
             ('', '', '', '', '')
         for pub_info in record['publication_info']:
+            if pub_info.get('hidden'):
+                continue
```
jacquerie: Sigh, more complexity in an already terrible filter : (

tests/unit/dojson/fixtures/test_hep_record.xml
```diff
@@ -197,6 +197,21 @@
...
+    <subfield code="y">2010</subfield>
+    <subfield code="m">note</subfield>
+    <subfield code="z">isbn</subfield>
+    <subfield code="x">pubinfo_freetext</subfield>
+    <subfield code="w">cnum</subfield>
+    <subfield code="t">confpaper_info</subfield>
+    <subfield code="r">reportNumber</subfield>
+    <subfield code="o">ConfAcr</subfield>
+    <subfield code="0">1</subfield>
+  </datafield>
```
jacquerie: Mh, maybe it would be better to write tests in the new style rather than complicating this fixture.

inspirehep/dojson/hep/fields/bd76x78x.py
kaplun: eheh, when I added my field then I lost the original one.

inspirehep/dojson/hep/fields/bd76x78x.py
kaplun: Exactly. It's going to `<datafield tag="773" ind1="1" ind2=" ">`.

inspirehep/modules/theme/jinja2filters.py
kaplun: It's simply skipping hidden pubnotes, since they are hidden.

tests/unit/dojson/fixtures/test_hep_record.xml
kaplun: What do you mean by new style?

jacquerie: I mean this style: https://github.com/inspirehep/inspire-next/blob/dd8042ce0f1953e2e764b9c19ac4df196f5d5dce/tests/unit/dojson/test_dojson_conferences.py#L33-L50

A snippet with the MARCXML that you're considering, the expected result, two lines of boilerplate for running the code and the assertion. It's a style introduced by @bittirousku and @tomaszgy, and I think it's a pretty neat idea.

kaplun: I was about to change it, but here I believe it has a value in its form, because it test also it's interaction WRT the non hidden publication in the same record, and it very practically allows to test the json->marc transformation as well.

jacquerie:
This really begs the question: will we ever have per field ACL?

kaplun:
It's not that this field is secret information, it's simply that they have introduced a flag to say: this is an alternative publication info that can be used to match the record, but is no longer recommended, hence should not be displayed. How else would you model this?

jacquerie:
Oh! I imagined it was "we know that this record is going to be published here, but we don't want to reveal this to the general public yet, because reasons".

kaplun:
We are going to have technical metadata indeed that is more related to workflows but this should go in the audit log table and available only through holding pen.

kaplun:
IMHO this one is ready to be merged

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -634,6 +634,11 @@
                     "curated_relation": {
                         "type": "boolean"
                     },
+                    "hidden": {
+                        "default": false,
+                        "description": "Whether these publication info are hidden because (e.g. because of legacy reasons)",
```
jacquerie: IMO "hidden" needs no explanation.

jacquerie:
I still see no point in this. This is just porting a limitation of legacy to next, where we have better tools (ES analyzers) to solve the problem this is supposed to solve.

For example: https://inspirehep.net/search?ln=en&p=7731_%3A**&of=text&ot=773__c&rg=1200. Do we really need to add a new `publication_info` if all we need is to be able to search with page ranges with an R prepended to them?

kaplun:
If not like this then: we need to have:
* [ ] dojson ignore hidden pubnotes
* [ ] augment mapping/enhancers in ES to also index with and without R-prefix iff (for the time being) journal is Phys. Rev.

annetteholtkamp:
This implies that a developer has to get active whenever an exception like the R for Phys Rev pops up. A hidden field can be populated by other people. To be discussed in a larger group I guess. At least include Kirsten and Florian since they maintain this field right now I think.

- Annette

kaplun:
@inspirehep/inspire-content

kaplun:
@ksachs @fschwenn indeed @jacquerie point makes sense. If we have to include a consistent alias (e.g. optional R for Phys Rev) it's more robust to do it once and for all at indexing time, than to expect that all records have to be amended. Amending all records would corresponds to either lots of manual work or a bibcheck rule to be written, which would anyway involve development and operations.

ksachs:
Consider also exceptions of the exception. Metadata don't need to be logical.

Whatever you set up should not only work for:
For 773__p:"Phys.Rev." 773__c:"R22-27" should also be indexed as "22-27".
There might also be the case:
For this set of records 773__v:"12" 773__n:"07" should be treated as 773__v:"1207", except for ...

IMHO there is absolutely NO way to code this mess. 
Additional metadata for indexing, that are not displayed can easily deal with ANY case.

jacquerie:
Currently `publication_info` has no `hidden` key: https://github.com/inspirehep/inspire-schemas/blob/bdae2b7b50a68480fc0ac6452cb7508a8bad1c8c/inspire_schemas/records/hep.yml#L1057-L1171. If we still want this, we need to add it to the schema before proceeding with this.

CC: @michamos

ksachs:
we need the feature

michamos:
OK, I will add the `hidden` flag to the schema. About the rapid communication issue, shouldn't we also add a new field to the `publication_info`, something like `journal_section`, that would contain `R`? This would take care of one very common use of these hidden pubnotes.

ksachs:
I don't under stand what you mean. Can you give an example?

michamos:
currently, we have
```
00533378 773__ $$cR5923-R5926$$pPhys.Rev.$$vE62$$y2000
000533378 7731_ $$c5923-5926$$pPhys.Rev.$$vE62$$y2000
```
what I suggest is a way to avoid having a hidden pubnote in this (very common) case, by translating it to
```json
{
    "publication_info": [
        {
            "page_start": "5923",
            "page_end": "5926",
            "journal_title": "Phys.Rev.",
            "journal_volume": "E62",
            "journal_section": "R",
            "year": "2000"
        }
    ]
}
```
and let the search engine deal with optionally prepending the `R` to the page range.

ksachs:
No way!
We had that discussion in September (this thread).
Put in the metadata what need to be indexed. No fiddling around with special cases for indexing.
How do you want to treat
000697133 7731_ $$c948-979$$pAdv.Theor.Math.Phys.$$v12$$y2008
000697133 773__ $$c945-979$$pAdv.Theor.Math.Phys.$$v12$$y2008

michamos:
I am not proposing to implement this *instead* of hidden pubnotes, but *in addition* to it. `R` is *not* part of the page strictly speaking, and I am proposing to properly model this case in the metadata. Your latest example would be implemented with a hidden pubnote.

ksachs:
R is part of the official article number and the DOI.
Phys. Rev. E 62, R5923  (from the Phys.Rev. web-page)
https://doi.org/10.1103/PhysRevE.62.R5923

michamos:
You're right, now `R5923` really looks like an article ID instead of a page. Even in the PDF, the page numbers have disappeared. Nevermind then, `hidden` is sufficient.

jacquerie:
Waiting for https://github.com/inspirehep/inspire-schemas/pull/137 to be merged.

michamos:
the schemas part is done, now needs to wait for `inspire-schemas` to get bumped to at least `32.1.1`.

jacquerie:
We could also backport this to `31.x`...

jacquerie:
This has happened in #2279 by backporting the required schema change to `inspire-schemas==31.1.0` (https://github.com/inspirehep/inspire-schemas/issues/143).

PR #1362, opened 2016-07-21
Remove regexp for full and alternative name fields

harunurhan:
Remove full_name and alternative_name regexp which does not match with all existing records.

### DIFF
--- a/inspirehep/modules/records/jsonschemas/records/hep.json
+++ b/inspirehep/modules/records/jsonschemas/records/hep.json
@@ -124,7 +124,6 @@
                     },
                     "alternative_name": {
                         "description": "FIXME: needed for import only",
-                        "format": ".+, .+",
                         "title": "Alternative author name",
                         "type": "string"
                     },
@@ -175,7 +174,6 @@
                     },
                     "full_name": {
                         "description": "author name as it appears in the paper",
-                        "format": ".+, .+",
                         "title": "Author name",
                         "type": "string"
                     },


jacquerie:
Mh. I believe you should remove all modifications to mappings, since you are using JSON Schema syntax, which is probably not supported by Elasticsearch.

harunurhan:
@jacquerie Addressed your comment, also removed changes for other `format` related stuff which will also be removed soon because of #1363

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -175,7 +175,7 @@
                     },
                     "full_name": {
                         "description": "author name as it appears in the paper",
-                        "format": ".+, .+",
+                        "pattern": ".+, .+",
```
jacquerie: What about [Jimmy](https://inspirehep.net/author/profile/Jimmy.1)?

harunurhan: Actually, the regexp was already there, I assumed all existing people would match the regexp. So, since we have Jimmy and possible others like him, it may be better idea remove this validation.

jacquerie: And I agree with this.

harunurhan:
@jacquerie Addressed your comment.

jacquerie:
BTW @harunurhan we have a few rules on how to format commit message, which are a subset of the Invenio ones: http://invenio.readthedocs.io/en/latest/technology/git.html#r2-remarks-on-commit-log-messages (and are not enforced as strictly).

For an example of how you could have formatted this commit see: https://github.com/inspirehep/inspire-next/pull/1366/commits/8f2daf3ee778a6b0791c6f8eb8635da670676dea.

harunurhan:
@jacquerie Thanks 👍 , I knew that we have some rules and was actually waiting for somebody to tell me because couldn't find the link

PR #1367, opened 2016-07-25
theme: evict ES from filters

jacquerie:
Doesn't yet do what it says on the tin.


inspirehep/modules/search/views.py
```diff
@@ -64,6 +64,8 @@ def search():
         ctx['search_api'] = '/api/journals/'
         return render_template('search/search_journals.html', **ctx)
     if collection == 'jobs':
+        from inspirehep.utils.search import get_number_of_records
```
jacquerie: This import has to be local, otherwise we would get a circular import.

inspirehep/modules/records/wrappers.py
```diff
@@ -19,16 +18,36 @@
...
\ No newline at end of file
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+from invenio_records.api import Record
+from invenio_search.api import RecordsSearch
+
+
+class JobsRecord(Record):
```
jacquerie: This is a concept I wanted to have for quite some time. Note here how it neatly wraps the querying behaviour and delegates the presentation behaviour to the template, while previously the filter was doing both.

jacquerie:
Now does what it says on the tin, and is ready to review. BTW this fixes the "Search 0 articles" you occasionally see on https://qa.inspirehep.net.

jacquerie:
This is waiting for the solution to #1368 to be pushed and merged, since it will also touch a few templates.

mihaibivol:
LGTM :)

jacquerie:
I fast-tracked this so that @mihaibivol's nightly build looks nicer : )

Will wait for green and merge.

PR #1370, opened 2016-07-27
migrator: better error logging

### LINKED ISSUE #1365: migrator: better error handling
Currently we only check for validation errors when migrating chunnks

We need a consistent way to account for:

* Errors when transforming XML strings into dicts (in this case we do not have a recid)
* Exceptions occuring during `dojson` conversion (in this case we __do__ have a recid and we can use the same format as we use for ValidationErrors)

mihaibivol:
* Refactors migration logic into a single function
* Wraps every step that might fail due to corrupt data in try: except:
  blocks
* Adds logging messages that are aggregated by Sentry

Closes #1365

### DIFF
--- a/inspirehep/celery.py
+++ b/inspirehep/celery.py
@@ -1,11 +1,34 @@
 # -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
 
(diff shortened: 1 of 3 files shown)


mihaibivol:
For deployment we also need to set `LOGGING_SENTRY_CELERY = True`

inspirehep/modules/migrator/tasks.py
```diff
@@ -204,31 +203,13 @@ def create_index_op(record):
...
-                    db.session.merge(prod_record)
-                    continue
-
-                index_queue.append(create_index_op(record))
-
-                prod_record.valid = True
-                db.session.merge(prod_record)
+           with db.session.begin_nested():
+                record = migrate_and_insert_record(raw_record)
+                if record:
```
jacquerie: This would be a nice place to add optional index population.

mihaibivol:
@jacquerie this is ready for review now and can be integrated :)

inspirehep/modules/migrator/tasks.py
```diff
@@ -176,7 +177,6 @@ def continuous_migration():
                         errors = "ValidationError: Record {0}: {1}".format(
                             recid, e
                         )
-                        current_app.logger.warning(errors)
```
jacquerie: Why are you already removing this? Isn't this part of #1371?

inspirehep/celery.py
```diff
@@ -9,3 +9,6 @@
 from .factory import create_app
 
 celery = create_celery_app(create_app())
+# Trigger an app log message upon import. Somehow this makes Sentry logging
+# work with `get_task_logger(__name__)`.
+celery.flask_app.logger.info('Created Celery app')
```
jacquerie: Since you are here... what about a license and `__future__` imports in this file?

jacquerie:
LGTM, but needs a rebase and a minor touch.

inspirehep/modules/migrator/tasks.py
mihaibivol: I didn't know how to test the continous migration. This should be replaced with ` migrate_and_insert_record(raw_record)` anyway

PR #1372, opened 2016-07-27
WIP tests: API acceptance tests

jacquerie:


.travis.yml
```diff
@@ -35,8 +36,10 @@ env:
     - DOCKER_VERSION=1.10.3-0~trusty
     - DOCKER_COMPOSE_VERSION=1.6.2
   matrix:
+    - SUITE=pep8
```
mihaibivol: can we also add another variable like `NEED_SETUP=true` in the matrix so that for the `PEP8` Build we don't do the whole build step and just pull python-base on `docker-compose run`.

mihaibivol: we could have something like:

```
- `${NEED_SETUP} && travis_retry whatever` on every build step
```

jacquerie:
Not so urgent, so I'm going to shelve this for a while. I'll just "cherry-pick" the part about separating PEP8 from the other builds.

PR #1373, opened 2016-07-27
WIP travis: add PEP8 build

jacquerie:
Salvaged from #1372. Currently doesn't include https://github.com/inspirehep/inspire-next/pull/1372#discussion_r72454840 because it was slightly trickier to implement than I tought.


jacquerie:
There are two ideas behind this:

1. Repurpose `./run-tests.sh` for something useful (inspired by the Invenio one)
2. Avoid having to click on a build just to discover that you just have PEP8 failures.

mihaibivol:
What's the purpose of the new `run_tests.sh` should it be used by devs to run things locally?

jacquerie:
...I don't know. I haven't decided: on one hand maybe we should have there all the `check-manifest`, `pydocstyle`, unit tests, `building docs`, like the Invenio people do. On the other hand, what's the point of that if it isn't part of our build? People are not going to run it. On the third hand (?!), maybe we can repurpose it as a Docker helper.

PR #1374, opened 2016-07-28
schemas: various improvements

jacquerie:


jacquerie:
Mh, changing the main schema names to singular is slightly more challenging than it looks. This is ready to go in the meantime.

PR #1376, opened 2016-07-28
theme: fix count shown on authors landing page

jacquerie:
Whoops. The actual collection name to pass is `hepnames`, not `authors`, otherwise we fall back on the HEP count.


mihaibivol:
LGTM for quickfix :)

PR #1378, opened 2016-07-29
dojson: refactor earliest_date

jacquerie:
Salvaged from #1364 since it was growing too much in scope.

### DIFF
--- a/inspirehep/dojson/hep/receivers.py
+++ b/inspirehep/dojson/hep/receivers.py
@@ -22,40 +22,32 @@
 
 """Contains signal receivers for HEP dojson processing."""
 
+from __future__ import absolute_import, division, print_function
+
+from itertools import chain
+
 from invenio_records.signals import before_record_insert, before_record_update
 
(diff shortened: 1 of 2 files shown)


jacquerie:
Looks straightforward enough, but if somebody wants to review...

inspirehep/dojson/hep/receivers.py
```diff
@@ -22,40 +22,32 @@
...
 from invenio_records.signals import before_record_insert, before_record_update
 
 from inspirehep.utils.date import create_earliest_date
+from inspirehep.utils.helpers import force_force_list
+from inspirehep.utils.record import get_value
 
 
 @before_record_insert.connect
 @before_record_update.connect
 def earliest_date(sender, *args, **kwargs):
```
kaplun: Uh! I now realize this should not be stored in the DB, but only sent to ES. The earliest date is something that should be dynamically computed after all other related metadata dates.

PR #1382, opened 2016-08-01
setup: pin invenio-oauthclient

jacquerie:
Attempts to fix https://travis-ci.org/inspirehep/inspire-next/jobs/148804909#L1610.


jacquerie:
Currently a pin rather than a bump because of https://travis-ci.org/inspirehep/inspire-next/jobs/148825809#L1553. @Panos512, can you have a look? Install `invenio-oauthclient==1.0.0a8` to see this error.

PR #1387, opened 2016-08-01
schemas: phd_advisors -> advisors

jacquerie:
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601538/

PR #1392, opened 2016-08-02
WIP schemas: aggregating ID schemas

liamkirsh:


inspirehep/modules/records/jsonschemas/records/elements/id.json
```diff
@@ -0,0 +1,80 @@
...
+                    ],
+                    "type": "string"
+                },
+                "value": {
+                    "pattern": "JACoW-\\d+",
+                    "type": "string"
+                }
+            },
+            "type": "object"
+        },
```
jacquerie: What about adding HAL?

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -177,48 +177,18 @@
...
-                        "description": "ORCID Id when available",
-                        "format": "orcid",
-                        "pattern": "\\d{4}-\\d{4}-\\d{4}-\\d{4}",
-                        "title": "ORCID ID",
-                        "type": "string"
-                    },
-                    "other_ids": {
-                        "description": "Just in case...",
+                    "ids": {
+                        "description": "Contains all identifiers, both internal and external.",
```
jacquerie: IMO this can go away, `ids` is pretty clear...

jacquerie:
Did you run `scripts/prettifyjson`?

jacquerie:
There might be some more places where we need identifiers. One is mentioned in #1387, others will require a careful look at the schemas.

inspirehep/modules/records/jsonschemas/records/elements/id.json
```diff
@@ -0,0 +1,94 @@
...
+        {
+            "properties": {
+                "type": {
+                    "enum": [
+                        "ORCID"
+                    ],
+                    "type": "string"
+                },
+                "value": {
+                    "pattern": "\\d{4}-\\d{4}-\\d{4}-\\d{4}",
```
kaplun: Here we need also to allow for `X` as checksum character.

inspirehep/modules/records/jsonschemas/records/elements/id.json
```diff
@@ -0,0 +1,94 @@
...
+                    "type": "string"
+                }
+            },
+            "type": "object"
+        },
+        {
+            "properties": {
+                "type": {
+                    "enum": [
+                        "HAL"
```
kaplun: As discussed IRL with @jacquerie this is actually not needed, since HAL IDs will be known only for institutions, not for people.

inspirehep/modules/records/jsonschemas/records/elements/id.json
```diff
@@ -0,0 +1,94 @@
...
+                    "type": "string"
+                },
+                "value": {
+                    "type": "string"
+                }
+            },
+            "type": "object"
+        },
+        {
+            "properties": {
```
kaplun: This makes any ID that don't match a pattern but match an enum to fall into the fallback. So basically everything.

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -177,49 +177,18 @@
...
-                        "description": "Legacy INSPIRE ID when available",
-                        "pattern": "INSPIRE-\\d+",
-                        "title": "INSPIRE ID",
-                        "type": "string"
+                    "ids": {
+                        "items": {
+                            "$ref": "elements/id.json"
+                        },
+                        "type": "array",
+                        "uniqueItems": true
```
kaplun: Note: we might check for the corresponding mapping if it's needed to make this as a nested value. Very likely not, because there is zero chance of conflict of values across different providers (i.e. ORCIDs are very different than JACOW etc.)

inspirehep/modules/records/jsonschemas/records/elements/id.json
liamkirsh: I suggested to @jacquerie to have the fallback only allow types that aren't in the preceding list, but IIRC I think he said it's better to be lenient with the values, at least initially.

jacquerie: Yeah... I was wrong : )

jacquerie:
BTW @liamkirsh you don't need to worry about this PR anymore, I'm currently integrating it in another PR I'm doing. Sadly I can't keep this commit as-is, but I will have to credit you as Co-authored-by.

PR #1395, opened 2016-08-03
schemas: thesis_supervisor -> thesis_supervisors

jacquerie:
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601705/

Doesn't yet handle all cases present in the data, but I'm going to fix those when they arrive in Sentry.

### DIFF
--- a/inspirehep/dojson/hep/fields/bd70x75x.py
+++ b/inspirehep/dojson/hep/fields/bd70x75x.py
@@ -24,37 +24,59 @@
 
 from __future__ import absolute_import, division, print_function
 
+from six.moves import zip_longest
+
 from dojson import utils
+
+from inspirehep.utils.helpers import force_force_list
+from inspirehep.utils.record import get_value
 
 from ..model import hep, hep2marc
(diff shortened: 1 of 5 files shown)


inspirehep/dojson/hep/fields/bd70x75x.py
```diff
@@ -24,37 +24,59 @@
...
+
+    FIXME: handle the presence of multiple 700__a.
+    FIXME: handle identifiers from 701__i and 701__j."""
+    def _get_affiliations(value):
+        result = []
+
+        institutions = force_force_list(value.get('u'))
+        recids = force_force_list(value.get('z'))
+
+        for value, recid in zip_longest(institutions, recids):
```
jacquerie: Oh right @mihaibivol, I wanted to point out this to you because I remember that you had a similar problem and used just `zip`.

mihaibivol: I zipped only if we had the same length in the data.

PR #1398, opened 2016-08-03
tests: ignore inspirehep module

david-caro:
* As we don't have any tests mixed in with the code, we can skip
  crawling inside the inspirehep module and shave off a few seconds
  and a bunch of useless logs from the test runs.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -23,7 +23,7 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
 [pytest]
-addopts = --pep8 --ignore=docs --ignore=node_modules --cov=inspirehep --cov-report=term-missing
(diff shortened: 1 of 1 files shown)


jacquerie:
My suspicion is that this will make the `pep8` build useless, so this PR should also add an alternative for running `pycodestyle` (`pep8`) on `inspirehep`.

jacquerie:
Oh, well, I was wrong: https://travis-ci.org/inspirehep/inspire-next/jobs/149514918.

david-caro:
do you know if there's any reason why we are running the test through distutils? (python setup.py test)

jacquerie:
1. Similarity with Invenio
2. Some separation between test deps and app deps

...but I don't care about either, so we could just use pytest. This IMO
requires to clean the mess in setup.py, which is why it hasn't happened
yet...

PR #1403, opened 2016-08-04
records: activity report document type addition

jmartinm:


inspirehep/modules/records/receivers.py
```diff
@@ -128,6 +128,9 @@ def populate_inspire_document_type(recid, json, *args, **kwargs):
                 elif element['primary'].lower() == 'report':
                     inspire_doc_type.append(element['primary'].lower())
                     break
+                elif element['primary'].lower() == 'activityreport':
+                    inspire_doc_type.append('activity report')
+                    break
```
jmartinm: Indeed, I felt quite sad while adding it.. ;) I am sure there is a cleaner way of writing this code with a `map` but went for the quick option.

PR #1407, opened 2016-08-05
WIP schemas: ORCID with letters

Dziolas:
* Changes ORCID field pattern to support letters

Signed-off-by: Wojciech Ziółek <wojciech.ziolek@cern.ch>

### DIFF
--- a/inspirehep/modules/records/jsonschemas/records/hep.json
+++ b/inspirehep/modules/records/jsonschemas/records/hep.json
@@ -197,7 +197,7 @@
                     "orcid": {
                         "description": "ORCID Id when available",
                         "format": "orcid",
-                        "pattern": "\\d{4}-\\d{4}-\\d{4}-\\d{4}",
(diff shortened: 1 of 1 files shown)


Dziolas:
@jmartinm Do you want me also to update dojson to remove "orcid:" from the beginning of string?

jacquerie:
Uhm, http://support.orcid.org/knowledgebase/articles/116780-structure-of-the-orcid-identifier says that only the last character might be an X, while all others should be digits. Did they change from this? In any case, our current pattern is wrong.

BTW have a look at #1392 to see where identifiers are goings.

jmartinm:
Do you have the example with the letters @Dziolas ?

No worries about DOJSON, we will clean it up after https://github.com/inspirehep/inspire-next/pull/1392 is merged.

Dziolas:
Ah sorry, I didn't check that. I just saw the las character being X and I assumed that there might be more. I will change it.

Dziolas:
Closing as Jacopo created new pull request already.

PR #1409, opened 2016-08-05
references: use mget instead of a parsed query

jmartinm:
* Fixes issue were the generated query is too long which throws a
  recursion depth error.

Signed-off-by: Javier Martin Montull <javier.martin.montull@cern.ch>

PR #1411, opened 2016-08-08
search: use search classes

jmartinm:


inspirehep/modules/search/api.py
```diff
@@ -0,0 +1,37 @@
...
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
```
jacquerie: Wrong license.

jmartinm:
This is a starting point to centralize the API for searching our collections.

This means that the `index` and `doc_type` is only specified here and there will be no low level calls to ES client hardcoding these parameters.

These classes inherit from Elasticsearch DSL (http://elasticsearch-dsl.readthedocs.io/en/latest/search_dsl.html#the-search-object) so all operations should be possible to be made using them.

inspirehep/modules/search/api.py
```diff
@@ -0,0 +1,163 @@
...
+
+from invenio_search.api import RecordsSearch
+
+
+class LiteratureSearch(RecordsSearch):
+    """Elasticsearch-dsl specialized class to search in Literature database."""
+
+    class Meta:
+        index = 'records-hep'
+        doc_types = 'hep'
```
jacquerie: How do we plan to inject these at configuration time?

jmartinm: hmm, what do you mean? The values can be 'hardcoded' here and this will be the only place where the index and doc type is defined.

`invenio-records-rest` knows how to extract those values from the `Meta` class just by passing the search class - https://github.com/inspirehep/inspire-next/pull/1411/files#diff-3f822b17443888990c9f70f9f89d66f6R287

What use case do you have in mind?

jacquerie: Oh, right, we could use everywhere what ES on demand requires us to use. It _could_ be cool for having distinct indices for tests and development, I guess, but it's less important.

jmartinm:
Still some current_search_client usages to be amended but you can already have a look at the changes so far in case there are comments/concerns

inspirehep/modules/search/api.py
```diff
@@ -0,0 +1,196 @@
...
+        try:
+            return self.get_record(uuid).execute().hits[0].to_dict()
+        except IndexError:
+            return {}
+
+
+class LiteratureSearch(RecordsSearch, SearchMixin):
+    """Elasticsearch-dsl specialized class to search in Literature database."""
+
+    class Meta:
```
david-caro: Shouldn't this inherit from object?

jmartinm: I am not sure if it is needed for the internal Meta class, was following the example from the parent class (https://github.com/inveniosoftware/invenio-search/blob/master/invenio_search/api.py#L49)

inspirehep/modules/search/api.py
```diff
@@ -0,0 +1,196 @@
...
+            "abstract.raw^4",
+            "author^10",
+            "author.raw^15",
+            "reportnumber^10",
+            "eprint^10",
+            "doi^10"
+        ]
+
+
+class AuthorsSearch(RecordsSearch, SearchMixin):
```
david-caro: This looks like a good case for a real python metaclass usage (so instead of repeating the same class definition over and over you can just specify the custom fields) :)

inspirehep/modules/search/api.py
david-caro: Yep, I don't really think that anyone will be subclassing this one, so it should be ok yes

inspirehep/modules/search/api.py
jmartinm: Can you put an example of how it would look like with a real metaclass? I am not that familiar with them.

david-caro: Something like:

```python
class MyMeta(type):
    def __new___(cls, name, parents, attrs):
        class Meta:
            fields = attrs.get('fields', ('*', ))
            facets = attrs.get('facets', {})
            doc_types = attrs.get('doc_types', name.lower().replace('seach', 1))

        attrs['Meta'] = Meta
        return super(InterfaceMeta, cls).__new__(cls, name, parents, dct)

class FancySearch(RecordsSearch, SearchMixin):
    __metaclass__ = MyMeta
    index = 'custom-index'
```

That should be equivalent (raw and untested) to:

```python
    class FancySearch(RecordsSearch, SearchMixin):

        class Meta:
            index = 'custom-index'
            doc_types = 'fancy'
            fields = ('*', )
            facets = {}
````

But it makes it way easier to repeat for new classes, as you don't have to add all the boilerplate.

Comment (jmartinm) [inline inspirehep/modules/search/api.py]: Nice, actually it is not much boilerplate, only the attributes to overwrite need to be specified in the `Meta` class. I have updated the PR to just leave the necessary ones so a class looks like:

```python
class InstitutionsSearch(RecordsSearch, SearchMixin):
    """Elasticsearch-dsl specialized class to search in Institutions database."""

    class Meta:
        index = 'records-institutions'
        doc_types = 'institutions'
```

inspirehep/modules/references/view_utils.py
```diff
@@ -85,7 +82,8 @@ def references(self):
                 else:
                     # Easier to use record macros over reference records.
                     if 'publication_info' in reference:
-                        reference['publication_info'] = [reference['publication_info']]
+                        reference['publication_info'] = [
+                            reference['publication_info']]
                     else:
                         reference['publication_info'] = []
```
jacquerie: This looks like a `force_force_list`, but I swear I remember already making this comment on this piece of code...

jmartinm: You made it here https://github.com/inspirehep/inspire-next/pull/1409 , that one is still pending as it might also be affected by this PR.

jmartinm:
Alright, tests passing now.

Just fighting now to convert [this query](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/theme/views.py#L326-L353) to use ES DSL.

jmartinm:
> Just fighting now to convert this query to use ES DSL.

And done.

PR #1414, opened 2016-08-08
utils: proper logging in record_getter

jacquerie:
@mihaibivol, whenever you have time to have a look at this...

### DIFF
--- a/inspirehep/utils/record_getter.py
+++ b/inspirehep/utils/record_getter.py
@@ -22,7 +22,7 @@
 
 """Resource-aware json reference loaders to be used with jsonref."""
 
-from __future__ import absolute_import, print_function
+from __future__ import absolute_import, division, print_function
 
 from functools import wraps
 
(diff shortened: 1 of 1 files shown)


inspirehep/utils/record_getter.py
```diff
@@ -55,9 +55,9 @@ def wrapper(*args):
         try:
             return f(*args)
         except Exception as e:
-            current_app.logger.error('Error in retrieving record {0} '
-                                     'caused by {1}'.format(repr(args),
-                                                            repr(e)))
+            current_app.logger.error(
+                'Error in retrieving record %s caused by %s',
```
mihaibivol: I would do something like `'Error in retrieving record %s caused by {}'.format(repr(e))` so that we have a different event for each cause of not finding an recid.

mihaibivol: Which in term will be the same thing as calling `logger.exception('Can't load recid %s', args)`. This thing will also give more context to the exception so that we filter out `PID` related things and not found errors from actual exceptions.

jacquerie:
Updated after review.

mihaibivol:
Shall We Get This Merged?

jacquerie:
Let's Get This Merged

PR #1415, opened 2016-08-08
dojson: normalize degree_type

jacquerie:
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601706/

This is the partial status of the `502__b` field (I interrupted a query that had been running for a few hours...):
```
   1 Dubna, JINR
   1 Licencia
   1 Maitrise Communication
   1 PDF
   1 Ph.D Thesis
   1 master
   2 Internship report
   2 PHD
   2 Rapport de stage
   3 phD
   3 thesis
   4 Licentiate
   4 Mainz U.
   6 Licenciate
   7 Phd
  85 Habilitation
  99 Laurea
 160 Bachelor
 412 Diploma
1063 Thesis
1159 Master
13047 PhD
```

This PR adds a little normalization to the DoJSON rule so that less values will result in invalid records. It's not a complete solution because I don't know if we should ignore the rest of the cases, or we want to have "Internship report"s and similar entries.

### DIFF
--- a/inspirehep/dojson/hep/fields/bd5xx.py
+++ b/inspirehep/dojson/hep/fields/bd5xx.py
@@ -91,12 +91,29 @@ def hidden_note2marc(self, key, value):
 
 
 @hep.over('thesis', '^502..')
-@utils.filter_values
 def thesis(self, key, value):
     """Get Thesis Information."""
+    DEGREE_TYPES_MAP = {
+        'RAPPORT DE STAGE': 'Internship Report',
+        'INTERNSHIP REPORT': 'Internship Report',
+        'DIPLOMA': 'Diploma',
+        'BACHELOR': 'Bachelor',
+        'LAUREA': 'Laurea',
+        'MASTER': 'Master',
+        'THESIS': 'Thesis',
+        'PHD': 'PhD',
+        'PDF': 'PhD',
+        'PH.D. THESIS': 'PhD',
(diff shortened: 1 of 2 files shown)


kaplun:
@inspirehep/inspire-content which values should we accept for Thesis type?

annetteholtkamp:
I’ve corrected a lot of the wrong entries. 
Normally we don’t take internship reports. I’ve kept 2 since they seem interesting. 

- Annette

jacquerie:
> Normally we don’t take internship reports. I’ve kept 2 since they seem interesting.

Ok, will add "Internship Report" to the enum, and a conversion from "Rapport de stage" to it.

What about adding a catch all "Other"?

inspirehep/dojson/hep/fields/bd5xx.py
```diff
@@ -91,12 +91,32 @@ def hidden_note2marc(self, key, value):
...
+        'Laurea': 'Laurea',
+        'master': 'Master',
+        'Master': 'Master',
+        'thesis': 'Thesis',
+        'Thesis': 'Thesis',
+        'phD': 'PhD',
+        'Phd': 'PhD',
+        'PhD': 'PhD',
+        'PHD': 'PhD',
+        'PDF': 'PhD',
```
kaplun: What about upper-casing the input first, so that you automatically catch all case-related inconsistencies?

PR #1419, opened 2016-08-10
travis: fix build

jacquerie:
After https://www.traviscistatus.com/incidents/2p40l49r3yxd the Travis people updated the Ubuntu Trusty images to a newer version:

> We've determined that an image cleanup process malfunctioned and deleted the stable build environment images. Due to various factors our only option at this point is roll forward with newer build environment images, for both Precise and Trusty.

1. This new Trusty image ships with the `1.12.0-0~trusty` version of `docker-engine`. Our Travis script was trying to downgrade it to `1.10.3-0~trusty`, which was failing: https://travis-ci.org/inspirehep/inspire-next/jobs/151099253#L422. The reason why escapes me, but I just decided to bump the `docker-engine` we require on Travis.
2. This new Trusty image ships *without* a `docker-compose` command. Our Travis script was trying to delete it, which was failing: https://travis-ci.org/inspirehep/inspire-next/jobs/151100896#L288. The fix here is just to add a `-f` flag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,8 +33,8 @@ services:
 env:
   global:
     - DOCKER_DATA="$HOME/docker_data"
-    - DOCKER_VERSION=1.10.3-0~trusty
-    - DOCKER_COMPOSE_VERSION=1.6.2
+    - DOCKER_VERSION=1.12.0-0~trusty
+    - DOCKER_COMPOSE_VERSION=1.8.0
   matrix:
     - SUITE=pep8
     - SUITE=unit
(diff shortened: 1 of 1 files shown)


jacquerie:
The build is now fixed. A full post-mortem is contained in the first comment to this PR.

Strictly speaking the workaround in https://graysonkoonce.com/managing-docker-and-docker-compose-versions-on-travis-ci/ that we used is no longer needed, since Travis now ships with a recent enough version of Docker. But I decided to keep it, since it feels more robust to future changes.

I'm also not sure if the image already contains a `docker-compose` command somewhere else than `/usr/local/bin` (we could run a dummy build with `which docker-compose` to determine that), but I decided to keep the `rm` command because it felt more robust.

People running Docker on their laptops (@davidcaro, @mihaibivol, @jmartinm, and maybe someone else): upgrade Docker to 1.12.0 and Docker-Compose to 1.8.0 to be on par with Travis.

PR #1423, opened 2016-08-10
dojson: earliest_date in ES only

kaplun:
* Amends earliest_date receiver to be triggered only at index time,
  rather than before_record_* time, thus adding earliest_date bit
  only in ES (it's not a property that is supposed to be touched
  by humans).

### DIFF
--- a/inspirehep/dojson/hep/receivers.py
+++ b/inspirehep/dojson/hep/receivers.py
@@ -26,16 +26,15 @@ from __future__ import absolute_import, division, print_function
 
 from itertools import chain
 
-from invenio_records.signals import before_record_insert, before_record_update
+from invenio_indexer.signals import before_record_index
 
 from inspirehep.utils.date import create_earliest_date
 from inspirehep.utils.helpers import force_force_list
 from inspirehep.utils.record import get_value
 
 
-@before_record_insert.connect
-@before_record_update.connect
-def earliest_date(sender, *args, **kwargs):
+@before_record_index.connect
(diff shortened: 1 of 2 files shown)


jmartinm:
👍

inspirehep/dojson/hep/receivers.py
```diff
@@ -26,15 +26,14 @@
...
 
 from inspirehep.utils.date import create_earliest_date
 from inspirehep.utils.helpers import force_force_list
 from inspirehep.utils.record import get_value
 
 
-@before_record_insert.connect
-@before_record_update.connect
+@before_record_index.connect
 def earliest_date(sender, *args, **kwargs):
```
jacquerie: The two signals have different signatures. I believe this has to be `def earliest_date(sender, json *args, **kwargs)`.

jacquerie: ...And consequently stuff should be modified on `json`, not on `sender`, in the body of the method.

kaplun: Mmh... But I checked https://github.com/inveniosoftware/invenio-indexer/blob/master/invenio_indexer/signals.py#L33

Isn't the docstring implying that json is the first argument passed?

kaplun: Answering myself: nope :cry: 
https://github.com/inveniosoftware/invenio-indexer/blob/master/invenio_indexer/signals.py#L33

PR #1426, opened 2016-08-10
mappings: add not_analyzed to uuid

jacenkow:
* Updates HEP schema to include uuid field in authors object.


jacenkow:
Resolves problems in #1353, which led to not finding signatures by their `uuid`s.

PR #1429, opened 2016-08-11
holdingpen: hide actions if workflow complete

david-caro:


jmartinm:
By looking at the code, looks 👍

david-caro:
I've just found that there's some cases where the workflow is finished but it's not marked as completed, I'm investigating though I don't really think that the issue should be fixed here though.

jmartinm:
Ok, just give the green light here when you think its ready and we can merge this PR.

david-caro:
Found and fixed the issue somewhere else in the invenio-workflows code, this can be mereged :)

PR #1434, opened 2016-08-11
holdingpen: hide resume if pending approval

david-caro:
* It makes no sense to resume a workflow if it requires manual
  approval, so just hide the button.


david-caro:
I think that the current condition is the correct one, because when we halt a workflow it's because it needs human intervention, otherwise it is in ERROR, WAITING (if waiting to be started/run) or COMPLETED. This affects any workflow (currently authors and articles).

david-caro:
moved to a bigger pr (with some other small fixes #1443 )

PR #1226, opened 2016-06-20
WIP migrator: Holding Pen import scripts

jalavik:
* Will add migration scripts for INSPIRE 2 (labs) Holding Pen to INSPIRE 3.
* Adds general integration with invenio-migrator

PR #1435, opened 2016-08-11
Workflows migration

jmartinm:
Superseeds https://github.com/inspirehep/inspire-next/pull/1226

### DIFF
--- a/inspirehep/modules/authors/dojson/fields/updateform.py
+++ b/inspirehep/modules/authors/dojson/fields/updateform.py
@@ -170,8 +170,8 @@ def institution_history(self, key, value):
             "status": "current" if position["current"] else "",
             "start_date": position["start_year"],
             "end_date": position["end_year"],
-            "email": position["email"],
-            "old_email": position["old_email"],
+            "email": position.get("email", ""),
+            "old_email": position.get("old_email", ""),
             "rank": position["rank"] if position["rank"] != "rank" else ""
         })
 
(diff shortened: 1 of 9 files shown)


jmartinm:
Rebased and updated this PR. I have tested successfully to load into the DB the dump from labs.

The legacy HP entries do not go into ElasticSearch because their `data_type` does not match the configuration of types that go into ES:

```python
In [1]: app.config['WORKFLOWS_UI_DATA_TYPES']
Out[1]:
{'authors': {'search_index': 'holdingpen-authors', 'search_type': 'authors'},
 'hep': {'search_index': 'holdingpen-hep', 'search_type': 'hep'}}
```

This is a good thing though because sending them to ES causes errors due to the data model inside the workflow objects not being compatible with the [mappings](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/mappings/holdingpen/authors.json).

I think we can update the Holding Pen mappings so that they reflect the new data model and for older HP entries they will be in the database but will not appear in the HP search.

jmartinm:
After checking with @jalavik we need to:

* [ ] Merge the commits related to the dump functionality that are in `production` branch into this PR.

* [ ] Remove current 500 record limit (https://github.com/inspirehep/inspire-next/commit/896241ab9674a7105386dba3a07af04f904a1c7b#diff-eb226742d3dff161b6730019a569a325R63). We need to be careful not to load too many rows into memory and run out of it.

kaplun:
So this PR can in principle be merged?

jmartinm:
Not yet, I am currently working on it to solve existing issues on the dumping and loading.

jmartinm:
This is now ready as far as I can see.

Issue #1431, opened 2016-08-11
Facet count mismatch after searching for author


jmartinm:
![facet](https://cloud.githubusercontent.com/assets/580727/17581760/f64661d6-5fa7-11e6-8566-f424097d5e53.gif)

At first sight it seems to happen only after searching for an author and then using any facet.

Could be related to https://github.com/inspirehep/inspire-next/issues/1425

Issue #1371, opened 2016-07-27
global: correct logging


mihaibivol:
As part of #1370 we experimented with the current install of `Sentry` and discovered how we should write logging messages so that `Sentry` aggregates them into a single event without spamming everyone (that much).

We will be trying to use `Sentry` as the main point of monitoring the health of the legacy migration and if it's not to spammy it would be cool to actually use it on QA.

TODOs:

* [ ] Define Sentry Friendly Log messages
* [ ] Fix all logging messages to be Sentry Friendly

jacquerie:
Duplicated Sentry entries caused by improper logging:

* https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601703/
* https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601692/
* https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601849/
* https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601838/
* https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601837/
* [...]

mihaibivol:
## Which logger to use:

Use per module loggers:

```
logger = logging.getLogger(__name__)
```

or, in Celery tasks

```python
logger = get_task_logger(__name__)
```

## Log in a Sentry Friendly way.


### Exceptions
If you catch an exception always log like this:

```python
try:
   failing_code()
except ExceptionClass:
    logger.exception('Some message')
    # treatment of exception
```

The logger wil automatically output the exception info so there's no need to be explicit about it

### Errors

For making Sentry group your exceptions by the message you have to use the following convetion:

```python
logger.error('I have an error %s %s', arg1, arg2)
```

Behind the scenes, the logger will do:
```
message % args
```

### Things that do not work

```python
logger.error('I have an error {} {}', arg1, arg2)
```
The logger wants to do `msg % args`

```python
logger.error('I have an error {} {}'.format(arg1, arg2))
```
Sentry won't know how to group since it won't receive any pattern

```python
logger.error('I have an error', arg1, arg2)
```
The logger won't do the `msg % args` thing

kaplun:
@mihaibivol can you put this piece of documentation into the official docs?

mihaibivol:
[...]

PR #1436, opened 2016-08-11
search: authors exact search

### LINKED ISSUE #1425: author search with and without quotes gives different results
author:A.D.Johnson.1 and author:"A.D.Johnson.1" give different results at the moment. This should not be the case and the behaviour should be the same as in production.

@Panos512 can you have a look?

Panos512:
* Converts `author-search` without double quotes to an `exact-match` search.
  (closes #1425 and #1431)

### DIFF
--- a/inspirehep/modules/records/mappings/records/hep.json
+++ b/inspirehep/modules/records/mappings/records/hep.json
@@ -111,13 +111,13 @@
                             ],
                             "type": "string"
                         },
-                        "inspire_bai": {
-                            "copy_to": [
-                                "author",
-                                "exactauthor"
-                            ],
-                            "index": "not_analyzed",
-                            "type": "string"
+                        "ids": {
+                            "properties": {
+                                "value": {
+                                    "index": "not_analyzed",
+                                    "type": "string"
+                                }
+                            }
                         },
                         "name_suggest": {
                             "analyzer": "simple",
--- a/inspirehep/modules/search/query_factory.py
+++ b/inspirehep/modules/search/query_factory.py
@@ -75,7 +75,7 @@ def inspire_query_factory():
             ))
         finally:
             if current_app.debug:
-                current_app.logger.info(json.dumps(query.to_dict(), indent=4))
+                current_app.logger.debug(json.dumps(query.to_dict(), indent=4))
             return query
 
     return invenio_query
--- a/inspirehep/modules/search/walkers/elasticsearch.py
+++ b/inspirehep/modules/search/walkers/elasticsearch.py
@@ -26,6 +26,7 @@
 from operator import or_
 
 from elasticsearch_dsl import Q
+from flask import current_app
 
 from invenio_query_parser.ast import (AndOp, DoubleQuotedValue, EmptyQuery,
                                       GreaterEqualOp, GreaterOp, Keyword,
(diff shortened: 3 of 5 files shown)


inspirehep/modules/search/walkers/elasticsearch.py
```diff
@@ -118,11 +118,16 @@ def query(keyword):
 
             fields = self.get_fields_for_keyword(keyword, mode='a')
             if fields == ['authors.full_name', 'authors.alternative_name']:
```
jmartinm: I am wondering here, is it more robust to check these 2 fields, or better to check the `author` key (from https://github.com/inspirehep/inspire-next/blob/master/inspirehep/config.py#L167).

In that way we can change where the `author` keyword will look but always keep the exact match behaviour.

kaplun: This is part of a series of refactoring and design decisions that we need to sit down and make straight. Hopefully part of @jacquerie's future journey in the mapping world.

Meanwhile, :+1: to @jmartinm comment.

inspirehep/modules/search/query_factory.py
```diff
@@ -75,6 +75,7 @@ def invenio_query(pattern, search):
             ))
         finally:
             if current_app.debug:
+                current_app.logger.info(query)
```
jacquerie: Isn't one of these two lines redundant?

inspirehep/modules/search/walkers/elasticsearch.py
```diff
@@ -117,12 +119,17 @@ def query(keyword):
...
-                    Q("match", authors__name_variations=str(node.value)),
-                    Q("match", authors__full_name=str(node.value)),
-                    Q("match", authors__inspire_bai=str(node.value))
-                ])
+            if fields == current_app.config['SEARCH_ELASTIC_KEYWORD_MAPPING']['author']:
+                return Q(
+                    'bool',
+                    must=Q('bool', should=[
+                        Q("match", authors__name_variations=str(node.value)),
+                        Q("term", authors__inspire_bai=str(node.value))
```
jacquerie: This no longer works: BAIs were moved to the `ids` field.

inspirehep/modules/search/query_factory.py
kaplun: shouldn't it be `current_app.logger.debug(query)`?

jacquerie: Ah, yeah, the log level should definitely be `debug`.

inspirehep/modules/search/walkers/elasticsearch.py
Panos512: so, should we make `ids` not analyzed in order to be able to match them with `term` queries? 

Also we don't need to copy to `author` and `exactauthor` anymore?

kaplun: IDS indeed should never be analyzed. They have to be taken as a literal string.

inspirehep/modules/records/mappings/records/hep.json
```diff
@@ -111,13 +111,13 @@
...
-                            ],
-                            "index": "not_analyzed",
-                            "type": "string"
+                        "ids":{
+                          "properties": {
+                            "value": {
+                              "index": "not_analyzed",
+                              "type": "string"
+                            }
+                          }
```
jacquerie: This change should be in its own commit.

jacquerie: Oh, and run `scripts/prettifyjson`.

inspirehep/modules/search/query_factory.py
```diff
@@ -75,7 +75,7 @@ def invenio_query(pattern, search):
             ))
         finally:
             if current_app.debug:
-                current_app.logger.info(json.dumps(query.to_dict(), indent=4))
+                current_app.logger.debug(json.dumps(query.to_dict(), indent=4))
```
jacquerie: Not relevant for this PR, but please read this writeup by @mihaibivol on how to do proper logging: https://github.com/inspirehep/inspire-next/issues/1371.

inspirehep/modules/search/walkers/elasticsearch.py
```diff
@@ -27,6 +27,8 @@
 
 from elasticsearch_dsl import Q
```
jacquerie: No space needed here.

inspirehep/modules/search/walkers/elasticsearch.py
```diff
@@ -27,6 +27,8 @@
```
jacquerie: It would be cool to add a preparatory commit that adds the `__future__` block and fixes the license. See: https://github.com/inspirehep/inspire-next/pull/1500/commits/e399aa155af1131c222c54909e671df630fe3d4d

jacquerie:
:+1:, it's almost there.

Panos512:
@jacquerie done! 👯

PR #1437, opened 2016-08-11
schemas: make australia part of australasia

jacquerie:
As per https://github.com/inspirehep/inspire/issues/199#issuecomment-239185808.

### DIFF
--- a/inspirehep/dojson/jobs/fields/bd1xx.py
+++ b/inspirehep/dojson/jobs/fields/bd1xx.py
@@ -96,7 +96,7 @@ def regions(self, key, value):
         'AF': 'Africa',
         'Africa': 'Africa',
         'Asia': 'Asia',
-        'Australia': 'Australia',
+        'Australia': 'Australasia',
         'Australasia': 'Australasia',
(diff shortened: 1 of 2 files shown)


kaplun:
@inspirehep/inspire-content Can't we have this occasion to simply improve the model to ask for supporting a specific location? (possibly up to a (list of) full addresses - and corresponding geocoding coordinates)?

Then the region can be simply derived from it. In this way we would be consistent with the Conference and Insittution database and we could place jobs right on the map.

jacquerie:
This would be cool, but we haven't generally solved the problem of integrating a geocoding solution in INSPIRE, so I believe that it can be pushed back...

kaplun:
Pushed back by whom? In principle we can happily use Google APIs since we are not receiving more than a handful of conference entries per month :)

What I am proposing here is simply to prepare the schema to welcome a list of addresses and corresponding geocoding to be ready for when we introduce forms to let user select such pieces of info.

But yep this is the subject of a new Github issue.

kaplun:
LGTM

PR #1441, opened 2016-08-12
search: enable _all in secondary collections

jmartinm:
* Enables _all in ES mappings for secondary collections and makes this
  field the default one to search on when no keywords are specified.s

### DIFF
--- a/inspirehep/modules/records/mappings/records/authors.json
+++ b/inspirehep/modules/records/mappings/records/authors.json
@@ -2,7 +2,7 @@
     "mappings": {
         "authors": {
             "_all": {
-                "enabled": false
+                "enabled": true
             },
             "date_detection": false,
             "dynamic_templates": [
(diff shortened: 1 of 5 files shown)


jmartinm:
This will return results in all collections (which is currently not the case) and for secondary collections should be quite accurate (using default ES).

inspirehep/modules/records/mappings/records/authors.json
```diff
@@ -2,7 +2,7 @@
     "mappings": {
         "authors": {
             "_all": {
-                "enabled": false
+                "enabled": true
```
jacquerie: This will probably be reverted, because we expect people to want high relevance when searching for people.

jacquerie:
ES reference: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-all-field.html.

This is a good idea for having _some_ results, but not the best results. It might even be enough for the five minor collections (Conferences, Experiments, Institutions, Jobs, Journals), and it is certainly better than nothing on these five.

I don't expect this to be enough for Authors, but we can care later.

PR #1443, opened 2016-08-12
Holdingpen fixes

david-caro:
Some fixes to adapt to the latest workflows for invenio 3

### DIFF
--- a/inspirehep/modules/workflows/tasks/submission.py
+++ b/inspirehep/modules/workflows/tasks/submission.py
@@ -261,7 +261,7 @@ def send_robotupload(url=None,
                 )
             )
 
-        if in_production_mode():
+        if not in_production_mode():
             return
 
         result = make_robotupload_marcxml(
--- a/inspirehep/modules/workflows/tasks/submission.py
+++ b/inspirehep/modules/workflows/tasks/submission.py
@@ -25,6 +25,7 @@
 from __future__ import absolute_import, division, print_function
 
 import os
+import logging
 from functools import wraps
 
 from flask import current_app, render_template
@@ -33,6 +34,9 @@ from retrying import retry
 
 from ....utils.tickets import get_instance, retry_if_connection_problems
 from .actions import in_production_mode
+
+
+LOGGER = logging.getLogger(__name__)
 
 
 @retry(
@@ -250,15 +254,11 @@ def send_robotupload(url=None,
 
         if current_app.debug:
             # Log what we are sending
-            current_app.logger.debug(
-                "Going to robotupload {mode} to {url}:\n{marcxml}\n".format(
-                    url=url,
-                    marcxml=marcxml,
-                    callback_url=combined_callback_url,
-                    mode=mode,
-                    nonce=obj.id,
-                    priority=5,
-                )
+            LOGGER.debug(
+                "Going to robotupload %s to %s:\n%s\n",
+                mode=mode,
+                url=url,
+                marcxml=marcxml,
             )
 
         if not in_production_mode():
--- a/inspirehep/modules/workflows/tasks/submission.py
+++ b/inspirehep/modules/workflows/tasks/submission.py
@@ -31,7 +31,8 @@ from flask import current_app, render_template
 from invenio_accounts.models import User
 from retrying import retry
 
-from inspirehep.utils.tickets import get_instance, retry_if_connection_problems
+from ....utils.tickets import get_instance, retry_if_connection_problems
+from .actions import in_production_mode
 
 
 @retry(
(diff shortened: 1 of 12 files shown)


inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -248,8 +248,8 @@ def _send_robotupload(obj, eng):
         marc_json = marcxml_processor.do(data)
         marcxml = legacy_export_as_marc(marc_json)
 
-        if not url and current_app.debug:
-            # Log what we would send
+        if current_app.debug:
+            # Log what we are sending
             current_app.logger.debug(
                 "Going to robotupload {mode} to {url}:\n{marcxml}\n".format(
```
jacquerie: Mh, time to fix this logger? See: #1371

setup.py
```diff
@@ -178,7 +178,6 @@
             'hepnames2marc = inspirehep.dojson.hepnames2marc:hepnames2marc',
         ],
         'invenio_base.api_apps': [
-            'inspire_theme = inspirehep.modules.theme:INSPIRETheme',
```
jacquerie: I'm not sure if this is correct, because some Jinja filters are run when sending stuff from the API... let's see what the test suite says.

david-caro: Actually, without this you can't run a search on the main page (it fails to find the template), though none of the tests fail :/

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -248,28 +252,26 @@ def _send_robotupload(obj, eng):
...
-                    callback_url=combined_callback_url,
-                    mode=mode,
-                    nonce=obj.id,
-                    priority=5,
-                )
-            )
-            return
-        else:
-            result = make_robotupload_marcxml(
+        if current_app.debug:
```
kaplun: Is this condition needed since it's anyway `LOGGER.debug()`?

inspirehep/modules/workflows/templates/inspire_workflows/index.html
```diff
@@ -77,13 +74,34 @@ <h3 class="text-center header-text">
...
         section_title="arxiv harvests">
       </holding-pen-dashboard-item>
 
     </div>
+
+    <div class="col-md-3 col-sm-6">
+
+      <holding-pen-dashboard-item
+        template="{{url_for('static', filename='js/inspire_workflows_ui/templates/dashboard_stat_block.html')}}"
+        filter_string="source=Alpha experiment"
```
kaplun: Uh? What is this dashboard?

kaplun: Leftover of a test?

david-caro: Not really, this is one of the sources that are crawled by the crawler, I added it because otherwise I would not be able to see the submissions on the dashboard (I think @eamonnmag is planning on generating them on the fly soon).

inspirehep/modules/workflows/tasks/submission.py
david-caro: I'm not sure, maybe @jacquerie knows?

jacquerie: AFAIK application in debug mode and log level of the logger are independent. They just share the same name.

inspirehep/modules/workflows/templates/inspire_workflows/index.html
kaplun: There are nearly 20 sources. It might then not be too feasible to then present all of them separately.
E.g. many of them are actually simply harvesting of theses from different websites, so maybe they can be presented in one place.

Also the Alpha experiment is going to publish probably a handful of publications and that's it. So the box is going to have a very brief life.

How difficult will it be to amend these boxes? Could it be worth pinging catalogers to ask what they need? @eamonnmag?

setup.py
jacquerie: Ach, that's because we just ask the page to be rendered, but we make no assertion on the data that populates it.

inspirehep/modules/workflows/templates/inspire_workflows/index.html
eamonnmag: It's like this now, so as separate dashboard items, so that you can choose which blocks to add, this is as configurable as it gets for now. 
The submission area below will hold all stats however. I will get around to adding this ASAP. I already have the data, just need to write the visualization.

david-caro: +1 for asking

inspirehep/modules/workflows/tasks/submission.py
kaplun: But in principle we could connect them, by simply tuning the logger depending on the current_app.debug flag. In this way your code doesn't have to check all the times if you are in debug mode or not. It's the logger that decides how to filter.

I think we should leverage on `invenio-logging`:
https://github.com/inveniosoftware/invenio-logging/blob/master/invenio_logging/fs.py#L46

kaplun: Better:
https://github.com/inveniosoftware/invenio-logging/blob/master/invenio_logging/sentry.py#L79

david-caro: That looks like it requires a pr of it's own to change it everywhere.

kaplun:
I guess we should integrate this branch is good to go, right?

david-caro:
There's still a bunch of things in the air (like populating the keywords) and making sure that the file buckets and such are used correctly, but I guess that this can go in for now even if it's not functionally complete

kaplun:
OK, so I'll merge it.

PR #1453, opened 2016-08-17
WIP schemas: refresh authors schema

jacquerie:


inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -223,30 +199,6 @@ def native_name2marc(self, key, value):
     }
 
 
-@hepnames.over('private_current_emails', '^595..')
-@utils.for_each_value
-def private_current_emails(self, key, value):
```
kaplun: Ahem, maybe better adding it to the schema.

kaplun: Better put them with other emails with a flag that they should stay hidden.

inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -361,34 +313,6 @@ def prizes2marc(self, key, value):
     }
 
 
-@hepnames.over('_public_note', '^680..')
-@utils.for_each_value
-def _public_note(self, key, value):
```
kaplun: There are actually public notes. We should add it back to the schema. Actually this is yet a meta-field that is valuable in all schemas.

inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -361,34 +313,6 @@ def prizes2marc(self, key, value):
...
-
-@hepnames2marc.over('680', '^_public_note$')
-@utils.for_each_value
-def _public_note2marc(self, key, value):
-    return {
-        'i': value
-    }
-
-
-@hepnames.over('_curators_note', '^667..')
```
kaplun: This is a non public note. We should also add it as a meta-field to all schemas. They are useful for cataloguers to write down notes.

jacquerie:
All those fields (except `public_notes`) are not used for searching: they are either metadata useful for cataloguers or data that we want to be hidden. Adding those means dealing with new Sentry issues due to the inevitable crappiness of the data, pushing back the "Enabling Search" milestone. They belong to the "All Editing on Labs" milestone, so I believe we should deal with them later.

jacquerie:
Oh, and `public_notes` should be dealt once and for all in all collections that need it in `inspirehep/dojson/common/base.py`.

inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -43,30 +43,6 @@
 NON_DIGIT = re.compile('[^\d]+')
 
 
-@hepnames.over('acquisition_source', '^541[10_].')
```
david-caro: Can you elaborate a bit? (maybe create an issue if not there yet, I tend to forget stuff :older_man: )

jacquerie:
> All those fields (except public_notes) are not used for searching: they are either metadata useful for cataloguers or data that we want to be hidden. Adding those means dealing with new Sentry issues due to the inevitable crappiness of the data, pushing back the "Enabling Search" milestone. They belong to the "All Editing on Labs" milestone, so I believe we should deal with them later.

Let me correct this: I am ok with adding them back to the schema, as long as someone else takes care of correcting the data issues that will pop up on Sentry.

jacquerie:
Mostly superseded by #1596 and #1623

PR #1457, opened 2016-08-17
dojson: make other_names flat

jacquerie:
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601596/

Previously the rule was producing an array of arrays, which did not
conform to the schema. Now the rule produces a flat array.


jacquerie:
Ready for review.

PR #1458, opened 2016-08-17
dojson: split multiple thesis supervisors

### LINKED ISSUE #1259: Thesis supervisor to be merged into authors
By integrating [Casrai CRediT](http://dictionary.casrai.org/Contributor_Roles) into the author role field, we can actually conflate the thesis supervisor into the author field (by specifying the `supervision` role).

In this way the `thesis_supervisor` field (i.e. 701) can be simplified.

jacquerie:
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601938/

A `701` field can contain multiple `a` subfields. A reasonable
interpretation is that they all share the same affiliation(s)
contained in the `u` subfields of the field. This code assumes
this and creates a `thesis_supervisor` per `701__a` field it
finds.

By using the `contributor_roles` added in cf74d6 we can specify
the fact that an author contributed to some work as a supervisor.
This means that the `thesis_supervisor` key is redundant, and
can be merged inside of `authors`.  (Closes #1259)

Also adds some logic to the BibTeX exporter in order not to add
the thesis supervisors as authors. This addition required some
refactoring to the `get_author` method, which removed the support
for multiple `full_name`s, since it should never happen anyway.

### DIFF
--- a/inspirehep/dojson/hep/fields/bd70x75x.py
+++ b/inspirehep/dojson/hep/fields/bd70x75x.py
@@ -36,12 +36,16 @@ from ...utils import get_record_ref
 
 
 @hep.over('thesis_supervisors', '^701..')
-@utils.for_each_value
 def thesis_supervisors(self, key, value):
     """Thesis supervisors.
 
-    FIXME: handle the presence of multiple 700__a.
     FIXME: handle identifiers from 701__i and 701__j."""
+    def _get_thesis_supervisor(a_value, value):
+        return {
+            'affiliations': _get_affiliations(value),
+            'full_name': a_value,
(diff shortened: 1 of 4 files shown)


kaplun:
While doing that please see #1259

jacquerie:
Ok, will add another commit on top of this to handle that.

jacquerie:
Ready for review.

jacquerie:
Ready for review.

jacquerie:
I fat-fingered a force push to `origin`, which is why the build looks as failing. I'm still going to merge because the PR build passed.

PR #1460, opened 2016-08-18
dojson: email -> emails

jacquerie:
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601540/

### DIFF
--- a/inspirehep/dojson/hep/fields/bd1xx.py
+++ b/inspirehep/dojson/hep/fields/bd1xx.py
@@ -27,8 +27,6 @@ from __future__ import absolute_import, division, print_function
 import logging
 import re
 
-from six.moves import zip_longest
-
 from dojson import utils
 
 from ..model import hep, hep2marc
@@ -59,11 +57,17 @@ def authors(self, key, value):
             u_values = force_force_list(value.get('u'))
             z_values = force_force_list(value.get('z'))
 
-            for u_value, z_value in zip_longest(u_values, z_values):
-                result.append({
-                    'record': get_record_ref(z_value, 'institutions'),
-                    'value': u_value,
-                })
+            # XXX: we zip only when they have the same length, otherwise
+            #      we might match a value with the wrong recid.
+            if len(u_values) == len(z_values):
+                for u_value, z_value in zip(u_values, z_values):
+                    result.append({
+                        'record': get_record_ref(z_value, 'institutions'),
+                        'value': u_value,
+                    })
+            else:
+                for u_value in u_values:
+                    result.append({'value': u_value})
 
             return result
 
--- a/inspirehep/dojson/hep/fields/bd1xx.py
+++ b/inspirehep/dojson/hep/fields/bd1xx.py
@@ -145,7 +145,7 @@ def authors(self, key, value):
             'affiliations': _get_affiliations(value),
             'alternative_names': force_force_list(value.get('q')),
             'curated_relation': value.get('y') == '1',
-            'email': value.get('m'),
+            'emails': force_force_list(value.get('m')),
             'full_name': _get_full_name(value),
             'ids': _get_ids(value),
             'record': _get_record(value),
@@ -179,7 +179,7 @@ def authors2marc(self, key, value):
             'q': value.get('alternative_names'),
             'i': value.get('inspire_id'),
             'j': value.get('orcid'),
-            'm': value.get('email'),
+            'm': value.get('emails'),
             'u': affiliations,
             'x': get_recid_from_ref(value.get('record')),
             'y': value.get('curated_relation')
--- a/inspirehep/dojson/jobs/fields/bd1xx.py
+++ b/inspirehep/dojson/jobs/fields/bd1xx.py
@@ -144,13 +144,22 @@ def institutions(self, key, value):
     a_values = force_force_list(value.get('a'))
     z_values = force_force_list(value.get('z'))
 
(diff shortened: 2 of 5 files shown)


tests/unit/dojson/test_dojson_hep.py
```diff
@@ -733,6 +733,52 @@ def test_authors_from_100__a_u_x_w_y_z_with_malformed_x():
...
+        '</datafield>'
+    )  # record/413614/export/xme
+
+    expected = [
+        {
+            'affiliations': [
+                {
+                    'record': {
+                        '$ref': 'http://localhost:5000/api/institutions/903393',
+                    },  # XXX: wrong, fix incoming.
```
jacquerie: Just noticed that the thing I did with `zip_longest` is completely wrong, and @mihaibivol was right in zipping only when the `len`s match.

jacquerie:
First commit replaces the uses of `zip_longest` with `zip`, and zips only when the two parallel lists of values and recids have the same length: otherwise we might end up matching a value with the wrong recid.

Second commit allows an author of a paper to have more than one email, since that's true in the [data](https://github.com/inspirehep/inspire-next/blob/25296a555b052951455acee34b09cdecfa63702f/inspirehep/dojson/current_marcxml_usage/hep-repeatable.txt#L111).

Ready for review.

inspirehep/dojson/hep/fields/bd1xx.py
```diff
@@ -59,11 +57,15 @@ def _get_affiliations(value):
...
-                    'value': u_value,
-                })
+            if len(u_values) == len(z_values):
+                for u_value, z_value in zip(u_values, z_values):
+                    result.append({
+                        'record': get_record_ref(z_value, 'institutions'),
+                        'value': u_value,
+                    })
+            else:
+                for u_value in u_values:
```
kaplun: I know why you are doing it, but I am not sure I will remember it in 1 month. Can you add a comment to explain this?

kaplun:
:+1: for me

inspirehep/dojson/hep/fields/bd1xx.py
jacquerie: Future @kaplun! Is the git commit message enough? https://github.com/inspirehep/inspire-next/pull/1460/commits/0cf0914772d730000d22014cb74d64128b2b4f2a : )

inspirehep/dojson/hep/fields/bd1xx.py
```diff
@@ -59,11 +57,15 @@ def _get_affiliations(value):
...
+                    result.append({'value': u_value})
```
david-caro: this is missing the case where `v_value` is longer than `u_value` no? (that seemed to be taken care of before)

jacquerie: `z_values` can never be longer than `u_values`, because they are derived from them via a search, so we don't need to treat that case.

jacquerie:
BTW the first commit highlights the fact that we could use a DoJSON util in INSPIRE to generally solve the problem of populating an affiliation.

inspirehep/dojson/hep/fields/bd1xx.py
kaplun: Meh, I usually log at git log only when I want to blame, not to understand what some code does :stuck_out_tongue:

inspirehep/dojson/jobs/fields/bd1xx.py
```diff
@@ -144,13 +144,20 @@ def institutions(self, key, value):
...
+                'curated_relation': record is not None,
+                'name': a_value,
+                'record': record,
+            })
+    else:
+        for a_value in a_values:
+            institutions.append({
+                'curated_relation': False,
+                'name': a_value,
+            })
```
david-caro: Same here, the case where z_values is longer is not handled (not sure if that's intended, that's why I point it out)

david-caro:
Make sure that this is not a miss: https://github.com/inspirehep/inspire-next/pull/1460/files#diff-8f850f5dde975afda8a95a3023d5005cR84

david-caro:
dammit... let me get the line

david-caro:
https://github.com/inspirehep/inspire-next/pull/1460/files#diff-8f850f5dde975afda8a95a3023d5005cR84

david-caro:
Meeeh...  if the link does not work for you, it's file bd1xx.py line 84.

inspirehep/dojson/hep/fields/bd1xx.py
david-caro: :+1: for good behavior!

jacquerie:
About line 84 in jobs/bd1xx.py: yeah, that's wrong, and it doesn't pop up on Sentry because we import only the public Jobs.

kaplun:
@jacquerie you should just use also the prodsync on the nightly. There shouldn't be any difference. (beside hidden fields and record)

jacquerie:
> you should just use also the prodsync on the nightly. There shouldn't be any difference. (beside hidden fields and record)

@kaplun in order to do that we have to solve the problem of connecting that box to AFS

jacquerie:
Apologies for rushing this, but I didn't want to waste a nightly.

kaplun:
Well for the time being just copy once the prodsync.tar file it's going to be good for a while

david-caro:
Apology accepted :bowtie:

PR #1463, opened 2016-08-18
API: extend /authors endpoint

jacenkow:
* Adds a new entry point in RECORDS_REST_ENDPOINTS.

* Adds serializers to authors module.

### DIFF
--- a/inspirehep/modules/authors/rest/__init__.py
+++ b/inspirehep/modules/authors/rest/__init__.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Record serialization."""
+
+from __future__ import absolute_import, division, print_function
+
+from inspirehep.modules.authors.rest.citations import AuthorAPICitations
+from inspirehep.modules.authors.rest.coauthors import AuthorAPICoauthors
+from inspirehep.modules.authors.rest.publications import AuthorAPIPublications
+from inspirehep.modules.authors.rest.stats import AuthorAPIStats
+from inspirehep.modules.records.serializers.response import (
+    record_responsify_nocache,
+)
+
+citations_v1 = AuthorAPICitations()
+citations_v1_response = record_responsify_nocache(citations_v1,
+                                                  'application/json')
+
+coauthors_v1 = AuthorAPICoauthors()
+coauthors_v1_response = record_responsify_nocache(coauthors_v1,
+                                                  'application/json')
+
+publications_v1 = AuthorAPIPublications()
+publications_v1_response = record_responsify_nocache(publications_v1,
+                                                     'application/json')
+
+stats_v1 = AuthorAPIStats()
+stats_v1_response = record_responsify_nocache(stats_v1,
+                                              'application/json')
--- a/inspirehep/modules/authors/rest/citations.py
+++ b/inspirehep/modules/authors/rest/citations.py
@@ -1,0 +1,123 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+import json
+
+from inspirehep.modules.search import LiteratureSearch
+
+
+class AuthorAPICitations(object):
+    """API endpoint for author collection returning citations."""
+
+    def serialize(self, pid, record, links_factory=None):
+        """Return a list of citations for a given author recid.
+
+        :param pid:
+            Persistent identifier instance.
+
+        :param record:
+            Record instance.
+
+        :param links_factory:
+            Factory function for the link generation, which are added to
+            the response.
+        """
+        author_pid = pid.pid_value
+        citations = []
+
+        search = LiteratureSearch().query({
+            "match": {
+                "authors.recid": author_pid
+            }
+        }).params(
+            _source=[
+                "authors.recid",
+                "control_number",
+                "self",
+            ]
+        )
+
+        # For each publication co-authored by a given author...
+        for result in search.scan():
+            result_source = result.to_dict()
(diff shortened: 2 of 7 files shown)


jacenkow:
So after the bloodiest single-day battle seems I got the minimum configuration to extend our API.

Note that we cannot leave `list_route` empty, since by default it will point to [`/records/`](https://pythonhosted.org/invenio-records-rest/api.html#invenio_records_rest.config.RECORDS_REST_ENDPOINTS). However, I cannot start the application not having `search_serializers` set to any value, which not even set is `invenio_records_rest.serializers':json_v1_search` by default?! I don't have a clue why.

I don't really like the fact of using serialisers to serve our use cases, since they are meant to translate one object to another, but AFAI Invenio doesn't support APIs.

CC: @kaplun, @jacquerie

jacquerie:
> I don't really like the fact of using serialisers to serve our use cases, since they are meant to translate one object to another, but AFAI Invenio doesn't support APIs.

I agree, but at least now we have something to build APIs the Invenio™ way :+1:

jacenkow:
@jacquerie, so Invenio guys said that they don't use `restful.py` anymore as it was in the previous system but `views.py`, which is meant for blueprints. I would go with `restful.py`, since the new serialisers are not quite serialisers 😉

jacenkow:
Endpoints
- [ ] `/citations`
- [ ] `/co-authors`
- [x] `/publications`
- [ ] `/stats`

inspirehep/modules/authors/rest/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+from __future__ import absolute_import, division, print_function
+
+from inspirehep.modules.authors.rest.coauthors import AuthorAPICoauthors
+from inspirehep.modules.authors.rest.publications import AuthorAPIPublications
+from inspirehep.modules.records.serializers.response import (
+    record_responsify_nocache,
+)
+
+coauthors_v1 = AuthorAPICoauthors()
+coauthors_v1_response = record_responsify_nocache(coauthors_v1,
```
kaplun: I think you should move this initialization blurbs directly in the corresponding submodules and only import here the globals you allocate there.

(you should also add a `__all__` information).

inspirehep/modules/authors/rest/coauthors.py
```diff
@@ -0,0 +1,80 @@
...
+
+        search = LiteratureSearch().query({
+            "match": {
+                "authors.recid": author_pid
+            }
+        }).params(
+            _source=[
+                "authors.full_name",
+                "authors.recid",
+                "authors.record",
```
kaplun: What about the BAI?

inspirehep/modules/authors/rest/coauthors.py
```diff
@@ -0,0 +1,80 @@
...
+            ]
+        )
+
+        for result in search.scan():
+            result_source = result.to_dict()['authors']
+
+            for author in result_source:
```
kaplun: Mmh here you are doing the aggregation in Python but I believe you should be able to do it more efficiently in ES itself.

inspirehep/modules/authors/rest/__init__.py
jacenkow: Seems suprising for me as well, however this is what Invenio guys do https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/serializers/__init__.py

inspirehep/modules/authors/rest/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+    record_responsify_nocache,
+)
+
+citations_v1 = AuthorAPICitations()
+citations_v1_response = record_responsify_nocache(citations_v1,
+                                                  'application/json')
+
+coauthors_v1 = AuthorAPICoauthors()
+coauthors_v1_response = record_responsify_nocache(coauthors_v1,
+                                                  'application/json')
```
jacquerie: Forgot two lines here.

inspirehep/modules/authors/rest/citations.py
```diff
@@ -0,0 +1,125 @@
...
+
+                # Get status if a citer is published.
+                # FIXME: As discussed with Sam, we should have a boolean flag
+                #        for this type of information.
+                try:
+                    if "Published" in [i['primary'] for i in
+                                       nested_result_source['collections']]:
+                        citation['published_paper'] = True
+                    else:
+                        citation['published_paper'] = False
```
jacquerie: This can just be `citation['published_paper'] = 'Published' in ...`

inspirehep/modules/authors/rest/coauthors.py
jacquerie: BAI is in `authors.ids`. It's not part of the mappings, I'm not sure if ES will complain for that.

tests/integration/test_author_api.py
```diff
@@ -0,0 +1,128 @@
...
+    The functionality is provided by invenio-records-rest, which is not the
+    subject of these tests.
+    """
+    with app.test_client() as client:
+        assert client.get(
+            '/api/authors/983220').status_code == 200
+
+        request = json.loads(
+            client.get('/api/authors/983220').data)
+        print(request)
```
jacquerie: Leftover.

tests/integration/test_author_api.py
```diff
@@ -0,0 +1,128 @@
...
+        assert request['id'] == 983220
```
jacquerie: A Flask `Response` object should have a `json` property:
```python
response = client.get('/api/authors/983220')

assert response.json['id'] == 983220
```

tests/integration/test_author_api.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def test_author_citations_endpoint(app):
+    """This test checks for the total count of objects and the availability
+    of required keys and their type.
+    """
+    with app.test_client() as client:
+        assert client.get(
+            '/api/authors/983220/citations/').status_code == 200
```
jacquerie: Mh. Just noticed this:
```
% http --headers http://inspire.jacenkow.com:5000/api/authors/984446/publications/
HTTP/1.1 200 OK

% http --headers http://inspire.jacenkow.com:5000/api/authors/984446/publications
HTTP/1.1 301 MOVED PERMANENTLY
```
(Note the missing `/` at the end of the URL). I think we want URLs without a trailing slash?

tests/integration/test_author_api.py
```diff
@@ -0,0 +1,128 @@
...
+        assert len(request) == 30
+        assert isinstance(request[0]['title'], six.string_types)
+        assert isinstance(request[0]['collaborations'][0], six.string_types)
+        assert isinstance(request[0]['journal']['record']['$ref'],
+                          six.string_types)
+        assert isinstance(request[0]['record']['$ref'], six.string_types)
+        assert isinstance(request[0]['citations'], int)
+        assert isinstance(request[0]['date'], six.string_types)
+        assert isinstance(request[0]['type'], six.string_types)
+        assert isinstance(request[0]['id'], int)
```
jacquerie: Idea: use JSON Schema to specify the format and the types of the response, and assert that the response respects the schema (this is actually how most people use JSON Schema).

jacquerie: But you can leave it out of this PR.

tests/integration/test_author_api.py
jacenkow: Unfortunately it doesn't have.

```python
In [1]: dir(response)
Out[1]: 
['__call__',
 '__class__',
 '__delattr__',
 '__dict__',
 '__doc__',
 '__enter__',
 '__exit__',
 '__format__',
 '__getattribute__',
 '__hash__',
 '__init__',
 '__module__',
 '__new__',
 '__reduce__',
 '__reduce_ex__',
 '__repr__',
 '__setattr__',
 '__sizeof__',
 '__str__',
 '__subclasshook__',
 '__weakref__',
 '_ensure_sequence',
 '_get_mimetype_params',
 '_on_close',
 '_status',
 '_status_code',
 'accept_ranges',
 'add_etag',
 'age',
 'allow',
 'autocorrect_location_header',
 'automatically_set_content_length',
 'cache_control',
 'calculate_content_length',
 'call_on_close',
 'charset',
 'close',
 'content_encoding',
 'content_language',
 'content_length',
 'content_location',
 'content_md5',
 'content_range',
 'content_type',
 'data',
 'date',
 'default_mimetype',
 'default_status',
 'delete_cookie',
 'direct_passthrough',
 'expires',
 'force_type',
 'freeze',
 'from_app',
 'get_app_iter',
 'get_data',
 'get_etag',
 'get_wsgi_headers',
 'get_wsgi_response',
 'headers',
 'implicit_sequence_conversion',
 'is_sequence',
 'is_streamed',
 'iter_encoded',
 'last_modified',
 'location',
 'make_conditional',
 'make_sequence',
 'mimetype',
 'mimetype_params',
 'response',
 'retry_after',
 'set_cookie',
 'set_data',
 'set_etag',
 'status',
 'status_code',
 'stream',
 'vary',
 'www_authenticate']
```

jacquerie:
LGTM, this has some warts, but the benefits of this vastly outweigh them. Besides, we need to merge this so that the work on the Authors profile Angular application is unblocked.

kaplun:
Beware of the WIP in the commit log then

kaplun:
(and the missing ending dot in my above comment).

PR #1465, opened 2016-08-19
jsonschema: simple titles

### LINKED ISSUE #1257: Simplifying titles

kaplun:
* Merge title_variations (210) and titles_old(247) into titles with
  the convention that the main title is the first in the list and all
  the others are variations or previous versions. (closes #1257)

### DIFF
--- a/inspirehep/dojson/hep/fields/bd20x24x.py
+++ b/inspirehep/dojson/hep/fields/bd20x24x.py
@@ -24,111 +24,69 @@
 
 from __future__ import absolute_import, division, print_function
 
-from dojson import utils
+import langdetect
 
 from ..model import hep, hep2marc
 
+from inspirehep.dojson.utils import force_single_element
 from inspirehep.utils.helpers import force_force_list
 
 
-@hep.over('title_translations', '^242[10_][0_]')
-@hep.over('titles_old', '^247[10_][10_]')
-@utils.for_each_value
-@utils.filter_values
-def title_translations_old(self, key, value):
-    """Translation of Title by Cataloging Agency."""
-    return {
-        'source': value.get('9'),
-        'title': value.get('a'),
-        'subtitle': value.get('b')
-    }
-
-
-@hep2marc.over('210', '^title_variations$')
-@hep2marc.over('242', '^title_translations$')
-@hep2marc.over('247', '^titles_old$')
-@utils.for_each_value
-@utils.filter_values
-def title_trans_var_old2marc(self, key, value):
-    """Translation of Title by Cataloging Agency."""
-    return {
-        'a': value.get('title'),
-        'b': value.get('subtitle'),
-        '9': value.get('source'),
-    }
-
-
-@hep.over('title_variations', '^210[10_][0_]')
-def title_variations(self, key, value):
-    """Title variations."""
-    def get_value(existing):
-        if not isinstance(value, (tuple, list)):
-            values = [value]
-        else:
-            values = value
-        out = []
-        for val in values:
-            out.append({
-                'title': val.get('a'),
-                'subtitle': val.get('b'),
-                'source': val.get('9'),
-            })
-        cleaned_titles = existing + out
-        return cleaned_titles
-
-    if 'title_variations' in self:
-        titles = get_value(self['title_variations'])
-    else:
-        titles = get_value([])
-
-    return titles
-
-
-@hep.over('titles', '^24[56][10_][0_]')
+@hep.over('titles', '^(210|24[2567])[10_][0_]')
 def titles(self, key, value):
-    """Title Statement."""
-    def get_value(existing):
-        if not isinstance(value, (tuple, list)):
-            values = [value]
-        else:
-            values = value
-        out = []
-        for val in values:
-            out.append({
-                'title': val.get('a'),
-                'subtitle': val.get('b'),
-                'source': val.get('9'),
-            })
-        cleaned_titles = existing + out
-        return cleaned_titles
+    def is_main_title(key):
+        return key.startswith('245')
 
-    if 'titles' in self:
-        titles = get_value(self['titles'])
-    else:
-        titles = get_value([])
+    def is_translated_title(key):
+        return key.startswith('242')
 
-    return titles
-
-
-@hep2marc.over('245', '^titles$')
-def titles2marc(self, key, value):
-    """Title Statement for 245/246."""
-    arxiv_246_title = None
-    titles = []
-
-    for title in force_force_list(value):
-        transformed_title = {
-            'a': title.get('title'),
-            'b': title.get('subtitle'),
-            '9': title.get('source'),
+    titles = self.setdefault('titles', [])
+    values = force_force_list(value)
+    for val in values:
+        title_obj = {
+            'title': val.get('a'),
+            'subtitle': force_single_element(val.get('b')),  # FIXME: #1484
+            'source': val.get('9'),
         }
-        if title.get('source', '').lower() == 'arxiv':
-            arxiv_246_title = transformed_title
+        if is_main_title(key):
+            titles.insert(0, title_obj)
+        elif is_translated_title(key):
+            title = val.get('a')
+            if title:
+                lang = langdetect.detect(title)
+                if lang:
+                    title_obj['language'] = lang
+                    self.setdefault('title_translations', []).append(title_obj)
         else:
-            titles.append(transformed_title)
+            titles.append(title_obj)
 
-    if len(titles) == 0 and arxiv_246_title is not None:
-        titles.append(arxiv_246_title)
-    if arxiv_246_title is not None:
-        self['246'] = arxiv_246_title
     return titles
+
+
+@hep2marc.over('246', '^titles$')
+def titles2marc(self, key, value):
+    def get_transformed_title(val):
+        return {
+            'a': val.get('title'),
+            'b': val.get('subtitle'),
+            '9': val.get('source'),
+        }
+
+    values = force_force_list(value)
+    if values:
+        # Anything but the first element is the main title
+        self['245'] = [get_transformed_title(values[0])]
+    return [get_transformed_title(value) for value in values[1:]]
+
+
+@hep2marc.over('242', r'^title_translations$')
+def title_translations2marc(self, key, value):
+    def get_transformed_title(val):
+        return {
+            'a': val.get('title'),
+            'b': val.get('subtitle'),
+            '9': val.get('source'),
+        }
+
+    values = force_force_list(value)
+    return [get_transformed_title(value) for value in values]
--- a/inspirehep/modules/records/jsonschemas/records/hep.json
+++ b/inspirehep/modules/records/jsonschemas/records/hep.json
@@ -498,10 +498,10 @@
         },
         "languages": {
             "items": {
-                "title": "Language",
-                "type": "string",
+                "format": "ISO 639-1",
                 "pattern": "\\w{2}",
-                "format": "ISO 639-1"
+                "title": "Language",
+                "type": "string"
             },
             "type": "array"
         },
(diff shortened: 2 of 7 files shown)


kaplun:
cc: @eamonnmag @jacquerie

kaplun:
:( Seems not to load records. I don't understand.

kaplun:
While locally for me I don't have any problem

kaplun:
cc: @jacquerie

tests/unit/orcid/test_orcid_converter.py
```diff
@@ -139,7 +139,7 @@ def test_successfull_conversion():
                                                                        'external-identifier-type': 'ARXIV'}]},
                 'journal-title': 'Nuovo Cim.',
                 'short-description': 'Abstract',
-                'title': {'subtitle': 'Subtitle', 'title': 'Title3'},
+                'title': {'subtitle': 'Subtitle', 'title': 'Title'},
```
kaplun: Amending this because in the new definition, the main title is simply the first in 245

inspirehep/dojson/hep/fields/bd20x24x.py
```diff
@@ -58,77 +55,38 @@ def title_trans_var_old2marc(self, key, value):
...
+    titles = self.setdefault('titles', [])
+    values = force_force_list(value)
+    for val in values:
+        obj = {
+            'title': val.get('a'),
+            'subtitle': val.get('b'),
+            'source': val.get('9'),
+        }
+        if key.startswith('245'):
+            # This is the main title
```
jacquerie: Better use a helper than a comment.

inspirehep/dojson/hep/fields/bd20x24x.py
```diff
@@ -32,7 +32,6 @@
 
 
 @hep.over('title_translations', '^242[10_][0_]')
```
jacquerie: Why leave the title translations out of this refactoring?

inspirehep/utils/record.py
```diff
@@ -29,36 +29,18 @@
...
-            if not chosen_subtitle:
-                chosen_subtitle = title.get('subtitle')
-        else:
-            if 'subtitle' in title:
-                chosen_subtitle = title.get('subtitle')
-    return chosen_subtitle
+    if record.get('titles', []):
+        return record['titles'][0].get('subtitle', '')
+    else:
+        return ""
```
jacquerie: Can be expressed in terms of `get_value` in fewer lines.

inspirehep/utils/record.py
```diff
@@ -29,36 +29,18 @@
...
-    for title in record.get('titles', []):
-        if 'source' in title and 'subtitle' in title and title.get('source') != 'arXiv':
-            return title.get('subtitle')
-        elif 'source' in title and 'subtitle' in title and title.get('source') == 'arXiv':
-            if not chosen_subtitle:
-                chosen_subtitle = title.get('subtitle')
-        else:
-            if 'subtitle' in title:
-                chosen_subtitle = title.get('subtitle')
-    return chosen_subtitle
```
jacquerie: The logic is clearly not equivalent, though...

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -803,24 +800,21 @@ def test_title_translations(marcxml_to_json, json_to_marc):
...
+    for variation in json_to_marc['246']:
+        source = variation.get('9')
+        if source:
+            assert source in get('source')
+        subtitle = variation.get('b')
+        if subtitle:
+            assert subtitle in get('subtitle')
+        title = variation.get('a')
+        if title:
+            assert title in get('title')
```
jacquerie: Needs proper tests. These are too easy to pass while being incorrect.

jacquerie:
Relevant error on Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601664/.

inspirehep/dojson/hep/fields/bd20x24x.py
kaplun: Because I think we need to keep titles in English and the translated version in separate fields, so that we can apply different mappings in ES. I know it's making the schema unnecessarily complex, but I don't see how I would be able e.g. to apply a specific language dependent stemming without separating the fields.

inspirehep/utils/record.py
kaplun: The new logic corresponds to the new proposed semantic, where the preferred title is simply the first in the list, regardless of it being arXiv or not. json-merger, will append arXiv titles, while inserting publisher ones :)

inspirehep/utils/record.py
kaplun: You mean `get_value['titles[0].subtitle']`?

inspirehep/dojson/hep/fields/bd20x24x.py
jacquerie: We can have everyhing in one field, and use this guy: https://github.com/jprante/elasticsearch-langdetect.

kaplun: As discussed IRL, we have decided to keep them separate and have `titles`, `titles_en`, `titles_fr`.

jacquerie:
Oh, and there's also the case of multiple subtitles: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601664/

inspirehep/dojson/hep/fields/bd20x24x.py
kaplun: See also: https://app.asana.com/0/2345936208876/170763976041252

kaplun: Mmh. This would depend on some specific dojson enhancement: inveniosoftware/dojson#172 so that we can write the rule that decide whether 245 is translated or not, depending on 041.

kaplun: I learn from @annetteholtkamp that the main title (245) is not necessarily the foreign language when a language for a document is set.

kaplun: Depends on #1483

inspirehep/dojson/hep/fields/bd20x24x.py
```diff
@@ -24,111 +24,71 @@
...
-
-@hep.over('title_variations', '^210[10_][0_]')
-def title_variations(self, key, value):
-    """Title variations."""
-    def get_value(existing):
-        if not isinstance(value, (tuple, list)):
-            values = [value]
+@hep.over('titles', '^(210|24[2567])[10_][0_]')
+def titles(self, key, value):
+    """All sorts of titles"""
```
jacquerie: We should start killing docstrings in DoJSON rules because they are either incorrect or useless.

inspirehep/dojson/hep/fields/bd20x24x.py
```diff
@@ -24,111 +24,71 @@
...
+            'subtitle': val.get('b'),
+            'source': val.get('9'),
+        }
+        if is_main_title(key):
+            titles.insert(0, obj)
+        elif is_translated_title(key):
+            title = val.get('a')
+            if title:
+                lang = langdetect.detect(title)
+            self['title.{0}'.format(lang)] = obj
```
jacquerie: `lang` here is undefined if `title` was `None.

inspirehep/utils/record.py
```diff
@@ -29,36 +29,18 @@
...
-            chosen_title = title.get('title')
-
-    if not chosen_title and len(record.get('titles', [])) > 0:
-        chosen_title = record['titles'][0]['title']
-
-    return chosen_title
+    if record.get('titles', []):
+        return record['titles'][0].get('title', '')
+    else:
+        return ""
```
jacquerie: All this can just be `return get_value('titles[0].title', default='')`

inspirehep/utils/record.py
jacquerie: `get_value(record, 'titles[0].subtitle', default='')`

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -1,6 +1,11 @@
 {
     "$schema": "http://json-schema.org/schema#",
     "description": "An article or thesis or book or...",
+    "patternProperties": {
+        "^title\\.\\w{2}$": {
+            "$ref": "elements/title.json"
```
jacquerie: A good example of why it's better to have singular schema elements : )

inspirehep/dojson/hep/fields/bd20x24x.py
```diff
@@ -24,111 +24,71 @@
...
+    def is_main_title(key):
+        return key.startswith('245')
+
+    def is_translated_title(key):
+        return key.startswith('242')
+
+    titles = self.setdefault('titles', [])
+    values = force_force_list(value)
+    for val in values:
+        obj = {
```
jacquerie: I'm sure this can have a better name... maybe `title_obj`?

jacquerie:
Overall this is good (it's always cool to remove code), but this needs proper tests, and to treat the case of multiple subtitles in some way.

inspirehep/modules/records/jsonschemas/records/hep.json
kaplun: Yep when I wrote this I also realized that the record-editor won't be happy as well. So probably we should delegate the difference to elasstiocsearch

kaplun:
> and to treat the case of multiple subtitles in some way.

As mentioned IRL there were only 3 records with multiple subtitles and they've been amended.

kaplun:
IMHO, Ready (let's see Travis though)

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -901,41 +901,34 @@ def test_corporate_author(marcxml_to_json, json_to_marc):
...
+        if value:
+            candidate[key] = value
+
+    title_variations = marcxml_to_json['titles'][1:]
+    for variation in json_to_marc['246']:
+        candidate = {}
+        build_candidate(candidate, 'source', variation.get('9'))
+        build_candidate(candidate, 'title', variation.get('a'))
+        build_candidate(candidate, 'subtitle', variation.get('b'))
+        assert candidate in title_variations
```
jacquerie: Not an improvement, and introduces a pattern which never appears anywhere else in the codebase. You can just delete this test.

tests/unit/utils/test_utils_record.py
```diff
@@ -56,7 +56,7 @@ def test_get_abstract_returns_empty_string_when_no_titles():
 
 
 def test_get_title_returns_empty_string_when_titles_is_empty():
-    empty_titles = Record({'titles': []})
+    empty_titles = Record({'foo': 'bar'})
```
jacquerie: This makes the test name wrong, which means that you changed the behaviour of `get_title` when a record has a `titles` key with value `[]`. If you think the test makes no longer sense, delete it, rather than leaving it with a broken name.

In fact, `get_title({'titles': []})` now returns `[]`, which is not even a string and definitely not a sensible value for a title, which means that my smart `get_value(record, 'titles[0].subtitle', default='')` wasn't so smart after all.

tests/unit/dojson/test_dojson_hep.py
kaplun: This is an improvement: it test that any alternative title input is well represented in the correct structure (up to subfield level) in the json titles. The previous implementation didn't test for correlation across subfields of the same field.

Additionally I don't agree with the argument that introducing for the first time a pattern implies it's wrong.

kaplun: I don't see how to easily improve it because the `title_variations` field necessarily is a super-set of the MARC `246` field. So I can't easily build things using `assert ... == ...`. However I still see a value in testing, this so I prefer to keep the test.

inspirehep/utils/record.py
jmartinm: Just saw this, and for the existing records, the order will be wrong? And the templates will not show the correct title?

kaplun: What do you mean existing records? What this PR does is to put the 245 (i.e. the main title) at the beginning of the `titles` field, and I have consequently amended the `get_(sub)title` functions to reflect this. So everything should be correct. Or I am not understanding what you mean :-)

jmartinm: My question was more in the direction of @jacquerie's. Where has this logic been moved to? The fact that titles that come from a publisher should be first and titles from arXiv should be after.

PR #1468, opened 2016-08-19
WIP schemas: don't restrict names with enum or format

jacquerie:
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601678/ and https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601681/.

Motivation: https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

### DIFF
--- a/inspirehep/modules/records/jsonschemas/records/authors.json
+++ b/inspirehep/modules/records/jsonschemas/records/authors.json
@@ -106,32 +106,15 @@
             "description": "Contains name information.",
             "properties": {
                 "numeration": {
-                    "enum": [
-                        "",
-                        "Jr.",
-                        "Sr.",
-                        "I",
-                        "II",
-                        "III",
-                        "IV",
-                        "VI",
-                        "VII",
-                        "VIII"
-                    ],
                     "type": "string"
                 },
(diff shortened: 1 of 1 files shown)


jacquerie:
@kaplun @david-caro WDYT?

kaplun:
Yup! What I would love jsonschema allowed was "warning" levels. E.g. one could set a pattern meaning it would be great the value does respect it, too bad if it doesn't. In this way we can provide hints for the editor, without blocking.

david-caro:
Looks good to me, though the 'hints' I don't think they should go with the schema itself, unless they are actually tested by it, it feels to me more like frontend ui feature.

kaplun:
What do you mean by "hints"? 

On another thing, you might mash preferred name into names, with the convention that the preferred is the first one in the list.

kaplun:
@jacquerie I guess this is still a WIP, right?
* [ ] mash preferred name into names, with the convention that the preferred is the first one in the list.

jacquerie:
Moved to https://github.com/inspirehep/inspire-schemas/issues/27.

PR #1472, opened 2016-08-22
utils: add test for error raised by get_es_records

liamkirsh:
When get_es_records is called with an empty list for the parameter `recids`, it throws a `RequestError: TransportError(400, u'action_request_validation_exception', u'Validation Failed: 1: no documents to get;')` error. This adds a test for that.

### DIFF
--- a/tests/unit/utils/test_utils_record_getter.py
+++ b/tests/unit/utils/test_utils_record_getter.py
@@ -21,6 +21,7 @@
 # or submit itself to any jurisdiction.
 
 import pytest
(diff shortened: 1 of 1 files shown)


jacquerie:
@jmartinm, is this the behaviour that we want in `get_es_records`? I'm tempted to say "no", but for now it doesn't really matter...

jmartinm:
Did not see this one before. The error message at least is quite clear.

I am for adding this test for the moment.

PR #1469, opened 2016-08-19
schema: merge keyword fields

david-caro:
* Merge the free_keywords and thesaurus_terms from marcxml to
  keywords field in jsonschema.
* Separate the energy_level out of the keywords to it's own field.

Signed-off-by: David Caro <david@dcaro.es>

PR #1476, opened 2016-08-23
api: extend /authors endpoint

jacquerie:
I rewrote the tests and found a weird bug: sometimes `null` is considered a keyword. For now I'm not raising an error by expanding the response schema to allow for `null` values there, but clearly something's off.

Closes #1463

### DIFF
--- a/inspirehep/modules/authors/rest/stats.py
+++ b/inspirehep/modules/authors/rest/stats.py
@@ -1,0 +1,125 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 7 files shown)


inspirehep/modules/authors/rest/stats.py
```diff
@@ -0,0 +1,124 @@
...
+                    statistics['types'][publication_type] = 1
+
+            # Get fields.
+            for field in result_source.get('facet_inspire_subjects', []):
+                fields.add(field)
+
+            # Get keywords.
+            for term in result_source.get('thesaurus_terms', []):
+                if term.get('keyword') is not "* Automatic Keywords *":
+                    keywords.append(term.get('keyword'))
```
jacquerie: Ah, there it is: if `term.get('keyword')` returns `None` it is still appended. This happens when a member of `thesaurus_terms` does not contain a `keyword` key, which happens for instance when it contains an `energy_range`, the thing that @david-caro is solving in #1469.

jacquerie: Also note this thing about Python:
```python
In [1]: s = 'foo'

In [2]: t = ''.join(['f', 'o', 'o'])

In [3]: s == t
Out[3]: True

In [4]: s is t
Out[4]: False
```
comparing strings with `is` is risky in very subtle ways.

jacquerie:
> I rewrote the tests and found a weird bug: sometimes null is considered a keyword. For now I'm not raising an error by expanding the response schema to allow for null values there, but clearly something's off.

Solved this.

PR #1478, opened 2016-08-23
travis: dummy PR to find out which tests are slow

jacquerie:
Uses the [`--durations`](http://doc.pytest.org/en/latest/usage.html#profiling-test-execution-duration) argument in py.test to measure the time that each test takes. The objective is to split the tests in the way I described in https://github.com/inspirehep/inspire-next/pull/1353#issuecomment-237524229.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -47,11 +47,11 @@ services:
       service: service_base
     volumes_from:
       - static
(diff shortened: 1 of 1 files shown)


jacquerie:
The result is that we can add a `workflows` build and save ~2:30 minutes on the build. Not amazing, but not bad either.

Issue #1399, opened 2016-08-03
tests: avoid distutils, just use pytest


jacquerie:
As mentioned by @david-caro [here](https://github.com/inspirehep/inspire-next/pull/1398#issuecomment-237292346), we could avoid passing through `distutils` to run the tests. Running `python setup.py test` ensures that the test dependencies are installed, but we already install the `[tests]` dependencies in the `pip` step of the Travis build: https://github.com/inspirehep/inspire-next/blob/97a8a8d40c3829582a610df168d226d70a9aa79a/docker-compose.deps.yml#L30. So we could just use `pytest` invocations, which would avoid the ugly `--addopts` in the invocations.

PR #1481, opened 2016-08-24
travis: speed up the build by splitting it

jacquerie:
Implements the idea I described in https://github.com/inspirehep/inspire-next/pull/1353#issuecomment-237524229 using the timings from https://github.com/inspirehep/inspire-next/pull/1478.

I believe this a nice time to solve #1399 as well.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -61,10 +61,23 @@ services:
       - redis
     depends_on:
       - worker
+  workflows:
+    extends:
+      service: service_base
+    command: python setup.py test --addopts="tests/integration/workflows"
+    volumes_from:
+      - static
+    links:
+      - database
+      - indexer
+      - rabbitmq
+      - redis
+    depends_on:
+      - worker
   integration:
     extends:
       service: service_base
-    command: python setup.py test --addopts="tests/integration --ignore tests/integration/disambiguation"
+    command: python setup.py test --addopts="tests/integration --ignore tests/integration/disambiguation --ignore tests/integration/workflows"
     volumes_from:
       - static
     links:
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -41,17 +41,17 @@ services:
       service: service_base
     volumes_from:
       - static
-    command: python setup.py test --addopts="inspirehep"
+    command: py.test inspirehep
   unit:
     extends:
       service: service_base
     volumes_from:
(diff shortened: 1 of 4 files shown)


docker-compose.test.yml
```diff
@@ -64,7 +64,20 @@ services:
...
+    volumes_from:
+      - static
+    links:
+      - database
+      - indexer
+      - rabbitmq
+      - redis
+    depends_on:
+      - worker
+  workflows:
```
jacquerie: @mihaibivol, any idea on how to DRY these guys?

mihaibivol: Sadly nop https://docs.docker.com/compose/extends/#/extending-services

jacquerie: Ok : (

jacquerie:
Ready for review, but since this is clearly making the build faster for everyone I think this should be on top of the queue...

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -243,7 +243,7 @@ def fake_magpie_api_request(url, data):
 def test_harvesting_arxiv_workflow_rejected(
     mocked_api_request_beard_block, mocked_api_request_magpie,
     mocked_api_request_beard, mocked_download,
-    app, record_oai_arxiv_plots):
+    small_app, record_oai_arxiv_plots):
```
kaplun: Can you document the difference between `small_app` Vs. `app` so that it's clear when to use which?

kaplun:
Other than my little comment, I trust whatever this branch does.

tests/integration/workflows/test_arxiv_workflow.py
jacquerie: I'll write it here, because I don't know where to write this: use `small_app` whenever you need all the components of the application (DB, ES, Celery...) to be there, but you don't need the records. Use `app` when you need all the demo records to be loaded.

kaplun: Thanks! It could have been written in https://github.com/inspirehep/inspire-next/blob/master/tests/integration/conftest.py#L39

kaplun: Alternatively we could rename `app` into `app_with_records`, and `small_app` into `app_without_records`.

PR #1483, opened 2016-08-24
dojson: language support

kaplun:
* Normalizes languages using ISO 639-1.

### DIFF
--- a/inspirehep/modules/records/jsonschemas/records/hep.json
+++ b/inspirehep/modules/records/jsonschemas/records/hep.json
@@ -514,7 +514,9 @@
         "languages": {
             "items": {
                 "title": "Language",
-                "type": "string"
+                "type": "string",
+                "pattern": "\\w{2}",
+                "format": "ISO 639-1"
             },
             "type": "array"
         },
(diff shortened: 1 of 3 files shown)


inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -514,7 +514,9 @@
         "languages": {
             "items": {
                 "title": "Language",
-                "type": "string"
+                "type": "string",
+                "pattern": "\\w{3}",
+                "format": "ISO 639-2"
```
jacquerie: Leftover : (

kaplun: Thanks!

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -116,6 +116,77 @@ def test_external_system_numbers(marcxml_to_json, json_to_marc):
...
+    snippet = (
+        '<datafield tag="041" ind1=" " ind2=" ">'
+        '  <subfield code="a">Italian</subfield>'
+        '</datafield>'
+    )
+
+    expected = ["it"]
+    result = clean_record(hep.do(create_record(snippet)))
+
+    assert expected == result['languages']
```
jacquerie: I just had an epiphany: it's easy to write tests that ensure that the rule satisfies the schema! You just need to fetch the schema, drill down to the right point and then call `validate` from `jsonschema` on that subschema. Nothing to fix here, I just wanted to write it down so I don't forget it.

kaplun: But in these tests we are additionally testing that the content is correctly propagated (i.e. rule might produce valid schema, but maybe loose data during the translation).

PR #1487, opened 2016-08-25
dojson: rewrite positions rule

jacquerie:
* Unifies `email` and `old_email` in a single field called `emails`,
  with the convention that the first emails are the most current.

* Consolidates the treatment of `status`/`current`, and moves
  `curated_relation` where it belongs (near the institution).

Signed-off-by: Jacopo Notarstefano <jacopo.notarstefano@cern.ch>

### DIFF
--- a/inspirehep/dojson/hepnames/fields/bd1xx.py
+++ b/inspirehep/dojson/hepnames/fields/bd1xx.py
@@ -263,65 +263,58 @@ setattr(hidden_notes2marc, '__extend__', True)
 
 @hepnames.over('positions', '^371..')
 @utils.for_each_value
-@utils.filter_values
 def positions(self, key, value):
-    """Position information.
-
-    In dates field you can put months in addition to years. In this case you
-    have to follow the convention `mth-year`. For example: `10-2012`.
-    """
-    curated_relation = False
+    """Positions that an author held during their career."""
+    curated = False
+    current = False
     recid = None
-    status = ''
-    recid_status = force_force_list(value.get('z'))
-    for val in recid_status:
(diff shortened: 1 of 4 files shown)


jacquerie:
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601591/ and https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601592/.

jacquerie:
Ready for review.

inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -263,65 +263,58 @@ def hidden_notes2marc(self, key, value):
...
+        if el.lower() == 'current':
+            current = True
         else:
-            try:
-                recid = int(val)
-                curated_relation = True
-            except ValueError:
-                pass
+            curated = el.isdigit()
+            if curated:
```
kaplun: This reads a bit weird. It took me a while: it's not because it's curated that you can assign it. But finally it's correct :)

kaplun:
So we are still keeping the separate emails that are independent from institutions, right?

jacquerie:
> So we are still keeping the separate emails that are independent from institutions, right?

This PR is not touching those, but, as I remarked in #1453, nobody is using them, so porting them to the new schema has a very low priority.

jacquerie:
I only amended the use in the template, because [`convert_for_form`](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/authors/views/holdingpen.py#L66-L142) is hilariously broken

Issue #1416, opened 2016-08-09
RFC API: /authors endpoint


jacenkow:
As @jacquerie  requested, the API endpoint for `/authors` should be changed to meet the following requirements:

* `/authors/<pidpath(authors):pid_value>`
    Should return the whole metadata for a given `pid` value.
    I believe there is no need to change.

    See [Darien Robert Wood](http://localhost:5000/api/authors/983220) as an example.

* `/authors/<pidpath(authors):pid_value>/affiliations`
    Should return a list of affiliations for a given `pid` value.

    However, do we really need this, since it is exactly the same process as calling
    `/authors/<pidpath(authors):pid_value>` and accessing `['positions']` key?

* `/authors/<pidpath(authors):pid_value>/citations`
    Should return a list of citations for a given `pid` value.

    For example,
    ```python
    {
        "date": "2016-08-09",
        "reference": {
            "record": {
                "$ref": "http://localhost:5000/api/literature/1356276"
            }
        },
        "citation": {
            "record": {
                "$ref": "http://localhost:5000/api/literature/1406194"
            }
        }
    }
    ```

* `/authors/<pidpath(authors):pid_value>/coauthors`
    Should return a list of co-authors for a given `pid` value.

    For example,
    ```python
    {
        "date": "2015-03-26",
        "record": {
            "$ref": "http://localhost:5000/api/literature/1356276"
        },
        "author": {
            "record": {
                "$ref": "http://localhost:5000/api/authors/990458"
            }
        }
    }
    ```

    However, I believe it would be more useful to group co-authors, by keys representing
    `pid` values like in the following example,

    ```python
    {
        "990458": {
            "name": "Sanchez-Hernandez, Alberto",
            "publications": [
                {
                    "$ref": "http://localhost:5000/api/literature/1356276"
                },
                ...
            ],
            "record": {
                "$ref": "http://localhost:5000/api/authors/990458"
            },
            "shared_publications_count": 550,
        }
    }
    ```

* `/authors/<pidpath(authors):pid_value>/publications`
    Should return a list of publications for a given `pid` value.

    For example,
    ```python
    {
        "date": "2015-03-26",
        "title": "Combined Measurement of the Higgs Boson Mass in pp Collisions at √s=7 and 8 TeV with the ATLAS and CMS Experiments",
        "journal": {
            "title": "Phys.Rev.Lett.",
            "record": {
                "$ref": "http://localhost:5000/api/journals/1214495"
            }
        },
        "record": {
[...]

PR #1488, opened 2016-08-25
API: improve /authors endpoint

jacenkow:
* Adds ids of journals to /publications endpoint.

* Restructures citee becoming a parent of an array of citers at
  /citations endpoint.

### DIFF
--- a/inspirehep/modules/authors/rest/citations.py
+++ b/inspirehep/modules/authors/rest/citations.py
@@ -44,7 +44,7 @@ class AuthorAPICitations(object):
             the response.
         """
         author_pid = pid.pid_value
-        citations = []
+        citations = {}
 
         search = LiteratureSearch().query({
             "match": {
@@ -61,11 +61,14 @@ class AuthorAPICitations(object):
         # For each publication co-authored by a given author...
         for result in search.scan():
             result_source = result.to_dict()
+
+            recid = int(result_source['control_number'])
             authors = set([i['recid'] for i in result_source['authors']])
+            citations[recid] = {}
 
             nested_search = LiteratureSearch().query({
                 "match": {
-                    "references.recid": result_source['control_number']
+                    "references.recid": recid
                 }
             }).params(
                 _source=[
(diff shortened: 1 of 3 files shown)


jacenkow:
CC: @jacquerie

inspirehep/modules/authors/rest/citations.py
```diff
@@ -92,12 +99,14 @@ def serialize(self, pid, record, links_factory=None):
                 except KeyError:
                     nested_authors = set()
 
-                citation = {}
-                citation['citee'] = {'record': citee}
-                citation['citer'] = {'record': nested_result_source['self']}
-
-                # If at least one author is shared, it's a self-citation.
-                citation['self_citation'] = len(authors & nested_authors) > 0
+                citation = dict(
```
jacquerie: Here you're using `dict` constructors to avoid shadowing Python's `id`? In fact, I believe the should be `_id` or `id_`, I don't remember which one the ES Python library is doing.

tests/integration/test_author_api.py
```diff
@@ -126,12 +136,14 @@ def test_api_authors_publications(app):
                     'type': 'string'
                 },
                 'id': {'type': 'integer'},
-                'journals': {
+                'journal': {
```
jacquerie: Whoops. In fact, I should've had `additionalProperties: false`.

tests/integration/test_author_api.py
```diff
@@ -42,35 +42,45 @@ def test_api_authors_root(app):
...
-                'self_citation': {'type': 'boolean'}
-            },
-            'type': 'object'
-        },
-        'type': 'array'
+                    }
+                }
+            }
+        }
     }
```
jacquerie: I'm having some trouble picturing what does a response look like in this schema. Can you write an example in https://github.com/inspirehep/inspire-next/issues/1416 ?

tests/integration/test_author_api.py
```diff
@@ -42,35 +42,45 @@ def test_api_authors_root(app):
...
-                    'properties': {
-                        'record': {
-                            'properties': {'$ref': {'type': 'string'}},
-                            'type': 'object'
+                'citers': {
+                    'type': 'array',
+                    'items': {
+                        'type': 'object',
+                        'properties': {
+                            'date': {'type': 'string'},
```
jacquerie: `format` should stay here (even if for now it's useless).

jacquerie:
The thing we discussed about adding the API schemas to `invenio/modules/records/jsonschemas` is harder than it looks (you need a lot of components to fetch a schema with refs), so I will keep it off this PR for now.

Issue #1486, opened 2016-08-25
RFC Reference matching algorithm


kaplun:
Let's assume we have workflow "reference matching":
- [ ] if not raw reference:
  - [ ] run refextract
  - [ ] store refextract ref as raw ref of type refextract
- [ ] for every reference
  - [ ] translate reference into mini-record
  - [ ] try to match mini-record with existing one
  - [ ] if sure match
    - [ ] record the match
  - [ ] elif ambiguous match and CORE:
  - [ ] request cataloguer to check
  - [ ] else:
    - [ ] create stub-record
    - [ ] record the match

annetteholtkamp:
What do we define as minimal metadata necessary for creating a stub record?

- Annette

kaplun:
I'd say: 
* arXiv ID (because we can fetch the matadata on the fly from arXiv
* DOI (same reason from Crossref)
* ISBN (because we can fetch the book on the fly)
* journal, volume/issue, (year), pages (because it allows to unambiguously match it)
* reportnumber + title + author

StellaCh:
@jacquerie shall we close this?

jacquerie:
> @jacquerie shall we close this?

We shall.

PR #1490, opened 2016-08-25
jsonschema: references

kaplun:

### DIFF
--- a/inspirehep/modules/records/jsonschemas/records/hep.json
+++ b/inspirehep/modules/records/jsonschemas/records/hep.json
@@ -665,7 +665,44 @@
         },
         "references": {
             "items": {
-                "$ref": "elements/reference.json"
+                "properties": {
+                    "curated_relation": {
+                        "type": "boolean"
+                    },
+                    "raw_refs": {
+                        "items": {
+                            "properties": {
+                                "position": {
+                                    "description": "The position of the reference as it appears on the document",
+                                    "type": "string"
+                                },
+                                "schema": {
+                                    "description": "E.g. text, or Elsevier XML, or refexract",
+                                    "type": "string"
+                                },
+                                "source": {
+                                    "type": "string"
+                                },
+                                "value": {
+                                    "type": "string"
+                                }
+                            },
+                            "required": [
+                                "value",
+                                "schema"
+                            ],
+                            "type": "object"
+                        },
+                        "type": "array"
+                    },
+                    "record": {
+                        "$ref": "elements/json_reference.json"
+                    },
+                    "reference": {
+                        "$ref": "elements/reference.json"
+                    }
+                },
+                "type": "object"
             },
             "type": "array",
             "uniqueItems": true
--- a/inspirehep/modules/records/jsonschemas/records/elements/reference.json
+++ b/inspirehep/modules/records/jsonschemas/records/elements/reference.json
@@ -47,9 +47,6 @@
             "type": "array",
             "uniqueItems": true
         },
-        "curated_relation": {
-            "type": "boolean"
-        },
         "dois": {
             "items": {
                 "pattern": "10\\.\\d+(\\.\\d+)?/.+",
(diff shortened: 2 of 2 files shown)


kaplun:
@david-caro @mihaibivol @jacquerie this is a proposal for a final schema to handle references in all its glory. To be used with #1486

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -513,10 +513,10 @@
         },
         "languages": {
             "items": {
-                "title": "Language",
-                "type": "string",
+                "format": "ISO 639-1",
                 "pattern": "\\w{2}",
-                "format": "ISO 639-1"
+                "title": "Language",
+                "type": "string"
```
kaplun: Sorry this should have gone into an other commit

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -684,7 +684,44 @@
         },
         "references": {
             "items": {
-                "$ref": "elements/reference.json"
+                "properties": {
+                    "curated_relation": {
+                        "type": "boolean"
+                    },
+                    "raw_refs": {
```
kaplun: References are an aggregated list of raw references (e.g. say that we are putting together arXiv and publisher)

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -684,7 +684,44 @@
...
+                        "items": {
+                            "properties": {
+                                "position": {
```
kaplun: The position is specific to a given raw reference with respect to its source.

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -684,7 +684,44 @@
...
+                                    "description": "The position of the reference as it appears on the document",
+                                    "type": "string"
+                                },
+                                "schema": {
+                                    "description": "E.g. text, or Elsevier XML, or refexract",
+                                    "type": "string"
+                                },
+                                "source": {
+                                    "type": "string"
+                                },
```
kaplun: Source is different than schema, because e.g. we could have an XML directly from Elsevier, but also a ref-extract coming from the same Elsevier PDF.

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -684,7 +684,44 @@
...
+                            "required": [
+                                "value",
+                                "schema"
+                            ],
+                            "type": "object"
+                        },
+                        "type": "array"
+                    },
+                    "record": {
+                        "$ref": "elements/json_reference.json"
```
kaplun: This is the mini-record as designed by @mihaibivol . I am not yet fully convinced it's worth putting it in the schema because I believe it shouldn't be editable by cataloguers, rather generated automatically from raw_refs. Cataloguers should focus only on matching existing references.

inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -684,7 +684,44 @@
...
+                    },
+                    "reference": {
+                        "$ref": "elements/reference.json"
+                    }
```
kaplun: This is the mini-record as designed by @mihaibivol . I am not yet fully convinced it's worth putting it in the schema because I believe it shouldn't be editable by cataloguers, rather generated automatically from raw_refs. Cataloguers should focus only on matching existing references.

inspirehep/modules/records/jsonschemas/records/hep.json
jacquerie: Yes, this definitely goes in a different commit.

inspirehep/modules/records/jsonschemas/records/hep.json
jacquerie: That schema would have to be edited as well then, because `record` and `curated_relation` are already there.

kaplun:
@mihaibivol actually have we ever integrated your mini-records?

PR #1491, opened 2016-08-25
api: restructure schemas of /authors/citations and /authors/publications

jacquerie:
Closes #1488

### DIFF
--- a/tests/integration/test_author_api.py
+++ b/tests/integration/test_author_api.py
@@ -136,12 +136,14 @@ def test_api_authors_publications(app):
                     'type': 'string'
                 },
                 'id': {'type': 'integer'},
-                'journals': {
+                'journal': {
                     'properties': {
+                        'id': {'type': 'integer'},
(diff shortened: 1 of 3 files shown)


tests/integration/test_author_api.py
```diff
@@ -45,28 +45,38 @@ def test_api_authors_citations(app):
...
-                'date': {
-                    'format': 'date',
-                    'type': 'string'
-                },
-                'published_paper': {'type': 'boolean'},
-                'self_citation': {'type': 'boolean'}
+                    'type': 'array'
+                }
             },
             'type': 'object'
```
jacquerie: @jacenkow note that there was a reason the schema had the types at the bottom: I expect these schemas to be moved in `inspirehep/modules/records/jsonschemas/api/authors`, where `scripts/prettifyjson` will run on them. So if they are already in a good order here the diff will be very clean when that happens (just replace the `"` with `'`).

PR #1493, opened 2016-08-26
authors: make the profiles great again

jacenkow:
* Updates the Angular.js apps to make use of new /authors API.

* Adds support for citations count including listing of
  publications and general statistics.

* Adds impact graphs for every listed publication.

* Updates collaborators to display authors instead of
  experiments (sorted by count of shared publications).

* Hides work/education box, when there is no data.

* Restrctures author apps.

### DIFF
--- a/inspirehep/modules/authors/bundles.py
+++ b/inspirehep/modules/authors/bundles.py
@@ -38,7 +38,8 @@ js = NpmBundle(
     depends=(
         'js/authors/author.js',
         'js/authors/profile.js',
-        'js/authors/publications.js'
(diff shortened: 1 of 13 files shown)


jacenkow:
image

CC: @kaplun @eamonnmag

kaplun:
This looks great! Just some UI notes that I already mentioned you IRL:
* would be great to have impact graphs clickable and opening directly into the corresponding record's impact graph
* you can sort by impact graph! What does that mean? @eamonnmag can we actually have a measure of impact that can be useful when sorting?

PR #1495, opened 2016-08-29
migrator: faster citation count implementation

kaplun:


kaplun:
@jacquerie this implementation should not take 1.5 days to compute :dancer:

jacquerie:
:+1:

The idea looks ok, but I think there's more to fix in this function. I will take care of merging this myself.

kaplun:
What do you want to fix? It works for me...

jacquerie:
> What do you want to fix?

The logic from line 271 to 281 (it can be greatly simplified).

PR #1496, opened 2016-08-29
dojson: ignore double longitudes

jacquerie:
Now that https://github.com/inspirehep/inspire/issues/170 was properly solved on Legacy data we can start ignoring this error on Nightly.

### DIFF
--- a/inspirehep/dojson/institutions/fields/bd1xx.py
+++ b/inspirehep/dojson/institutions/fields/bd1xx.py
@@ -34,7 +34,6 @@ from inspirehep.utils.helpers import force_force_list
 
 
 @institutions.over('location', '^034..')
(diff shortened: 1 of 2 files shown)


kaplun:
@jacquerie Note that that unfortunately it worked partially so in the dojson rule, you still have to skip all the single longitudes

kaplun:
To be more clear: we currently have records with a field having just a longitude, and a field having a longitude and a latitude.

jacquerie:
> To be more clear: we currently have records with a field having just a longitude, and a field having a longitude and a latitude.

Ok. I added a commit that doesn't produce a location if it isn't complete with both a latitude and a longitude.

david-caro:
LGTM

PR #1499, opened 2016-08-29
schema: allow list as publication info notes

david-caro:
* Model pulication info notes as a list with unique values (if not
  unique, duplicates will be ignored on parsing)


inspirehep/modules/records/jsonschemas/records/hep.json
```diff
@@ -651,7 +651,11 @@
                         "type": "string"
                     },
                     "note": {
```
jacquerie: This could be `notes`, to suggest that this is an array.

PR #1498, opened 2016-08-29
dojson: more languages

kaplun:
Signed-off-by: Samuele Kaplun <samuele.kaplun@cern.ch>

PR #1501, opened 2016-08-29
dojson: more languages

jacquerie:
Closes #1498

### DIFF
--- a/inspirehep/dojson/hep/fields/bd01x09x.py
+++ b/inspirehep/dojson/hep/fields/bd01x09x.py
@@ -40,27 +40,36 @@ from inspirehep.utils.helpers import force_force_list
 
 LANGUAGE_MAPPING = {
     "bulgarian": "bg",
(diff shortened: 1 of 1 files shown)


inspirehep/dojson/hep/fields/bd01x09x.py
```diff
@@ -70,7 +79,7 @@
 
 REVERSE_LANGUAGE_MAPPING = {v: k for k, v in LANGUAGE_MAPPING.iteritems()}
 
-RE_SPLIT_LANGUAGES = re.compile("\/| or | and |,|=")
+RE_SPLIT_LANGUAGES = re.compile("\/| or | and |,|=|\s+")
```
kaplun: Ouch I now realize it should be:
```python
RE_SPLIT_LANGUAGES = re.compile(r"\/| or | and |,|=|\s+")
```

PR #1504, opened 2016-08-30
WIP jsonschema: no additionalProperties

kaplun:

### DIFF
--- a/inspirehep/modules/records/jsonschemas/records/authors.json
+++ b/inspirehep/modules/records/jsonschemas/records/authors.json
@@ -1,8 +1,10 @@
 {
     "$schema": "http://json-schema.org/schema#",
+    "additionalProperties": false,
     "properties": {
         "advisors": {
             "items": {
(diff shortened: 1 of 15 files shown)


jacquerie:
This was a very optimistic commit : )

We will resuscitate this commit and apply it on top after we:

1. Solve all Sentry issues
2. Make the DoJSON rules and the schemas coherent for all collections

kaplun:
Ah sure :dancer: it's a WIP. But it's a tool to address your second point. 2. is not valid if this commit is included.

jacquerie:
> But it's a tool to address your second point. 2. is not valid if this commit is not included and passing.

Yup, that's the trick to do it: you add this commit, you WIP on top until you pass the validation, then you squash your WIPs and put this commit last : )

kaplun:
Exactly that's what I had in mind

jacquerie:
Moved to https://github.com/inspirehep/inspire-schemas/issues/26.

PR #1511, opened 2016-08-30
dojson: do not migrate 970 SPIRES SYSNOs

kaplun:

This should fix https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601734/


kaplun:
It has been requested to preserve SPIRES IDs

PR #1512, opened 2016-08-30
dojson: ignore Open Access redundant license

kaplun:
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601709/

### DIFF
--- a/inspirehep/dojson/hep/fields/bd5xx.py
+++ b/inspirehep/dojson/hep/fields/bd5xx.py
@@ -209,8 +209,12 @@ def funding_info2marc(self, key, value):
 @utils.filter_values
 def license(self, key, value):
     """Add Terms Governing Use and Reproduction Note."""
+    license_value = force_force_list(value.get('a'))
+    # We strip away redundant 'Open Access' string
+    license_value = [val for val in license_value if license_value != 'Open Access']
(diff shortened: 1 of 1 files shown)


inspirehep/dojson/hep/fields/bd5xx.py
```diff
@@ -209,8 +209,12 @@ def funding_info2marc(self, key, value):
 @utils.filter_values
 def license(self, key, value):
     """Add Terms Governing Use and Reproduction Note."""
+    license_value = force_force_list(value.get('a'))
+    # We strip away redundant 'Open Access' string
+    license_value = [val for val in license_value if license_value != 'Open Access']
+    license_value = force_single_element(license_value)
```
jacquerie: This discards data. Are we modeling it correctly? Software things can be dual-licensed, why not papers?

kaplun: I have checked. Only 17 records have these Open Access license (which is not a valid name for a license) and have also a valid CC-BY-* one. These are also the very same records having more than one license.

All the other papers have valid CC-* licenses.

david-caro:
dammit... I was working on this :)

kaplun:
There were no comment on Sentry... Oops!

PR #1514, opened 2016-08-31
dojson: multiple 970 SPIRES sysnos

kaplun:

Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601734/

### DIFF
--- a/inspirehep/dojson/common/base.py
+++ b/inspirehep/dojson/common/base.py
@@ -148,15 +148,16 @@ def spires_sysnos(self, key, value):
     value = force_force_list(value)
     new_recid = None
     for val in value:
-        if 'a' in val:
-            external_system_numbers.append(
-                {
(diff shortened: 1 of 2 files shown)


inspirehep/dojson/common/base.py
```diff
@@ -149,13 +149,15 @@ def spires_sysnos(self, key, value):
...
     for val in value:
         if 'a' in val:
-            external_system_numbers.append(
-                {
-                    "institute": "SPIRES",
-                    "value": val.get('a'),
-                    "obsolete": True
-                }
-            )
+            for sysno in force_force_list(val['a']):
```
jacquerie:
```python
if 'a' in val:
    for sysno in force_force_list(val['a']):
        ...
```
can just be
```python
for sysno in force_force_list(val.get('a')):
    ...
```

PR #1515, opened 2016-08-31
dojson: skip invalid DOIs

kaplun:

Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601552/


kaplun:
Looks like these fields are actually `Handle`s, rather than `DOI`s, so rather than failing porting them, we should rather tag them properly.

PR #1522, opened 2016-09-01
WIP warnings: suppress SQLAlchemy warnings

jacquerie:
With this PR we go from:
```
$ inspirehep shell
/Users/jacquerie/.virtualenvs/inspire/lib/python2.7/site-packages/sqlalchemy/ext/declarative/clsregistry.py:120: SAWarning: This declarative base already contains a class with the same class name and module name as flask_sqlalchemy.RecordMetadataVersion, and will be replaced in the string-lookup table.
  item.__name__

/Users/jacquerie/.virtualenvs/inspire/lib/python2.7/site-packages/sqlalchemy/orm/properties.py:194: SAWarning: On mapper Mapper|RecordMetadataVersion|records_metadata_version, primary key column 'records_metadata_version.transaction_id' is being combined with distinct primary key column 'records_metadata_version.transaction_id' in attribute 'transaction_id'.  Use explicit properties to give each column its own mapped attribute name.
  self.columns[0], self.key))

Python 2.7.12 (default, Jun 29 2016, 14:04:44)
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
IPython: 5.1.0
App: inspirehep [debug]
Instance: /Users/jacquerie/.virtualenvs/inspire/bin/../var/inspirehep-instance

In [1]:
```
to:
```
$ inspirehep shell
Python 2.7.12 (default, Jun 29 2016, 14:04:44)
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
IPython: 5.1.0
App: inspirehep [debug]
Instance: /Users/jacquerie/.virtualenvs/inspire/bin/../var/inspirehep-instance

In [1]:
```

### DIFF
--- a/inspirehep/modules/warnings/ext.py
+++ b/inspirehep/modules/warnings/ext.py
@@ -20,11 +20,7 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-"""Extension that suppresses warnings.
-
-By default Invenio-Base enables deprecation warnings. This extension
-disables them, unless ``DEBUG_SHOW_DEPRECATION_WARNINGS`` is ``True``.
-"""
+"""Extension that suppresses warnings."""
 
 from __future__ import absolute_import, division, print_function
 
@@ -44,11 +40,7 @@ class INSPIREWarnings(object):
         """Initialize the application."""
         self.init_config(app)
 
-        if not app.config['DEBUG_SHOW_DEPRECATION_WARNINGS']:
-            warnings.filterwarnings('ignore', category=DeprecationWarning)
-
         app.extensions['inspire-warnings'] = self
 
     def init_config(self, app):
         """Initialize the configuration."""
-        app.config.setdefault('DEBUG_SHOW_DEPRECATION_WARNINGS', False)
--- a/inspirehep/modules/warnings/ext.py
+++ b/inspirehep/modules/warnings/ext.py
@@ -20,11 +20,34 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-"""INSPIREWarnings module."""
+"""Warnings extension that removes deprecation warnings.
+
+Invenio-Base enables deprecation warnings and this extension removes
+this setting unless ``DEBUG_SHOW_DEPRECATION_WARNINGS`` is set to True.
+"""
 
 from __future__ import absolute_import, division, print_function
 
-from .ext import INSPIREWarnings
+import warnings
 
 
-__all__ = ("INSPIREWarnings", )
+class INSPIREWarnings(object):
+    """INSPIREWarnings extension implementation."""
+
+    def __init__(self, app=None):
+        """Initialize extension."""
+        self.app = app
+        if app is not None:
+            self.init_app(app)
+
+    def init_app(self, app):
+        """Initialize a Flask application."""
+        app.config.setdefault("DEBUG_SHOW_DEPRECATION_WARNINGS",
+                              False)
+
+        if not app.config['DEBUG_SHOW_DEPRECATION_WARNINGS']:
+            warnings.filterwarnings(
+                'ignore',
+                category=DeprecationWarning
+            )
+        app.extensions["inspire-warnings"] = self
--- a/inspirehep/modules/warnings/ext.py
+++ b/inspirehep/modules/warnings/ext.py
@@ -20,10 +20,10 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-"""Warnings extension that removes deprecation warnings.
+"""Extension that suppresses warnings.
 
-Invenio-Base enables deprecation warnings and this extension removes
-this setting unless ``DEBUG_SHOW_DEPRECATION_WARNINGS`` is set to True.
+By default Invenio-Base enables deprecation warnings. This extension
+disables them, unless ``DEBUG_SHOW_DEPRECATION_WARNINGS`` is ``True``.
 """
 
 from __future__ import absolute_import, division, print_function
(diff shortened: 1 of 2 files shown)


jacquerie:
@kaplun, @jmartinm: the first 3 commits refactor the `warnings` module to be a single place where we suppress warnings. It doesn't provide any helper to do that, but they should be easy to write. The fourth commit uses this refactored module to suppress SQLAlchemy warnings, since they appear to come from some unwanted interaction between Flask-SQLAlchemy and Invenio-Records, outside our control. We could also use this to suppress the `ExtDeprecationWarning`s.

kaplun:
@jacquerie but the warning in invenio-record is there for some reason right? What could go wrong if we ignore it? What Invenio developers should do to fix it? Is there some fix we can send upstream to fix it?

jmartinm:
I think there was some discussion on the Invenio chat. If you think that is is safe to supress SQLAlchemy warnings, the code LGTM.

kaplun:
@jmartinm I guess you refer to https://gitter.im/inveniosoftware?at=57c7e6dedc3c80b559a76ac8

Now unfortunately that discussion had been moved privately between @tiborsimko and @jacquerie so I am not aware of the conclusion.

So shall we hide in inspire-next invenio-specific warnings?

inspirehep/modules/warnings/ext.py
```diff
@@ -0,0 +1,65 @@
...
+        """Initialize the extension."""
+        if app:
+            self.init_app(app)
+
+    def init_app(self, app):
+        """Initialize the application."""
+        self.init_config(app)
+
+        if not app.config['DEBUG_SHOW_DEPRECATION_WARNINGS']:
+            warnings.filterwarnings('ignore', category=DeprecationWarning)
```
kaplun: I don't believe it's wise to silence deprecation warnings. In particular Invenio is going to use them more and more after releasing 1.0 and we risk to miss them.

jacquerie: Well, that's what this code has been doing for the past six months. They are going to break compatibility without warning anyway.

jacquerie:
> Now unfortunately that discussion had been moved privately between @tiborsimko and @jacquerie so I am not aware of the conclusion.

In fact, there wasn't any.

inspirehep/modules/warnings/ext.py
kaplun: Until 1.0 yes, after nope.

kaplun:
Ok. @jmartinm explained that indeed we were already silencing warnings. Sorry haven't understood that. But so, assuming it's nice not to have a cluttered screen and logs, we need a process to also have all the warnings so that we don't miss important one. 
What can be done in that case?

inspirehep/modules/warnings/ext.py
```diff
@@ -0,0 +1,65 @@
...
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Extension that suppresses warnings.
+
+By default Invenio-Base enables deprecation warnings. This extension
+disables them, unless ``DEBUG_SHOW_DEPRECATION_WARNINGS`` is ``True``.
+
+This extension also disables SQLAlchemy warnings, unless the variable
+``DEBUG_SHOW_SQLALCHEMY_WARNINGS` is set to ``True``.
```
jacquerie: @kaplun answering https://github.com/inspirehep/inspire-next/pull/1522#issuecomment-245849314.

kaplun: Yes, but what I mean is: how do we ensure someone will from time to time check them? inspire-nightly making a report?

jacquerie: Nah: if having them constantly repeated on your terminal every time you gave a `inspirehep` command didn't make you fix them for the past six months, how can a nightly report help with that?

kaplun: @jacquerie are these warnings telling us something important or not? The fact that so far we have ignored them, doesn't mean that it was the right thing to do. I can imagine that some warnings are really just noise. But others might be the anticipation of some imminent disaster.
Until now we were not using stuff on production for real, so no harm was caused. But as we approach putting on Invenio 3 on production we need to become more proactive towards them. 
Can you list on Github which warnings we have at the moment that you think it's safe to ignore? Are there warnings we should instead act upon (either in inspire-next repo or on upstream Invenio packages)?

david-caro: Is there any way to filter out just the ones we know are 'safe'? (for example, being already acted upon)

jacquerie: No breaking change in the Invenio API was ever anticipated by a warning, yet breaking API changes have happened all the time, and will continue to happen because they don't care about us. You just don't see them because they're caught by nightly and fixed on the spot before the next day.

The list of warnings that are safe to ignore is exactly this PR.

kaplun: @jacquerie they introduced in Invenio 2 deprecation warnings but given the need for reimplementing everything from scratch in Invenio 3, and given it's bleeding edge status (although the only code really runnable on production), and our pressure towards reaching quickly Invenio 3 they decided not to use deprecation warnings (which are slowing down development) until they make the stable release 1.0. I think it was a necessary evil. We can't ask for perfect situation and speed all in one step. Especially since INSPIRE itself was originally behind the request to have the hybrid of Invenio 2 to allow legacy and Flask to coexist which ended up slowing down reimplementation of Invenio. 

After 1.0 API will be declared stable and will evolve using deprecation warning techniques.

Also if they break the API and fix them in one day I think it's quick enough. Of course we have to be careful that if we deploy on production we must use pinned versions that we know they work.

We could use for help us this: https://github.com/inveniosoftware/requirements-builder

Regarding this PR, if indeed these Warnings are the only to go LGTM.

lnielsen: >  yet breaking API changes have happened all the time, and will continue to happen because they don't care about us. 

@jacquerie I am a bit hurt your statement. Personally I've invested an enormous amount of time in getting Invenio 3 to where it is today - both at work but also during evening, nights and weekends (take a look at top and bottom line of the contributions on my [GitHub profile](https://github.com/lnielsen) and you might get an idea). My job is not to develop Invenio 3, my job is to run Zenodo, like it's your job to run INSPIRE. However I (perhaps naively) believe that by contributing as much to Invenio 3 as I've done, that I'll get a better product/service in the end. I'm sorry if you feel that we don't care about you, however please try to see it from the other side that it is in fact quite the opposite. I could have decided to just focus on Zenodo, instead I've devoted much more than what is required of me to help build a product that (I hope) is also very useful to you guys and (hopefully) lets you focus on building a great service. 

As for the breaking changes during alpha development of Invenio 3, then I think this is normal and expectable. If we were to provide backward compatibility it would just delay the time at which we could finally declare Invenio 3 fully stable and make a final 3.0 release. And also, the breaking changes was not to bother you, but to get a better product and more clean APIs in the end. Last but not least, I don't feel comfortable declareing any module stable, unless it's been actually battle tested on a production site.
The main goal for Invenio 3 is to have a suite of stable modules with clean APIs and having breaking changes properly managed. This was actually the main reason for initially starting on Invenio 3, and I think it's an absolutely must for running a service on top of Invenio.

So, we do care about you, however we have services to run like you, so I'd like to ask for a bit of patience/understanding until we can finally release Invenio v3.0, or even better help us get there faster for the common good of all our services.

PR #1526, opened 2016-09-01
dojson: handle multiple experiments on authors

david-caro:
sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601676/

### DIFF
--- a/inspirehep/dojson/hepnames/fields/bd1xx.py
+++ b/inspirehep/dojson/hepnames/fields/bd1xx.py
@@ -393,48 +393,85 @@ def _curators_note2marc(self, key, value):
 
 
 @hepnames.over('experiments', '^693..')
-@utils.for_each_value
-def experiments(self, key, value):
+def experiments(self, key, values):
     """Information about experiments.
 
     Currently the ``id`` subfield stores the name of the experiment. It should
     be moved to storing id of the experiment as soon as experiments are
     available as records.
     """
-    try:
-        start_year = int(value.get('s'))
-    except (TypeError, ValueError):
-        start_year = None
-    try:
-        end_year = int(value.get('d'))
-    except (TypeError, ValueError):
-        end_year = None
+    # FIXME: change to use the flatten decorator once the new dojson version is
+    # out
 
-    return {
-        'id': value.get('i'),
-        'name': value.get('e'),
-        'start_year': start_year,
-        'end_year': end_year,
-        'status': value.get('z')
-    }
+    def _int_or_none(maybe_int):
+        try:
+            return int(maybe_int)
+        except (TypeError, ValueError):
+            return None
+
+    def _get_json_experiments(marc_dict):
+        start_year = _int_or_none(marc_dict.get('s'))
+        end_year = _int_or_none(marc_dict.get('d'))
+
+        names = force_force_list(marc_dict.get('e'))
+        recids = force_force_list(marc_dict.get('0'))
+        name_recs = zip(names, recids or [None] * len(names))
+
+        for name, recid in name_recs:
+            record = get_record_ref(recid, 'experiments')
+            yield {
+                'curated_relation': record is not None,
+                'current': (
+                    True if marc_dict.get('z', '').lower() == 'current'
+                    else False
+                ),
+                'end_year': end_year,
+                'name': name,
+                'record': record,
+                'start_year': start_year,
+            }
+
+    values = force_force_list(values)
+    json_experiments = self.get('experiments', [])
+    for experiment in values:
+        if experiment:
+            json_experiments.extend(_get_json_experiments(experiment))
+
+    return json_experiments
 
 
 @hepnames2marc.over('693', '^experiments$')
-@utils.for_each_value
-def experiments2marc(self, key, value):
+def experiments2marc(self, key, values):
     """Information about experiments.
 
     Currently the ``id`` subfield stores the name of the experiment. It should
     be moved to storing id of the experiment as soon as experiments are
     available as records.
     """
-    return {
-        'i': value.get('id'),
-        'e': value.get('name'),
-        's': value.get('start_year'),
-        'd': value.get('end_year'),
-        'z': value.get('status')
-    }
+    # FIXME: change to use the flatten decorator once the new dojson version is
+    # out
+
+    def _get_marc_experiment(json_dict):
+        marc = {
+            'e': json_dict.get('name'),
+            's': json_dict.get('start_year'),
+            'd': json_dict.get('end_year'),
+        }
+        status = 'current' if json_dict.get('current') else None
+        if status:
+            marc['z'] = status
+        recid = get_recid_from_ref(json_dict.get('record', None))
+        if recid:
+            marc['0'] = recid
+        return marc
+
+    marc_experiments = self.get('693', [])
+    values = force_force_list(values)
+    for experiment in values:
+        if experiment:
+            marc_experiments.append(_get_marc_experiment(experiment))
+
+    return marc_experiments
 
 
 @hepnames.over('advisors', '^701..')
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -49,12 +49,13 @@ def email_app():
         CELERY_EAGER_PROPAGATES_EXCEPTIONS=True,
         MAIL_SUPPRESS_SEND=True,
     )
+
     FlaskCeleryExt(app)
 
     InvenioMail(app, StringIO())
 
     with app.app_context():
-        yield app
+        return app
 
 
 @pytest.fixture(scope='session', autouse=True)
--- a/tests/unit/dojson/fixtures/test_hepnames_record.xml
+++ b/tests/unit/dojson/fixtures/test_hepnames_record.xml
@@ -114,15 +114,6 @@
     <datafield tag="678" ind1=" " ind2=" ">
         <subfield code="a">Nobel Prize Physics 2003</subfield>
     </datafield>
-    <datafield tag="693" ind1=" " ind2=" ">
-        <subfield code="e">SDSS</subfield>
-    </datafield>
-    <datafield tag="693" ind1=" " ind2=" ">
-        <subfield code="d">2020</subfield>
-        <subfield code="e">CERN-ALPHA</subfield>
-        <subfield code="s">2014</subfield>
-        <subfield code="z">current</subfield>
-    </datafield>
     <datafield tag="701" ind1=" " ind2=" ">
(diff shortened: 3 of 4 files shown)


jacquerie:
Nit: misspelled "authors" in the commit message.

inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -25,6 +25,7 @@
 from __future__ import absolute_import, division, print_function
 
 import re
```
jacquerie: Nit: space here, because `re` is a stdlib dependency, while `six` is vendored.

inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -397,50 +398,71 @@ def _curators_note2marc(self, key, value):
...
-@utils.for_each_value
-@utils.filter_values
-def experiments(self, key, value):
+def experiments(self, key, values):
     """Information about experiments.
 
     Currently the ``id`` subfield stores the name of the experiment. It should
     be moved to storing id of the experiment as soon as experiments are
     available as records.
     """
```
jacquerie: Is this docstring still relevant? I mean, `693__i` is basically an unused field: https://inspirehep.net/search?ln=en&cc=HepNames&of=text&ot=693__i&p=693__i:**.

inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -397,50 +398,71 @@ def _curators_note2marc(self, key, value):
...
+
+    def _get_json_experiments(marc_dict):
+        try:
+            start_year = int(marc_dict.get('s'))
+        except (TypeError, ValueError):
+            start_year = None
+        try:
+            end_year = int(marc_dict.get('d'))
+        except (TypeError, ValueError):
+            end_year = None
```
jacquerie: Maybe abstract this in a local helper, to avoid a repetition this close.

inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -397,50 +398,71 @@ def _curators_note2marc(self, key, value):
...
+
+        for name in force_force_list(marc_dict.get('e')):
+            yield {
+                'id': marc_dict.get('i'),
+                'name': name,
+                'start_year': start_year,
+                'end_year': end_year,
+                'status': marc_dict.get('z')
+            }
```
jacquerie: This doesn't fit in the schema that you should be generating: https://github.com/inspirehep/inspire-next/blob/90032ba4825c6c4dda78c9c2b1d3edf18692b9dc/inspirehep/modules/records/jsonschemas/records/authors.json#L65-L86

tests/unit/dojson/fixtures/test_hepnames_record.xml
```diff
@@ -123,6 +123,15 @@
         <subfield code="s">2014</subfield>
         <subfield code="z">current</subfield>
     </datafield>
+    <!-- test that repeated names on the same experiment are split into
+         different expeniment datafield by dojson
```
jacquerie: Nit: s/expeniment/experiments/

inspirehep/dojson/hepnames/fields/bd1xx.py
david-caro: True!

jacquerie:
Left a few nits, a suggestion on how to DRY a bit the code but more importantly something to fix, the fact that the record produced doesn't respect the schema.

inspirehep/dojson/hepnames/fields/bd1xx.py
david-caro: I'm very confused..., it did not fit the schema before either right? And it seems by far (no id... no status...)

jacquerie: Yep, it was pretty broken. In fact, I have a feeling this commit will become "schemas: rewrite experiments rule in authors" : )

tests/unit/dojson/test_dojson_hepnames.py
```diff
@@ -30,7 +30,10 @@
 from dojson.contrib.marc21.utils import create_record
 
 from inspirehep.dojson.hepnames import hepnames2marc, hepnames
-from inspirehep.dojson.utils import clean_record, get_recid_from_ref
+
+from inspirehep.dojson.utils import clean_record, get_record_ref
+
+from ..conftest import app
```
david-caro: I'm not sure how should I order these imports...

Another thing, in order to be able to import the `app` fixture to use it directly, I have to create `__init__.py` files in the tests dirs, is that ok? (missed them on this commit, if ok will send on next, if not ok, will have to rethink how the app context is created).

inspirehep/dojson/hepnames/fields/bd1xx.py
david-caro: dammit... missed it, will change

inspirehep/dojson/hepnames/fields/bd1xx.py
david-caro: @kaplun ^? (I think I already asked you, but I don't remember the answer :disappointed_relieved:  )

tests/unit/dojson/test_dojson_hepnames.py
jacquerie: From http://doc.pytest.org/en/latest/goodpractices.html#choosing-a-test-layout-import-rules:
> avoid “__init__.py” files in your test directories. This way your tests can run easily against an installed version of mypkg, independently from the installed package if it contains the tests or not.

tests/unit/dojson/test_dojson_hepnames.py
```diff
@@ -30,7 +30,10 @@
 from dojson.contrib.marc21.utils import create_record
 
 from inspirehep.dojson.hepnames import hepnames2marc, hepnames
-from inspirehep.dojson.utils import clean_record, get_recid_from_ref
+
```
jacquerie: Extra space, order is correct.

tests/unit/dojson/test_dojson_hepnames.py
```diff
@@ -55,16 +58,195 @@ def test_dates(marcxml_to_json, json_to_marc):
...
+                'e': 'SECOND-SIMULTANEOUS',
+                's': 2015,
+            },
+        ],
+    ),
+
+]
+
+
+@pytest.mark.parametrize("data", EXPERIMENTS_DATA, ids=idfn)
```
david-caro:
```
_____________________________________ test_experiments[simultaneous_experiments] ______________________________________

experiments_data = ('simultaneous_experiments', '\n            <datafield tag="693" ind1=" " ind2=" ">\n                <subfield code="d...{'0': 1, 'd': 2013, 'e': 'FIRST-SIMULTANEOUS', 's': 2015}, {'0': 2, 'd': 2013, 'e': 'SECOND-SIMULTANEOUS', 's': 2015}])

    def test_experiments(experiments_data):
        """Test if experiments is created correctly."""
        xml_snippet, expected_json, expected_marc = experiments_data[1:]
        if not xml_snippet.strip().startswith('<record>'):
            xml_snippet = '<record>%s</record>' % xml_snippet
    
        json_data = clean_record(hepnames.do(create_record(xml_snippet)))
        json_experiments = json_data['experiments']
        marc_experiments = hepnames2marc.do(json_data)['693']
    
        assert marc_experiments == expected_marc
>       assert json_experiments == expected_json
E       assert [{'curated_re...ANEOUS', ...}] == [{'curated_rel...ANEOUS', ...}]
E         At index 0 diff: {'start_year': 2015, 'name': 'FIRST-SIMULTANEOUS', 'record': {'$ref': 'http://localhost:5000/api/record/00001'}, 'end_year': 2013, 'current': False, 'curated_relation': True} != {'start_year': 2015, 'name': 'FIRST-SIMULTANEOUS', 'record': {'$ref': 'http://inspirehep.net/api/record/00001'}, 'end_year': 2013, 'current': False, 'curated_relation': True}
E         Use -v to get the full diff

tests/unit/dojson/test_dojson_hepnames.py:246: AssertionError
======================================== 4 failed, 36 passed in 11.43 seconds =========================================
```

there you go, it's more readable if you run with verbose, and even easier to debug if you run with --pdb (or --ipdb if ipdb is installed in the venv, or --bpdb), the original xml, the original parsed xml, the json result and the expectation and the parsed marc result and expectation are locally in that env too, so just a print of the one you want to check will show the contents.

Still trying to figure out how to use the 'app' fixture without the modules... does not seem an easy task... it looks very strange that pytest does not recommend reusing the code as modules...

david-caro: found an ugly way, using a lambda to run the get_ref function inside the fixture context...

tests/unit/conftest.py
```diff
@@ -49,12 +50,14 @@ def email_app():
         CELERY_EAGER_PROPAGATES_EXCEPTIONS=True,
         MAIL_SUPPRESS_SEND=True,
     )
+    app.config.update(PRESERVE_CONTEXT_ON_EXCEPTION=False,)
```
david-caro: I'll try removing this too, this is supposed to fix the 'popping wrong app context' issue, but it did not, the new code does not have that issue though, so it might not be needed

tests/unit/conftest.py
```diff
@@ -74,6 +77,7 @@ def app():
         TESTING=True,
     )
 
+    app.config.update(PRESERVE_CONTEXT_ON_EXCEPTION=False,)
```
david-caro: same here

tests/unit/dojson/test_dojson_hepnames.py
```diff
@@ -55,16 +56,200 @@ def test_dates(marcxml_to_json, json_to_marc):
...
     """Test if experiments is created correctly."""
-    assert (marcxml_to_json['experiments'][1]['name'] ==
-            json_to_marc['693'][1]['e'])
-    assert (marcxml_to_json['experiments'][1]['start_year'] ==
-            json_to_marc['693'][1]['s'])
-    assert (marcxml_to_json['experiments'][1]['end_year'] ==
-            json_to_marc['693'][1]['d'])
-    assert (marcxml_to_json['experiments'][1]['status'] ==
-            json_to_marc['693'][1]['z'])
+    mock_get_record_ref.side_effect = lambda x: x and 'mocked_recid_%s' % x
```
david-caro: this actually should not work... as you should be able to call get_record_ref with an extra parameter (the collection), will investigate

PR #1533, opened 2016-09-05
global: roles fix

Panos512:
* `roles` field has been deleted.
* `contributor_roles` is now used to store author roles.
* Solves https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601708/ .

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -1095,6 +1095,25 @@ CRAWLER_SETTINGS = {
 # Inspire mappings
 # ================
 
+INSPIRE_LEGACY_ROLES = {
+    'editing': [
+        'ed.',
+        'eds.',
+        'ed,,',
+        'eds',
+        'ed,',
+        'ed. et al.'
+    ],
+    'administration': [
+        'task force leader',
+        'resource manager',
+        'scientific coordinator',
+        'chairman',
+        'chair',
+        'workshop chair'
+    ]
+}
+
 INSPIRE_CATEGORIES = [
     'Accelerators',
     'Astrophysics',
--- a/inspirehep/dojson/hep/fields/bd1xx.py
+++ b/inspirehep/dojson/hep/fields/bd1xx.py
@@ -24,10 +24,16 @@
 
 from __future__ import absolute_import, division, print_function
 
+import six
 import logging
 import re
 
 from dojson import utils
+
+from flask import current_app
+
+from inspirehep.utils.record import get_value as utils_get_value
+from inspirehep.utils.helpers import force_force_list
 
 from ..model import hep, hep2marc
 from ...utils import (
@@ -141,6 +147,34 @@ def authors(self, key, value):
             if x_value and x_value.isdigit():
                 return get_record_ref(x_value, 'authors')
 
+        def _get_contributor_role(value):
+            if not value:
+                return None
+
+            if not isinstance(value, six.string_types):
+                value = set(value)
+
+            values = force_force_list(value)
+
+            contributor_roles = []
+            for value in values:
+                value = value.lower()
+                if value in current_app.config['INSPIRE_LEGACY_ROLES']['editing']:
+                    contributor_roles.append(
+                        {
+                            'schema': 'CRediT',
+                            'value': 'Writing - review & editing'
+                        }
+                    )
+                if value in current_app.config['INSPIRE_LEGACY_ROLES']['administration']:
+                    contributor_roles.append(
+                        {
+                            'schema': 'CRediT',
+                            'value': 'Project administration'
+                        }
+                    )
+            return contributor_roles
+
         return {
             'affiliations': _get_affiliations(value),
             'alternative_names': force_force_list(value.get('q')),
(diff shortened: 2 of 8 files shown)


inspirehep/dojson/hep/fields/bd1xx.py
```diff
@@ -141,6 +145,30 @@ def _get_record(value):
...
+                return None
+            if not isinstance(value, str):
+                value = set(value)
+                if len(value) is 1:
+                    value = value.pop()
+
+            if value in current_app.config['INSPIRE_LEGACY_ROLES']['editing']:
+                return [
+                    {
+                        'source': 'CRediT',
```
kaplun: I now realize this should have been `'scheme': 'CRediT'`. Because the source is where the data come from, while the schema is the namespace against which the value is written.

Can you amend also the json schema?

inspirehep/config.py
```diff
@@ -1095,6 +1095,28 @@ def _(x):
...
+        'Ed.',
+        'ed,,',
+        'eds',
+        'ed,',
+        'ed. et al.',
+        'et al.'
+    ],
+    'administration': [
+        'Task Force Leader',
+        'resource manager',
```
kaplun: @suenjedt shoudn't this be `Resources` rather than `Project administration`?

inspirehep/dojson/hep/fields/bd1xx.py
```diff
@@ -141,6 +145,30 @@ def _get_record(value):
             if x_value and x_value.isdigit():
                 return get_record_ref(x_value, 'authors')
 
+        def _get_contributor_role(value):
+            if not value:
+                return None
+            if not isinstance(value, str):
+                value = set(value)
+                if len(value) is 1:
+                    value = value.pop()
```
kaplun: I think here you should use one of the utilities `force_force_list()`, I believe we have mentioned IRL that one author can have a list of contributor roles, so you should force your input to be a list using the helper and then happily append all the mapped values.

inspirehep/config.py
```diff
@@ -1095,6 +1095,28 @@ def _(x):
...
+INSPIRE_LEGACY_ROLES = {
+    'editing': [
+        'ed.',
+        'eds.',
+        'Ed.',
+        'ed,,',
+        'eds',
+        'ed,',
+        'ed. et al.',
+        'et al.'
```
kaplun: I believe `et al.` shouldn't be there. It's due to when people are cut & pasting from a reference.

inspirehep/dojson/hep/fields/bd1xx.py
```diff
@@ -141,6 +146,33 @@ def _get_record(value):
...
+                return None
+
+            if not isinstance(value, str):
+                value = set(value)
+
+            values = force_force_list(value)
+
+            contributor_roles = []
+            for value in values:
+                if value in current_app.config['INSPIRE_LEGACY_ROLES']['editing']:
```
kaplun: What about making it more robust and force `value` and all the content of `INSPIRE_LEGACY_ROLES` to be lowercase, so that we are future proof.
Invenio 1 was case insensitive.

inspirehep/dojson/hep/fields/bd1xx.py
```diff
@@ -175,7 +207,8 @@ def get_value(value):
         ]
         return {
             'a': value.get('full_name'),
-            'e': value.get('role'),
+            # FIXME: We only add the first role to the xml
+            'e': utils_get_value(value, 'contributor_roles[0].value'),
```
kaplun: I think `utils_get_value(value, 'contributor_roles.value')` would return you the whole list of values. (hopefully)

Panos512: yeah but is dojson smart enough to make multiple xml tags with this list?

kaplun: I think so :)

inspirehep/dojson/hep/fields/bd1xx.py
```diff
@@ -141,6 +146,34 @@ def _get_record(value):
             if x_value and x_value.isdigit():
                 return get_record_ref(x_value, 'authors')
 
+        def _get_contributor_role(value):
+            if not value:
+                return None
+
+            if not isinstance(value, str):
```
jmartinm: for Python 3 compatibility `six.string_types`

inspirehep/modules/disambiguation/search.py
```diff
@@ -349,7 +349,7 @@ def get_signature(uuid):
             "authors.orcid",
             "authors.profile",
             "authors.recid",
-            "authors.role",
+            "authors.contributor_roles[0].value",
```
jmartinm: Does this work? Or when specifying `_source` you just put `"authors.contributor_roles.value"` ?

jmartinm:
Except the small comments, looks 👍

spirosdelviniotis:
Suenje says it is project administration (we looked it up in the taxonomy)

PR #1508, opened 2016-08-30
WIP setup: remove theme ext from API app

jacquerie:
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/601686/

* The theme extension adds a few error handlers to the
  `error_handler_spec` of the application. These error
  handlers try to render a template with custom Jinja2
  filters, which are not loaded in the API app, causing
  another exception that masks the true source of error.
  Therefore, removes the theme extension from the API.

Signed-off-by: Jacopo Notarstefano <jacopo.notarstefano@cern.ch>

PR #1539, opened 2016-09-05
global: theme extension removal from API

jmartinm:
* Removes the theme extension that should not be present in the API app.
  (closes #1508)

* Refactors publication_info jinja2 filter to separate the data generation
  from the template rendering.

* Makes use of LiteratureRecord in search serializer instead of jinja2
  filter to avoid dependency on templates. The dependency on the templates
  is the reason why theme is loaded into API at the moment.

### DIFF
--- a/inspirehep/modules/records/wrappers.py
+++ b/inspirehep/modules/records/wrappers.py
@@ -23,6 +23,7 @@
 from __future__ import absolute_import, division, print_function
 
 from inspirehep.utils.record import get_title
+from inspirehep.modules.records.json_ref_loader import replace_refs
 from inspirehep.modules.records.es_record import ESRecord
 from inspirehep.modules.search import JobsSearch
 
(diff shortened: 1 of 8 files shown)


jmartinm:
Some things that I am not sure yet in this PR:

- Since we use the same serializer code for all collections there is a need to detect if we are in Literature or not. At the moment, I added a check based on the `$schema` [see code](https://github.com/inspirehep/inspire-next/pull/1539/files#diff-c3d13284b39cf142017b2f2c5068964aR63)

- When we are in `serializers/json.py`, `record` is a python dict with the ES result, I am not sure about the complexity implication of wrapping it into `LiteratureRecord` [see code](https://github.com/inspirehep/inspire-next/pull/1539/files#diff-c3d13284b39cf142017b2f2c5068964aR64)

- Also a more general 'challenge' that we already had before is that when calling `record.conference_information` if there is a link to a conference there might be extra requests to ES to retrieve the linked records and extract info such as the Conference title.  [see code](https://github.com/inspirehep/inspire-next/pull/1539/files#diff-466969129ace4bfbb91ed718928d1e59R40)
This will impact request time. We need to see if it makes sense to denormalize the information we need (which introduces the challenge of keeping it up-to-date in the children when the parent changes).

jmartinm:
TODO:

* [ ] Render Publication information and Conference information on search results by modifying the templates in https://github.com/inspirehep/inspirehep-search-js

inspirehep/modules/records/wrappers.py
```diff
@@ -35,6 +36,63 @@ def title(self):
         """Get preferred title."""
         return get_title(self)
 
+    @property
+    def conference_information(self):
+        """."""
```
jacquerie: Missing docstring.

inspirehep/modules/records/wrappers.py
```diff
@@ -35,6 +36,63 @@ def title(self):
...
+                    "page_end": pub_info.get('page_end'),
+                    "artid": pub_info.get('artid'),
+                }
+            )
+
+        return conf_info
+
+    @property
+    def publication_information(self):
+        """."""
```
jacquerie: Also here.

inspirehep/modules/theme/jinja2filters.py
```diff
@@ -479,105 +479,76 @@ def json_dumps(data):
 
 @blueprint.app_template_filter()
 def publication_info(record):
-    """Displays inline publication and conference information"""
+    """
+    Displays inline publication and conference information.
+
+    The record is a LiteratureRecord instance
```
jacquerie:
```python
"""Displays inline publication and conference information.

Assumes that the record is an instance of `LiteratureRecord`.
"""
```

inspirehep/modules/records/wrappers.py
```diff
@@ -35,6 +36,63 @@ def title(self):
...
+        pub_info_list = []
+        for pub_info in self['publication_info']:
+            pub_info_list.append({
+                'journal_title': pub_info.get('journal_title', ''),
+                'journal_volume': pub_info.get('journal_volume', ''),
+                'year': str(pub_info.get('year', '')),
```
jacquerie: Python 3: use `six` here and in the following lines?

jmartinm:
Thanks for the comments, although some insights on the 'meat' of the PR - https://github.com/inspirehep/inspire-next/pull/1539#issuecomment-244760361 would be welcome :)

jmartinm:
Depends on https://github.com/inspirehep/inspirehep-search-js/pull/21

jacquerie:
> - Since we use the same serializer code for all collections there is a need to detect if we are in Literature or not. At the moment, I added a check based on the $schema

We should rather have a `JSONBriefSerializer` per collection. In fact, the code is currently trying to trim authors for records that might not even have authors, yuck.

jacquerie:
> - When we are in serializers/json.py, record is a python dict with the ES result, I am not sure about the complexity implication of wrapping it into `LiteratureRecord`

...I don't see the problem! Why do you think this would be a problem?

> - Also a more general 'challenge' that we already had before is that when calling record.conference_information if there is a link to a conference there might be extra requests to ES to retrieve the linked records and extract info such as the Conference title. see code
This will impact request time. We need to see if it makes sense to denormalize the information we need (which introduces the challenge of keeping it up-to-date in the children when the parent changes).
 
This is more challenging, but it's probably not the right time to tackle this: at most we're returning 10 records from a search page, which requires at most 10 additional ES queries, a negligible impact on performance. If in the future we want to allow bigger search pages, or API searches with arbitrary windows, this is going to be a problem.

jmartinm:
> We should rather have a JSONBriefSerializer per collection. In fact, the code is currently trying to trim authors for records that might not even have authors, yuck.

Yep, that would be the more explicit solution. The only thing is that it becomes quite verbose and we need some extra files so when I mentioned it in one standup that I would need one serializer per collection I remember seeing some panic faces.

I will have a look if that can be done with the minimal code added.

jmartinm:
> ...I don't see the problem! Why do you think this would be a problem?

I was wondering if wrapping the dictionary would add some overhead but it is probably ok.

```python
def a():
    x = {"hello": "world"}
    print(x["hello"])

def b():
   x = LiteratureRecord({"hello": "world"})
   print(x["hello"])

%timeit a()
100000 loops, best of 3: 16.7 us per loop

%timeit b()
10000 loops, best of 3: 25.7 us per loop
```

jmartinm:
No more WIP

Issue #1402, opened 2016-08-04
evict SERVER_NAME from config.py


mihaibivol:
`SERVER_NAME` is bad to have and rely on. So far the checklist for doing this includes:

* [ ] Remove all `url_for` calls that are not in views or templates.
* [ ] Change the `url_for` [call](https://github.com/inspirehep/inspire-next/blob/ef45a2dc90b964a76c0a7c70dfa3bc260f3c3615/inspirehep/modules/migrator/tasks.py#L324) needed to get the URL for a given schema file to [this](https://github.com/inveniosoftware/invenio-jsonschemas/blob/6986a1d4a6d80aa97e78d6799df4dfa8ca3200eb/invenio_jsonschemas/ext.py#L140)
* [ ] Don't store the `netloc` in jsonrefs and just present it for the user (e.g http://localhost:5000/smth/ becomes /smth/) but when somebody does a GET he seels http://localhost:5000/smth/
* [ ] Remove all uses of literal `SERVER_NAME`

* [ ] Find other cases that need to be set up

jmartinm:
This is still very relevant. We want to have the $ref URLs in the API be relative so that it facilitates, for example, loading a dump from Labs to QA.

PR #1540, opened 2016-09-05
tests: coverage of web part

kaplun:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,6 +94,11 @@ dump.rdb
 # KDevelop4
 .kdev4/
 *.kdev4
+*.kate-swp
 
 # Vim swapfiles
 .*.sw?
+
+# Selenium report
+assets
+selenium-report.html
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,7 @@ env:
     - SUITE=disambiguation
     - SUITE=workflows
     - SUITE=integration
+    - SUITE=acceptance
 
 matrix:
   fast_finish: true
@@ -61,19 +62,22 @@ before_install:
   - sudo mv docker-compose /usr/local/bin
 
 install:
-  - "travis_retry docker-compose pull"
-  - "travis_retry docker-compose build"
-  - "travis_retry docker-compose -f docker-compose.deps.yml run --rm pip"
-  - "docker-compose -f docker-compose.deps.yml run --rm assets"
+  - travis_retry docker-compose pull
+  - travis_retry docker-compose build
+  - travis_retry docker-compose -f docker-compose.deps.yml run --rm pip
+  - docker-compose -f docker-compose.deps.yml run --rm assets
 
 script:
   - "docker-compose -f docker-compose.test.yml run --rm ${SUITE}"
 
 after_script:
-  - "docker-compose -f docker-compose.test.yml kill"
+  # Killing via SIGTERM in order to trigger atexit and dump coverage information in WSGI
+  - "docker-compose -f docker-compose.test.yml kill -s SIGTERM"
   - "docker-compose -f docker-compose.test.yml rm -f"
 
 after_success:
+  - mv .coverage .coverage.tests
+  - coverage combine
   # Coverage report contains Docker paths. We replace them, so that we can run Coveralls.
   - sed -i 's@\"/code/@'"\"$(pwd)/"'@g' .coverage
   - coveralls
--- a/inspirehep/wsgi_with_coverage.py
+++ b/inspirehep/wsgi_with_coverage.py
@@ -20,9 +20,27 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-[run]
-omit =
-  inspirehep/celery.py
-  inspirehep/cli.py
-  inspirehep/wsgi.py
-  inspirehep/wsgi_with_coverage.py
+"""INSPIREHEP WSGI app instantiation with support for coverage.py.
+
+Uses the trick in http://stackoverflow.com/a/20689873/1407497 to instantiate
+a WSGI application that reports back information on test coverage.
+"""
+
+from __future__ import absolute_import, division, print_function
+
+import atexit
+import sys
+
+import coverage
+
+cov = coverage.Coverage(data_suffix=True)
+cov.start()
+
+from .wsgi import application  # noqa
+
+
+def save_coverage():
+    cov.stop()
+    cov.save()
+
+atexit.register(save_coverage)
--- a/tests/acceptance/conftest.py
+++ b/tests/acceptance/conftest.py
@@ -1,0 +1,75 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Pytest configuration for web tests."""
+
+from __future__ import absolute_import, division, print_function
+
+from os import environ
+from time import sleep
+
+import pytest
+from selenium import webdriver
+
+from invenio_db import db
+from invenio_search import current_search_client as es
+
+from inspirehep.config import SERVER_NAME
+from inspirehep.factory import create_app
+
+
+@pytest.fixture(scope='module')
+def small_app(request):
+    """Flask application fixture."""
+    app = create_app()
+    app.config.update({'DEBUG': True})
+
+    with app.app_context():
+        # Imports must be local, otherwise tasks default to pickle serializer.
+        from inspirehep.modules.migrator.tasks import migrate
+        from inspirehep.modules.fixtures.files import init_all_storage_paths
+        from inspirehep.modules.fixtures.users import init_users_and_permissions
+
+        db.drop_all()
+        db.create_all()
+
+        sleep(10)
+        _es = app.extensions['invenio-search']
+        list(_es.delete(ignore=[404]))
+        list(_es.create(ignore=[400]))
+
+        init_all_storage_paths()
+        init_users_and_permissions()
+
+        migrate('./inspirehep/demosite/data/demo-records-small.xml',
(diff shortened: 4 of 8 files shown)


kaplun:
@mihaibivol @jacquerie I am giving a try at introducing web-side coverage, so that basically we can happily introduce selenium tests and these would automatically include coverage of the corresponding web parts.

kaplun:
WDYT?

david-caro:
does it have any noticable impact on performance? (I guess not)

kaplun:
I guess if it does we can trigger it only for when we'll have a selenium suite.

kaplun:
(Let's see current performance of this PR)

david-caro:
It does not seem to have any noticeable impact, nice.

jacquerie:
Well, this also needs to setup a dummy X server and run at least one selenium test with coverage to be declared as a victory : )

Procfile
```diff
@@ -20,7 +20,7 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-web: gunicorn inspirehep.wsgi -c gunicorn.cfg
+web: gunicorn inspirehep.wsgi_with_coverage -c gunicorn.cfg
```
mihaibivol: Why this change? We don't use procfile on tests

mihaibivol: Did it produce something locally though? :)

mihaibivol:
We can create another subdirectory inside the tests path in which we just do a dummy query to web:5000.  Which reminds me. If we will start the web node inside the test suite we will have to correctly set SERVER_NAME so the page won't respond with a weird response. Alternatively it would be better to start evicting it #1402

Procfile
kaplun: Exactly that was the reason for it :dancer:

kaplun:
@mihaibivol you might need to explain me better what you mean here :+1: 
BTW I have kickstarted directory, with prototype integration with saucelab :dancer:

docker-compose.yml
```diff
@@ -40,7 +40,7 @@ services:
   web:
     extends:
       service: service_base
-    command: gunicorn -b 0.0.0.0:5000 -t 3600 --access-logfile "-" inspirehep.wsgi
+    command: gunicorn -b 0.0.0.0:5000 -t 3600 --access-logfile "-" inspirehep.wsgi_with_coverage
```
mihaibivol: This is only for development and nightly builds. The file you are looking for is `docker-compose.test.yml`

mihaibivol:
The way I see integration with selenium and coverage is kind of like this:

  * Make SERVER_NAME configurable via environment variables (we will need it later)
  * In `docker-compose.test.yml` add:
    *   an `acceptance` service which runs the tests avaiable in the web directory (similar to unit and everything else). Since this target runs in a container it can only ping other containers. In this case `http://web:5000` which will be another container. It will also use `http://selenium:4444` as a command executor. This will have links only to the `web` and `selenium` services for running the tests and probably links to everything else (similar with integration) so we can load some records from the per-session fixture.
    * add a new service called `selenium` that just runs the `selenium/standalone-firefox` Docker image.`
    * add a new service which actually runs the webserver called `web`. It will be linked exactly as the `integration` service but will have SERVER_NAME configured to `web:5000`. In this way the tests that run from acceptance which will call `http://web:5000`~~ will return 200s instead of 404s.

tests/web/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+
+import pytest
+
+from selenium import webdriver
+
+@pytest.yield_fixture(scope='session')
+def driver():
+    """Selenium driver."""
+
+    desired_cap = desired_cap = {
```
jacquerie: wat

kaplun: Ops

tests/web/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+        'platform': "Mac OS X 10.11",
+        'browserName': "firefox",
+        'version': "48",
+        'tunnel-identifier': environ['TRAVIS_JOB_NUMBER']
+    }
```
david-caro: this should be a matrix at some point right? (testing a few browsers and such...)

kaplun: In principle yes, although, I am not sure we want to have too much of it, as it would slow down Travis even further. Maybe we can have just have one mobile and one firefox.

kaplun:
P.s. I have just applied @mihaibivol suggestions for naming things properly.

.gitignore
```diff
@@ -94,6 +94,11 @@ dump.rdb
 # KDevelop4
 .kdev4/
 *.kdev4
+*.kate-swp
```
jacquerie: Unrelated.

.travis.yml
```diff
@@ -61,19 +62,22 @@ before_install:
...
 
 install:
-  - "travis_retry docker-compose pull"
-  - "travis_retry docker-compose build"
-  - "travis_retry docker-compose -f docker-compose.deps.yml run --rm pip"
-  - "docker-compose -f docker-compose.deps.yml run --rm assets"
+  - travis_retry docker-compose pull
+  - travis_retry docker-compose build
+  - travis_retry docker-compose -f docker-compose.deps.yml run --rm pip
+  - docker-compose -f docker-compose.deps.yml run --rm assets
```
jacquerie: Unrelated.

tests/acceptance/conftest.py
```diff
@@ -0,0 +1,77 @@
...
+
+"""Pytest configuration for web tests."""
+
+from __future__ import absolute_import, division, print_function
+
+from os import environ
+from time import sleep
+
+import pytest
+
```
jacquerie: No space here.

tests/acceptance/conftest.py
```diff
@@ -0,0 +1,77 @@
...
+    """Flask application fixture."""
+    app = create_app()
+    app.config.update({'DEBUG': True})
+
+    with app.app_context():
+        # Imports must be local, otherwise tasks default to pickle serializer.
+        from inspirehep.modules.migrator.tasks import migrate
+        from inspirehep.modules.fixtures.files import init_all_storage_paths
+        from inspirehep.modules.fixtures.users import \
+            init_users_and_permissions
```
jacquerie: `from inspirehep.modules.fixtures.users import init_users_and_permissions` (no need for a newline)

inspirehep/wsgi_with_coverage.py
```diff
@@ -0,0 +1,43 @@
...
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""INSPIREHEP WSGI app instantiation but with Coverage."""
+
+from __future__ import absolute_import, division, print_function
+
+# See: http://stackoverflow.com/a/20689873/1407497
```
jacquerie: This should be up in the docstring.

inspirehep/wsgi_with_coverage.py
```diff
@@ -0,0 +1,43 @@
...
+import sys
+
+cov = coverage.Coverage(data_suffix=True)
+cov.start()
+
+from .wsgi import application  # noqa
+
+
+def save_coverage():
+    print("saving coverage", file=sys.stderr)
```
jacquerie: I don't think this is useful.

docker-compose.yml
```diff
@@ -50,6 +50,8 @@ services:
       - indexer
       - rabbitmq
       - redis
+    depends_on:
+      - worker
```
jacquerie: Unrelated? What's this solving?

tests/acceptance/test_login.py
```diff
@@ -0,0 +1,38 @@
...
+    sign_in = selenium.find_element_by_link_text('Sign in')
+    assert sign_in
+    sign_in.click()
+    assert 'Please sign in to suggest content to INSPIRE' in selenium.page_source
+    selenium.get(os.environ['SERVER_NAME'] + '/login/?local=1')
+    email = selenium.find_element_by_id('email')
+    email.send_keys('admin@inspirehep.net')
+    password = selenium.find_element_by_id('password')
+    password.send_keys('123456')
+
```
jacquerie: The test stops at the crucial point before asserting the key thing : )

But it's okay, it's a PoC stub.

tests/acceptance/test_login.py
```diff
@@ -0,0 +1,38 @@
...
+import os
+
+
+def test_login(selenium, small_app):
+    sign_in = selenium.find_element_by_link_text('Sign in')
+    assert sign_in
+    sign_in.click()
+    assert 'Please sign in to suggest content to INSPIRE' in selenium.page_source
+    selenium.get(os.environ['SERVER_NAME'] + '/login/?local=1')
```
jacquerie: Not something to fix now, but I anticipate that this path generation will benefit from an helper.

inspirehep/wsgi_with_coverage.py
```diff
@@ -0,0 +1,43 @@
...
+# or submit itself to any jurisdiction.
+
+"""INSPIREHEP WSGI app instantiation but with Coverage."""
+
+from __future__ import absolute_import, division, print_function
+
+# See: http://stackoverflow.com/a/20689873/1407497
+import atexit
+import coverage
+import sys
```
jacquerie:
```
import atexit
import sys

import coverage
```

inspirehep/wsgi_with_coverage.py
```diff
@@ -0,0 +1,43 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""INSPIREHEP WSGI app instantiation but with Coverage."""
```
jacquerie: """INSPIREHEP WSGI app instantiation with support for coverage.py.

Uses the trick in http://stackoverflow.com/a/20689873/1407497 to instantiate
a WSGI application that reports back information on test coverage.
"""

jacquerie:
LGTM, but some cleanup is needed. I agree 100% with @mihaibivol's architecture.

tests/acceptance/test_login.py
kaplun: Exactly. Actually I have not yet figure out how to click the `Log in` button. Ouch. But in case there is no password field, then the test would also fail, so the `find_element_by*` are weak form of asserts.

docker-compose.yml
kaplun: @mihaibivol tentative to solve an error that was caused buy something else. Actually indeed a leftover. Well spot.

inspirehep/wsgi_with_coverage.py
kaplun: Indeed left over of debugging session.

tests/acceptance/conftest.py
kaplun: Was a cut & paste from an other folder. Will amend it.

kaplun:
This PR indeed implement @mihaibivol's architecture minus (or almost) what is left in #1402.

tests/acceptance/test_login.py
kaplun: Actually in a second step I was thinking of adding a selenium fixture with an already authenticated session. (In Invenio 1 we had such convenience for acceptance testing).

Issue #1523, opened 2016-09-01
migrator: mark records as deleted


jmartinm:
When migrating records from production, records that have `980:DELETED` should be marked as deleted in the new system.

This does not seem to be the case atm:

http://inspire-nightly.cern.ch/search?page=1&size=25&cc=literature&q=980:deleted

Note that `invenio-records` provides a `delete` method in the API - https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L276 so we can use that to `soft-delete` the record.
# Extra

When the record is deleted, accessing the record URL should do a redirection if the field `new_recid` is present to the new record id (as it is the case in production).

kaplun:
For the redirection this could done via the pidstore.

jmartinm:
Done https://github.com/inspirehep/inspire-next/issues/1541

PR #1541, opened 2016-09-06
general: support for deleted records 

spirosdelviniotis:

### DIFF
--- a/inspirehep/dojson/common/base.py
+++ b/inspirehep/dojson/common/base.py
@@ -190,13 +190,18 @@ def collections(self, key, value):
         return {
             'primary': force_single_element(value.get('a')),
             'secondary': force_force_list(value.get('b')),
-            'deleted': value.get('c'),
         }
+
+    def _is_deleted(value):
+        return value and value.lower() == 'deleted'
 
     collections = self.get('collections', [])
 
     values = force_force_list(value)
     for value in values:
+        if _is_deleted(value.get('c')):
+            self['deleted'] = True
+
         collections.append(_get_collection(value))
 
     return collections
@@ -210,8 +215,18 @@ def collections2marc(self, key, value):
     return {
         'a': value.get('primary'),
         'b': value.get('secondary'),
-        'c': value.get('deleted')
     }
+
+
+@hep2marc.over('980', '^deleted$')
+@hepnames2marc.over('980', '^deleted$')
+@utils.for_each_value
+def deleted2marc(self, key, value):
+    """Set Deleted value to marc xml."""
+    if value:
+        return {
+            'c': 'DELETED',
+        }
 
 
 @hep.over('deleted_records', '^981..')
--- a/inspirehep/dojson/common/base.py
+++ b/inspirehep/dojson/common/base.py
@@ -202,8 +202,8 @@ def collections(self, key, value):
     return collections
 
 
-@hep2marc.over('980', 'collections')
-@hepnames2marc.over('980', 'collections')
+@hep2marc.over('980', '^collections$')
+@hepnames2marc.over('980', '^collections$')
 @utils.for_each_value
 def collections2marc(self, key, value):
     """Collection this record belongs to."""
--- a/inspirehep/dojson/common/base.py
+++ b/inspirehep/dojson/common/base.py
@@ -186,20 +186,18 @@ def spires_sysnos2marc(self, key, value):
 @jobs.over('collections', '^980..')
 def collections(self, key, value):
     """Collection this record belongs to."""
-    value = force_force_list(value)
-
-    def get_value(value):
-        primary = force_single_element(value.get('a'))
+    def _get_collection(value):
         return {
-            'primary': primary,
-            'secondary': value.get('b'),
+            'primary': force_single_element(value.get('a')),
+            'secondary': force_force_list(value.get('b')),
             'deleted': value.get('c'),
         }
 
     collections = self.get('collections', [])
 
-    for val in value:
-        collections.append(get_value(val))
+    values = force_force_list(value)
+    for value in values:
+        collections.append(_get_collection(value))
 
     return collections
 
--- a/inspirehep/modules/migrator/tasks/records.py
+++ b/inspirehep/modules/migrator/tasks/records.py
@@ -55,7 +55,7 @@ from inspirehep.modules.pidstore.providers import InspireRecordIdProvider
 from inspirehep.modules.pidstore.minters import inspire_recid_minter
 from inspirehep.utils.dedupers import dedupe_list
 from inspirehep.utils.helpers import force_force_list
-from inspirehep.utils.record import get_value
+from inspirehep.utils.record import get_value, soft_delete_pidstore_for_record
 
 from ..models import InspireProdRecords
 
@@ -314,6 +314,9 @@ def record_upsert(json):
             record = Record.create(json, id_=None)
             # Create persistent identifier.
             inspire_recid_minter(str(record.id), json)
+
+        if json.get('deleted'):
+            soft_delete_pidstore_for_record(record.id)
 
         return record
 
--- a/inspirehep/modules/records/jsonschemas/records/authors.json
+++ b/inspirehep/modules/records/jsonschemas/records/authors.json
@@ -50,6 +50,9 @@
             "format": "date",
             "type": "string"
         },
+        "deleted": {
+            "type": "boolean"
+        },
         "email_addresses": {
             "description": "Contains current email addresses",
             "items": {
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -33,7 +33,8 @@ from invenio_records.signals import before_record_insert, before_record_update
 from inspirehep.dojson.utils import classify_field, get_recid_from_ref
 from inspirehep.utils.date import create_valid_date
 from inspirehep.utils.helpers import force_force_list
-from inspirehep.utils.record import get_value
+from inspirehep.utils.record_getter import get_db_record
+from inspirehep.utils.record import get_value, soft_delete_pidstore_for_record
 
 from .experiments import EXPERIMENTS_MAP
 from .signals import after_record_enhanced
@@ -334,3 +335,15 @@ def add_recids_and_validate(sender, json, *args, **kwargs):
     """Ensure that recids are generated before being validated."""
     populate_recid_from_ref(sender, json, *args, **kwargs)
     references_validator(sender, json, *args, **kwargs)
+
+
+@before_record_update.connect
+def check_if_record_is_going_to_be_deleted(sender, *args, **kwargs):
+    """Checks if 'deleted' field is set as True before updating.
+
+    If 'deleted' field exists and its value is True, before update,
+    then delete all the record's pidstores.
+    """
+    if sender.get('deleted'):
+        record = get_db_record('literature', int(sender.get('control_number')))
+        soft_delete_pidstore_for_record(record.id)
--- a/tests/integration/test_record.py
+++ b/tests/integration/test_record.py
@@ -1,0 +1,141 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+import pytest
+
+from dojson.contrib.marc21.utils import create_record
+from invenio_db import db
+from invenio_indexer.api import RecordIndexer
+from invenio_pidstore.models import PersistentIdentifier
+from invenio_search import current_search_client as es
+
+from inspirehep.dojson.hep import hep
+from inspirehep.modules.migrator.tasks.records import record_upsert
+from inspirehep.utils.record_getter import get_db_record
+
+
+@pytest.fixture(scope='function')
+def record_already_deleted_in_marcxml(app):
+    snippet = (
+        '<record>'
+        '  <controlfield tag="001">222</controlfield>'
+        '  <controlfield tag="005">20160913214552.0</controlfield>'
+        '  <datafield tag="980" ind1=" " ind2=" ">'
+        '    <subfield code="c">DELETED</subfield>'
+        '  </datafield>'
+        '  <datafield tag="980" ind1=" " ind2=" ">'
+        '    <subfield code="a">HEP</subfield>'
+        '  </datafield>'
+        '</record>'
+    )
+
+    with app.app_context():
+        json_record = hep.do(create_record(snippet))
+        json_record['$schema'] = 'http://localhost:5000/schemas/records/hep.json'
+
+        with db.session.begin_nested():
+            record = record_upsert(json_record)
+            if record:
+                r = RecordIndexer()
+                r.index(record)
+
+        db.session.commit()
+
+    yield
+
+    with app.app_context():
+        ri = RecordIndexer()
+
+        record = get_db_record('literature', 222)
+        ri.delete(record)
+        record.delete(force=True)
+
+        pid = PersistentIdentifier.get('literature', 222)
+        PersistentIdentifier.delete(pid)
+
+        object_uuid = pid.object_uuid
+        PersistentIdentifier.query.filter(object_uuid == PersistentIdentifier.object_uuid).delete()
+        db.session.commit()
+
+
+@pytest.fixture(scope='function')
+def record_not_yet_deleted(app):
+    snippet = (
+        '<record>'
+        '  <controlfield tag="001">333</controlfield>'
+        '  <controlfield tag="005">20160913214552.0</controlfield>'
+        '  <datafield tag="980" ind1=" " ind2=" ">'
+        '    <subfield code="a">HEP</subfield>'
+        '  </datafield>'
+        '</record>'
+    )
+
+    with app.app_context():
+        json_record = hep.do(create_record(snippet))
+        json_record['$schema'] = 'http://localhost:5000/schemas/records/hep.json'
+
+        with db.session.begin_nested():
+            record = record_upsert(json_record)
+            if record:
+                ri = RecordIndexer()
+                ri.index(record)
+
+        es.indices.refresh('records-hep')
+        db.session.commit()
+
+    yield
+
+    with app.app_context():
+        record = get_db_record('literature', 333)
+
+        ri = RecordIndexer()
+        ri.delete(record)
+        record.delete(force=True)
+
+        pid = PersistentIdentifier.get('literature', 333)
+        PersistentIdentifier.delete(pid)
+
+        object_uuid = pid.object_uuid
+        PersistentIdentifier.query.filter(object_uuid == PersistentIdentifier.object_uuid).delete()
+        db.session.commit()
+
+
+def test_deleted_record_stays_deleted(app, record_already_deleted_in_marcxml):
+    with app.test_client() as client:
+        assert client.get('/api/literature/222').status_code == 410
+
+
+def test_record_can_be_deleted(app, record_not_yet_deleted):
+    with app.test_client() as client:
+        assert client.get('/api/literature/333').status_code == 200
+
+        record = get_db_record('literature', 333)
+        record['deleted'] = True
+        record.commit()
+        if record:
+            ri = RecordIndexer()
+            ri.index(record)
+        db.session.commit()
+
+        assert client.get('/api/literature/333').status_code == 410
--- a/tests/unit/dojson/test_dojson_common_base.py
+++ b/tests/unit/dojson/test_dojson_common_base.py
@@ -436,7 +436,6 @@ def test_collections_from_980__double_a_double_b():
                 'Something',
                 'Something else',
             ],
-            'c': None,
         },
     ]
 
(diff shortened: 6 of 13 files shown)


inspirehep/utils/helpers.py
```diff
@@ -78,3 +78,12 @@ def force_force_list(data):
     elif isinstance(data, (tuple, set)):
         return list(data)
     return data
+
+def is_deleted(data):
```
kaplun: This is not worth to be put in `.helpers`. You can embed it directly into the `collection` rule.

inspirehep/utils/helpers.py
```diff
@@ -78,3 +78,12 @@ def force_force_list(data):
...
+    """Check if data is deleted.
+    """
+
+    if data.lower() == 'deleted':
```
kaplun: You have to check that `data` is actually different than `None`. (this is what is making the whole PR to crash BTW :dancer: )
So you can do e.g. `if data and data.lower() == 'deleted'.

inspirehep/utils/helpers.py
```diff
@@ -78,3 +78,12 @@ def force_force_list(data):
...
+        return True
+    else:
+        return False
```
kaplun: You can make it even more compact:
```
def is_deleted(data):
    return data and data.lower() == 'deleted'
```

kaplun: One more tiny comment. For consistency: we are calling the argument `value` rather than `data` :dancers: :

tests/unit/dojson/test_dojson_common_base.py
```diff
@@ -382,3 +382,39 @@ def test_urls_from_marcxml_multiple_8564():
     result = clean_record(hep.do(create_record(snippet)))
 
     assert expected == result['urls']
+
+
+def test_for_980_case():
```
kaplun: Yep, better calling `test_collections_deleted_in_980()`

tests/unit/dojson/test_dojson_common_base.py
```diff
@@ -382,3 +382,39 @@ def test_urls_from_marcxml_multiple_8564():
...
+    expected_json = [
+        {
+            'deleted': True,
+            'primary': 'Test_a1',
+            'secondary': 'Something',
+        },
+    ]
+    expected_marc = [
+        {
+            'c': True,
```
kaplun: Be careful: here it should be `'c': 'DELETED'` so something is not correct right now.

inspirehep/utils/helpers.py
david-caro: Outside this module you mean? The previous function here uses 'data'.

kaplun:
Overall the idea was to have a dedicated flag `deleted` to be added to the json schema, so not putting the flag within `collections`.

tests/unit/dojson/test_dojson_common_base.py
```diff
@@ -25,7 +25,7 @@
 from dojson.contrib.marc21.utils import create_record
 
 from inspirehep.dojson.hep import hep
-from inspirehep.dojson.hepnames import hepnames
+from inspirehep.dojson.hepnames import hepnames,hepnames2marc
```
david-caro: missing space after ',' (pep8 issue...)

kaplun:
* [ ] Amend to set the deleted flag in `self['deleted'] = True`
* [ ] Amend all the JSONSchemas to support `deleted` property.

kaplun:
ah! @spirosdelviniotis one more thing: we follow for the commit logs certain conventions detailed here:
http://invenio.readthedocs.io/en/latest/technology/git.html#r2-remarks-on-commit-log-messages

inspirehep/dojson/common/base.py
```diff
@@ -190,14 +190,18 @@ def collections(self, key, value):
...
-            'primary': primary,
-            'secondary': value.get('b'),
-            'deleted': value.get('c'),
+            'primary': force_single_element(value.get('a')),
+            'secondary': force_force_list(value.get('b')),
         }
 
+    if not self.get('deleted', False):
+        self['deleted'] = is_deleted(value[0].get('c'))
+
```
kaplun: As we discussed IRL this could be safer as in:
```python
    for val in value:
        if is_deleted(val.get('c')):
            self['deleted'] = True
        collections.append(get_value(val))
```

david-caro: Yep, sorry, this should actually go inside the get_value, though it would be nicer outside... just after the appent, something like 'update_deleted' or similar

inspirehep/dojson/common/base.py
```diff
@@ -209,16 +213,31 @@ def get_value(value):
...
+@hep2marc.over('980', 'deleted')
+@hepnames2marc.over('980', 'deleted')
+def deleted2marc(self, key, value):
+    """Deleted this record belongs to."""
+
+    return [{
+        'c': "DELETED" if key == 'deleted' and value else None
+    }]
+
+
```
kaplun:
```python
def deleted2marc(self, key, value):
    """Deleted this record belongs to."""
    if key == 'deleted' and value:
        return {
            'c': 'DELETED',
        }
```

inspirehep/dojson/common/base.py
```diff
@@ -209,16 +213,33 @@ def get_value(value):
...
+            value_b = value["secondary"]
+        else:
+            value_b = None
+
     return {
-        'a': value.get('primary'),
-        'b': value.get('secondary'),
-        'c': value.get('deleted')
+        'a': value['primary'],
+        'b': value_b,
```
kaplun:
```python
return {
    'a': value.get('primary'),
    'b': value.get('secondary'),
}
```

inspirehep/dojson/common/base.py
```diff
@@ -209,16 +213,33 @@ def get_value(value):
...
     }
 
 
+@hep2marc.over('980', 'deleted')
+@hepnames2marc.over('980', 'deleted')
```
kaplun: Add `@utils.for_each_value` otherwise it will conflict with the above function.

inspirehep/dojson/common/base.py
```diff
@@ -209,16 +213,33 @@ def get_value(value):
 @hep2marc.over('980', 'collections')
```
kaplun: Please add `'^collection$'`

inspirehep/dojson/common/base.py
david-caro: I thought it was not a regexp...

kaplun:
@spirosdelviniotis I think this PR is still missing:
> * [ ]  Amend all the JSONSchemas to support deleted property.

In every `inspirehep/modules/records/jsonschemas/records/*.json` there should be a new property to represent the deleted field you just introduced.

kaplun:
Regarding commit log, please follow the Invenio convention:
http://invenio.readthedocs.io/en/latest/technology/git.html#r2-remarks-on-commit-log-messages

kaplun:
For an introduction of what jsonschema is (useful for your life): https://spacetelescope.github.io/understanding-json-schema/

kaplun:
BTW you should mention this is to resolve #1523

kaplun:
So, to implement also an integration test we need to:
* [ ] create a record in the system (e.g. using `inspirehep.migrator.tasks.record_upsert()`)
* [ ] amend the created record by setting the `deleted` property to `true`.
* [ ] ensure the record is correctly set as deleted, i.e. validation passes.

At this point the record is not really deleted, right? Next steps will be to ensure that:
* ES filters it out
* pidstore redirects to the merged record (if any) or present a 410.

kaplun:
So to complete this PR I'd propose that you create the three above points, and an additional test that will fail that tests that the record is inaccessible.

To test an HTTP url in flask you can simply use: http://flask.pocoo.org/docs/0.11/testing/#other-testing-tricks

kaplun:
Note that testing that accessing the deleted record actually return 410 will actually fail, since there is not yet the code to implement that.

For that reason you will need to decorate your test with:
```python
@pytest.mark.xfail(reason='deletion of record not yet fully implemented')
def test_foo()
```

kaplun:
So far we have not yet such integration tests. So you can create a first file:
`tests/integration/test_records.py`

See sibling files for how to put copyright headers and similar :)

kaplun:
See e.g. `test_api_authors_root()`.

inspirehep/modules/disambiguation/receivers.py
```diff
@@ -92,7 +92,7 @@ def _needs_beard_reprocessing(authors_before, authors_after):
         return True
 
 
-@after_record_insert.connect
+# @after_record_insert.connect
```
jacquerie: Crude solution to a real problem: this receiver fires when a record is inserted, and then the next receiver fires after some other receiver on the same record indexing is run.

Since the disambiguation module is broken anyway, we should not be firing these messy signals.

kaplun: What do you propose to solve it?

inspirehep/modules/migrator/tasks.py
```diff
@@ -300,6 +300,13 @@ def create_record(record):
 
 def record_upsert(json):
     """Insert or update a record."""
+
+    def soft_delete_record():
+        if 'deleted' in json.keys():
+            if json['deleted']:
```
kaplun:
```
if json.get('deleted'):
```

inspirehep/modules/disambiguation/receivers.py
jacquerie: Revert the merging of the disambiguation module: it's broken, it causes bugs like this.

kaplun: That is not constructive and not generous. The whole module is not broken. It might have certain bad design decisions WRT handling record updates which do not justify removing the whole PR.

I am fine with detaching the signals as proposed in this PR, though.

We need to have disambiguation working for in 2 milestones. (i.e. after Invenio 3, and search)

tests/integration/test_record.py
```diff
@@ -45,85 +45,50 @@ def test_incoming_deleted_record(app):
                '</record>'
                '</collection>')
 
-    json_record = clean_record(hepnames.do(create_record(snippet)))
 
+def test_incoming_deleted_record(app):
+    json_record = clean_record(hepnames.do(create_record(snippet)))
```
kaplun: Please rebase on top of latest master. @jacquerie has amended all calls to `clean_record`.

tests/integration/test_record.py
```diff
@@ -45,85 +45,50 @@ def test_incoming_deleted_record(app):
...
 
-    with app.test_client() as client:
-        response = client.get('/api/literature/1486323')
+    req_timeout = current_app.config['INDEXER_BULK_REQUEST_TIMEOUT']
+    es_bulk(
+        current_search_client,
+        index_queue,
+        stats_only=True,
+        request_timeout=req_timeout,
+    )
```
kaplun: This whole technique is using `es_bulk` for one record. In this case it might be overkill.
Just use `invenio_indexer.api.RecordIndexer().index(record)`

kaplun:
I see that the disambiguation is naturally failing due to the fact that you have commented out the receivers.

What is instead the error when the receivers are active? Can you re-enable them?

tests/integration/test_record.py
spirosdelviniotis: Requested changes done! :smile:

kaplun:
For the failing test:
```python
if _needs_beard_reprocessing(before_json['authors'], json['authors'])
```

is failing because it's checking for a key that could be missing. To fix it you can simply replace it:
```python
if _needs_beard_reprocessing(before_json.get('authors'), json('authors'))
```

and amend `_needs_beard_reprocessing` to check that if both are empty it should return `False` (no need to reprocess records that have no authors).

kaplun:
Can you rebase once more? @jacquerie has merged #1571 that should no longer trigger your failing test.

inspirehep/dojson/common/base.py
```diff
@@ -188,34 +188,51 @@ def collections(self, key, value):
...
+        return value and value.lower() == 'deleted'
+
     def get_value(value):
-        primary = force_single_element(value.get('a'))
         return {
-            'primary': primary,
-            'secondary': value.get('b'),
-            'deleted': value.get('c'),
+            'primary': force_single_element(value.get('a')),
+            'secondary': force_force_list(value.get('b')),
```
jacquerie: Unrelated, to be added to some preparatory commit.

inspirehep/dojson/common/base.py
```diff
@@ -188,34 +188,51 @@ def collections(self, key, value):
...
+
         collections.append(get_value(val))
 
     return collections
 
 
-@hep2marc.over('980', 'collections')
-@hepnames2marc.over('980', 'collections')
+@hep2marc.over('980', '^collections$')
+@hepnames2marc.over('980', '^collections$')
```
jacquerie: Unrelated, to be added to some preparatory commit.

inspirehep/dojson/common/base.py
```diff
@@ -188,34 +188,51 @@ def collections(self, key, value):
...
 @utils.for_each_value
 def collections2marc(self, key, value):
     """Collection this record belongs to."""
+
```
jacquerie: No reason for this extra space.

inspirehep/dojson/common/base.py
```diff
@@ -188,34 +188,51 @@ def collections(self, key, value):
...
         'a': value.get('primary'),
         'b': value.get('secondary'),
-        'c': value.get('deleted')
     }
 
 
+@hep2marc.over('980', '^deleted$')
+@hepnames2marc.over('980', '^deleted$')
+@utils.for_each_value
+@utils.ignore_value
```
jacquerie: Useless, remove this.

inspirehep/dojson/common/base.py
```diff
@@ -188,34 +188,51 @@ def collections(self, key, value):
...
+def deleted2marc(self, key, value):
+    """Deleted this record belongs to."""
+
```
jacquerie: Extra space.

inspirehep/dojson/common/base.py
```diff
@@ -188,34 +188,51 @@ def collections(self, key, value):
...
-        'c': value.get('deleted')
     }
 
 
+@hep2marc.over('980', '^deleted$')
+@hepnames2marc.over('980', '^deleted$')
+@utils.for_each_value
+@utils.ignore_value
+def deleted2marc(self, key, value):
+    """Deleted this record belongs to."""
```
jacquerie: This docstring doesn't mean a thing...

inspirehep/modules/migrator/tasks.py
```diff
@@ -300,6 +300,12 @@ def create_record(record):
 
 def record_upsert(json):
     """Insert or update a record."""
+
```
jacquerie: Extra space.

inspirehep/modules/migrator/tasks.py
```diff
@@ -300,6 +300,12 @@ def create_record(record):
...
+    def soft_delete_record():
+        if json.get('deleted'):
+            for pid in PersistentIdentifier.query.filter(PersistentIdentifier.object_uuid == record.id).all():
```
jacquerie: Why not `.one`? Can it really happen that multiple records have the same UUID?

inspirehep/modules/records/jsonschemas/records/authors.json
```diff
@@ -1,6 +1,9 @@
 {
     "$schema": "http://json-schema.org/schema#",
     "properties": {
+        "deleted": {
+            "type": "boolean"
+        },
```
jacquerie: Run `scripts/prettifyjson` before committing, this is not at the right place.

tests/integration/test_record.py
```diff
@@ -0,0 +1,83 @@
...
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+from inspirehep.modules.migrator.tasks import record_upsert, migrate_and_insert_record
+from invenio_indexer.api import RecordIndexer
+from invenio_db import db
+from inspirehep.dojson.hepnames import hepnames, hepnames2marc
+from dojson.contrib.marc21.utils import create_record
+import pytest
```
jacquerie: Import blocks should actually be:
```python

import pytest

from dojson.contrib.marc21.utils import create_record
from invenio_db import db
from invenio_indexer.api import RecordIndexer

from inspirehep.dojson.hepnames import hepnames, hepnames2marc
from inspirehep.modules.migrator.tasks import record_upsert, migrate_and_insert_record
```

First `__future__` imports, then stdlib (there isn't any here), then vendored from outside of Invenio, then Invenio, then INSPIRE. A space before each block.

tests/integration/test_record.py
```diff
@@ -0,0 +1,83 @@
...
+               '	</datafield>'
+               '</record>'
+               '</collection>')
+
+
+def test_incoming_deleted_record(app):
+    json_record = hepnames.do(create_record(snippet))
+    json_record['$schema'] = "http://localhost:5000/schemas/records/hep.json"
+
+    record = insert_record_json(json_record)
```
jacquerie: This is modifying the DB state permanently without cleanup.

tests/integration/test_record.py
```diff
@@ -0,0 +1,83 @@
...
+               '    <controlfield tag="001">1486322</controlfield>'
+               '	<controlfield tag="005">20160913214552.0</controlfield>'
+               '     <datafield tag="980" ind1=" " ind2=" ">'
+               '         <subfield code="c">DELETED</subfield>'
+               '     </datafield>'
+               '	<datafield tag="980" ind1=" " ind2=" ">'
+               '		<subfield code="a">HEP</subfield>'
+               '	</datafield>'
+               '</record>'
+               '</collection>')
```
jacquerie: Spacing is very off, moreover this should at least be inside of a fixture.

tests/integration/workflows/fixtures/record.xml
```diff
@@ -0,0 +1,12 @@
...
+        <controlfield tag="001">1486322</controlfield>
+        <controlfield tag="005">20160913214552.0</controlfield>
+        <datafield tag="980" ind1=" " ind2=" ">
+            <subfield code="c">DELETED</subfield>
+        </datafield>
+        <datafield tag="980" ind1=" " ind2=" ">
+            <subfield code="a">HEP</subfield>
+        </datafield>
+    </record>
+</collection>
```
jacquerie: Not used anywhere.

tests/integration/test_record.py
```diff
@@ -0,0 +1,83 @@
...
+        with db.session.begin_nested():
+            record = record_upsert(json_record)
+            if record:
+                r = RecordIndexer()
+                r.index(record)
+        db.session.commit()
+    finally:
+        db.session.close()
+
+    return record
```
jacquerie: This is application logic, I don't understand why this is in a test.

tests/unit/dojson/test_dojson_common_base.py
```diff
@@ -380,3 +380,48 @@ def test_urls_from_marcxml_multiple_8564():
     result = hep.do(create_record(snippet))
 
     assert expected == result['urls']
+
+
+def test_collections_deleted_in_980():
+    """Test collections deleted in 980 case.
+
+    """
```
jacquerie: Docstring means nothing.

tests/unit/dojson/test_dojson_common_base.py
```diff
@@ -274,7 +274,7 @@ def test_urls_from_marcxml_856_with_two_u_single_y():
         {
             'description': 'Conference web page',
             'value': 'http://www.physics.unlv.edu/',
-         },
+        },
```
jacquerie: Unrelated, to be split in a preparatory commit.

inspirehep/modules/migrator/tasks.py
kaplun: No, it will happen that the same UUID have many PIDs. (e.g. recid, DOI, arXiv... or recid, ORCID, KAKEN...)

spirosdelviniotis:
All the changes are committed. :smile:

kaplun:
@spirosdelviniotis have you applied also @jacquerie requests?

inspirehep/modules/migrator/tasks/records.py
```diff
@@ -315,6 +320,8 @@ def record_upsert(json):
             # Create persistent identifier.
             inspire_recid_minter(str(record.id), json)
 
+        soft_delete_record()
+
```
kaplun: Somehow this reads strange: i.e. that you would always call `soft_delete_record()` and inside it decides whether to delete it or not.

I'd suggest that you move the condition out of `soft_delete_record()` so that it's clear what will happen.

kaplun:
Just a mini comment and then LGTM.

jmartinm:
The receiver that I mention in https://github.com/inspirehep/inspire-next/pull/1541#pullrequestreview-1094168 can be added to [receivers.py](https://github.com/inspirehep/inspire-next/blob/2adf87da921a168d558c36422c244a9b116a0cd4/inspirehep/modules/records/receivers.py)

tests/integration/test_record.py
```diff
@@ -0,0 +1,153 @@
...
+
+def test_record_can_be_deleted_with_method_commit(app, record_not_yet_deleted):
+    with app.test_client() as client:
+        assert client.get('/api/literature/333').status_code == 200
+        record = get_db_record('literature', 333)
+        record['deleted'] = True
+        record.commit()
+        db.session.commit()
+
+        assert client.get('/api/literature/333').status_code == 410
```
jacquerie: Merge these two tests, they test the same thing.

spirosdelviniotis: Those two tests are merged now. :smile:

inspirehep/modules/records/receivers.py
```diff
@@ -334,3 +335,15 @@ def add_recids_and_validate(sender, json, *args, **kwargs):
     """Ensure that recids are generated before being validated."""
     populate_recid_from_ref(sender, json, *args, **kwargs)
     references_validator(sender, json, *args, **kwargs)
+
+
+@before_record_update.connect
+def check_if_record_is_going_to_be_deleted(sender, *args, **kwargs):
```
jacquerie: Signature should be `def check_if_record_is_going_to_be_deleted(sender, json, *args, **kwargs):`

It doesn't matter right now, but I can guarantee that people are going to start copying the wrong signature, and then it will matter.

kaplun:
Great @spirosdelviniotis so just apply @jacquerie micro-comment https://github.com/inspirehep/inspire-next/pull/1541#discussion_r80233277, rebase and we shall merge your first PR. :tada:

inspirehep/modules/records/receivers.py
kaplun: ? Are you sure? `before_record_update()` is called in `invenio_records.api` with:
```python
before_record_update.send(self)
```

so there is no json argument.

jacquerie: I was wrong 💃

PR #1545, opened 2016-09-06
travis: enable fast_finish

kaplun:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,6 +42,9 @@ env:
     - SUITE=workflows
     - SUITE=integration
 
+matrix:
+  fast_finish: true
+
 python:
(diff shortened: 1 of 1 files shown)


jacquerie:
Docs for this feature: https://docs.travis-ci.com/user/customizing-the-build#Fast-Finishing

Are we planning to have the Selenium tests allowed to fail in the future, or is this just to report failures on PEP8 / Unit tests as soon as possible?

kaplun:
No I think selenium tests should not fail as much as regression one. It's more to fail early (hopefully not due to PEP8) so that Travis is freed up quickly and that a developer receive as quick as possible a negative response, if necessary.

PR #1517, opened 2016-08-31
conferences: make series and number one field in conferences

tomaszgy:
Signed-off-by: Tomasz Gargas <tomaszgy@gmail.com>

PR #1547, opened 2016-09-07
dojson: series and number in one field

kaplun:
* Makes series and number back into one field in conferences, when
  they are split in multiple.
  (closes #1517)

### DIFF
--- a/tests/unit/dojson/test_dojson_conferences.py
+++ b/tests/unit/dojson/test_dojson_conferences.py
@@ -565,12 +565,14 @@ def test_series_from_411__a():
         '</datafield>'
     )  # record/1430017
 
-    expected = [
-        'DPF Series',
-    ]
     result = clean_record(conferences.do(create_record(snippet)))
 
-    assert expected == result['series']
+    expected = [
+        {
+            'name': 'DPF Series'
+        },
+    ]
+    assert result['series'] == expected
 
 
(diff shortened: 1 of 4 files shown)


tests/unit/dojson/test_dojson_conferences.py
```diff
@@ -580,26 +582,35 @@ def test_series_number_from_411__n():
...
-    expected = 7
     result = clean_record(conferences.do(create_record(snippet)))
 
-    assert expected == result['series_number']
+    expected = [
+        {
+            'number': 7
+        },
+    ]
+    assert result['series'] == expected
```
jacquerie: This is asserting that you're creating invalid records, according to the schema above. Definitely not a good thing, since now we have 3.4k more invalid records: https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/603955/.

kaplun: @jacquerie is right. @tomaszgy Can you please amend your rules to not generate a lonely `number` when `name` is not available?

tomaszgy: Please see https://github.com/inspirehep/inspire-next/pull/1553.

tomaszgy: BTW, well spotted @jacquerie :smile:

PR #1550, opened 2016-09-07
workflows/authors: holdingpen fixes

### LINKED ISSUE #1535: Holding pen: detailed view for author New form
Filling all the form in https://labs.inspirehep.net/author/new creates the following detailed page in the Holding Pen:

screen shot 2016-09-05 at 2 21 55 pm

On master branch, the detailed page in the Holding Pen looks like:

screen shot 2016-09-05 at 2 23 12 pm

* [ ] Display missing fields: Email, Experiment, ...
* [ ] Display user comment: I assume the idea is to have it in the `Notes` box.
* [ ] Add `Review submission` button which should be a link to `/submit/author/newreview?objectid=<workflow_id>`

### LINKED ISSUE #1537: Author New review form fixes
How to reproduce
=============

1) Fill the form http://localhost:5000/submit/author/create

2) Visit http://localhost:5000/holdingpen and click `Halted` in the first box 'Author New'

3) There should be one entry, click on it and check the `id` in the URL (like `/holdingpen/<id>`)

4) Now access the review form (there should be a direct link on the detailed page that is missing at the moment - see https://github.com/inspirehep/inspire-next/issues/1535) so access directly the URL `http://localhost:5000/submit/author/newreview?objectid=1`

There is an error in the Jinja2 template `review_form.html`

```
BuildError: Could not build url for endpoint 'inspirehep_authors.holdingpenreview' with values ['approved', 'objectid']. Did you mean 'inspirehep_authors_holdingpen.newreview' instead?
```

After fixing that, check that the form renders with the same content that you filled in 1) and there should be some buttons at the end of the form like:

screen shot 2016-09-05 at 2 54 20 pm

Note
-----

The view code for the review page is [here](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/authors/views/holdingpen.py#L306) and the handler for the review form is [here](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/authors/views/holdingpen.py#L328)

Panos512:
* Displays missing fields on holdingpen detailed view
  (notes, experimens, native name, email address).
* Adds "Review Submission" button on holdingpen detailed view.
* "Review Submission" url now properly works.
* "Review Submission" form now populates with updated values.

closes #1535
closes #1537

### DIFF
--- a/inspirehep/modules/authors/forms.py
+++ b/inspirehep/modules/authors/forms.py
@@ -23,6 +23,7 @@
 from flask_babelex import gettext as _
 
 from wtforms import validators
+from wtforms.fields import Flags
 from wtforms.widgets import html_params, \
     HiddenInput, \
     HTMLString, \
@@ -274,13 +275,28 @@ class AuthorUpdateForm(INSPIREForm):
 
     """Author update form."""
 
-    # Hidden field to hold author id information
     bai = fields.StringField(
-        widget=HiddenInput()
+        label=_('Bai'),
+        description=u'e.g. M.Santos.1',
+        widget=HiddenInput(),
+        widget_classes="form-control",
+        validators=[validators.Optional(),
+                    RegexpStopValidator(
+                        "(\\w+\\.)+\\d+",
+                        message="A valid Bai is in the form of 'M.Santos.1'.",
+        )]
     )
 
     inspireid = fields.StringField(
-        widget=HiddenInput()
+        label=_('Inspireid'),
+        description=u'e.g. INSPIRE-0000000',
+        widget=HiddenInput(),
+        widget_classes="form-control",
+        validators=[validators.Optional(),
+                    RegexpStopValidator(
+                        "INSPIRE-\\d{8}",
+                        message="A valid Inspireid is in the form of 'INSPIRE-0000000'.",
+        )]
     )
 
     # Hidden field to hold record id information
@@ -519,6 +535,6 @@ class AuthorUpdateForm(INSPIREForm):
             self.orcid.validators = []
         if is_review:
             self.bai.widget = TextInput()
-            self.bai.widget_classes = "form-control"
+            self.bai.flags = Flags()
             self.inspireid.widget = TextInput()
-            self.inspireid.widget_classes = "form-control"
+            self.inspireid.flags = Flags()
--- a/inspirehep/modules/authors/views/holdingpen.py
+++ b/inspirehep/modules/authors/views/holdingpen.py
@@ -43,9 +43,12 @@ from flask_babelex import gettext as _
 from flask_breadcrumbs import register_breadcrumb
 from flask_login import login_required, current_user
 from werkzeug.datastructures import MultiDict
+from wtforms.widgets import HiddenInput
 
 from invenio_db import db
+
 from invenio_workflows import workflow_object_class, start, resume
+from invenio_workflows_ui.api import WorkflowUIRecord
 
 from inspirehep.dojson.utils import strip_empty_values
 from inspirehep.modules.forms.form import DataExporter
@@ -82,7 +85,7 @@ def convert_for_form(data):
     if "urls" in data:
         data["websites"] = []
         for url in data["urls"]:
-            if "description" not in url:
+            if not url.get('description'):
                 data["websites"].append({"webpage": url["value"]})
             else:
                 if url["description"].lower() == "twitter":
@@ -195,6 +198,7 @@ def new():
         data["bai"] = bai
 
     form = AuthorUpdateForm(data=data)
+
     ctx = {
         "action": url_for('.submitnew'),
         "name": "authorUpdateForm",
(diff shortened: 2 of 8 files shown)


inspirehep/modules/authors/views/holdingpen.py
```diff
@@ -308,10 +310,19 @@ def newreview():
     objectid = request.values.get('objectid', 0, type=int)
     if not objectid:
         abort(400)
-    workflow_object = workflow_object_class.get(objectid)
+    workflow_object = WorkflowUIRecord.get_record(objectid)['metadata']
```
jmartinm: `workflow_metadata` would be a better name.

inspirehep/modules/authors/views/holdingpen.py
```diff
@@ -308,10 +310,19 @@ def newreview():
...
+
+    # Converting json to populate form
+    convert_for_form(workflow_object)
+    workflow_object['comments'] = workflow_object['_private_note']
+    research_fields = workflow_object.pop('research_field')
+    final_research_fields = []
+    for field in research_fields:
+        final_research_fields.append(field['_term'])
+    workflow_object['research_field'] = final_research_fields
```
jmartinm:
```python
```

inspirehep/modules/workflows/static/js/inspire_workflows_ui/templates/details_decision_authors.html
```diff
@@ -1,4 +1,7 @@
 <div ng-if="vm.record._extra_data._action === 'author_approval' && vm.record._workflow.status === 'HALTED'">
+<button class="btn btn-info"
+        ng-click="Utils.redirectToReviewSubmission(vm.record.id)">Review Submission
```
jmartinm: I am wondering if we should have a `Utils.redirect` function and in this template that is specific to the author forms you could do `ng-click="Utils.redirect('/submit/author/newreview?objectid={{vm.record.id}})"`

jmartinm:
The review form is missing the bai and inspireid fields.

This is a screenshot from current labs:

screen shot 2016-09-09 at 10 14 50 am

Those fields should appear only on the review form (based on this code - https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/authors/forms.py#L517-L524)

inspirehep/modules/workflows/static/js/inspire_workflows_ui/templates/details_decision_authors.html
Panos512: It sounds like a better idea than having dedicated functions for every redirect. But where is the place to put that? (maybe somewhere in `modules/theme/static/js` ?)

jmartinm: In principle I was thinking of leaving it in the same place you have it now, just a more generic `Utils.redirect` instead of `Utils.redirectToReviewSubmission`

Panos512:
TODO:
 
- [ ]  `bai` and `inspireid` are also visible on submmition form

inspirehep/modules/authors/forms.py
```diff
@@ -274,13 +274,26 @@ class AuthorUpdateForm(INSPIREForm):
 
     """Author update form."""
 
-    # Hidden field to hold author id information
     bai = fields.StringField(
-        widget=HiddenInput()
```
jmartinm: Oh, these fields have to be hidden for the users. They should only be visible in the review form.

The following code https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/authors/forms.py#L517-L524 was doing the trick, can you check if it works?

jmartinm: I see you are doing a similar thing here https://github.com/inspirehep/inspire-next/pull/1550/files#diff-c8175080bad0020db2b4eea02db63e7cR201 if it is equivalent we could delete the other code.

Panos512: I'm doing it here https://github.com/Panos512/inspire-next/blob/a48fabae60ee52fa0a9da56af9b2fa2f3d143c09/inspirehep/modules/authors/views/holdingpen.py#L201 isn't this enough?

jmartinm: It might be enough, but as before it was being done here - https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/authors/forms.py#L517-L524

You should check if that code then it is not used any more and we should delete it. Or probably even better you could use the previous existing code and when creating the class `AuthorUpdateForm` set the flag `is_update` and `is_review` that should do the hiding of the fields.

jmartinm: Is there any benefit on changing this? Removing the `HiddenInput ` and then adding it in the `__init__` VS how it was before, hidden by default and then added.

jmartinm:
Can you rebase and will merge? Thanks.

PR #1554, opened 2016-09-09
forms: 1532 fix autocomplete authors

rikirenz:


jmartinm:
The PR contains 3 commits at the moment. You should get the latest `master` branch and apply your commit (one for the moment fixing the Institutions autocomplete) on top.

Then you can `push -f` to your fork `1532-fix-autocomplete-authors` and this PR should be correct.

jmartinm:
For the commit message first line, apart from the component name, we also write a short summary (max 50 chars). So instead of:

```
forms:
```

Something like:

```
forms: institutions autocomplete fix in author forms
```

rikirenz:
I am not sure about the last changes in the authors_typeahead.js file. I was not able to find the field status; I do not know if you have removed it or you have renamed it. Anyway the control for the institutions is still present.

inspirehep/modules/forms/static/js/forms/affiliations_typeahead.js
```diff
@@ -42,15 +42,10 @@ define([
             pattern = pattern + "affautocomplete:" + "/" + this + ".*/";
           })
 
-          return '/search?cc=Institutions&p=' + pattern + '&of=recjson&rg=100'
+          return '/api/institutions?q=' + pattern
```
kaplun: I believe pattern should be properly URL encoded.

inspirehep/modules/forms/static/js/forms/authors_typeahead.js
```diff
@@ -41,14 +41,10 @@ define([
             pattern = pattern + "authorautocomplete:" + "/" + this + ".*/";
           })
 
-          return '/search?cc=HepNames&p=' + pattern + '&of=recjson&rg=100'
+          return '/api/authors?q=' + pattern
```
kaplun: As above.

inspirehep/modules/forms/static/js/forms/experiments_typeahead.js
```diff
@@ -43,15 +43,10 @@ define([
             }
             pattern = pattern + "experimentautocomplete:" + "/" + this + ".*/";
           })
-
-          return '/search?cc=Experiments&p=' + pattern + '&of=recjson&rg=100'
+          return '/api/experiments?q=' + pattern
```
kaplun: Ditto

kaplun:
Be careful: `forms:institutions` is missing a space :dancer:

tests/acceptance/conftest.py
```diff
@@ -36,10 +36,13 @@
 from inspirehep.config import SERVER_NAME
 from inspirehep.factory import create_app
 
+import os
```
jacquerie: You already have `environ` imported above.

tests/acceptance/conftest.py
```diff
@@ -60,8 +63,7 @@ def small_app(request):
         init_all_storage_paths()
         init_users_and_permissions()
 
-        migrate('./inspirehep/demosite/data/demo-records-small.xml',
-                wait_for_results=True)
+        #migrate('./inspirehep/demosite/data/demo-records.xml.gz', wait_for_results=True)
```
jacquerie: Leftover.

tests/acceptance/test_author_create_typehead.py
```diff
@@ -0,0 +1,62 @@
+
```
jacquerie: Extra line here.

tests/acceptance/conftest.py
```diff
@@ -36,10 +36,13 @@
...
 from inspirehep.factory import create_app
 
+import os
 
-@pytest.fixture(scope='module')
+@pytest.fixture(scope='session')
 def small_app(request):
     """Flask application fixture."""
+
+
```
jacquerie: Extra lines here.

tests/acceptance/conftest.py
```diff
@@ -73,3 +75,15 @@ def selenium(selenium, small_app):
     selenium.maximize_window()
     selenium.get(environ['SERVER_NAME'])
     return selenium
+
+
+@pytest.fixture()
+def login(selenium):
```
jacquerie: Not sure if this is the best name, since this is doing the login (and logout) _as an admin_, and I imagine we will need at least another fixture for an unprivileged user. But YAGNI, so for now it can stay like this.

tests/acceptance/conftest.py
```diff
@@ -16,7 +16,7 @@
 # You should have received a copy of the GNU General Public License
 # along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
 #
-# In applying this licence, CERN does not waive the privileges and immunities
+# In applying this license, CERN does not waive the privileges and immunities
```
jacquerie: This is just british spelling vs. american spelling... It can stay like it was.

tests/acceptance/test_author_create_typehead.py
```diff
@@ -0,0 +1,62 @@
...
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+
+from __future__ import absolute_import, division, print_function
+from time import sleep
+
+from selenium.webdriver.common.keys import Keys
+
+import os
```
jacquerie: Imports should be:
```python

import os
from time import sleep

from selenium.webdriver.common.keys import Keys
```
the idea: first `__future__`, then stdlib, then vendored dependencies, then Invenio dependencies (not present here), then INSPIRE dependencies (not present here).

tests/acceptance/test_author_create_typehead.py
```diff
@@ -0,0 +1,62 @@
...
+
+def test_institutions_typehead(selenium, login):
+    """"Test the suggestions in the istitutions field"""
+    selenium.get(os.environ['SERVER_NAME'] + '/submit/author/create')
+    institution_field = selenium.find_element_by_id("institution_history-0-name")
+    institution_field.send_keys("CER")
+    sleep(2)
+    institution_field.send_keys(Keys.DOWN)
+    institution_field.send_keys(Keys.ENTER)
```
jacquerie: I know what these three line are because you explained me, but it could be obscure for someone else. Maybe make it clear by abstracting in a method, something like `force_autocomplete_event(field)`?

tests/acceptance/test_login.py
```diff
@@ -25,7 +25,7 @@
 import os
 
 
-def test_login(selenium, small_app):
+def test_login(selenium):
```
jacquerie: This change is correct, but probably belongs to a preparatory commit.

tests/acceptance/test_login.py
```diff
@@ -35,4 +35,4 @@ def test_login(selenium, small_app):
     email.send_keys('admin@inspirehep.net')
     password = selenium.find_element_by_id('password')
     password.send_keys('123456')
-
+    selenium.find_element_by_xpath("//button[@type='submit']").click()
```
jacquerie: Is this really related to this PR?

tests/acceptance/test_author_create_typehead.py
```diff
@@ -0,0 +1,62 @@
...
+from __future__ import absolute_import, division, print_function
+from time import sleep
+
+from selenium.webdriver.common.keys import Keys
+
+import os
+
+def test_institutions_typehead(selenium, login):
+    """"Test the suggestions in the istitutions field"""
+    selenium.get(os.environ['SERVER_NAME'] + '/submit/author/create')
```
jacquerie: As I anticipated in the previous PR this will probably benefit from an helper. Not 100% necessary that you add it in this PR.

tests/acceptance/test_login.py
kaplun: Could be committed separately but at least is adding the final bit that was missing to mine. Maybe would also be nice to test that the login was successful

rikirenz: Ok. I can check the label in the profile_status

tests/acceptance/conftest.py
rikirenz: Could you explain me what does it means "leftover" in this context?

tests/acceptance/test_author_create_typehead.py
rikirenz: good to know!

tests/acceptance/conftest.py
jacquerie: Sure. It means: "I think you made this modification as a temporary measure while you were exploring the problem, and now you left it inside of the commit. I think you should remove it."

tests/acceptance/test_author_create_typehead.py
```diff
@@ -31,30 +31,31 @@
...
 
 def test_experiments_typehead(selenium, login):
     """"Test the suggestions in the experiments field"""
     selenium.get(os.environ['SERVER_NAME'] + '/submit/author/create')
-    experiment_field = selenium.find_element_by_id("experiments-0-name")
-    force_autocomplete_event(experiment_field, "ATL")
-    assert 'ATLAS' in experiment_field.get_attribute("value")
+    field = selenium.find_element_by_id("experiments-0-name")
+    field.send_keys("ATL")
+    sleep(5)
```
kaplun: Have you already tried using explicit waits here? Seems to be the recommended way: http://stackoverflow.com/questions/30986604/is-it-best-practice-to-use-thread-sleep-or-explicit-wait-before-click-on-any-e

rikirenz:
I have pushed the selenium way for the wait. I think the tests will fail. Let me know if you think is wrong...
@kaplun @jacquerie

rikirenz:
As you can see from the errors it seems that typeahead is no able to recognised the event of click on the element.  Let me know, if for you is ok merge the previous commit or if you think is better wait for  another (more) elegant solution.

tests/acceptance/test_author_create_typehead.py
```diff
@@ -0,0 +1,66 @@
...
+    selenium.get(os.environ['SERVER_NAME'] + '/submit/author/create')
+    advisor_field = selenium.find_element_by_id("advisors-0-name")
+    force_autocomplete_event(advisor_field, "alexe", selenium)
+    assert 'Vorobyev, Alexey' in advisor_field.get_attribute("value")
+
+
+def force_autocomplete_event(field, chunk_text, selenium):
+    """Performs selection in a html text field"""
+    field.send_keys(chunk_text)
+    selenium.execute_script('$("#header-section").hide()')
```
jmartinm: I guess this what needed, what was the problem?

tests/acceptance/test_author_create_typehead.py
```diff
@@ -0,0 +1,66 @@
...
+    selenium.execute_script('$("#collections-section").hide()')
+    WebDriverWait(selenium, 10).until(EC.presence_of_element_located((By.CSS_SELECTOR, "div.tt-suggestion"))).click()
```
jmartinm: Each of the suggestions from typeahead seems to be in a `div` with class `tt-suggestion`.

Does this work or you need to get the first element on the list and trigger click ?

rikirenz:
@jmartinm 
Tips:

1.  If you wanna run the test locally you have to do as follows:

- `honcho start`

- open a new console and run `sh scripts/recreate_records`

- run the command that you can find in the docker-compose.test.yml in the acceptance section: `py.test --driver Remote --host localhost --port 4444 --capability browserName firefox --html=selenium-report.html tests/acceptance` (pay attention because you have to change selenium -> localhost)

2.  Actually I do not have idea how to reduce the loading time. The main idea was reduce the size of the file that we have in order to reproduce the test in a fast way with the data that are necessary. We can create a `demo-records-for-selenium-small.xml` with enough data to test all the web UI functionalities.

3. For the documentation there is no problem. I'll do it

kaplun:
> run the command that you can find in the docker-compose.test.yml in the acceptance section: py.test --driver Remote --host localhost --port 4444 --capability browserName firefox --html=selenium-report.html tests/acceptance (pay attention because you have to change selenium -> localhost)

Isn't this connecting to a remote Firefox? Personally, to run it locally, I have downloaded https://github.com/mozilla/geckodriver/releases and made it available in the path and then:

`py.test --driver Firefox --html=selenium-report.html tests/acceptance`

tests/acceptance/test_author_create_typehead.py
rikirenz: I have created test cases that are able to provide only one suggestion element. In fact I have tried in debug mode (with `pdb`) to select  all the elements available with the class `tt-suggestion`, and only one was present in the DOM. Anyway I will check again because I am not totally sure about it.

jacquerie:
You definitively understood the point about having commits that don't contain extraneous stuff, but now you got to the opposite extreme: commits that contain too little, and do not constitute a feature by themselves.

I'm going to integrate this with a better commit structure, so that you can have a look.

PR #1555, opened 2016-09-09
tests: remove deprecated yield_fixture uses

david-caro:
* Since py.test>2.10 the yield_fixture construct is deprecated, and
  the request.addfinalizer mechanism was replaced by using just yield
  instead of return in the fixture, see:
      http://doc.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code


jmartinm:
LGTM.

PR #1565, opened 2016-09-14
dojson: add post functions

jacquerie:
* First commit fixes some licenses, some imports, and some docstrings. Nothing exciting.
* ~~Second commit removes `current_marcxml_usage`, because its utility has been largely reduced since we have nightly. If you are still using it please let me know, so that I can remove this commit!~~
* ~~Third~~ Second commit renames `schema.py` to `model.py` for uniformity.
* ~~Fourth~~ Third commit is the meat of the PR: it replaces `SchemaOverdo` with `FilterOverdo`, which is a specialization of `Overdo` that allows a list of post functions (filters) to run after the DoJSON conversion finished. This means we no longer need to add a `clean_record` in every test, and also fixes a small bug where we ran `add_book_info` **after** `strip_empty_values` ran. Note that we can use closures to pass arguments to filters/post functions, as I did in `add_schema`.
* ~~Fifth~~ Fourth commit is a minor cleanup of a leftover use of `strip_empty_values`.


jacquerie:
Ready for review.

kaplun:
Mmh... `current_marcxml_usage` should maybe not stay in inspire-next, but should still exist. nightly doesn't tell you which MARC fields we have not ported, and which outliers we have not mapped. So it's definitively still valuable.

inspirehep/dojson/conferences/model.py
```diff
@@ -24,7 +24,7 @@
 
 from __future__ import absolute_import, division, print_function
 
-from ..schema import SchemaOverdo
+from ..model import SchemaOverdo
```
kaplun: Mmh `models` is used for SQLAlchemy.  But I guess there is no clash. In the end this is a model? Or more a mapping. Whatever name :dancer:

inspirehep/dojson/model.py
```diff
@@ -20,22 +20,41 @@
...
+
+def add_schema(schema):
+    def _add_schema(record, blob):
+        record['$schema'] = schema
+        return record
+
+    return _add_schema
+
+
+def clean_record(record, blob):
```
kaplun: I think this should be moved outside of dojosn, because we'll need it in general, everytime some record is inserted/updated even if from native JSON.

I'd register it as a `@before_record_update/insert`.

jacquerie: This won't work for the tests though, since no record is updated or inserted.

kaplun: Yep. I fear the only solution is to do it twice: once in migrator and once with signal. Maybe in migrator we can detach the signal, since we know it's already taking care in dojson.

kaplun: So, if you have not implemented it in your PR, at least add an issue not to forget that we need to clean incoming JSON records.

PR #1566, opened 2016-09-14
theme: fix invalid HTML in header

liamkirsh:
Removed a duplicate HTML attribute (ignored by the browser) and added a space betwen the URL attributes.

Signed-off-by: Liam Kirsh <liam.kirsh@cern.ch>


jacquerie:
The W3 validator is unhappy about some other things, and we could think add tests that check the validity of the HTML of this and other pages. I believe this is low priority right now, but thanks for this!

kaplun:
Ha! Memories... https://github.com/inveniosoftware/invenio/commit/38c1b948c51d56311701923f6ea0992c080fcc88

PR #1567, opened 2016-09-14
hal: add support for adding and updating HAL records

liamkirsh:


inspirehep/modules/hal/hal_api.py
```diff
@@ -0,0 +1,101 @@
...
+# or submit itself to any jurisdiction.
+
+from tempfile import TemporaryFile
+import zipfile
+
+from sword2 import Connection
+from sword2 import http_layer
+import httplib2
+
+from settings import HAL_USER, HAL_PASSWORD
```
kaplun: This should be called `config.py`, not `settings.py`.

inspirehep/modules/hal/hal_api.py
```diff
@@ -0,0 +1,101 @@
...
+
+COL_IRI = "https://api-preprod.archives-ouvertes.fr/sword/hal"
+PREPROD = "https://api-preprod.archives-ouvertes.fr/sword/"
```
kaplun: These 2 magic files should also be configurable and put in `config.py`. Actually you can prefill them with production values if in production, and pre-prod when the instance is in development mode.

inspirehep/modules/hal/hal_api.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def add_record(tei, document=None):
+    c = Connection("", user_name=HAL_USER,
+                   user_pass=HAL_PASSWORD,
+                   http_impl=HttpLib2Layer_IgnoreCert(".cache"))
```
kaplun: Ignoring the certificate should be configurable, since hopefully they will fix the issue one day :smile:

inspirehep/modules/hal/hal_api.py
```diff
@@ -0,0 +1,101 @@
...
+from sword2 import http_layer
+import httplib2
+
+from settings import HAL_USER, HAL_PASSWORD
+
+COL_IRI = "https://api-preprod.archives-ouvertes.fr/sword/hal"
+PREPROD = "https://api-preprod.archives-ouvertes.fr/sword/"
+
+
+def add_record(tei, document=None):
```
kaplun: What is `document`?

inspirehep/modules/hal/hal_api.py
```diff
@@ -0,0 +1,101 @@
...
+
+    if document:
+        tempFile = TemporaryFile()
+        with zipfile.ZipFile(tempFile,
+                             mode='w',
+                             compression=zipfile.ZIP_DEFLATED) as zf:
+            zf.writestr('meta.xml', tei)
+            zf.writestr('doc.pdf', document)
+        tempFile.seek(0)
+        payload = tempFile.read()
```
kaplun: This has the problem that you are reading the whole ZIP back in memory, so you are not gaining having stored it to disk.

Is there a way to pass the payload with an iterator?

kaplun: Answering myself: yes you can: https://github.com/swordapp/python-client-sword2/blob/master/sword2/connection.py#L727

inspirehep/modules/hal/hal_api.py
```diff
@@ -0,0 +1,101 @@
...
+        mimetype = "application/zip"
+        filename = "meta.xml"
+    else:
+        payload = tei
+        mimetype = "text/xml"
+        filename = "meta.xml"
```
kaplun: This is identical to the above. Looks like the only difference is `edit_(media)_iri` Vs. `col_iri`.

Please either write a single function, or move the common part into a helper function.

inspirehep/modules/hal/hal_api.py
```diff
@@ -0,0 +1,101 @@
...
+                       mimetype=mimetype,
+                       filename=filename,
+                       packaging="http://purl.org/net/sword-types/AOfr",
+                       in_progress=False)
+    return receipt
+
+
+class HttpLib2Layer_IgnoreCert(http_layer.HttpLib2Layer):
+    def __init__(self, cache_dir):
+        self.h = httplib2.Http(cache_dir,
```
kaplun: Do you actually have to specify the `cache_dir`? Probably yes...

inspirehep/modules/hal/hal_api.py
```diff
@@ -0,0 +1,101 @@
...
+                               timeout=30.0,
+                               ca_certs=None,
+                               disable_ssl_certificate_validation=True)
```
kaplun: As mentioned above, please make this flag configurable in `config.py`.

inspirehep/modules/hal/models.py
```diff
@@ -0,0 +1,34 @@
...
+from invenio_db import db
+
+
+class InspireHalRecords(db.Model):
+    __tablename__ = 'inspire_hal_records'
+
+    inspire_id = db.Column(db.String(160), primary_key=True)
+    hal_id = db.Column(db.String(160), primary_key=True)
+    version = db.Column(db.Integer)
+    date = db.Column(db.DateTime)
```
kaplun: Better called timestamp since it should contain the full time (up to the second)

inspirehep/modules/hal/models.py
```diff
@@ -0,0 +1,34 @@
...
+from __future__ import absolute_import, division, print_function
+
+from invenio_db import db
+
+
+class InspireHalRecords(db.Model):
+    __tablename__ = 'inspire_hal_records'
+
+    inspire_id = db.Column(db.String(160), primary_key=True)
+    hal_id = db.Column(db.String(160), primary_key=True)
```
kaplun: I don't think you need to have the `hal_id` to be `primary_key`. It's enough for it to be unique.

Actually now that I am thinking about it. The most clean thing would be to store here only the `hal_id`, and let the `pidstore` store the mapping between `inspire_id` and `hal_id`. But maybe it's for a future commit. I am not sure you have time to dive into pidstore :)

inspirehep/modules/hal/tasks.py
```diff
@@ -0,0 +1,69 @@
...
+from celery.utils.log import get_task_logger
+logger = get_task_logger(__name__)
+
+DATE_FORMAT = "%a, %d %b %Y %X %Z"
+
+
+def doc_type_should_be_sent_to_hal(record):
+    """If the record has at least one institution with a halID, return True."""
+    # FIXME: Figure out how to check if institutions are in HAL
+    return False
```
kaplun: Actually the criteria will be simpler. We shall introduce a flag such as `for_hal` in the schema, that the cataloguers will set to True in order to declare that a record has to be sent.

inspirehep/modules/hal/tei.py
```diff
@@ -32,9 +32,10 @@
 from inspirehep.utils.record_getter import get_es_records
 
 from .templates import template
+from .typology import hal_doctype, doctype_fallback
 
 
-def tei_response(record):
+def tei_response(record, has_attachment=False):
```
kaplun: What about calling it `json2tei`? or `convert_record_to_tei`. `tei_response` seems like the name of a variable containing the response from a service called `tei` :-)

inspirehep/modules/hal/tasks.py
```diff
@@ -0,0 +1,69 @@
...
+    """Sends records to hal."""
+    if True:#doc_type_should_be_sent_to_hal(sender):
+        logger.info("Sending to hal: ", inspire_id)
+        try:
+            hal_tei = tei_response(sender, attachment)
+            receipt = hal_api.add_record(hal_tei, attachment)
+            hal_log_record = InspireHalRecords(
+                inspire_id = inspire_id,
+                hal_id = receipt.id,
+                version = 1,
```
kaplun: So here you assume it's the first time we are sending the record to HAL, right? I assume the rest of the code will come later.

inspirehep/modules/hal/tasks.py
```diff
@@ -0,0 +1,69 @@
...
+def send_to_hal(sender, inspire_id, attachment=None):
+    """Sends records to hal."""
+    if True:#doc_type_should_be_sent_to_hal(sender):
+        logger.info("Sending to hal: ", inspire_id)
+        try:
+            hal_tei = tei_response(sender, attachment)
+            receipt = hal_api.add_record(hal_tei, attachment)
+            hal_log_record = InspireHalRecords(
+                inspire_id = inspire_id,
+                hal_id = receipt.id,
```
kaplun: here the `hal_id` will contain `v1`?

inspirehep/modules/hal/tasks.py
```diff
@@ -0,0 +1,69 @@
...
+                version = 1,
+                date = datetime.strptime(receipt.response_headers['date'], DATE_FORMAT)
+            )
+            with db.session.begin_nested():
+                db.session.add(hal_log_record)
+            db.session.commit()
+            logger.info("Successfully sent to hal: ",
+                        receipt.id)
+        except Exception as e:
+            print e.response.text, id
```
kaplun: Leftover

inspirehep/modules/hal/tasks.py
```diff
@@ -0,0 +1,69 @@
...
+            logger.error("Failed to push to hal: ", id)
```
kaplun: You might want to log also the exception. @jacquerie what is the recommended way with the new pattern for Sentry?

inspirehep/modules/hal/typology.py
```diff
@@ -0,0 +1,44 @@
...
+    'bookchapter': "COUV",
+    # Chapitre d'ouvrage / Book chapter
+    # 'review': "NOTE",
+    # Note de lecture / Book review
+    'published': "ART",
+    # Article dans une revue / Journal article
+    'lectures': "LECTURE",
+    # Cours / Course
+}
+doctype_fallback = "OTHER"  # Fallback: Autre publication / Other publication
```
kaplun: I think these magic values and constants should also be merged into `config.py`.

inspirehep/modules/hal/typology.py
```diff
@@ -0,0 +1,44 @@
...
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+hal_doctype = {
```
kaplun: better call it `INSPIRE_HAL_DOCTYPE_MAP`.

Note, it's a python convention to always type in upper case global variables.

inspirehep/modules/hal/templates/hal_tei.xml
```diff
@@ -109,6 +117,7 @@
               {% endfor %}#}
               <classCode scheme="halTypology" n="{{ typology }}"/>
             </textClass>
+            <abstract xml:lang="en">Test</abstract>
```
kaplun: Left over

inspirehep/modules/hal/tei.py
```diff
@@ -201,8 +183,8 @@ def _journal_data(pub_info):
 
 def _structure_data(struct):
     return {
-        'type': get_value(struct, "collections[1].primary", "").lower(),
-        # ^^ FIXME: This may not be one of the HAL accepted values:
+        'type': "institution",
+        # ^^ FIXME: Needs to choose from the HAL accepted values:
         # institution, department, laboratory or researchteam
```
kaplun: Here we miss to add the HAL ID of the institution, that is supposedly stored at the referenced recid.

inspirehep/modules/hal/hal_api.py
liamkirsh: I think it's read into memory anyway by the sword module, but I'll make the fix on our side. https://github.com/swordapp/python-client-sword2/blob/db1660d6a48fa1f6c69b97b3ece7916fdafc555c/sword2/http_layer.py#L71

inspirehep/modules/hal/hal_api.py
liamkirsh: Document is the PDF.

inspirehep/modules/hal/tasks.py
liamkirsh: No, the receipt.id field returned by Connection.create doesn't include the version.

inspirehep/modules/hal/tasks.py
jacquerie: There's [this](http://stackoverflow.com/a/19414701/374865) hack around it. Maybe @mihaibivol has some spare time to look at this?

inspirehep/modules/hal/hal_api.py
kaplun: Please, add then corresponding docstring to describe the function.

inspirehep/modules/hal/hal_api.py
kaplun: Seriously!? What are they waiting for to fix it.
OK we should send them a PR :+1:

kaplun: Mmh. The project seems dead. Maybe worth fork it and patch it directly locally.

kaplun: Can you amend `requirements.txt` to refer directly to our new forked repo: https://github.com/inspirehep/python-client-sword2

inspirehep/modules/hal/hal_api.py
liamkirsh: Yes, although it looks like the parameter is ignored when using http_layet.HttpLib2Layer in the sword2 module: https://github.com/swordapp/python-client-sword2/pull/6

Do you think it would make sense to hardcode in ".cache" in this class/move it to the config.py and remove cache_dir as a parameter?

inspirehep/modules/hal/tasks.py
liamkirsh: No, from my testing the receipt returned by `Connection.add_record` includes a location without a version.

inspirehep/modules/hal/hal_api.py
kaplun: Makes me really think the project is dead. Can you send a PR to our fork, and make INSPIRE to use our fork?

inspirehep/modules/hal/tei.py
liamkirsh: Should I add that as a FIXME?

kaplun: Yeah. Remember to collect all the FIXMEs in a github issue (e.g. with checkboxes :dancer: )

jacquerie:
Superseded by #1785.

PR #1569, opened 2016-09-15
relations: integrate with Inspire

tomaszgy:
Signed-off-by: Tomasz Gargas tomaszgy@gmail.com


inspirehep/config.py
```diff
@@ -1345,3 +1345,11 @@ def _(x):
...
 except ImportError:
     pass
+
+
+# Relations
+RELATIONS_NEO4J_USER = 'CHANGE_ME'
+RELATIONS_NEO4J_PASSWORD = 'CHANGE_ME'
+RELATIONS_NEO4J_HOST = os.environ.get(
+    "RELATIONS_NEO4J_HOST",
+    "localhost:7687")
```
jmartinm: This should go in `inspirehep/modules/relations/config.py`

inspirehep/modules/relations/builders/conferences/builder.py
```diff
@@ -0,0 +1,45 @@
...
+from inspirehep.modules.relations.model_updaters import (
+    set_property_of_central_node,
+    set_uid
+    )
+
+
+conferences = GraphModelBuilder(central_node_type=ConferenceNode)
+
+@conferences.element_processor('control_number')
+def recid(model, element):
```
jmartinm: As this is repeated for all collections, would it make sense to add it to the `common/builder.py` ?

inspirehep/modules/relations/command_producers.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+def produce_labels_string(labels):
+    return ''.join([':' + label for label in labels if label])
+
+
+def produce_properties_string(properties):
+    if properties:
+        properties_strings = []
+        for key, value in properties.items():
```
jmartinm: You could save one line with:

```python
for key, str(value) in properties.items():
```

inspirehep/modules/relations/command_producers.py
```diff
@@ -0,0 +1,218 @@
...
+            properties_strings.append(properties_string)
+
+            return '{{{}}}'.format(','.join(properties_strings))
+    else:
+        return ''
+
+def produce_element_block(labels=None, properties=None, variable_name=None):
+    if not labels: labels = []
+    if not properties: properties = {}
+    if not variable_name: variable_name = ''
```
jmartinm: `labels` and `properties` might be fine since ` produce_element_block(labels=[],..)` is bad idea in Python. But for `variable_name` you could just add a default value of empty string (and avoid this `if`)

inspirehep/modules/relations/migrate.py
```diff
@@ -0,0 +1,258 @@
...
+    for key in keys[:-1]:
+        try:
+            current_dict = current_dict.setdefault(key, dict())
+        except AttributeError:
+            raise TypeError(
+                'Wrong internal structure. ' + '.'.join(
+                    keys[:keys.index(key)]
+                    ) + ' is already something else than dict.'
+                )
+    return current_dict.setdefault(keys[-1], default)
```
jmartinm: Why is the `default` only used in `keys[-1]` ?

tomaszgy: If you have:
```python
keys = ['first_level', 'second_level', 'third']
```

and you call:
```python
nested_setdefault(my_dict, keys)
```

then you want `nested_setdefualt` to do basically this:

```python
my_dict['first_level']['second_level'].setdefault('third', default)
```
That's why `my_dict`, `my_dict['first_level']` and `my_dict['second_level']` need to be dictionaries.

I'll add documentation before I leave, sorry it's not here yet ;)

tomaszgy:
All right, guys, PR ready to review 😄  Please note that:

1. It assumes that Neo4j instance is already running which is a false assumption. @jmartinm, @david-caro and @jacquerie - may I ask you to add docker configuration for Neo4j (together with overwriting Neo's conf file and - probably - changing Neo's password)? Thanks in advance!
2. There are some minor things that for some reasons I wasn't able to solve. I wrote them down in https://github.com/inspirehep/inspire-next/issues/1652 so that they might be solved by someone else in the future.

Cheers!

kaplun:
@tomaszgy thanks a lot for having contributed this even beyond your staying at CERN!

jmartinm:
https://github.com/inspirehep/inspire-next/pull/1588 adds docker-compose configuration for neo4j

jacquerie:

PR #1570, opened 2016-09-15
utils: add get_record_type function

tomaszgy:

### DIFF
--- a/inspirehep/utils/record.py
+++ b/inspirehep/utils/record.py
@@ -24,6 +24,8 @@
 
 import re
 
+from inspirehep.modules.pidstore.providers import InspireRecordIdProvider
+
 SPLIT_KEY_PATTERN = re.compile('\.|\[')
 
 
@@ -105,6 +107,19 @@ def get_value(record, key, default=None):
         except KeyError:
             return default
     return value
+
+
+def get_record_type(record):
+    """Get the record's ``pid_type``.
+
+    Return the record's ``pid_type`` (e.g. literature or jobs)
+    from the record's ``$schema``.
+    """
+    schema = record.get('$schema')
+    if schema:
+        return InspireRecordIdProvider.schema_to_pid_type(schema)
+    else:
+        raise TypeError('The schema of this record does not exist.')
 
 
 def is_submitted_but_not_published(record):
--- a/tests/unit/utils/test_utils_record.py
+++ b/tests/unit/utils/test_utils_record.py
@@ -21,7 +21,12 @@
 
 from __future__ import absolute_import, print_function
 
-from inspirehep.utils.record import get_abstract, get_subtitle, get_title, get_value
+from inspirehep.utils.record import (
+    get_abstract,
+    get_record_type,
+    get_subtitle,
+    get_title,
+    get_value,)
 
 from invenio_records.api import Record
 
@@ -294,3 +299,29 @@ def test_get_value_returns_none_on_index_error():
     })
 
     assert get_value(single_title, 'titles.title[1]') is None
+
+
+def test_get_record_type_returns_type_for_existing_schema():
+    record = Record({
+        '$schema': 'http://localhost:5000/schemas/records/jobs.json'
+    })
+
+    assert get_record_type(record) == 'jobs'
+
+
+@pytest.mark.xfail(reason="No information available \
+                   about existing and non-existing schemas.")
+def test_get_record_type_raises_error_for_nonexisting_schema():
+    record = Record({
+        '$schema': 'does-not-exist.json'
+    })
+
+    with pytest.raises(TypeError):
+        get_record_type(record) is None
+
+
+def test_get_record_type_raises_error_when_no_type():
+    record = Record({})
+
+    with pytest.raises(TypeError):
+        get_record_type(record) is None


tests/unit/utils/test_utils_record.py
```diff
@@ -294,3 +300,9 @@ def test_get_value_returns_none_on_index_error():
     })
 
     assert get_value(single_title, 'titles.title[1]') is None
+
+
+def test_get_record_type_returns_none_when_no_type():
+    record = Record({})
+
+    assert get_record_type(record) is None
```
jacquerie: Missing tests for when the record has a valid (existing) schema and an invalid (non existing) schema.

tests/unit/utils/test_utils_record.py
```diff
@@ -21,7 +21,13 @@
 
 from __future__ import absolute_import, print_function
 
-from inspirehep.utils.record import get_abstract, get_subtitle, get_title, get_value
+from inspirehep.utils.record import (
+    get_abstract,
+    get_record_type,
+    get_subtitle,
+    get_title,
+    get_value
```
jacquerie: Missing comma.

tests/unit/utils/test_utils_record.py
```diff
@@ -21,7 +21,13 @@
...
+    )
```
jacquerie: Extra whitespace.

inspirehep/utils/record.py
```diff
@@ -107,6 +109,15 @@ def getitem(k, v, default):
     return value
 
 
+def get_record_type(record):
+    """Get record's type e.g. 'literature', 'jobs' etc."""
+    schema = record.get('$schema')
+    if schema:
+        return InspireRecordIdProvider.schema_to_pid_type(schema)
+    else:
+        return None
```
jacquerie:
```
else:
    return None
```
is not needed, because it will be returned anyway by the control flow reaching the end of the function.

Yet, is it really a sensible return value? I can already imagine the errors.

inspirehep/utils/record.py
```diff
@@ -22,6 +22,8 @@
 
 """Helpers for handling records."""
 
+from inspirehep.modules.pidstore.providers import InspireRecordIdProvider
```
jacquerie: Wrong order, must go after the stdlib imports.

inspirehep/utils/record.py
```diff
@@ -107,6 +109,15 @@ def getitem(k, v, default):
     return value
 
 
+def get_record_type(record):
+    """Get record's type e.g. 'literature', 'jobs' etc."""
```
jacquerie:
```
"""Get the record's ``pid_type``.

Return the record's ``pid_type`` (e.g. literature or jobs)
from the record's ``$schema``.
"""

File: tests/unit/utils/test_utils_record.py (hunk 1 above)
Comment (tomaszgy) [inline tests/unit/utils/test_utils_record.py]: How can I know that the schema does not exist? Just by looking at `inspirehep/modules/records/jsonschemas/records` directory? Or is the list of schemas which we're using defined somewhere else?

Comment (jacquerie) [inline tests/unit/utils/test_utils_record.py]: No need to overcomplicate this: just put a `$schema` value of `'does-not-exist.json'` and document what your implementation is doing in this case.

File: inspirehep/utils/record.py (hunk 4 above)
Comment (tomaszgy) [inline inspirehep/utils/record.py]: Fair point, `None`'s not the best option. What's the alternative though - raising a custom exception? Empty string?

File: tests/unit/utils/test_utils_record.py (hunk 1 above)
Comment (tomaszgy) [inline tests/unit/utils/test_utils_record.py]: Sure, I can do that and the test will fail because the implementation of the function (to be more exact - implementation of the `InspireRecordPidProvider.schema_to_pid_type()` on which it depends) has no knowledge about existing and non-existing schemas - it's a simple url parsing.

It migh have not been clear in my comment - I'd just like to solve the problem which is going to make not-written-yet test fail :smile: .

See [`InspireRecordPidProvider.schema_to_pid_type()` definition](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/pidstore/providers.py#L87).

File: inspirehep/utils/record.py (hunk 4 above)
Comment (jacquerie) [inline inspirehep/utils/record.py]: Exception is probably the best, because it needs to be handled, so the caller knows that they should be careful.

File: tests/unit/utils/test_utils_record.py (hunk 1 above)
Comment (jacquerie) [inline tests/unit/utils/test_utils_record.py]: One more reason to document that with a test! How could one know that, without looking at that code in depth? For the implementation see: http://doc.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

Comment (tomaszgy) [inline tests/unit/utils/test_utils_record.py]: Will document it, sure thing.

Also, I can leave the test with xfail but I think it'll be better to have the function repaired. In order to do that I need a mapping `schemas - pid types` or at least predefined list of schemas (`config.py`?). Any idea where to get it from?

File: tests/unit/utils/test_utils_record.py [hunk 7]
```diff
@@ -294,3 +299,29 @@ def test_get_value_returns_none_on_index_error():
...
+
+@pytest.mark.xfail(reason="No information available \
+                   about existing and non-existing schemas.")
+def test_get_record_type_raises_error_for_nonexisting_schema():
+    record = Record({
+        '$schema': 'does-not-exist.json'
+    })
+
+    with pytest.raises(TypeError):
+        get_record_type(record) is None
```
jmartinm: The record has `$schema` on it. Why does `TypeError` get raised?

jacquerie:
Superseded by #1612

PR #1571, opened 2016-09-16
disambiguation: detach receivers

jacquerie:
* First commit retouches some imports in the `disambiguation` module.
* Second commit detaches the receivers from the signals. The rationale is in the commit message:

    > On record migration the `append_new_record_to_queue` receiver fires,
  which adds the record to the `DisambiguationRecord` queue. Then, when
  the same record is indexed, `append_updated_record_to_queue` fires,
  which will try adding the same record to the queue. In doing so, it
  will shoot back a query at ES to try to get a previous version of the
  record, which will probably return a 404 because this is the first
  time we're inserting the record. This generates undue pressure on ES
  and achieves nothing, therefore this PR disable these signals.

### DIFF
--- a/inspirehep/modules/disambiguation/__init__.py
+++ b/inspirehep/modules/disambiguation/__init__.py
@@ -22,13 +22,10 @@
 
 """Inspire wrapper for Beard author disambiguation service."""
 
-from __future__ import (
-    absolute_import,
-    division,
-    print_function,
-)
+from __future__ import absolute_import, division, print_function
 
 from .ext import InspireDisambiguation
 from .tasks import disambiguation_daemon
+
 
 __all__ = ('InspireDisambiguation', 'disambiguation_daemon')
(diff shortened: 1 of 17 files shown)


kaplun:
Can we update not to query ES, in case of record insert?

jacquerie:
> Can we update not to query ES, in case of record insert?

The logic would still be wrong, because it would check twice if the record should be added to the queue.

kaplun:
I don't see which filenames you have fixed. Can you detail?

jacquerie:
> I don't see which filenames you have fixed. Can you detail?
```
% git diff HEAD~1 | ag rename
rename from tests/unit/disambiguation/test_logic.py
rename to tests/unit/disambiguation/test_disambiguation_logic.py
rename from tests/unit/disambiguation/test_receivers.py
rename to tests/unit/disambiguation/test_disambiguation_receivers.py
```
for the well known problem in which all test names must be different otherwise py.test crashes during collection.

Issue #1577, opened 2016-09-19
Authorization


kaplun:
* [ ] Ensure restricted records are not accessible #1288
* [ ] Ensure restricted collections are not accessible
* [ ] Ensure holding pen is accessible to cataloguer only
* [ ] Ensure admin interface is accessible to superadmins only
* [ ] Ensure POST/PUT/DELETE APIs are not available
* [ ] Ensure deleted records are not accessible #1541

kaplun:
@jmartinm can you help coordinating this part? You are aware of status of collections/records. @spirosdelviniotis is taking care of deleted records in #1541

jmartinm:
Closed by https://github.com/inspirehep/inspire-next/pull/1589

We just need to make sure to assign the correct roles and actions when we deploy Invenio 3.

PR #1589, opened 2016-09-23
global: add record/collection permissions

### LINKED ISSUE #919: Make use of invenio-collections ?
Previously we were making use of https://github.com/inveniosoftware/invenio-collections that was connecting to the `before_record_update` signal to always add a `_collections` field to a record.

This field was used, for example, to know if a certain user had permission to see a collection.

The status of `invenio-collections` needs to be evaluated and decided if we make use of it.

A concrete task to test is to make hidden collections not show in a general HEP search. With our current demo records 427 results are obtained but only 426 should be shown as one is part of the hidden collection `CDF Internal Notes`.

See [this test](https://github.com/inspirehep/inspire-next/commit/dcff4ded8184e64c1e1aed5398280694fa0f87eb#diff-1200a5bdb3e2ab7c11ec659179d2474aR54)

jmartinm:

Closes #919

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -125,6 +125,114 @@ scripts
 USERPROFILES_EXTEND_SECURITY_FORMS = False
 USERPROFILES_SETTINGS_TEMPLATE = 'inspirehep_theme/accounts/settings/profile.html'
 
+# Collections
+# ===========
+COLLECTIONS_DELETED_RECORDS = '{dbquery} AND NOT deleted:True'
+"""Enhance collection query to exclude deleted records."""
+
+COLLECTIONS_QUERY_PARSER = 'inspirehep.modules.search.parser:Main'
+"""User search query lexical parser."""
+
+COLLECTIONS_QUERY_WALKERS = [
+    'inspirehep.modules.search.walkers.pypeg_to_ast:PypegConverter',
+]
+"""Modules to create the query AST."""
+COLLECTIONS_USE_PERCOLATOR = False
+"""Define which percolator you want to use.
+
+Default value is `False` to use the internal percolator.
+You can also set True to use elasticsearch to provide percolator resolver.
+NOTE that ES percolator uses high memory and there might be some problems
+when creating records.
+"""
+
+INSPIRE_COLLECTIONS_DEFINITION = [
+    {
+        "query": "collections.primary:HEP",
+        "name": "Literature"
+    },
+    {
+        "query": "collections.primary:H1-INTERNAL-NOTE",
+        "name": "H1 Internal Notes"
+    },
+    {
+        "query": "collections.primary:INSTITUTION",
+        "name": "Institutions"
+    },
+    {
+        "query": "collections.primary:HERMES-INTERNAL-NOTE",
+        "name": "HERMES Internal Notes"
+    },
+    {
+        "query": "collections.primary:ZEUS-INTERNAL-NOTE",
+        "name": "ZEUS Internal Notes"
+    },
+    {
+        "query": "collections.primary:JOB",
+        "name": "Jobs"
+    },
+    {
+        "query": "collections.primary:JOBHIDDEN",
+        "name": "Jobs Hidden"
+    },
+    {
+        "query": "collections.primary:CONFERENCES",
+        "name": "Conferences"
+    },
+    {
+        "query": "collections.primary:D0-INTERNAL-NOTE",
+        "name": "D0 Internal Notes"
+    },
+    {
+        "query": "collections.primary:EXPERIMENT",
+        "name": "Experiments"
+    },
+    {
+        "query": "collections.primary:DATA",
+        "name": "Data"
+    },
+    {
+        "query": "collections.primary:ZEUS-PRELIMINARY-NOTE",
+        "name": "ZEUS Preliminary Notes"
+    },
+    {
+        "query": "collections.primary:JOURNALS",
+        "name": "Journals"
+    },
+    {
+        "query": "collections.primary:H1-PRELIMINARY-NOTE",
+        "name": "H1 Preliminary Notes"
+    },
+    {
+        "query": "collections.primary:D0-PRELIMINARY-NOTE",
+        "name": "D0 Preliminary Notes"
+    },
+    {
+        "query": "595__c:CDS collections.primary:HEP",
+        "name": "For CDS"
+    },
+    {
+        "query": "collections.primary:CDF-INTERNAL-NOTE",
+        "name": "CDF Internal Notes"
+    },
+    {
+        "query": "collections.primary:HIDDEN",
+        "name": "HIDDEN"
+    },
+    {
+        "query": "collections.primary:CDF-NOTE",
+        "name": "CDF Notes"
+    },
+    {
+        "query": "collections.primary:HEPNAMES",
+        "name": "Authors"
+    },
+    {
+        "query": "collections.primary:HALhidden",
+        "name": "HAL Hidden"
+    }
+]
+
 # Search
 # ======
 
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -137,11 +137,12 @@ COLLECTIONS_QUERY_WALKERS = [
     'inspirehep.modules.search.walkers.pypeg_to_ast:PypegConverter',
 ]
 """Modules to create the query AST."""
+
 COLLECTIONS_USE_PERCOLATOR = False
 """Define which percolator you want to use.
 
 Default value is `False` to use the internal percolator.
-You can also set True to use elasticsearch to provide percolator resolver.
+You can also set True to use ElasticSearch to provide percolator resolver.
 NOTE that ES percolator uses high memory and there might be some problems
 when creating records.
 """
@@ -704,14 +705,14 @@ RECORDS_UI_ENDPOINTS = dict(
         route='/literature/<pid_value>',
         template='inspirehep_theme/format/record/'
                  'Inspire_Default_HTML_detailed.tpl',
-        record_class='inspirehep.modules.records.wrappers:LiteratureRecord'
+        record_class='inspirehep.modules.records.wrappers:LiteratureRecord',
     ),
     authors=dict(
         pid_type='authors',
         route='/authors/<pid_value>',
         template='inspirehep_theme/format/record/'
                  'authors/Author_HTML_detailed.html',
-        record_class='inspirehep.modules.records.wrappers:AuthorsRecord'
+        record_class='inspirehep.modules.records.wrappers:AuthorsRecord',
     ),
     data=dict(
         pid_type='data',
@@ -722,31 +723,31 @@ RECORDS_UI_ENDPOINTS = dict(
         pid_type='conferences',
         route='/conferences/<pid_value>',
         template='inspirehep_theme/format/record/Conference_HTML_detailed.tpl',
-        record_class='inspirehep.modules.records.wrappers:ConferencesRecord'
+        record_class='inspirehep.modules.records.wrappers:ConferencesRecord',
     ),
     jobs=dict(
         pid_type='jobs',
         route='/jobs/<pid_value>',
         template='inspirehep_theme/format/record/Job_HTML_detailed.tpl',
-        record_class='inspirehep.modules.records.wrappers:JobsRecord'
+        record_class='inspirehep.modules.records.wrappers:JobsRecord',
     ),
     institutions=dict(
         pid_type='institutions',
         route='/institutions/<pid_value>',
         template='inspirehep_theme/format/record/Institution_HTML_detailed.tpl',
-        record_class='inspirehep.modules.records.wrappers:InstitutionsRecord'
+        record_class='inspirehep.modules.records.wrappers:InstitutionsRecord',
     ),
     experiments=dict(
         pid_type='experiments',
         route='/experiments/<pid_value>',
         template='inspirehep_theme/format/record/Experiment_HTML_detailed.tpl',
-        record_class='inspirehep.modules.records.wrappers:ExperimentsRecord'
+        record_class='inspirehep.modules.records.wrappers:ExperimentsRecord',
     ),
     journals=dict(
         pid_type='journals',
         route='/journals/<pid_value>',
         template='inspirehep_theme/format/record/Journal_HTML_detailed.tpl',
-        record_class='inspirehep.modules.records.wrappers:JournalsRecord'
+        record_class='inspirehep.modules.records.wrappers:JournalsRecord',
     )
 )
 
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -415,6 +415,7 @@ RECORDS_REST_ENDPOINTS = dict(
         default_media_type='application/json',
         max_result_window=10000,
         search_factory_imp='inspirehep.modules.search.query:inspire_search_factory',
+        read_permission_factory_imp="inspirehep.modules.records.permissions:record_read_permission_factory",
         record_class='inspirehep.modules.records.es_record:ESRecord'
     ),
     literature_db=dict(
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -693,63 +693,59 @@ RECORDS_REST_ENDPOINTS = dict(
     ),
 )
 
+
+RECORDS_UI_DEFAULT_PERMISSION_FACTORY = \
+    "inspirehep.modules.records.permissions:record_read_permission_factory"
+
 RECORDS_UI_ENDPOINTS = dict(
     literature=dict(
         pid_type='literature',
         route='/literature/<pid_value>',
         template='inspirehep_theme/format/record/'
                  'Inspire_Default_HTML_detailed.tpl',
-        record_class='inspirehep.modules.records.wrappers:LiteratureRecord',
-        permission_factory_imp='invenio_records_rest.utils:allow_all',
+        record_class='inspirehep.modules.records.wrappers:LiteratureRecord'
     ),
     authors=dict(
         pid_type='authors',
         route='/authors/<pid_value>',
         template='inspirehep_theme/format/record/'
                  'authors/Author_HTML_detailed.html',
-        record_class='inspirehep.modules.records.wrappers:AuthorsRecord',
-        permission_factory_imp='invenio_records_rest.utils:allow_all',
+        record_class='inspirehep.modules.records.wrappers:AuthorsRecord'
     ),
     data=dict(
         pid_type='data',
         route='/data/<pid_value>',
-        template='inspirehep_theme/format/record/Data_HTML_detailed.tpl',
-        permission_factory_imp='invenio_records_rest.utils:allow_all',
+        template='inspirehep_theme/format/record/Data_HTML_detailed.tpl'
     ),
     conferences=dict(
         pid_type='conferences',
         route='/conferences/<pid_value>',
         template='inspirehep_theme/format/record/Conference_HTML_detailed.tpl',
-        record_class='inspirehep.modules.records.wrappers:ConferencesRecord',
-        permission_factory_imp='invenio_records_rest.utils:allow_all',
+        record_class='inspirehep.modules.records.wrappers:ConferencesRecord'
     ),
     jobs=dict(
         pid_type='jobs',
         route='/jobs/<pid_value>',
         template='inspirehep_theme/format/record/Job_HTML_detailed.tpl',
-        record_class='inspirehep.modules.records.wrappers:JobsRecord',
-        permission_factory_imp='invenio_records_rest.utils:allow_all',
+        record_class='inspirehep.modules.records.wrappers:JobsRecord'
     ),
     institutions=dict(
         pid_type='institutions',
         route='/institutions/<pid_value>',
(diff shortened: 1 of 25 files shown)


jmartinm:
Addresses https://github.com/inspirehep/inspire-next/issues/1577

One problem I see at the moment is that when I enable invenio-collections I am getting ES timeouts locally when migrating records (possibly due to some extra computation happening in `invenio-collections` receivers)

kaplun:
I see from https://github.com/inveniosoftware/invenio-collections/blob/master/invenio_collections/config.py#L24 that by adjusting `COLLECTIONS_DELETED_RECORDS` we can directly have ES to exclude Deleted records, so then @spirosdelviniotis don't have to explicitly delete records in ES.

jmartinm:
Another side effect I just realised. As `invenio-collections` listens to `before_record_insert` signal when creating a new record in the HoldingPen, the receiver runs (which is probably ok) but it is causing an error since in the HoldingPen instead of passing around a `Record` object we are passing a `dict`

```
:15:38 worker.1    | [2016-09-23 13:15:38,553: ERROR/MainProcess] Task invenio_workflows.tasks.start[e8913b38-46fb-487b-a088-4bbf83b73c3d] raised unexpected: AttributeError("'dict' object has no attribute 'dumps'",)
13:15:38 worker.1    | Traceback (most recent call last):
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
13:15:38 worker.1    |     R = retval = fun(*args, **kwargs)
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/lib/python2.7/site-packages/flask_celeryext/app.py", line 43, in __call__
13:15:38 worker.1    |     return Task.__call__(self, *args, **kwargs)
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/lib/python2.7/site-packages/celery/app/trace.py", line 438, in __protected_call__
13:15:38 worker.1    |     return self.run(*args, **kwargs)
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/src/invenio-workflows/invenio_workflows/tasks.py", line 77, in start
13:15:38 worker.1    |     return text_type(run_worker(workflow_name, data, **kwargs).uuid)
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/src/invenio-workflows/invenio_workflows/worker_engine.py", line 53, in run_worker
13:15:38 worker.1    |     engine.process(objects, **kwargs)
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/lib/python2.7/site-packages/workflow/engine.py", line 390, in process
13:15:38 worker.1    |     self._process(objects)
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/lib/python2.7/site-packages/workflow/engine.py", line 547, in _process
13:15:38 worker.1    |     obj, self, callbacks, exc_info
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/src/invenio-workflows/invenio_workflows/engine.py", line 363, in Exception
13:15:38 worker.1    |     obj, eng, callbacks, exc_info
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/lib/python2.7/site-packages/workflow/engine.py", line 970, in Exception
13:15:38 worker.1    |     reraise(*exc_info)
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/lib/python2.7/site-packages/workflow/engine.py", line 529, in _process
13:15:38 worker.1    |     self.run_callbacks(callbacks, objects, obj)
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/lib/python2.7/site-packages/workflow/engine.py", line 481, in run_callbacks
13:15:38 worker.1    |     self.execute_callback(callback_func, obj)
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/lib/python2.7/site-packages/workflow/engine.py", line 564, in execute_callback
13:15:38 worker.1    |     callback(obj, self)
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/src/inspire/inspirehep/modules/workflows/tasks/actions.py", line 175, in emit_record_signals
13:15:38 worker.1    |     before_record_insert.send(obj.data)
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/lib/python2.7/site-packages/blinker/base.py", line 267, in send
13:15:38 worker.1    |     for receiver in self.receivers_for(sender)]
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/src/invenio-collections/invenio_collections/receivers.py", line 112, in __call__
13:15:38 worker.1    |     matcher=self.matcher)
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/src/invenio-collections/invenio_collections/receivers.py", line 88, in get_record_collections
13:15:38 worker.1    |     for collections in matcher(collections, record):
13:15:38 worker.1    |   File "/Users/jmartinm/.virtualenvs/invenio3/src/invenio-collections/invenio_collections/percolator.py", line 99, in _find_matching_collections_externally
13:15:38 worker.1    |     body = {"doc": record.dumps()}
13:15:38 worker.1    | AttributeError: 'dict' object has no attribute 'dumps'
```

kaplun:
As discussed IRL we can drop `emit_record_signals` since all the use-cases are already taken care during real record insertion/indexing.

jmartinm:
> One problem I see at the moment is that when I enable invenio-collections I am getting ES timeouts locally when migrating records (possibly due to some extra computation happening in invenio-collections receivers)

This problem is now solved by not using percolators (`COLLECTIONS_USE_PERCOLATOR=False`) and instead using 'internal matching'.

jmartinm:
Second commit fixes https://github.com/inspirehep/inspire-next/pull/1589#issuecomment-249166778

jmartinm:
TODO:

* [ ] Fix landing page query. Currently shows the number of records in the index (and once searching we only search with `Literature` collection) so numbers don't match.
* [ ] Restrict collections. Collections that are currently restricted should still be restricted.

inspirehep/modules/fixtures/users.py
```diff
@@ -46,5 +46,19 @@ def init_users_and_permissions():
...
             role=superuser_role
         ))
-        db.session.add(ActionRoles(action='admin-access', role=superuser_role))
+        db.session.add(ActionRoles(
+            action='admin-access',
+            role=superuser_role)
+        )
+
+        # Create a user that has access to restricted collection
+        restricted_user = ds.create_user(
```
jmartinm: This is just for testing at the moment, will add this user and permission to a test instead.

inspirehep/modules/search/api.py
```diff
@@ -77,12 +83,34 @@ def mget(self, uuids, **kwargs):
         return [document['_source'] for document in documents['docs']]
 
 
+def inspire_filter():
+    """Filter applied to all queries."""
+    if request:
+        collection = request.values.get('cc', 'Literature')
+
+        all_restricted_collections = set(
```
jmartinm: This would be better cached to avoid doing a DB query on every request.

jmartinm:
Ready for comments. Now that Travis is passing, will add some extra tests regarding permissions/collections.

jacquerie:
Last commit (`tests: test_record app fix`) should go first, otherwise you are breaking `git bisect`.

inspirehep/config.py
```diff
@@ -135,6 +135,114 @@ def _(x):
...
+
+Default value is `False` to use the internal percolator.
+You can also set True to use elasticsearch to provide percolator resolver.
+NOTE that ES percolator uses high memory and there might be some problems
+when creating records.
+"""
+
+COLLECTIONS_DEFINITION = [
+    {
+        "dbquery": "collections.primary:HEP",
```
kaplun: Since this config is under our control, you could write here `query` instead of `dbquery`, so that its sexier.

inspirehep/config.py
```diff
@@ -135,6 +135,114 @@ def _(x):
...
+COLLECTIONS_USE_PERCOLATOR = False
+"""Define which percolator you want to use.
+
+Default value is `False` to use the internal percolator.
+You can also set True to use elasticsearch to provide percolator resolver.
+NOTE that ES percolator uses high memory and there might be some problems
+when creating records.
+"""
+
+COLLECTIONS_DEFINITION = [
```
kaplun: INSPIRE specific configuration variable should be prefixed by `INSPIRE_`, (ala `ZENODO_`).

inspirehep/modules/records/permissions.py
```diff
@@ -0,0 +1,105 @@
...
+    user_roles = [r.name for r in current_user.roles]
+    if 'superuser' in user_roles:
+        return True
+
+    all_restricted_collections = set(
+        [
+            a.argument for a in ActionUsers.query.filter_by(
+                action='view-restricted-collection').all()
+        ]
+    )
```
kaplun: Shouldn't this be computed only once, for performance reason. We could cache it in Redis for a minute or so.

setup.py
```diff
@@ -184,6 +185,11 @@
             'hepnames = inspirehep.dojson.hepnames:hepnames',
             'hepnames2marc = inspirehep.dojson.hepnames2marc:hepnames2marc',
         ],
+        'invenio_access.actions': [
+            'view_restricted_collection'
+            ' = inspirehep.modules.records.permissions:'
+            'action_view_restricted_collection',
+        ],
```
kaplun: This seems extreme indentation :dancer:

tests/integration/test_record.py
```diff
@@ -132,4 +133,5 @@ def test_record_can_be_deleted(app, record_not_yet_deleted):
             ri.index(record)
         db.session.commit()
 
+    with app.test_client() as client:
         assert client.get('/api/literature/333').status_code == 410
```
kaplun: Do you actually have any test for collection integration?

jmartinm: Not yet, will be added today :)

jmartinm:
The last commit adds first tests related to permissions - https://github.com/inspirehep/inspire-next/pull/1589/commits/bd9e5e571a189f6c8fb112625cd12cff13327ff0

The search filter has no tests yet, will add those next.

kaplun:
What about:

> * [ ] Fix landing page query. Currently shows the number of records in the index (and once searching we only search with Literature collection) so numbers don't match.

Is this solved by this PR as well?

jmartinm:
> Is this solved by this PR as well?

Good point, not yet ;)

jmartinm:
This PR is now ready to be reviewed. The functionality added is covered by tests.

inspirehep/config.py
```diff
@@ -125,6 +125,114 @@ def _(x):
 USERPROFILES_EXTEND_SECURITY_FORMS = False
 USERPROFILES_SETTINGS_TEMPLATE = 'inspirehep_theme/accounts/settings/profile.html'
 
+# Collections
+# ===========
+COLLECTIONS_DELETED_RECORDS = '{dbquery} AND NOT collections.primary:"DELETED"'
```
kaplun: Currently, after @spirosdelviniotis work the  `DELETED` information corresponds to a flag being set to True in `deleted`.

I am not sure how to formulate this filter in Invenio query syntax though. (i.e. `deleted != True`)

inspirehep/modules/records/mappings/records/journals.json
```diff
@@ -204,6 +204,10 @@
                     ],
                     "tokenizer": "standard",
                     "type": "custom"
+                },
+                "lowercase_analyzer": {
+                    "filter": "lowercase",
+                    "tokenizer": "keyword"
```
kaplun: Question: since you ignore case in matching collection in ES, are you anyway then presenting them in the HTML and URLs in their canonical form or you are reusing the case that was provided as input by the user?

inspirehep/modules/records/permissions.py
```diff
@@ -0,0 +1,148 @@
...
+        restricted_collections = set(
+            [
+                a.argument for a in ActionUsers.query.filter_by(
+                    action='view-restricted-collection').all()
+            ]
+        )
+        if restricted_collections:
+            cache.set(
+                'restricted_collections',
+                pickle.dumps(restricted_collections),
```
kaplun: Please use `protocol ` `-1` to have the most efficient protocol being chosen.

inspirehep/modules/search/api.py
```diff
@@ -77,12 +89,32 @@ def mget(self, uuids, **kwargs):
...
+        user_roles = [r.name for r in current_user.roles]
+        if 'superuser' in user_roles:
+            user_coll = all_restricted_collections
+        else:
+            user_coll = user_collections
+
+        query = Q('match', _collections=collection)
+
+        for collection in list(all_restricted_collections - user_coll):
+            query = query & ~Q('match', _collections=collection)
```
kaplun: Have you tried using actually ES filtering?

inspirehep/modules/records/permissions.py
```diff
@@ -0,0 +1,105 @@
...
+        record_collections = set(record['_collections'])
+        restricted_coll = all_restricted_collections & record_collections
+        if restricted_coll:
+            for collection in restricted_coll:
+                if not DynamicPermission(
+                        ParameterizedActionNeed(
+                            "view-restricted-collection",
+                            collection)
+                ).can():
+                    return False
```
jacquerie: I think something like
```python
def _cant_view(collection):
    return not DynamicPermission(
        ParametrizedActionNeed(
            'view-restricted-collection', collection)).can()

return any(map(_cant_view, restricted_coll))
```
reads nicer, but you're free to disagree : )

tests/integration/test_permissions.py
```diff
@@ -0,0 +1,311 @@
...
+    pid = PersistentIdentifier.get('literature', '123')
+    db.session.delete(pid)
+    record.delete(force=True)
+    es.delete(index='records-hep', doc_type='hep', id=pid.object_uuid)
+    db.session.commit()
+
+
+@pytest.fixture(scope='function')
+def restricted_record(app):
+    # Create a restricted collection
```
jacquerie: Nit: I think the rest of the code is hammering enough the fact that this is a restricted collection... : )

jmartinm:
After https://github.com/inspirehep/inspire-next/pull/1622 is merged I will change the way I read/write into Redis as Flask-cache allows for writing `sets` into Redis (I guess with some internal magic)

inspirehep/modules/records/permissions.py
```diff
@@ -0,0 +1,148 @@
...
```
jacquerie: I feel something like this reads better:
```python
def _cant_view(collection):
    return not DynamicPermission(
        ParametrizedActionNeed(
            'view-restricted-collection', collection)).can()

return any(map(_cant_view, restricted_coll))
```
It's ok if you disagree : )

tests/integration/test_permissions.py
```diff
@@ -0,0 +1,337 @@
...
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+
+from __future__ import absolute_import, division, print_function
+
+import pickle
```
jacquerie: Nit: space here, because `pickle` is stdlib.

tests/integration/test_permissions.py
```diff
@@ -0,0 +1,337 @@
...
+import pytest
+
```
jacquerie: Nit: ...and no space here, because `pytest` and `flask_security` are both vendored...

tests/integration/test_permissions.py
```diff
@@ -0,0 +1,337 @@
...
+from invenio_db import db
+from invenio_indexer.api import RecordIndexer
+from invenio_pidstore.models import PersistentIdentifier
+from invenio_records.api import Record
+from invenio_search import current_search_client as es
+
+from inspirehep.modules.pidstore.minters import inspire_recid_minter
+from inspirehep.modules.search.api import LiteratureSearch
+from inspirehep.utils.cache import cache
+
```
jacquerie: Nit: ...one extra space here.

tests/integration/test_permissions.py
```diff
@@ -0,0 +1,337 @@
...
+])
+def test_record_public_detailed_read(app, sample_record, users, user_info, status):
+    """Test that a record can be read by everybody through web interface."""
+    with app.test_client() as client:
+        if user_info:
+            # Login as user
+            login_user_via_session(client, email=user_info['email'])
+
+        result = client.get("/literature/123")
+        assert result.status_code == status
```
jacquerie: Since you are not going to use `result` further IMO `assert client.get('/literature/123').status_code == status` is enough.

tests/integration/test_permissions.py
```diff
@@ -0,0 +1,337 @@
...
+    (dict(email='admin@inspirehep.net'), 200),
+])
+def test_record_public_api_read(app, api_client, sample_record, users, user_info, status):
+    """Test that a record can be read by everybody through the API."""
+    if user_info:
+        # Login as user
+        login_user_via_session(api_client, email=user_info['email'])
+
+    result = api_client.get("/literature/123")
+    assert result.status_code == status
```
jacquerie: Same thing here.

tests/integration/test_permissions.py
```diff
@@ -0,0 +1,337 @@
...
+def test_record_restricted_detailed_read(app, restricted_record, users, user_info, status):
+    """Test permissions of restricted record accessing detailed view."""
+
+    with app.test_client() as client:
+        if user_info:
+            # Login as user
+            login_user_via_session(client, email=user_info['email'])
+
+        result = client.get("/literature/222")
+        assert result.status_code == status
```
jacquerie: Here.

tests/integration/test_permissions.py
```diff
@@ -0,0 +1,337 @@
...
+    # admin user
+    (dict(email='admin@inspirehep.net'), 200),
+])
+def test_record_restricted_api_read(app, api_client, restricted_record, users, user_info, status):
+    """Test permissions of a restricted record through the API."""
+
+    if user_info:
+        login_user_via_session(api_client, email=user_info['email'])
+    result = api_client.get("/literature/222")
+    assert result.status_code == status
```
jacquerie: And here.

inspirehep/config.py
jmartinm: @spirosdelviniotis will test this out.

Maybe something like:

```
COLLECTIONS_DELETED_RECORDS = '{dbquery} AND NOT deleted:True'
```

works.

inspirehep/modules/records/permissions.py
jmartinm: I will use https://github.com/inspirehep/inspire-next/pull/1622 so will not do pickling in the end myself.

inspirehep/modules/search/api.py
jmartinm: The query returned here is added to a ES filter query. You can see the output in this test https://github.com/inspirehep/inspire-next/pull/1589/files#diff-8eedb14f28e016fe0e021789e25878abR336

inspirehep/config.py
```diff
@@ -125,6 +125,114 @@ def _(x):
...
+"""User search query lexical parser."""
+
+COLLECTIONS_QUERY_WALKERS = [
+    'inspirehep.modules.search.walkers.pypeg_to_ast:PypegConverter',
+]
+"""Modules to create the query AST."""
+COLLECTIONS_USE_PERCOLATOR = False
+"""Define which percolator you want to use.
+
+Default value is `False` to use the internal percolator.
```
jmartinm: This docstring is copy/pasted from `invenio-collections`. Maybe the phrase is a bit confusing since they called it 'internal percolator'. If we think of `percolator` just as the ES one, then the config makes sense `COLLECTIONS_USE_PERCOLATOR` `True` or `False`

inspirehep/config.py
```diff
@@ -585,63 +694,59 @@ def _(x):
     ),
 )
 
+
+RECORDS_UI_DEFAULT_PERMISSION_FACTORY = \
+    "inspirehep.modules.records.permissions:record_read_permission_factory"
```
jmartinm: :)

inspirehep/modules/records/permissions.py
jmartinm: Yes, looks nice, will change it for the sake of fanciness.

inspirehep/modules/records/mappings/records/journals.json
jmartinm: We use them in the UI in the canonical form, like `Authors`. But then you can do `/search?cc=Authors` or `/search?cc=authors`

jmartinm:
Comments addressed and changed `redis` cache get/set to use Flask-cache (https://github.com/inspirehep/inspire-next/commit/16e3b4a678f2cd7b2717e121f46882c8561898ae)

Please, review one more or time and ready to 🚢

PR #1600, opened 2016-09-23
WIP

### LINKED ISSUE #1595: get_es_record crash when ES not yet refreshed
As pointed out in https://github.com/inveniosoftware/invenio-records-rest/issues/84#issuecomment-249178606 our implementation of `ESRecord.updated()` in https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/es_record.py#L62 need to be corrected to return `None` when the record hasn't yet been processed by ES (e.g. when it has just been inserted, hence no 404 is returned, but we can't really compute the `updated` attribute.

spirosdelviniotis:
Closes #1595

### DIFF
--- a/inspirehep/modules/records/es_record.py
+++ b/inspirehep/modules/records/es_record.py
@@ -28,6 +28,8 @@ from __future__ import absolute_import, division, print_function
 
 import arrow
 
+from datetime import datetime
+
 from elasticsearch.exceptions import NotFoundError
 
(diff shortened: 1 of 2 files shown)


jacquerie:
Why close? This looked ok: it just needed a little integration work.

spirosdelviniotis:
I have already created a new branch with better name and I commit separately the changes in a new PR.

PR #1601, opened 2016-09-26
Reorder travis

kaplun:


jacquerie:
Oh, interesting idea. I'm not sure if builds are guaranteed to start in the order they're written, but it appears it happened in this case.

kaplun:
I think it's pretty consistent :dancer: at least I checked previous builds and they clearly represented order in `travis.yml` and from top to bottom :)

Mmh. Doesn't improve much. The bottleneck stays `integration` (it's true that even with the previous order Travis was starting it at the very beginning because it was fourth, and Travis is running 4 jobs in parallel).

kaplun:
Ah no wait. It's not true. I think the clock of Travis is not trustworthy.

kaplun:
OK. Travis actually starts 5 jobs at the same time from top to bottom, competing with other build on going. First build come, first served. So indeed having integration at the top is the best way to have it end earlier thus not adding to the max time of the build.

PR #1602, opened 2016-09-26
forms: fix authors autocomplete

jacquerie:
@rikirenz, some things changed from the integrated PR:

* I squashed the first two commits  because they implement the same idea: switching from using a `small_app` in the acceptance tests to a `full_app`, or, as we call it, `app`.
* I squashed together the three commits that fix the suggestions parsers and the three tests commit since they basically do the same thing, except for different fields. Note here that this signals a refactoring opportunity: the fact that you had to something three times means that the code is not DRY (https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).
* I put the commit that implements the feature first, so that `git bisect` won't break when driven by the acceptance test suite.

Then, some style comments:

* People usually write commit titles in the imperative mode, as if they are instructing the code base to do something: not `changed foo` but `change foo`.
* We cargo cult (https://en.wikipedia.org/wiki/Cargo_cult_programming) from Invenio the signoff of commits (eg: `Signed-off-by: Riccardo Candido <riccardocandido@gmail.com>`). There's absolutely no reason to do this except aesthetic consistency (I think that only the Linux kernel uses this Git feature). 

CC: @spirosdelviniotis, you might want to read this as well.

Closes #1554

### DIFF
--- a/inspirehep/modules/forms/static/js/forms/affiliations_typeahead.js
+++ b/inspirehep/modules/forms/static/js/forms/affiliations_typeahead.js
@@ -39,18 +39,13 @@ define([
             if (index != 0) {
               pattern = pattern + " AND ";
             }
-            pattern = pattern + "affautocomplete:" + "/" + this + ".*/";
+            pattern = pattern + "affautocomplete:" + "/" + encodeURIComponent(this) + ".*/";
           })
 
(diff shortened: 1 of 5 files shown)


rikirenz:
Anyway the documentation is missing I am going to do a new pull request now.

PR #1603, opened 2016-09-26
records: get_es_record fix

### LINKED ISSUE #1595: get_es_record crash when ES not yet refreshed

spirosdelviniotis:
* Fixes get_es_record crash.
* Closes #1595 .
* Minor test refactoring.

### DIFF
--- a/tests/integration/test_record.py
+++ b/tests/integration/test_record.py
@@ -121,6 +121,47 @@ def record_not_yet_deleted(app):
         db.session.commit()
 
 
+@pytest.fixture(scope='function')
+def record_not_es_refreshed(app):
+    snippet = (
+        '<record>'
+        '  <controlfield tag="001">444</controlfield>'
+        '  <controlfield tag="005">20160913214552.0</controlfield>'
+        '  <datafield tag="980" ind1=" " ind2=" ">'
+        '    <subfield code="a">HEP</subfield>'
+        '  </datafield>'
+        '</record>'
+    )
+
+    with app.app_context():
+        json_record = hep.do(create_record(snippet))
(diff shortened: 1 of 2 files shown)


tests/integration/test_record.py
```diff
@@ -139,3 +180,8 @@ def test_record_can_be_deleted(app, record_not_yet_deleted):
         db.session.commit()
 
         assert client.get('/api/literature/333').status_code == 410
+
+
+def test_record_can_be_immediately_es_updated(app, record_not_es_refreshed):
+    with app.test_client() as client:
+        assert client.get('/api/literature/444').status_code == 200
```
jacquerie: I really appreciate that you thought of documenting this with a test 👍 

On the other hand, I think it's better if we reuse the previous fixture and remove the index refresh, rather than defining a new "non-refreshed" fixture.

jacquerie: Ah, it's what @kaplun asked you to do [here](https://github.com/inspirehep/inspire-next/issues/1595#issuecomment-249190827) 💃

kaplun: There is a difference: the refresh is really needed if one wants to see what ES has actually indexed. Instead this particular test is a regression test to demonstrate that visiting a record being indexed in this very moment doesn't lead to a 500.

PR #1613, opened 2016-09-28
general: upstream invenio-pidstore

### LINKED ISSUE #1579: Upstream pidstore
We are currently using a forked version of pidstore due to: inveniosoftware/invenio-pidstore#63 which is not yet been integrated due to missing proof that this hasn't a performance impact on MySQL.
- Is there any volunteer to run such a benchmark? @jacquerie maybe it's something you can quickly sketch up with a notebook?
- Alternatively in order to remove our fork we should pick up shorter PIDs and introduce a translation mapping.

kaplun:
- No longer uses INSPIRE fork of invenio-pidstore.
- Introduces new get_pid_type_for() helper to map from endpoints
  to corresponding pid_types, now that pid_types are constrained in
  length.
  (closes #1579)

### DIFF
--- a/inspirehep/modules/pidstore/providers.py
+++ b/inspirehep/modules/pidstore/providers.py
@@ -31,9 +31,16 @@ import os
 import requests
 from flask import current_app
 
+from invenio_records_rest.proxies import current_records_rest
 from invenio_pidstore.models import PIDStatus, RecordIdentifier
-
 from invenio_pidstore.providers.base import BaseProvider
+
+from inspirehep.config import RECORDS_UI_ENDPOINTS
+
+
+_PID_TYPE_TO_ENDPOINT = {value['pid_type']: key
+                         for key, value in RECORDS_UI_ENDPOINTS.items()
+                         if '_' not in key}
 
 
 def _get_next_pid_from_legacy():
@@ -84,8 +91,18 @@ class InspireRecordIdProvider(BaseProvider):
 
     @staticmethod
     def schema_to_pid_type(schema):
-        pid_type = os.path.splitext(os.path.basename(schema))[0]
-        if pid_type == 'hep':
+        endpoint = os.path.splitext(os.path.basename(schema))[0]
+        if endpoint == 'hep':
             # FIXME: temp hack until we rename hep.json to literature.json
-            return 'literature'
-        return pid_type
+            endpoint = 'literature'
+        return get_pid_type_for(endpoint)
+
+
+def get_pid_type_for(endpoint):
+    """Get the pid_type code corresponding to the recid of the given endpoint."""
+    return RECORDS_UI_ENDPOINTS[endpoint]['pid_type']
+
+
+def get_endpoint_from_pid_type(pid_type):
+    """Return the endpoint corresponding to the given pid_type."""
+    return current_records_rest.default_endpoint_prefixes[pid_type]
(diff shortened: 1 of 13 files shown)


inspirehep/modules/pidstore/providers.py
```diff
@@ -84,8 +86,13 @@ def create(cls, object_type=None, object_uuid=None, **kwargs):
...
             # FIXME: temp hack until we rename hep.json to literature.json
-            return 'literature'
-        return pid_type
+            endpoint = 'literature'
+        return get_pid_type_for(endpoint)
+
+
+def get_pid_type_for(endpoint):
+    """Get the pid_type code corresponding to the recid of the given endpoint."""
+    return RECORDS_REST_ENDPOINTS[endpoint]['pid_type']
```
jacquerie: This is called with input coming from the users (e.g. the `/ajax/references` view function), therefore I'd recommend wrapping it in a `try/except` with a default.

jacquerie:
[This helper] (used in the API) is using the opposite mapping from the one you are now introducing. I'd say we should "invert" the dictionary with a dictionary comprehension.

jacquerie:
[This](https://github.com/inspirehep/inspire-next/blob/3da459572ad2a15e1699cf40009b0e68f4769edb/inspirehep/modules/theme/views.py#L573) is another place where you could use the inverse mapping.

jacquerie:
Put it in WIP because it's missing a few uses of the `pid_type` names, but yeah, I like this idea of reducing the custom packages :+1:

inspirehep/modules/pidstore/providers.py
kaplun: shouldn't the caller code be wrapped? (i.e. the one that is directly receiving the input from the user?)

kaplun:
> This helper is using the opposite mapping from the one you are now introducing. I'd say we should also have in the config the "inverted" dictionary generated by a dictionary comprehension.

The link behind *This helper* is broken. What did you want to link?

inspirehep/modules/pidstore/providers.py
kaplun: In particular there is no good default, and it would hide wrong usages. I'd prefer to bubble up the exception to the calling code which is really aware of the origin of the `endpoint` argument and should know how to act on a `KeyError`.

kaplun:
> but yeah, I like this idea of reducing the custom packages :+1:

This one was really a blocker IMHO because our fork is changing the SQLAlchemy model and wouldn't therefore allow for an easy upgrade to the upstream one.

inspirehep/modules/pidstore/providers.py
jacquerie: In this case I'd say it's equivalent, because this is not a whitelist/sanitization/security issue, just a potential source of exceptions on Sentry.

jacquerie:
> The link behind This helper is broken.

Whoops, fixed.

kaplun:
I think I have implemented your requests. 

P.s. OT why haven't you used the Github review for this PR? Just by chance or you don't like it?

jacquerie:
> OT why haven't you used the Github review for this PR? Just by chance or you don't like it?

I love it! But in this case my comments were about things that were **missing** from your PR, therefore I had no way of commenting on those...

kaplun:
Gaah: it's failing only on acceptance tests. I wonder what I am missing at this point?

kaplun:
Sent a new version using upstream `invenio-record-rests` `pid_type` to `endpoint_prefix` mapping.

jacquerie:
What's the status on this? This was a good thing.

kaplun:
Ah nothing, I need focus on it to make it work. I might do it today.

PR #1614, opened 2016-09-29
tests: simple search acceptance tests

kaplun:

### DIFF
--- a/tests/acceptance/test_search.py
+++ b/tests/acceptance/test_search.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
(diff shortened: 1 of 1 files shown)


tests/acceptance/test_search.py
```diff
@@ -0,0 +1,74 @@
...
+    selenium.find_element_by_link_text('Authors').click()
+    assert 'Search 100 authors' in selenium.page_source
+    search_button = selenium.find_element_by_id('search-form-button').click()
+    assert 'Found 100 results.' in selenium.page_source
+
+
+@pytest.mark.xfail(reason='we need first to implement integration of data')
+def test_data(selenium):
+    selenium.find_element_by_link_text('Data').click()
+    assert 'Search 100 data sets' in selenium.page_source
```
jacquerie: @eamonnmag, can you point us to an endpoint we can use to get the number of records in HEPData?

jacquerie:
These tests could be rewritten to use a spy (see: https://pypi.python.org/pypi/pytest-mock/1.2) on `render_template` to assert that it was called with the proper arguments. This would also mean that these tests would be moved to the integration suite, and merged with the existing ones. Or do we think it's important to test e2e these pages?

jmartinm:
I am checking to try to reduce PRs open. Is this still relevant?

kaplun:
I think it has its value, in the sense that it tries to see that the splash pages and regular searches are not crashing in any given PR.

kaplun:
I shall rebase it.

jacquerie:

PR #1615, opened 2016-09-29
orcid: invenio-orcid integration

### LINKED ISSUE #1247: Port to invenio-orcid
After #1093 is integrated, and when `inveniosoftware/invenio-orcid` is ready, we should port to use it.

Panos512:
* Adds invenio-orcid dependency.
* Uses invenio-orcid for record pushing to orcid instead of
  the orcid module of inspire.

Closes #1247


inspirehep/modules/orcid/ext.py
```diff
@@ -40,15 +38,3 @@ def __init__(self, app=None, **kwargs):
     def init_app(self, app, **kwargs):
         """Initialize application object."""
         app.extensions['inspire-orcid'] = self
```
kaplun: Do you need still this file at all?

kaplun:
I think this is no longer WIP right? :+1:

Panos512:
@kaplun yeap 👍

inspirehep/modules/orcid/ext.py
Panos512: Not sure tbh. Isn't this needed to create the extension (in order to declare the receivers etc.)? And also we are gonna need it for #1318. I can try unplugging the extensions although.

Panos512:
@kaplun I thinks it's ready.

PR #1617, opened 2016-09-29
WIP relations: SQLAlchemy approach

kaplun:

### DIFF
--- a/inspirehep/modules/relations/__init__.py
+++ b/inspirehep/modules/relations/__init__.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2015, 2016 CERN.
+#
+# INSPIRE is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Relations."""
+
+from flask import current_app
+
+from werkzeug.urls import url_parse
+from invenio_db import db
+from invenio_records.signals import after_record_insert, after_record_update
+
+from .models import Relations
+
+
+def uri_to_uuid(uri):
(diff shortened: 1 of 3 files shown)


kaplun:
@jacquerie @eamonnmag this is a prototype of table based relations. WDYT?

inspirehep/modules/relations/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+    if len(path_parts) < 2:
+        current_app.logger.error('Bad JSONref URI: {0}'.format(uri))
+        return None
+
+    record_type = path_parts[-2]
+    recid = path_parts[-1]
+    return PersistentIdentifier.get(record_type, recid).id
+
+
+def crawl_json_for_ref(json, root='', relation=''):
```
kaplun: Recursively crawl a json looking for all the `$ref`s and keeping track of the corresponding JSON path (e.g. `authors[0]`) and relation (e.g. `authors`).

inspirehep/modules/relations/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+from flask import current_app
+
+from werkzeug.urls import url_parse
+from invenio_db import db
+from invenio_records.signals import after_record_insert, after_record_update
+
+from .models import Relations
+
+def uri_to_uuid(uri):
```
kaplun: This transform a URI from `$ref` in the corresponding `uuid` to be used in SQL.

inspirehep/modules/relations/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+                    ))
+    return ret
+
+
+@after_record_insert.connect
+@after_record_update.connect
+def upsert_relation(json):
+    with db.session.begin_nested():
+        id_from = uri_to_uuid(json['$self'])
+        Relations.query().filter(Relations.id_from == id_from).delete()
```
kaplun: First deletes all outgoing relations from this record and then re-upload them.

kaplun:
The main thing missing with such approach Vs. the proposed Neo4J one is that I can not filter records via proporties that are not relations. 

E.g. how can I consider only those publication from a given year?

jacquerie:
A similar design was proposed long ago by several people (among which me). It was a bad idea, and it still is: it is essentially reimplementing a graph database in a relational one.

kaplun:
That ii think is what @eamonnmag was proposing yesterday. I mean at some point we have to give up on "this is bad because it implements feature X which is instead for tool Y" since this is incompatible with "we are adding service over service and this is bad".

Why is this approach bad? I'd it really a problem to reimplement? I guess there are predominance implications but are we hitting them at our scale? Compared we the benefit of being able to do joins directly in the db and have transactions consistency?

Alternatively the same technique could be used to populate Neo4j but there we will have to extend it with additional properties if we want to allow for further filtering.

kaplun:
P.s. and including me https://github.com/inveniosoftware/invenio/issues/2573 :smile:

jacquerie:

Issue #1616, opened 2016-09-29
Clean up environment variables and update docker-compose files


jmartinm:
Since Invenio already supports automatically loading config from the OS environment (see comment https://github.com/zenodo/zenodo/blob/master/zenodo/config.py#L28) we don't need to add  `os.environment.get` in our [configuration](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/config.py).

* [ ] Remove the appearances of `os.environment.get`
* [ ] Update docker-compose files to use `APP_<variable_name>` in the environment section.

wurbanski:
Hello,

I'm here from Hacktoberfest, going to have a jab at it ;-)

wurbanski:
Please have a look at #1621.

Also, one thing I noticed (and it really made me twist my brains around) is that the variable name ```SERVER_NAME``` is being reused in different contexts in [docker-compose.test.yml](https://github.com/wurbanski/inspire-next/blob/79057cea55fa69c5444bc3f65820fd54621beac3/docker-compose.test.yml) (as app config once and as test config the second time). Maybe it would be a good idea to rename the variable used for tests to ```TEST_SERVER_NAME``` instead? Of course this would have to be changed in tests as well :)

kaplun:
Closed in  #1621

PR #1621, opened 2016-10-03
Clean up environment variables and update docker-compose files #1616

wurbanski:
* Removed os.environment.get from config
* Updated docker-compose files


jmartinm:
Thanks for your contribution @wurbanski

PR #1624, opened 2016-10-04
WIP global: adds invenio_query_parser dependency

Panos512:
WIP until inveniosoftware/invenio-query-parser#62 is merged.


setup.py
```diff
@@ -58,6 +58,7 @@
     'invenio-logging>=1.0.0a3',
     'invenio-mail>=1.0.0a4',
     'invenio-oauthclient>=1.0.0a8',
+    'invenio-query-parser>=0.6.1.dev20160418',
```
jmartinm: Checking the [commits](https://github.com/inveniosoftware/invenio-query-parser/commits/master) in `invenio-query-parser` this might be `'invenio-query-parser>=0.6'`

inspirehep/modules/search/walkers/printer.py
```diff
@@ -27,7 +27,7 @@
 from invenio_query_parser.walkers import printer
 
 from .. import parser
-from ..ast import SpiresOp
+from invenio_query_parser.contrib.spires.ast import SpiresOp
```
jmartinm: At the moment we have on `inspire-next` all the code that used to be in `contrib` folder (such as https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/search/converter.py)

If we have push rights in `invenio-query-parser` and we can easily modify the `contrib` content then this change should be good. But then you can also cleanup all the files that will not be used anymore from https://github.com/inspirehep/inspire-next/tree/master/inspirehep/modules/search

jacquerie:

PR #1629, opened 2016-10-07
Fix autocompletion journals literature

rikirenz:
I have found this bug during my tests. It is necessary in order to push the new tests.

### DIFF
--- a/inspirehep/modules/records/mappings/records/journals.json
+++ b/inspirehep/modules/records/mappings/records/journals.json
@@ -181,6 +181,17 @@
                         "journalautocomplete"
                     ],
                     "type": "string"
+                },
+                "short_titles": {
+                    "properties": {
+                        "title": {
+                            "copy_to": [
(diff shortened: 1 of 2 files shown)


inspirehep/modules/records/mappings/records/journals.json
```diff
@@ -181,6 +181,17 @@
...
+                "short_titles": {
+                    "properties": {
+                        "title": {
+                            "copy_to": [
+                                "journalautocomplete"
+                            ],
+                            "type": "string"
+                        }
+                    },
+                    "type": "object"
```
jacquerie: As I was saying yesterday, amending the mappings now is somewhat of a fool's errand, because the schemas are still in flux. On the other hand, the test that you're going to add is going to enforce that this feature is preserved, so it's okay.

jmartinm:
👍  Looks good if it goes together with the Selenium test

PR #1635, opened 2016-10-11
audit: dump only valid records.

Panos512:
* Audit table dumping now only dumps data related to specific workflows types.


Panos512:
@jmartinm ping :dancing_men:

PR #1636, opened 2016-10-11
WIP

kaplun:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,7 +77,9 @@ after_script:
 
 after_success:
   - mv .coverage .coverage.tests
+  - ls .coverage*
   - coverage combine
+  - ls .coverage*
   # Coverage report contains Docker paths. We replace them, so that we can run Coveralls.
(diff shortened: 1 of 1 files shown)


jacquerie:
What did you discover from this? Can we close?

kaplun:
The name of coverage files. I wasn't sure if the combine was doing the right thing. Apparently it does.

kaplun:
Actually: just checked the selenium test:
```
travis_time:end:1c8f648d:start=1476200399447458407,finish=1476200399456708514,duration=9250107
[0Ktravis_fold:end:after_success.1
[0Ktravis_fold:start:after_success.2
[0Ktravis_time:start:03def000
[0K$ ls .coverage*
.coveragerc  .coverage.tests
```
I have here the proof that coverage is not created. I know the reason. Coverage would be created only when stopping gunicorn cleanly (e.g. sending a SIGTERM).

What is the best way to do this in Docker within Travis?

kaplun:
Mmh. Shit I was already doing that :(

PR #1651, opened 2016-10-14
travis: pin selenium image

jacquerie:


jacquerie:
The build that failed was the one I mistakenly sent to `origin` instead of `jacquerie` (it failed because I interrupted it). The PR build was passing.

PR #1654, opened 2016-10-17
requirements: invenio-record-editor addition

jmartinm:


jmartinm:
* [ ] Add to search results templates and detailed templates an `Edit` link if a cataloger is logged in. For search results, the information about the logged in user needs to be available in the angular template.

PR #1655, opened 2016-10-17
Change node-sass version to 3.8.0 at INSTALL.rst

zzacharo:
@kaplun Yes i will review it according to the requirements.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -114,7 +114,7 @@ We build assets using `npm`. Make sure you have installed it system wide.
 .. code-block:: console
 
     (inspirehep)$ sudo npm update
-    (inspirehep)$ sudo npm install -g node-sass clean-css requirejs uglify-js
+    (inspirehep)$ sudo npm install -g node-sass@3.8.0 clean-css requirejs uglify-js
 
 
 .. note::


kaplun:
Hi @zzacharo can you amend the commit log of your PR in order to respect the requirements described here:
http://invenio.readthedocs.io/en/latest/technology/git.html#r2-remarks-on-commit-log-messages

In particular:
* you should avoid using a verb in the subject
* you can prefix the subject with the area of improvement: e.g. `INSTALL: `
* Add a bullet list where you describe in detail what this commit does and why.
* Add in parenthesis that it closes the corresponding github issue.

For example of other commit logs just try:
`git log`

jacquerie:
> * you should avoid using a verb in the subject

This just makes commit messages awkward, with absolutely no benefit.

> * you can prefix the subject with the area of improvement: e.g. `INSTALL: `

Better: `docs: `

> * Add a bullet list where you describe in detail what this commit does

No, the commit is already saying that. Explain _why_, if it's not obvious from the title.


All in all:

```
docs: recommend node-sass@3.8.0

* More recent versions of `node-sass` cause issue #1642, therefore
   recommend developers to use this particular version (closes #1644).

Signed-off-by: ...
```

PR #1656, opened 2016-10-17
 packaging: fix pidstore package name 

david-caro:
~~* There was a leftover statement using logging module directly instead
  of the logger object.~~

* This might lead to installing both, upstream as invenio_pidstore and
  another as invenio-pidstore generating strange conflicts.


kaplun:
Ahem... the commit log is unrelated...

david-caro:
Hmm, the `hub` command is misbehaving... xd

david-caro:
Actually, I'm wrong, this is not the reason why both packages got installed together... or why it failed to remove after... Will have to debug further to find the real reason.

PR #1659, opened 2016-10-19
api: add citesummary

### LINKED ISSUE #1304: Citation summary
Legacy supported a mode where several metrics where computed based on the records matching a query. The same functionality should be available on Labs, possibly implemented as an API entry-point allowing for spitting out JSON.

See: https://inspirehep.net/search?ln=en&p=author%3AB.Storaci.1&of=hcs2

jacquerie:
Introduces the new API module, were all things API should live. Note that the folder structure allows for multiple API versions: the same should be added to our `invenio-records-rest` configuration and possibly upstream (I'll expand on this later).

As a PoC for the API module, implements the `citesummary` endpoint for `literature`, `authors`, `institutions` using ES queries.

Addresses #1304
Closes #1572

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -419,6 +419,29 @@ RECORDS_REST_ENDPOINTS = dict(
         read_permission_factory_imp="inspirehep.modules.records.permissions:record_read_permission_factory",
         record_class='inspirehep.modules.records.es_record:ESRecord'
     ),
+    literature_citesummary=dict(
+        default_media_type='application/json',
+        item_route=(
+            '/literature'
+            '/<pid(literature,record_class="inspirehep.modules.records.es_record:ESRecord"):pid_value>'
+            '/citesummary'),
+        list_route='/literature/',
+        max_result_window=10000,
+        pid_fetcher='inspire_recid_fetcher',
+        pid_minter='inspire_recid_minter',
+        pid_type='literature',
+        record_serializers={
+            'application/json': (
+                'inspirehep.modules.api.v1.literature:citesummary_response')
+        },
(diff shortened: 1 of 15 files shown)


kaplun:
It hasn't failed :hooray:

inspirehep/config.py
```diff
@@ -419,6 +419,30 @@ def _(x):
...
         record_class='inspirehep.modules.records.es_record:ESRecord'
     ),
+    literature_citesummary=dict(
+        default_media_type='application/json',
+        item_route=(
+            '/literature'
+            '/<pid(literature,record_class="inspirehep.modules.records.es_record:ESRecord"):pid_value>'
+            '/citesummary'),
+        list_route='/literature/',
+        max_result_window=10000,
```
kaplun: What practical implications has this number?

jacquerie: This might mean that we under report on exactly the two papers with more than 10k citations. I'm not sure if we can push this number a little higher and fix this issue, or we would have to live with that...

inspirehep/config.py
```diff
@@ -647,6 +693,28 @@ def _(x):
...
+        },
+        search_class='inspirehep.modules.search:InstitutionsSearch',
+        search_factory_imp='inspirehep.modules.search.query:inspire_search_factory',
+        search_serializers={
+            'application/json': (
+                'invenio_records_rest.serializers:json_v1_search'),
+            'application/vnd+inspire.brief+json': (
+                'invenio_records_rest.serializers:json_v1_search'),
+        },
+    ),
```
jacquerie: Done.

PR #1660, opened 2016-10-19
records: propagate links for merged records

### LINKED ISSUE #484: Merging records should propagate links

spirosdelviniotis:
- Closes #484.
- Add propagate links for merged records via celery task.
- Added `'$ref'` field in order to make them not analyzed.

Signed-off-by: Spiros Delviniotis spyridon.delviniotis@cern.ch

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -1601,6 +1601,14 @@ ARXIV_TO_INSPIRE_CATEGORY_MAPPING = {
     "stat.TH": "Other"
 }
 
+# Ignored fields for merging records
+# ===========================
+IGNORE_REFS_IN_FIELDS = ['deleted_recid', 'new_recid']
+
+# Json schemas
+# ===========================
+INSPIRE_JSON_SCHEMAS = ['hep', 'authors', 'conferences', 'experiments', 'institutions', 'jobs', 'journals']
+
 # Import local config file if it is present.
 try:
     from inspirehep.config_local import *
--- a/inspirehep/modules/records/mappings/records/hep.json
+++ b/inspirehep/modules/records/mappings/records/hep.json
@@ -117,7 +117,8 @@
                                     "index": "not_analyzed",
                                     "type": "string"
                                 }
-                            }
+                            },
+                            "type": "object"
                         },
                         "name_suggest": {
                             "analyzer": "simple",
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -28,7 +28,7 @@ import six
 from flask import current_app
 
 from invenio_indexer.signals import before_record_index
-from invenio_records.signals import before_record_insert, before_record_update
+from invenio_records.signals import before_record_insert, before_record_update, after_record_update
 
 from inspirehep.dojson.utils import classify_field, get_recid_from_ref
 from inspirehep.modules.pidstore.utils import get_pid_type_from_schema
@@ -390,3 +390,21 @@ def populate_experiment_suggest(sender, json, *args, **kwargs):
                 'payload': {'$ref': get_value(json, 'self.$ref')},
             },
         })
+
+
+@after_record_update.connect
+def check_if_record_has_merged(sender, *args, **kwargs):
+    """Checks if the record is merged after update.
+
+        If 'deleted' and 'new_record' field exists and them value is True, after update,
+        then propagate links to involved records.
+    """
+    from inspirehep.modules.records.tasks import replace_old_ref_for_record
+
+    deleted = sender.get('deleted')
+    new_recid = get_recid_from_ref(sender.get('new_record'))
+    if deleted and new_recid:
+        old_record_ref = str(sender.get('self').get('$ref'))
+        new_record_ref = sender.get('new_record').get('$ref')
+
+        replace_old_ref_for_record.delay(old_record_ref=old_record_ref, new_record_ref=new_record_ref)
--- a/inspirehep/modules/records/tasks.py
+++ b/inspirehep/modules/records/tasks.py
@@ -1,0 +1,153 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Celery tasks provided by Record module."""
+
+from __future__ import absolute_import, division, print_function
+
+from celery import current_app, shared_task
+from celery.utils.log import get_task_logger
+from elasticsearch.helpers import scan
+
+from invenio_search import current_search_client as es
+from invenio_db import db
+from invenio_records.api import Record
+
+from inspire_schemas import api
+from inspirehep.modules.search.utils import dotter
+
+logger = get_task_logger(__name__)
+
+
+@shared_task(ignore_result=True)
+def replace_old_ref_for_record(old_record_ref, new_record_ref):
+    ignore_refs_in_fields = current_app.conf.get('IGNORE_REFS_IN_FIELDS')
+    schemas_names = current_app.conf.get('INSPIRE_JSON_SCHEMAS')
+    possible_fields = get_ref_fields(ignore_refs_in_fields, schemas_names)
+    old_control_number = get_recid_from_url(old_record_ref)
+
+    record_uuids = es_search_pointing_ids_from_ref_in_terms(possible_fields, old_control_number)
+
+    with db.session.begin_nested():
+        for rec_uuid in record_uuids:
+            rec_db = Record.get_record(rec_uuid)
+            find_and_replace_old_link_with_new_link_in_record(rec_db, new_record_ref, old_record_ref)
+            rec_db.commit()
+    db.session.commit()
+
+
+def get_recid_from_url(url):
+    try:
+        res = int(url.split('/')[-1])
+    except ValueError:
+        res = None
+    return res
+
+
+def find_ref_fields_in_schema(schema_name):
+    def construct_str(string):
+        result_list = []
+        for splited_path in string.split('.'):
+            if '_records' in splited_path:
+                result_list.append(splited_path[:-len('_records')] + '_recid')
+                break
+            elif '_record' in splited_path:
+                result_list.append(splited_path[:-len('_record')] + '_recid')
+                break
+            elif 'records' in splited_path:
+                result_list.append(splited_path[:-len('records')] + 'recid')
+                break
+            elif 'record' in splited_path:
+                result_list.append(splited_path[:-len('record')] + 'recid')
+                break
+
+            result_list.append(splited_path)
+        return '.'.join(result_list)[1:]
+
+    schema = api.load_schema(schema_name)
+    all_fields_dotted = dotter(schema.get('properties'), '', [])
+    fields = []
+    for field_dotted in all_fields_dotted:
+        if 'record' in field_dotted:
+            contructed_field = construct_str(field_dotted)
+            fields.append(contructed_field)
+
+    return fields
+
+
+def get_ref_fields(ignore_refs_in_fields, schemas_names):
+    if schemas_names is None:
+        schemas_names = ['hep', 'authors', 'conferences',
+                         'experiments', 'institutions',
+                         'jobs', 'journals']
+    fields_to_check_links = []
+
+    for schema_name in schemas_names:
+        fields_to_check_links += find_ref_fields_in_schema(schema_name)
+
+    set_fields_to_check_links = set(fields_to_check_links)
+    set_bad_fields = set(ignore_refs_in_fields)
+
+    return list(set_fields_to_check_links - set_bad_fields)
+
+
+def es_search_pointing_ids_from_ref_in_terms(fields_recid, old_control_number):
+    body = {
+        "query": {
+            "bool": {
+                "should": [
+                    {
+                        "term": {
+                            _key: {"value": old_control_number}}
+                    }
+                    for _key in fields_recid]
+            }
+        },
+        "fields": [
+            "_id"
+        ]
+    }
+
+    scroll = scan(es, query=body, index="records-*")
+
+    for response in scroll:
+        rec_uuid = response['_id']
+
+        yield rec_uuid
+
+
+def find_and_replace_old_link_with_new_link_in_record(record, new_link, old_link):
+    bad_fields = ['deleted_recid', 'new_recid']
+
+    if isinstance(record, dict):
+        for key, value in record.items():
+            if key not in bad_fields:
+                if isinstance(value, dict):
+                    for key_ref in value:
+                        if key_ref == '$ref':
+                            if old_link == value.get('$ref'):
+                                value['$ref'] = new_link
+
+            find_and_replace_old_link_with_new_link_in_record(value, new_link, old_link)
+    elif isinstance(record, list):
+        for item in record:
+            find_and_replace_old_link_with_new_link_in_record(item, new_link, old_link)
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -38,13 +38,14 @@ from inspirehep.factory import create_app
 @pytest.fixture(scope='session')
 def app():
     """Flask application fixture."""
-    app = create_app()
-    app.config.update(
-        {
-            'DEBUG': True,
-            'WTF_CSRF_ENABLED': False
-        }
-    )
+    app = create_app(DEBUG=True,
+                     WTF_CSRF_ENABLED=False,
+                     TESTING=True,
+                     TEST_RUNNER='celery.contrib.test_runner.CeleryTestSuiteRunner',
+                     CELERY_ALWAYS_EAGER=True,
+                     CELERY_RESULT_BACKEND="cache",
+                     CELERY_CACHE_BACKEND="memory",
+                     CELERY_EAGER_PROPAGATES_EXCEPTIONS=True)
 
     with app.app_context():
         # Imports must be local, otherwise tasks default to pickle serializer.
--- a/tests/unit/records/test_records_tasks.py
+++ b/tests/unit/records/test_records_tasks.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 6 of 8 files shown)


spirosdelviniotis:
I tried to an older version, `'SQLAlchemy<1.1.2'`. When I reinstalled the requirements it the version was `SQLAlchemy (1.0.15)` . The error was exactly the same... :disappointed:

spirosdelviniotis:
I testify with my eyes (@kaplun writing here).

inspirehep/config.py
```diff
@@ -69,6 +69,8 @@ def _(x):
 CELERY_ACCEPT_CONTENT = ['json', 'msgpack', 'yaml']
 CELERY_TIMEZONE = 'Europe/Amsterdam'
 CELERY_DISABLE_RATE_LIMITS = True
+CELERY_ALWAYS_EAGER = True
+CELERY_EAGER_PROPAGATES_EXCEPTIONS = True
```
kaplun: I assume these are year just for testing purposes. We should ship INSPIRE with eager mode on by default. Please remove it in the final PR.

inspirehep/modules/records/mappings/records/authors.json
```diff
@@ -23,6 +23,20 @@
                     "analyzer": "lowercase_analyzer",
                     "type": "string"
                 },
+                "advisors": {
```
kaplun: So, as suggested by @jacquerie at standup, it is true that finally you can search in ES just the recid. So, since this will always contain an integer, ES will have correctly mapped by default, so there won't be a need for touching all these mappings in your PR.

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,136 @@
...
+
+    scroll = scan(es, query=body, index="records-*")
+
+    record_uuids = []
+    for response in scroll:
+        _rec_uuid = response['_id']
+        if _rec_uuid:
+            record_uuids.append(_rec_uuid)
+
+    return record_uuids
```
kaplun: So given the above note on using recids you can basically amend the above to search for `old_recid` instead of `old_link`.

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,136 @@
...
+
+    schema = api.load_schema(collection)
+    all_fields_dotted = dotter(schema.get('properties'), '', [])
+    for field_dotted in all_fields_dotted:
+        if '$ref' in field_dotted:
+            if not any(field in field_dotted for field in fields):
+                if not any(bad_field in field_dotted for bad_field in bad_fields):
+                    fields_dotted.append(field_dotted)
+
+    return [remove_dot(str(field)) for field in fields_dotted]
```
kaplun: So here's the tricky part: you shall amend this function to return you all those `_recid` fields.
Basically, a naïve approach I would do is to walk each schema and look for anyfield called `record` or `*_record` and return a corresponding path with `recid` or `*_recid`.

E.g.: http://inspire-nightly.cern.ch/api/literature/1484423

You can see there is `publication_info.conference_record` so your function should return `publication_info.conference_recid`.

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,136 @@
...
+        for key, value in record.items():
+            if key not in bad_fields:
+                if isinstance(value, dict):
+                    for key_ref in value.keys():
+                        if '$ref' == str(key_ref):
+                            if old_link == value.get('$ref'):
+                                value['$ref'] = new_link
+
+            find_and_replace_old_link_with_new_link_in_record(value, new_link, old_link)
+        return
```
kaplun: No need for this return given the below `elif`.

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,136 @@
...
+        _rec_uuid = response['_id']
+        if _rec_uuid:
+            record_uuids.append(_rec_uuid)
+
+    return record_uuids
+
+
+def find_and_replace_old_link_with_new_link_in_record(record, new_link, old_link):
+    if isinstance(record, dict):
+        for key, value in record.items():
```
kaplun: you can use `iteritems()` which is faster.

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,136 @@
...
+            if key not in bad_fields:
+                if isinstance(value, dict):
+                    for key_ref in value.keys():
```
kaplun: You can simply use: `for key-ref in value`.

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,136 @@
...
+                        if '$ref' == str(key_ref):
```
kaplun: No need to cast.

kaplun: And I swap: variable on the left and value on the right. Since in this way you can read it more naturally:

*If the variable key_ref has value $ref*, instead of *if the value $ref has(?) the variable key_ref.

inspirehep/modules/records/receivers.py
```diff
@@ -361,3 +361,21 @@ def check_if_record_is_going_to_be_deleted(sender, *args, **kwargs):
         else:
             record = get_db_record(collection, control_number)
             soft_delete_pidstore_for_record(record.id)
+
+
+@after_record_update.connect
+def check_if_record_has_merged(sender, *args, **kwargs):
```
jmartinm: I think it could be more accurate to have the trigger of when records are merged by listening to changes in the `pidstore_redirect` table. (using SQLAlchemy events - http://docs.sqlalchemy.org/en/latest/orm/events.html#sqlalchemy.orm.events.MapperEvents.after_update)


```python
from sqlalchemy import event

# standard decorator style
@event.listens_for(SomeClass, 'after_update')
def receive_after_update(mapper, connection, target):
    "listen for the 'after_update' event"

    # ... (event handling logic) ...
```

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,164 @@
...
+def get_recid_from_url(url):
+    try:
+        res = int(url.split('/')[-1])
+    except ValueError:
+        res = None
+    return res
+
+
+def find_ref_fields_in_schema(collection, sum_fields):
+    bad_fields = ['deleted_recid', 'new_recid']
```
spirosdelviniotis: I could pass it as an argument to the function `find_ref_fields_in_schema` .

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def _construct_str(string):
```
spirosdelviniotis: The mission of that function is to translate paths to needed paths. For example, `something_records` -> `something_recid` . That will be the probable field to search the needed recid.

inspirehep/modules/records/tasks.py
kaplun: Yeah, you can also read it as a global form inspirehep.config. In this way we can amend it in case of the schema evolving. Also I wouldn't call it `bad_fields` :) It's more `ignore_refs_in_fields`

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,164 @@
...
+            "_id"
+        ]
+    }
+
+    scroll = scan(es, query=body, index="records-*")
+
+    record_uuids = []
+    for response in scroll:
+        _rec_uuid = response['_id']
+        if _rec_uuid:
```
spirosdelviniotis: You are right! :dancer:

inspirehep/modules/records/receivers.py
spirosdelviniotis: I'll keep it as it's right now because of "race conditions" in SQLalchemy.
I can investigate it more but I don't think that's the proper solution.

spirosdelviniotis:
PR updated :smile:

inspirehep/modules/records/mappings/records/hep.json
```diff
@@ -117,7 +117,8 @@
                                     "index": "not_analyzed",
                                     "type": "string"
                                 }
-                            }
+                            },
+                            "type": "object"
```
kaplun: This should come in a preparatory commit.

inspirehep/modules/records/receivers.py
```diff
@@ -390,3 +390,21 @@ def populate_experiment_suggest(sender, json, *args, **kwargs):
                 'payload': {'$ref': get_value(json, 'self.$ref')},
             },
         })
+
+
+@after_record_update.connect
+def check_if_record_has_merged(sender, *args, **kwargs):
```
kaplun: -> `check_if_record_is_merged` or `_has_been_merged` :)

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+from invenio_records.api import Record
+
+from inspire_schemas import api
+from inspirehep.modules.search.utils import dotter
+
+logger = get_task_logger(__name__)
+
+
+@shared_task(ignore_result=True)
+def replace_old_ref_for_record(old_record_ref, new_record_ref):
```
kaplun: Please, add docstring explaining the parameter and what it does.

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+
+    with db.session.begin_nested():
+        for rec_uuid in record_uuids:
+            rec_db = Record.get_record(rec_uuid)
+            find_and_replace_old_link_with_new_link_in_record(rec_db, new_record_ref, old_record_ref)
+            rec_db.commit()
+    db.session.commit()
+
+
+def get_recid_from_url(url):
```
kaplun: This should go in some helpers. Not sure right now the best module. It might has well been that it already exists. (cc @jmartinm @jacquerie )

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+        res = int(url.split('/')[-1])
+    except ValueError:
+        res = None
+    return res
+
+
+def find_ref_fields_in_schema(schema_name):
+    def construct_str(string):
+        result_list = []
+        for splited_path in string.split('.'):
```
kaplun: typo: `splited` -> `splitted`.

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+            if '_records' in splited_path:
+                result_list.append(splited_path[:-len('_records')] + '_recid')
+                break
```
kaplun: No need for `break` since it's a set of `if..elif`

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+            elif '_record' in splited_path:
+                result_list.append(splited_path[:-len('_record')] + '_recid')
+                break
+            elif 'records' in splited_path:
+                result_list.append(splited_path[:-len('records')] + 'recid')
+                break
+            elif 'record' in splited_path:
+                result_list.append(splited_path[:-len('record')] + 'recid')
+                break
+
```
kaplun: ... assuming here you write `else`

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+    fields = []
+    for field_dotted in all_fields_dotted:
+        if 'record' in field_dotted:
+            contructed_field = construct_str(field_dotted)
+            fields.append(contructed_field)
+
+    return fields
+
+
+def get_ref_fields(ignore_refs_in_fields, schemas_names):
```
kaplun: Add docstring.

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+    if schemas_names is None:
+        schemas_names = ['hep', 'authors', 'conferences',
```
kaplun: You can read it from the config variable that you have added.

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+from inspirehep.modules.search.utils import dotter
+
+logger = get_task_logger(__name__)
+
+
+@shared_task(ignore_result=True)
+def replace_old_ref_for_record(old_record_ref, new_record_ref):
+    ignore_refs_in_fields = current_app.conf.get('IGNORE_REFS_IN_FIELDS')
+    schemas_names = current_app.conf.get('INSPIRE_JSON_SCHEMAS')
+    possible_fields = get_ref_fields(ignore_refs_in_fields, schemas_names)
```
kaplun: This can not be possibly called each time. Otherwise the system goes down. Should be done at initialization time.

@jacquerie, @jmartinm what do you recommend? Shall create yet another internal extension, or there is some common place where this function can be called and its result be stored for the duration of the process?

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+                         'jobs', 'journals']
+    fields_to_check_links = []
+
+    for schema_name in schemas_names:
+        fields_to_check_links += find_ref_fields_in_schema(schema_name)
+
+    set_fields_to_check_links = set(fields_to_check_links)
+    set_bad_fields = set(ignore_refs_in_fields)
+
+    return list(set_fields_to_check_links - set_bad_fields)
```
kaplun: No need to cast to `list`. `set` is iterable as well.

tests/unit/records/test_records_tasks.py
```diff
@@ -0,0 +1,87 @@
...
+            },
+            u'artid': u'45',
+            u'year': 2008,
+            u'page_start': u'45'
+        }],
+        u'self': {
+            u'$ref': u'http://localhost:5000/api/$ref/333'
+        },
+        u'$schema': u'$ref://localhost:5000/schemas/records/hep.json'
+    }
```
jacquerie: This record contains a lot more stuff than what we're interested in testing. Can't it just be
```
json_record = {
    'publication_info': [
        {'parent_record': {'$ref': http://localhost:5000/ref/literature/222}},
    ],
}
```
?

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+
+    scroll = scan(es, query=body, index="records-*")
+
+    for response in scroll:
+        rec_uuid = response['_id']
+
+        yield rec_uuid
+
+
+def find_and_replace_old_link_with_new_link_in_record(record, new_link, old_link):
```
jacquerie: `def update_links(record, new_link, old_link):`

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,153 @@
...
+    bad_fields = ['deleted_recid', 'new_recid']
```
jacquerie: Looks the same as https://github.com/inspirehep/inspire-next/pull/1660/files#diff-3f822b17443888990c9f70f9f89d66f6R1606

jacquerie: But why is it a blacklist, anyway? This means that if some field with a `$ref` is added you need to remember to update this field if needed, otherwise data loss happens. If it's a whitelist then at worst you lose an update, which is going to be done by the background task anyway.

kaplun: ? It's a blacklist because in general we should always update all fields, besides these very 2. It's improbable that more are going to be added. While it's very likely that more $ref are added for which we want to handle updates of links

tests/integration/conftest.py
```diff
@@ -38,13 +38,14 @@
...
-        }
-    )
+    app = create_app(DEBUG=True,
+                     WTF_CSRF_ENABLED=False,
+                     TESTING=True,
+                     TEST_RUNNER='celery.contrib.test_runner.CeleryTestSuiteRunner',
+                     CELERY_ALWAYS_EAGER=True,
+                     CELERY_RESULT_BACKEND="cache",
+                     CELERY_CACHE_BACKEND="memory",
+                     CELERY_EAGER_PROPAGATES_EXCEPTIONS=True)
```
kaplun: This should be separated into a preparatory commit.

tests/unit/records/test_records_tasks.py
kaplun: I guess also `control_number` and `$schema` and `self` are necessary, since these are all meta-metadata that are involved into this process.

kaplun: Actually this test is also to ensure that `journal_$ref` is not considered.

jacquerie: But then there's no assertion about that.

tests/unit/records/test_records_tasks.py
```diff
@@ -0,0 +1,87 @@
...
+    expected = 'http://localhost:5000/api/record/111'
+    old_link = 'http://localhost:5000/api/literature/222'
+
+    find_and_replace_old_link_with_new_link_in_record(json_record, expected, old_link)
+    test_field = json_record.get('publication_info')[0].get('parent_record').get('$ref')
+
+    assert test_field == expected
+
+
+def test_json_walker_input_containing_words_with_ref():
```
jacquerie: If you change `find_and_replace_old_link_with_new_link_in_record` -> `update_links` then this can become a sentence:
```python
def test_update_links_ignores_other_words_containing_ref():
    pass
```

inspirehep/modules/records/tasks.py
jacquerie: https://github.com/inspirehep/inspire-next/blob/a7a87f540af391ba556e7cc40b67a5c226d169f2/inspirehep/dojson/utils/__init__.py#L95-L107

inspirehep/modules/records/tasks.py
jmartinm: I think the correct place is in extension initialization. You could reuse the `search` one if it fits there (https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/search/ext.py#L43) which I am not sure.

Just add an extra line:

```python
self.ref_fields = get_ref_fields(ignore_refs_in_fields, schemas_names)
```

And then access it:

```python
current_app.extensions['inspire-search'].ref_fields
```

Note that the extensions get initialized in some 'random' order so you might not have all the config available at that time so the related configs can be moved next to the `ext.py` https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/search/config.py

inspirehep/modules/records/tasks.py
kaplun: I was almost sure I was wrong!

inspirehep/modules/records/tasks.py
spirosdelviniotis: The function get_recid_from_ref doesn't worked for me because the argument for the task is URL (e.g. `http://localhost:5000/api/record/111`) and that function's argument is something like: `{'$ref': 'http://localhost:5000/api/record/111'}`.
If it's needed I could make a transformation to the argument before calling like:
```
arg = {'$ref': old_record_ref}
get_recid_from_ref(arg)
```
But it doesn't look so good to me.

kaplun: @spirosdelviniotis IMHO you can amend the definition of `get_recid_from_ref` to accept a URL (and consequently adapt all the calls) in a preparatory commit.

jacquerie: @kaplun that looks like a bad idea.

@spirosdelviniotis You just need to stop one step before while doing record traversal, and use the util on that.

jacquerie: BTW this is yet another case in which a whitelist would help: since you don't have to do record traversal to find and replace you can directly call the util on the path that you already have.

jacquerie: The fact that we have an helper for the latter, but not for the former proves otherwise, no?

kaplun: Not necessarily: simply that it was implemented the first time when there was a blob at hand. 
But @spirosdelviniotis is right that it's ugly to create a dummy blob when you don't have it.

inspirehep/modules/records/tasks.py
spirosdelviniotis: @jacquerie (@kaplun writing here): you convinced me about the whitelist usage, and about the fact that potentially we can just use it to iterate over the list and directly fetch the concerned fields, using `get_value()`. However, how do you do in case of (sub-)lists? e.g. say you want to fetch all the `$ref` in `authors`?

PR #1661, opened 2016-10-19
All the changes needed to be able to use hepcrawl

david-caro:
- Using the latest workflows changes made by Jan
- Only talking into account the manual flows and wps and arXiv spiders
- Things that still need checking:
  - [x] Review article flows
  - [x] Review author flows
  - ~~[ ] wps spider ingestion~~ #1681 
  - ~~[ ] audit logs/tracing~~ #1682
  - [x] spider deployment automation/scripting
  - ~~[ ] deploy on nightly :sparkles:~~ #1683 

**Things that still need fixing (so far, coarse description):**

General:
- ~~[ ] From any dashboard page there's no link to itself, or refresh button (minor)~~ #1684
- [x] Delete button does not work (Fixed with the latest versions of workflows)
- ~~[ ] Halted records have no 'message' on why~~ #1685
- ~~[ ] In the global dashboard, instead of showing the total number of flows per type as the main big number, use the number of flows not completed.~~ #1686
- ~~[ ] Restarting a completed flow does nothing (the flow remains accepted/rejected and does not ask for anything).~~ #1687
- ~~[ ] In the flow detail page you can't see the current flow status when it's 'preparing'.~~ #1688
- [x] The 'most recent' ordering seems to not be working (investigating).
- ~~[ ] If there's an error in one of the steps, and you restart the flow, the stack trace error and the message on the detail page do not go away even if the retry passed.~~ same as #1687

Articles:
- [x] Keywords don't get from holding pen to inspire when accepting a record
- ~~[ ] There's no distinction between guessed keywords and manually curated ones~~ #1689
- [x] ~~There's an utf-related error shown on manual literature submission (does not seem to break anything though)~~ **CAN'T REPRODUCE**
- [x] In the detailed view in the holding pen, the 'subject areas' table shows 'score' but that's actually the schema, 'ARXIV' for example
- ~~[ ] In the detailed view, you can manually add any subject area, even if it's not really recognized, breaking the schema.~~ #1690
- ~~[ ] You can add empty named keywords.~~ same as #1690
- [x] After completing a workflow, it still allows editing the object in the detailed view.
- [x] If you change a workflow from the detailed view, it overrides any other status (that is an issue if you edit the record for example while it's preparing, as it overrides the status set by the task in the meantime, and renders the workflow stuck and unrestartable)
- [x] Files don't get passed to the robotupload

Authors:
- [x] Accepting an author does not actually save it.
- ~~[ ] Author submissions have, in the subjects table,  a 'score' that is not actually used for anything.~~ #1691
- [x] #1668 When submitting an author, the `accept` button is not like the others
- [x] Can't accept an author (models related error)
- [x] #1668 When hitting the button 'review submission' it redirects to a non-existing page (http://localhost:5000/submit/author/newreview?objectid={{vm.record.id}})
- [x] There is no author submission button from inspire
- ~~[ ] When going to the author submission form, it highlights the 'literature' tab.~~ #1692
- ~~[ ] The search button in the search bar does not work.~~ #1693
- ~~[ ] When modifying an advisor's data, if you don't hit the tick button it does not get saved.~~ #1694
- ~~[ ] When modifying the year of an experiment, you can input any data.~~ #1690
- ~~[ ] When accepting an author, the message shows 'accepted as non core'.~~ #1695


david-caro:
@salmele @jmartinm @jacquerie 

Here's a small list of things I'm noticing while doing a first overview of the flows will start fixing all the things above by order of priority (blockers, quick + medium impact, any other).

Feel free to write your ideas or if you want to work on any of them (not all are specifically workflow issues ;) ) put your name near it or something so we don't overlap.

Thanks!!!

david-caro:
btw. if you need help setting an env with this up, let me know, It's not trivial... yet

kaplun:
As discussed in IRL the main blocker before deployment is:
> * [ ] Accepting an author does not actually save it.

All the others are minor nuisance that can be fixed even after deployment

david-caro:
:+1: I'll add as issues everything that I don't fix now

jmartinm:
@Panos512 could have a look at the Author form issues, I remember, for example:

> When hitting the button 'review submission' it redirects to a non-existing page (http://localhost:5000/submit/author/newreview?objectid={{vm.record.id}})

was fixed at some point by him, maybe there is some regression.

> Accepting an author does not actually save it.

Save where? :) For a new author, at the end of the workflow a `robotupload` needs to be send to production with the marcxml of the author to be inserted (in --insert mode). Is that what is missing?

david-caro:
> Save where? :) For a new author, at the end of the workflow a robotupload needs to be send to production with the marcxml of the author to be inserted (in --insert mode). Is that what is missing?

Yes, but for non-production env it should save it to database so you can know what is actually being saved (we are doing that already for the authors). This can be alleviated improving the logs/audits too.
Btw. Is there any way to use robotupload on nightly/locally instead? Or is it only available on legacy? (that would be the best)

kaplun:
Robot upload is only available on legacy, but we could actually use inspireheptest... that we can happily bombard :-)

jmartinm:
> Robot upload is only available on legacy, but we could actually use inspireheptest... that we can happily bombard :-)

Yes, in the past we were trying the whole process of creating a record with inspireheptest. And during tests using httpretty to mock the response from robotupload.

Panos512:
@jmartinm  Yeah  I remember this was fixed, but maybe I  missed something. I'll take a look.

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -321,3 +348,8 @@ def user_pdf_get(obj, eng):
         else:
             obj.data['fft'] = [fft]
         obj.log.info("PDF file added to FFT.")
+
+
+def remove_references(obj, eng):
+    obj.log.info(obj.data)
```
david-caro: this should be debug

inspirehep/modules/authors/tasks.py
```diff
@@ -48,6 +48,9 @@ def formdata_to_model(obj, formdata):
     # ======
     # Schema
     # ======
+    if '$schema' not in data and '$schema' in obj.data:
+        data['$schema'] = obj.data.get('$schema')
```
jmartinm: Since `'$schema' in obj.data` we probably don't need `.get()`

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -368,7 +368,7 @@ def prepare_files(obj, eng):
 
     def _get_fft(url, name):
         return {
-            'url': url,
+            'url': os.path.realpath(url),
```
kaplun: :+1:  Are we actually sure on QA this is on AFS?

PR #1668, opened 2016-10-25
workflows: fixes author's review submissions

Panos512:
* Fixes a bug that didn't allow reviewing author's submissions.
* Accept button of workflow's detailed view now has correct styling.

### DIFF
--- a/inspirehep/modules/authors/views/holdingpen.py
+++ b/inspirehep/modules/authors/views/holdingpen.py
@@ -314,16 +314,23 @@ def newreview():
     objectid = request.values.get('objectid', 0, type=int)
     if not objectid:
         abort(400)
-    workflow_metadata.extra_data['is-update'] = True
+
     workflow_metadata = WorkflowUIRecord.get_record(objectid)['metadata']
+
+    # Initialise extra_data
+    workflow_metadata['extra_data'] = workflow_metadata.get('extra_data', {})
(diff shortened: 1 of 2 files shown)


Panos512:
WIP Because:

- [ ]  Research fields don't pass to the form.

inspirehep/modules/authors/views/holdingpen.py
```diff
@@ -314,16 +314,23 @@ def newreview():
     objectid = request.values.get('objectid', 0, type=int)
     if not objectid:
         abort(400)
-    workflow_metadata.extra_data['is-update'] = True
+
     workflow_metadata = WorkflowUIRecord.get_record(objectid)['metadata']
 
+    # Initialise extra_data
```
david-caro: (I mean that I have this fix also on my branch, just swapping the lines, without having to initialize it)

david-caro: No need, I can rebase on top of this (the smaller the pr the better). If you were getting it empty, maybe we should do the extra check and initialize only if not there :/, something like:
```
workflow_metadata['extra_data'] = workflow_metadata.get('extra_data', {})
```

Panos512: done 😉

david-caro:
This shows up other errors down the pipeline, but by itself fixes what it should :+1:

PR #1679, opened 2016-10-27
dojson: fix old email roundtrip

kaplun:

### DIFF
--- a/inspirehep/dojson/hepnames/fields/bd1xx.py
+++ b/inspirehep/dojson/hepnames/fields/bd1xx.py
@@ -260,7 +260,7 @@ def positions(self, key, value):
         'curated_relation': curated,
     }
 
-    emails = [el for el in force_force_list(value.get('m'))]
+    emails = [el for el in force_force_list(value.get('m', 'noemail@inspirehep.net'))]
     emails.extend([el for el in force_force_list(value.get('o'))])
 
     _rank = value.get('r')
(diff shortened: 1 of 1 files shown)


kaplun:
Cc: @jmartinm

jacquerie:
A test would be cool here.

kaplun:
Yeah, this is a workaround to the fact that we tried to simplify our structure in JSONSchema, but we have to de-simplify it when send it back to legacy.

Hence the insertion of a canary value to be sure that something that on legacy was tagged as an old email doesn't suddenly became a current email due to the roundtrip.

jmartinm:
In any case the output will still be broken in some cases, for example in this record https://inspirehep.net/record/1019084/export/xm there are 2 current emails for a position, so when we export it as marcxml the second one will become and old email (subfield `$m`)

kaplun:
Ỳeah that's why the simplification in the first place:

Let's just have a list of current emails, and a list of previous emails. However: what shall we do?

jmartinm:
While inspirehep is the master of the data (and data model) I am not sure we can do this kind of lossy conversion if we want to have forms (like literature and authors) that import from INSPIRE, modify the content and push back to INSPIRE.

This kind of simplifications would be better once we do the final migration of data into the system and Labs becomes master.

jmartinm:
this will not be needed after this PR - https://github.com/inspirehep/inspire-next/pull/1697

PR #1680, opened 2016-10-28
setup: upstream invenio-pidstore

### LINKED ISSUE #1579: Upstream pidstore

jacquerie:
Let's see if I got all uses...

Closes #1579
Closes #1613


jacquerie:
> Let's see if I got all uses...

I got all uses.

PR #1697, opened 2016-10-31
dojson: keep old_emails in positions

jmartinm:
* To avoid losing information when converting from/to legacy marcxml
  keep for the moment the old and new emails on different fields.

### DIFF
--- a/inspirehep/dojson/hepnames/fields/bd1xx.py
+++ b/inspirehep/dojson/hepnames/fields/bd1xx.py
@@ -289,7 +289,7 @@ def positions2marc(self, key, value):
         't': value.get('end_date'),
         'm': value.get('emails'),
         'o': value.get('old_emails'),
-        'z': value.get('current'),
+        'z': 'Current' if value.get('current') else None,
     }
 
 
--- a/inspirehep/dojson/hepnames/fields/bd1xx.py
+++ b/inspirehep/dojson/hepnames/fields/bd1xx.py
@@ -261,7 +261,7 @@ def positions(self, key, value):
     }
 
     emails = [el for el in force_force_list(value.get('m'))]
-    emails.extend([el for el in force_force_list(value.get('o'))])
+    old_emails = [el for el in force_force_list(value.get('o'))]
 
     _rank = value.get('r')
     rank = classify_rank(_rank)
@@ -269,6 +269,7 @@ def positions(self, key, value):
     return {
         'institution': institution if institution['name'] else None,
         'emails': emails,
+        'old_emails': old_emails,
         '_rank': _rank,
         'rank': rank,
         'start_date': value.get('s'),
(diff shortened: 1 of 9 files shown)


jmartinm:
Depends on https://github.com/inspirehep/inspire-schemas/pull/32

jmartinm:
Added a new commit that fixes the update form so that we don't lose information when importing data from production and also so that we don't send back bad data to inspirehep.

On the other hand it is not ideal (same as nowadays on Labs) since to simplify the UI we extract one of the emails marked as current (`$$m` subfield) and add it into the form as `Public email`.

If the user wants to remove that email address, the update marcxml to production will not arrive correctly inside of the `positions` array. Instead all the `positions` will be kept the same.

So as it is, at the moment it is only useful as far as I can see when a user does not have any emai address and wants to add one. Then on the update marcxml there will be an extra `371__m` field with that email address and we will add it into the record.

I am not sure if there is a way to show in the form the emails as a list but still keep them somehow updated inside the `positions` array so that the update sent to production makes sense.

It would be easy if I add the email addresses close to each institution history entry but that makes the form quite cluttered and not readable (blame it on our current data model)

jmartinm:
Tests pass by the way, a new review would be good :)

inspirehep/dojson/hepnames/fields/bd1xx.py
```diff
@@ -281,16 +282,14 @@ def positions(self, key, value):
...
         'a': value.get('institution', {}).get('name'),
         'r': value.get('_rank'),
         's': value.get('start_date'),
         't': value.get('end_date'),
-        'm': emails[0] if emails else None,
-        'o': emails[1:] if emails else None,
-        'z': value.get('current'),
+        'm': value.get('emails'),
+        'o': value.get('old_emails'),
+        'z': 'Current' if value.get('current') else None,
```
david-caro: good catch

inspirehep/modules/authors/views/holdingpen.py
```diff
@@ -43,14 +43,12 @@
 from flask_breadcrumbs import register_breadcrumb
 from flask_login import login_required, current_user
 from werkzeug.datastructures import MultiDict
-from wtforms.widgets import HiddenInput
```
david-caro: this was unused?

inspirehep/modules/authors/views/holdingpen.py
```diff
@@ -43,14 +43,12 @@
...
 
 from invenio_db import db
 
 from invenio_workflows import workflow_object_class, start, resume
 from invenio_workflows_ui.api import WorkflowUIRecord
 
-from inspirehep.dojson.utils import strip_empty_values
```
david-caro: same here

inspirehep/modules/authors/dojson/fields/updateform.py
```diff
@@ -180,12 +180,12 @@ def institution_history(self, key, value):
             continue
         positions.append({
             "institution": {'name': position["name"]},
-            "status": "current" if position["current"] else "",
+            "current": 'Current' if position["current"] else None,
```
david-caro: The status one made more sense imo, I'd try to keep that one instead if possible.

inspirehep/modules/authors/views/holdingpen.py
jmartinm: yep, unused imports that were bothering me :P

david-caro: :+1:  (they bother me too :stuck_out_tongue_winking_eye: )

jmartinm:
Pushed last version, I think I finally managed to have a more or less elegant solution so that we can present to the user all their current emails (to be able to modify them) and the MarcXML update sent to legacy will be understandable to apply the change (more info in commit msg)

PR #1672, opened 2016-10-26
citesummary: adds cite summary view

eamonnmag:
- ADD Adds citesummary endpoint.
- TODO: Needs data endpoint available @jacquerie.

Signed-off-by: Eamonn Maguire eamonnmag@gmail.com

PR #1699, opened 2016-11-01
theme: add citesummary view

### LINKED ISSUE #1304: Citation summary

jacquerie:
Closes #1304 
Closes #1672

### DIFF
--- a/inspirehep/modules/theme/bundles.py
+++ b/inspirehep/modules/theme/bundles.py
@@ -71,12 +71,6 @@ js = NpmBundle(
     }
 )
 
-detailedjs = NpmBundle(
-    "js/detailed_record_init.js",
-    filters="requirejs",
-    output="gen/detailed_record.%(version)s.js",
-)
-
 css = NpmBundle(
     "scss/inspirehep.scss",
     filters="node-scss, cleancss",
@@ -106,3 +100,23 @@ holding_pen_css = NpmBundle(
     output="gen/inspirehep.holding.%(version)s.css",
     depends="scss/**/*.scss"
 )
+
+citesummary_js = NpmBundle(
+    "js/inspire_citesummary_init.js",
+    filters='requirejs',
+    output="gen/citesummary.%(version)s.js",
+    npm={
+        "inspire-citesummary-js": "~0.0.5"
+    }
+)
+
+citesummary_css = NpmBundle(
+    "node_modules/dc/dc.min.css",
+    "node_modules/inspire-citesummary-js/examples/style.css",
+    filters='cleancss',
+    output="gen/citesummary.%(version)s.css",
+    npm={
+        "dc": "~2.0.0-beta.32"
+    }
+)
+
--- a/inspirehep/modules/theme/static/js/inspire_citesummary_init.js
+++ b/inspirehep/modules/theme/static/js/inspire_citesummary_init.js
@@ -1,0 +1,28 @@
+/*
+ * This file is part of INSPIRE.
+ * Copyright (C) 2014 CERN.
+ *
+ * INSPIRE is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * INSPIRE is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * In applying this licence, CERN does not waive the privileges and immunities
+ * granted to it by virtue of its status as an Intergovernmental Organization
+ * or submit itself to any jurisdiction.
+ */
+
+require([
+  'citesummary'
+], function() {
+  // Loaded libraries that run on every page
+
+});
--- a/inspirehep/modules/theme/static/js/settings.js
+++ b/inspirehep/modules/theme/static/js/settings.js
@@ -35,14 +35,19 @@ require.config({
     'bootstrap-datetimepicker': 'node_modules/eonasdan-bootstrap-datetimepicker/src/js/bootstrap-datetimepicker',
     'bootstrap-multiselect': 'node_modules/bootstrap-multiselect/dist/js/bootstrap-multiselect',
     'bucketsjs': 'node_modules/buckets-js/dist/buckets',
+    'citesummary': 'node_modules/inspire-citesummary-js/dist/inspire-citesummary-js',
     'clipboard': 'node_modules/clipboard/dist/clipboard',
+    'crossfilter': 'node_modules/crossfilter2/crossfilter.min',
     'd3': 'node_modules/d3/d3',
     'd3-tip': 'node_modules/d3-tip/index',
     'd3wrapper': 'js/d3-wrapper',
+    'dc': 'node_modules/inspire-citesummary-js/src/lib/dc',
+    'dc-legend2': 'node_modules/inspire-citesummary-js/src/lib/dc.legend2',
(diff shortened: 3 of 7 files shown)


jacquerie:
It works (http://inspire-nightly.cern.ch/experiments/1110643/citesummary), but I probably did some atrocities with `require.js`, so I want @jmartinm's opinion on this.

inspirehep/modules/theme/static/js/inspire_citesummary_init.js
```diff
@@ -0,0 +1,28 @@
+/*
+ * This file is part of INSPIRE.
+ * Copyright (C) 2014 CERN.
```
jacquerie: Bad year.

inspirehep/modules/theme/static/js/settings.js
```diff
@@ -132,10 +137,31 @@ require.config({
...
+    'dc-legend2': {
+      deps: [
+        'dc'
+      ]
+    },
     'holding-pen-controllers': {
       deps: [
         'angular',
-     ],
+      ],
```
jacquerie: Sigh. To be split in its own commit.

inspirehep/modules/theme/static/scss/base/core.scss
```diff
@@ -183,29 +182,120 @@ a[ng-click] {
...
+  -o-animation: loading-bar-spinner 400ms linear infinite;
+  animation: loading-bar-spinner 400ms linear infinite;
 }
 
-
 .modal-body {
-    max-height: calc(100vh - 212px);
-    overflow-y: auto;
+  max-height: calc(100vh - 212px);
+  overflow-y: auto;
```
jacquerie: Style changes above this line will be split in another commit.

inspirehep/modules/theme/static/scss/base/core.scss
```diff
@@ -183,29 +182,120 @@ a[ng-click] {
...
+    }
+
+    &.journals {
+      background-color: $journals-collection;
+
+      .cs-back-link {
+        padding-bottom: 15px;
+        border-bottom: 1px solid $asbestos;
+      }
+    }
```
jacquerie: I wonder if there's a way to use mixins to DRY these.

inspirehep/modules/theme/templates/inspirehep_theme/citations/citation_summary.html
```diff
@@ -0,0 +1,77 @@
+{#
+# This file is part of INSPIRE.
+# Copyright (C) 2015, 2016 CERN.
```
jacquerie: Bad year.

inspirehep/modules/theme/templates/inspirehep_theme/citations/citation_summary.html
```diff
@@ -0,0 +1,77 @@
...
+# License, or (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
```
jacquerie: Bad license.

inspirehep/modules/theme/views.py
```diff
@@ -182,6 +182,11 @@ def data():
     )
 
 
+@blueprint.route('/<string:collection>/<int:recid>/citesummary', methods=['GET',])
+def citation_summary(collection, recid):
+    """ View for citation summary for a record """
+    return render_template('inspirehep_theme/citations/citation_summary.html', collection=collection, recid=recid)
+
```
jacquerie: Add a space here.

inspirehep/modules/theme/bundles.py
```diff
@@ -71,12 +71,6 @@
     }
 )
 
-detailedjs = NpmBundle(
-    "js/detailed_record_init.js",
-    filters="requirejs",
-    output="gen/detailed_record.%(version)s.js",
-)
```
jacquerie: Also this cleanup should go in its own commit.

jmartinm: Why is this removed? seems to be still in `setup.py` https://github.com/inspirehep/inspire-next/pull/1699/files#diff-2eeaed663bd0d25b7e608891384b7298R239

jmartinm:
> It works (http://inspire-nightly.cern.ch/experiments/1110643/citesummary), but I probably did some atrocities with require.js, so I want @jmartinm's opinion on this.

Nothing that calls my attention at first sight (apart from the removal of `detailedjs`)

Something specific that you did and are not sure if it is correct?

jacquerie:
> Something specific that you did and are not sure if it is correct?

https://github.com/inspirehep/inspire-next/pull/1699/files#diff-b389934115eaa8ec351610c9ddb7496fR154

As far as I understand, `dc` does not do this exporting on its own, but just defines a `dc` variable. This configuration puts that `dc` variable in the global scope of a page that uses `citesummary.js`, which sounds pretty yucky to me.

jmartinm:
> As far as I understand, dc does not do this exporting on its own, but just defines a dc variable. This configuration puts that dc variable in the global scope of a page that uses citesummary.js, which sounds pretty yucky to me.

I see, actually checking the source of `dc.js` it seems like it supports requirejs. There is the following code:

```javascript
if(typeof define === "function" && define.amd) {
        define(["d3", "crossfilter"], _dc);
    } else if(typeof module === "object" && module.exports) {
        var _d3 = require('d3');
        var _crossfilter = require('crossfilter');
        // When using npm + browserify, 'crossfilter' is a function,
        // since package.json specifies index.js as main function, and it
        // does special handling. When using bower + browserify,
        // there's no main in bower.json (in fact, there's no bower.json),
        // so we need to fix it.
        if (typeof _crossfilter !== "function") {
            _crossfilter = _crossfilter.crossfilter;
        }
        module.exports = _dc(_d3, _crossfilter);
    } else {
        this.dc = _dc(d3, crossfilter);
    }
```

So in theory just by adding it on `require` should work. Although I assume you tried and that wasn't the case.

jmartinm:
I can take this branch and give it a try locally to see first hand. Will report here my findings.

jmartinm:
To me it seems to work without the `dc` entry in the settings.js `shim` config. What was the error you were seeing?

I only get every now and then an error when brushing:

```
Error: <g> attribute transform: Trailing garbage, "translate(0,NaN)"
```

inspirehep/modules/theme/bundles.py
jmartinm: Oh I see 👍

jacquerie:
> Although I assume you tried and that wasn't the case.

Yup. Even on Nightly. What configuration are you using? `DEBUG`, `ASSETS_DEBUG`...?

jmartinm:
> Yup. Even on Nightly. What configuration are you using? DEBUG, ASSETS_DEBUG...?

I was using all production style. No `DEBUG` or `ASSETS_DEBUG`.

jacquerie:
Another bug I spotted: the `citesummary`view should handle gracefully the case in which a paper has no citations. For example: http://inspire-nightly.cern.ch/api/literature/1483042/citesummary results in http://inspire-nightly.cern.ch/literature/1483042/citesummary (check the error in the console). This can happen in real data (it's not an artifact of a non-full migration): https://inspirehep.net/record/1483042

jacquerie:
Another bug I spotted: http://inspire-nightly.cern.ch/literature/451647/citesummary (again, check the console).

kaplun:
Before we deploy this we need to solve performance issues.

jacquerie:
Postponed, will be resurrected if necessary.

PR #1703, opened 2016-11-03
tools: strip empty lines in authorlist

### LINKED ISSUE #1702: Authorlist: cannot parse affiliations when empty lines
The new authorlist tool to convert lists of affiliations into MARC fails when empty lines are present.

Slightly changing the second example into
```
F. Lastname1, F.M. Otherlastname1,2

1 CERN
2 Otheraffiliation
```
gives the error message
```
Sorry! Seems there is a problem somewhere. Please see error message: ('Cannot parse affiliation ', u'')
```

jacquerie:
Allows an arbitrary number of blank lines between the authors line and the affiliations by stripping them before processing.

Closes #1702


jacquerie:
@bittirousku, note that `tests/unit/tools/test_utils.py` should have been `tests/unit/tools/tests_tools_utils.py`. This is because of how test collection works in py.test: if two test files in the same suite share the same name but live in different folders you get an error when collecting. So we need to namespace the tests by redundantly adding the folder name to the filename. Ugly, but it works.

jacquerie:
Waiting for @bittirousku to review.

kaplun:
I discovered from @annetteholtkamp that the input catalogers put come from cut & pasting from PDFs. So we might want to do even more clean up. 

Is there some quick PDF library available for stripping/cleaning that we can clean up at this stage?

kaplun:
Actually we can use refextract which is installed: https://github.com/inspirehep/refextract/blob/master/refextract/documents/pdf.py#L434 and/or https://github.com/inspirehep/refextract/blob/master/refextract/documents/text.py#L164

bittirousku:
@jacquerie Thanks! I sort of knew the test naming thing, but didn't see the problem here before.

jacquerie:
> Thanks! I sort of knew the test naming thing, but didn't see the problem here before.

Yeah, this is because we've been doing this everywhere, so the first time that you add a `test_utils.py` nothing happens, because it's the only one, but the second time someone does it everything blows up.

jacquerie:
> I discovered from @annetteholtkamp that the input catalogers put come from cut & pasting from PDFs. So we might want to do even more clean up. 

First they give us the failures, then we fix them, like it has happened here. Otherwise we might be fixing the wrong thing.

kaplun:
@jacquerie what I mean is that we have developed already along the year a set of heuristic to undo many typically (LaTeX/)PDF quirks that, given the information the cataloguer is cut & pasting from PDF, we can already apply at no cost (since it's already written and tested) and would anticipate going back and fort between the cataloguer and the developer and deployement...

kaplun:
They will come with ligature issue ffi Vs fi, etc.

To me it's a total loss of time of at least three person, that can be saved by simply applying experience.

PR #1709, opened 2016-11-06
setup: pin elasticsearch-dsl

jacquerie:
Otherwise https://travis-ci.org/inspirehep/inspire-next/jobs/173476792 happens.


jacquerie:
Note that `elasticsearch==2.2.0` fails: https://travis-ci.org/inspirehep/inspire-next/jobs/173663654.

PR #1710, opened 2016-11-07
authors: form output schema compliant

jmartinm:
* Makes sure the output of the author creation form is compliant with
  the json schema, to be able to save the author locally.

### DIFF
--- a/inspirehep/modules/authors/dojson/fields/updateform.py
+++ b/inspirehep/modules/authors/dojson/fields/updateform.py
@@ -186,7 +186,7 @@ def institution_history(self, key, value):
             continue
         positions.append({
             "institution": {'name': position["name"]},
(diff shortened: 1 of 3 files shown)


inspirehep/modules/authors/dojson/fields/updateform.py
```diff
@@ -186,7 +186,7 @@ def institution_history(self, key, value):
             continue
         positions.append({
             "institution": {'name': position["name"]},
-            "current": 'Current' if position["current"] else None,
+            "current": 'Current' if position["current"] else False,
```
jacquerie: This looks weird, because this has type `[bool, string]`.

jmartinm: You are right, in the schema is defined as `boolean` so should be failing when passing a string.

PR #1713, opened 2016-11-07
pidstore: remove invenio-records-rest from api app

jmartinm:
Signed-off-by: Javier Martin Montull <javier.martin.montull@cern.ch>

PR #1717, opened 2016-11-08
setup: remove invenio-records-rest from apps

### LINKED ISSUE #1711: Click on search results gives JSON view
I assume after the introduction of https://github.com/inspirehep/inspire-next/commit/f6c29308cad90ac3c70cab52f0db7c466246fb06 this is happening.

When clicking on any search result, instead of the detailed view, the JSON view is being displayed.

jacquerie:
Closes #1711
Closes #1713
Closes #1718

### DIFF
--- a/inspirehep/modules/pidstore/utils.py
+++ b/inspirehep/modules/pidstore/utils.py
@@ -24,20 +24,31 @@
 
 from __future__ import absolute_import, division, print_function
 
+from flask import current_app
+from six import iteritems
 from six.moves.urllib.parse import urlsplit
 
-from invenio_records_rest import current_records_rest
+
+def _get_pid_type_endpoint_map():
+    pid_type_endpoint_map = {}
+    for key, value in iteritems(current_app.config['RECORDS_REST_ENDPOINTS']):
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -42,7 +42,7 @@ def app():
     app.config.update(
         {
             'DEBUG': True,
-            'WTF_CSRF_ENABLED': False
+            'WTF_CSRF_ENABLED': False,
```
jmartinm: This is not needed. Just a leftover trying to make it work with `invenio-db[postgresql,versioning]==1.0.0b3`

tests/integration/test_permissions.py
```diff
@@ -84,6 +84,7 @@ def sample_record(app):
     db.session.delete(pid)
     record.delete(force=True)
     es.delete(index='records-hep', doc_type='hep', id=pid.object_uuid)
+    es.indices.refresh('records-hep')
```
jmartinm: Same with these 2 changes, they are not needed in principle, was trying to debug the problems with `invenio-db[postgresql,versioning]==1.0.0b3`

PR #1719, opened 2016-11-08
tests: invenio-db==1.0.0b3 support

jmartinm:
* Makes sure `transaction` table is cleaned properly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -116,13 +116,13 @@ extras_require = {
         'Sphinx>=1.3',
     ],
     'postgresql': [
-        'invenio-db[postgresql,versioning]>=1.0.0a6,<=1.0.0b2',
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -72,6 +72,8 @@ def app():
         es.indices.refresh('records-hep')  # Makes sure that all citation counts were added.
 
         yield app
+        db.session.remove()
+        db.drop_all()
```
jacquerie: Do we really need these two?

jmartinm: It wasn't needed it seems. Just pushed a new version removing these lines.

jmartinm:
Tests now passing. Ok to merge?

PR #1722, opened 2016-11-08
WIP: use recid as pid_type

jmartinm:
Giving a try if it is possible to have a single `pid_type` called `recid` shared by all our endpoints.

Tests are supposed to fail for the moment.

### DIFF
--- a/inspirehep/modules/hal/tei.py
+++ b/inspirehep/modules/hal/tei.py
@@ -122,7 +122,7 @@ def _parse_structures(record):
                 continue
 
     try:
-        records = get_es_records('ins', recids)
+        records = get_es_records('institutions', recids)
     except RequestError:
         records = []
 
--- a/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_detailed.tpl
+++ b/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_detailed.tpl
@@ -131,7 +131,7 @@
       ) {
         DataTables.attachTo(document, {
           'recid': "{{ record.control_number }}",
-          'collection': "literature",
+          'endpoint': "literature",
           {% if record['series'] %}
             'seriesname' : "{{ record['series'][0] }}"
           {% endif %}
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -354,9 +354,24 @@ SEARCH_ELASTIC_KEYWORD_MAPPING = {
 # Records
 # =======
 RECORDS_REST_ENDPOINTS = dict(
-    literature=dict(
+    record=dict(
         default_endpoint_prefix=True,
-        pid_type='lit',
+        pid_type='recid',
+        list_route='/record/',
+        item_route=(
+            '/record'
+            '/<pid(recid,record_class="inspirehep.modules.records.es_record:ESRecord"):pid_value>'),
+        record_serializers={
+            'application/json': ('invenio_records_rest.serializers'
+                                 ':json_v1_response'),
+        },
+        search_serializers={
+            'application/json': ('invenio_records_rest.serializers'
+                                 ':json_v1_search'),
+        }
+    ),
+    literature=dict(
+        pid_type='recid',
         pid_minter='inspire_recid_minter',
         pid_fetcher='inspire_recid_fetcher',
         search_class='inspirehep.modules.search:LiteratureSearch',
@@ -413,7 +428,7 @@ RECORDS_REST_ENDPOINTS = dict(
         list_route='/literature/',
         item_route=(
             '/literature'
-            '/<pid(lit,record_class="inspirehep.modules.records.es_record:ESRecord"):pid_value>'),
+            '/<pid(recid,record_class="inspirehep.modules.records.es_record:ESRecord"):pid_value>'),
         default_media_type='application/json',
         max_result_window=10000,
         search_factory_imp='inspirehep.modules.search.query:inspire_search_factory',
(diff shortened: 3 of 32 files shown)


jmartinm:
Things seem to mostly work. One blocker I see at the moment is that the record class receives from `pidstore` only the `object_uuid` https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/es_record.py#L49 and to fetch that single UUID from ElasticSearch we need to know what index to fetch from.

I am not aware yet of other side effects.

One good thing is that, since all the records would have the same `pid_type` we would avoid the extra query to the database in the [record getter](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/utils/record_getter.py)

kaplun:
Can't we fetch the from `_all` index?

jmartinm:
> Can't we fetch the from _all index?

There is no such a thing for the [GET API](http://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch.Elasticsearch.get). `index` parameter is mandatory and there is no `_all`

inspirehep/modules/hal/tei.py
```diff
@@ -122,7 +122,7 @@ def _parse_structures(record):
                 continue
 
     try:
-        records = get_es_records('ins', recids)
+        records = get_es_records('recid', recids)
```
jacquerie: I know that this is just a PoC, but this refactor should remove this parameter from all these helper functions, rather than forcing people to tediously type `recid` everywhere.

inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_detailed.tpl
```diff
@@ -131,7 +131,7 @@
       ) {
         DataTables.attachTo(document, {
           'recid': "{{ record.control_number }}",
-          'collection': "literature",
+          'endpoint': "literature",
```
jacquerie: Ah-ah! I missed this guy in my PR.

inspirehep/modules/hal/tei.py
jmartinm: Yes, the `recid` will be implicit and people don't have to think about it.

On the other hand, you will have to pass a more high level endpoint in order to know what search class to use (and what index). So this call would be:

```
records = get_es_records('literature', recids)
```

Where literature is the key in `RECORDS_REST_ENDPOINTS`

jmartinm:
Current challenges:

1) With the current config, `current_records_rest.default_endpoint_prefixes` has inside `{'recid': 'literature'}` so the links provided by the API ([see here](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L450) always point to `/literature` no matter what collection we are in. This does not have any visible effect when using the web UI, but it would be no good for a API user.

2) Support `/record/<recid>`. Here we will need to fetch the record from DB, have a function `schema_to_endpoint` and redirect to the appropriate `/<endpoint>/<recid>`.

One option for 1) would be to introduce an extra entry in `RECORDS_REST_ENDPOINTS`:

```python
record=dict(
        default_endpoint_prefix=True,
        pid_type='recid',
        list_route='/record/',
        item_route=(
            '/record'
            '/<pid(recid,record_class="inspirehep.modules.records.es_record:CustomRecordClass"):pid_value>'),
        record_serializers={
            'application/json': ('invenio_records_rest.serializers'
                                 ':json_v1_response'),
        },
        search_serializers={
            'application/json': ('invenio_records_rest.serializers'
                                 ':json_v1_search'),
        }
    ),
```

that would act as our default prefix.

Then the API response would have:

```
"links": {
    "next": "http://localhost:5000/api/record/?sort=-mostrecent&q=&page=2&size=25", 
    "self": "http://localhost:5000/api/record/?sort=-mostrecent&q=&page=1&size=25"
  }
```

But such a generic API endpoint would not know what search class to use (so what index from ES)

jmartinm:
Also worth mentioning, some advantages of this PR:

1) Higher level API for developers. When interacting with ES APIs from `record_getter` you would call it like `get_es_records('literature', recids)` using the endpoint names defined in `RECORDS_REST_ENDPOINTS`. (NOTE that we can also do this with current separation of pid_types)

2) `pidstore_pid` table would have all our IDs with `pid_type` `recid` which would work like currently on legacy, ensuring that a given id is unique (and accessible with `/record/XXX`)

3) Simplified/easier to understand code.

jmartinm:
So @jacquerie , do you think the advantages are worth it? Or seeing that there are things to work around we might as well stay as we are? As I don't see that many advantages, I am happy to throw it away.

jacquerie:
The thing I don't understand is how to overcome this issue:

> 1) With the current config, `current_records_rest.default_endpoint_prefixes` has inside `{'recid': 'literature'}` so the links provided by the API ([see here](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/views.py#L450) always point to `/literature` no matter what collection we are in. This does not have any visible effect when using the web UI, but it would be no good for a API user.

We **must** record somewhere where this record lives, so that we can query the appropriate ES index. The PIDStore looks like the most sensible place for this. So yeah, I'm leaning towards throwing this PR away. On the other hand, https://github.com/inspirehep/inspire-next/pull/1722#discussion_r87065876 must be salvaged from it!

jmartinm:
> We must record somewhere where this record lives, so that we can query the appropriate ES index. The PIDStore looks like the most sensible place for this. So yeah, I'm leaning towards throwing this PR away. On the other hand, #1722 (comment) must be salvaged from it!

Yep, agreed.

Will create a PR with that small fix.

PR #1607, opened 2016-09-26
WIP Selenium tests for literature and author forms

rikirenz:
Closes #1593

PR #1730, opened 2016-11-10
tests: add acceptance tests for forms

### LINKED ISSUE #1593: Selenium tests for literature and author forms
# Literature suggestion form

1) Access http://localhost:5000/submit/literature/create
2) Test the form validators. Form is defined in [this file](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/literaturesuggest/forms.py) and an example of [field with validator](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/literaturesuggest/forms.py#L234)
3) Submit the form and get success page.

Now you should see in the console logs that the worker starts doing some computation (executing the workflow) and after some seconds the record should be in the Holding Pen. Let's see if that is true:

4) Access http://localhost:5000/holdingpen/ and check that the dashboard shows the new submission

screen shot 2016-09-23 at 11 04 35 am

5) Click on the Halted link and you will get to the Holding Pen listing:

screen shot 2016-09-23 at 11 04 58 am

6) Click on the holding pen entry and do some assertions on the detailed page, e.g, check that the reject, accept and core buttons are there and check that the metadata you filled in the form is there: title, authors, subject and so on.

7) Click on Accept and make sure you get the right message on the page:

screen shot 2016-09-23 at 11 08 56 am
# Author suggestion form

1) Go to http://localhost:5000/submit/author/create

2) Fill the form and test validators. Form is defined in [this file](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/authors/forms.py)

3) Submit and get success page.

4) Go to http://localhost:5000/holdingpen/ and confirm the dashboard shows correct info.

screen shot 2016-09-23 at 11 14 21 am

5) Click on Halted to see the entry that was just submitted

screen shot 2016-09-23 at 2 48 52 pm

6) Go into the detailed page and make sure the 4 action buttons are there and that the metadata corresponds to the metadata entered in the form.

screen shot 2016-09-23 at 2 49 23 pm

7) Clicking on Accept, Reject or Core should show a success message.

8) Clicking on Review Submission should go back to the original form.
- There should be 2 extra fields in the form (that are only in the review): inspireid and bai

9) Try submitting the review form.
# Author update form

Same as author suggestion form until accessing the detailed page in the Holding Pen.

The detailed page should not have any action button.

jacquerie:
Closes #1593
~~Closes #1607~~
~~Closes #1677~~
~~Closes #1740~~

- [x] Clean up `docs/tests.rst`
- [x] Use `jsonify` when it makes sense
- [x] ~~Use fixtures when it makes sense~~


.travis.yml
```diff
@@ -72,12 +72,12 @@ script:
...
+  - "docker-compose -f docker-compose.test.yml kill -s SIGHUP web"
+  - "docker-compose -f docker-compose.test.yml kill"
   - "docker-compose -f docker-compose.test.yml rm -f"
 
 after_success:
-  - mv .coverage .coverage.tests
-  - coverage combine
   # Coverage report contains Docker paths. We replace them, so that we can run Coveralls.
   - sed -i 's@\"/code/@'"\"$(pwd)/"'@g' .coverage
+  - cat .coverage
```
jmartinm: was this for debugging or we want to see that file on travis output?

docs/tests.rst
```diff
@@ -23,10 +23,95 @@
...
 =======================
 
+How to Write Selenium Tests
+---------------------------
+
+Selenium Test Framework
+~~~~~~~~~~~~~~~~~~~~~~~
+
+INSPIRE's Selenium tests are written using an in-house framework called BAT
+(``inspirehep/bat``). The framework is made of four main components:
```
jmartinm: Should this be a link to the module code?

docs/tests.rst
```diff
@@ -23,10 +23,95 @@
...
+                'institution_history-0-name', institution)
+
+        return ArsenicResponse(_write_institution)
+
+in the same module represents the action of filling the autocomplete field
+of id ``institution_history-0-name`` with the content of the ``institution``
+variable.
+
+Note that the latter method returns a closure over ``expected_data`` and
+``institutions`` which is going to be used by an ``has_error`` call to determine
```
jmartinm: `institutions` -> `institution`

.travis.yml
jacquerie: Debugging. It's part of the last WIP commit. Yesterday I tried to really have coverage on Selenium tests, but couldn't make it work.

docs/tests.rst
```diff
@@ -23,10 +23,95 @@
...
+
+ArsenicResponse
+~~~~~~~~~~~~~~~
+
+As mentioned above, an ``ArsenicResponse`` wraps a closure that is going to be
+used by an ``has_error`` call to determine if the action executed
+successfully.
+
+
 How to Run Selenium Tests
```
jmartinm: Here it might be good to mention to first install dependencies ` pip install '.[tests]'`

docs/tests.rst
```diff
@@ -23,10 +23,95 @@
...
 -------------------------
 
-
 Via Docker
```
jmartinm: With the command mentioned below to run tests:

```
SERVER_NAME="http://localhost:5000" py.test --driver Firefox --html=selenium-report.html tests/acceptance
```

on Mac OS I get:

```
WebDriverException: Message: 'geckodriver' executable needs to be in PATH.
```

jmartinm: To make it work I had to install it:

```
❯ brew install geckodriver
```

docs/tests.rst
jacquerie: TBH this entire section should be rewritten.

inspirehep/bat/arsenic.py
```diff
@@ -0,0 +1,80 @@
...
+class Arsenic(object):
+    _instance = None
+
+    def __new__(cls, *args):
+        if cls._instance is None:
+            cls._instance = object.__new__(cls, *args)
+        return cls._instance
+
+    def __init__(self, *args):
+        if len(args) != 0:
```
jmartinm: ` if args:`

jacquerie: Whoops! I told @rikirenz and then forgot to fix it ._.

docs/tests.rst
jmartinm: There is also no mention of what version of Firefox is expected. If I run the tests locally on Mac OS with Firefox `48.0.2` and geckodriver `0.11.1` I get all errors like:

```
NoSuchElementException: Message: Unable to locate element: [id="skipImportData"]
NoSuchElementException: Message: Unable to locate element: [id="arxiv_id"]
 NoSuchElementException: Message: Unable to locate element: [id="doi"]
```

@rikirenz

jmartinm: Well I see in `docker-compose.test.yml`

```
image: selenium/standalone-firefox:2.53.1-beryllium
```

Not sure what would be equivalent in Mac OS.

inspirehep/bat/pages/holding_panel_literature_list.py
```diff
@@ -0,0 +1,78 @@
...
+)
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support.ui import WebDriverWait
+
+from ..arsenic import Arsenic, ArsenicResponse
+
+
+def go_to():
+    Arsenic().get(os.environ['SERVER_NAME'] + '/holdingpen/list/?page=1&size=10&source=submission&workflow_name=HEP&status=HALTED')
```
jmartinm: I am not sure about having to rely on `SERVER_NAME` - see https://github.com/inspirehep/inspire-next/issues/1402

Can it work just with:

```
Arsenic().get('/holdingpen/list/?page=1&size=10&source=submission&workflow_name=HEP&status=HALTED')
```
?

tests/acceptance/test_author_creation.py
```diff
@@ -0,0 +1,174 @@
...
+    'advisors-degree_type': 'PhD',
+    'comments': 'Some comments about the author',
+    'subject-0': 'ACC-PHYS',
+    'subject-1': 'ASTRO-PH'
+}
+
+
+def test_institutions_typehead(login):
+    create_author.go_to()
+    assert create_author.write_institution('cer', 'CERN').has_error()
```
jmartinm: By just looking at the assertion why would it have an error?

tests/acceptance/test_author_creation.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def test_experiments_typehead(login):
+    create_author.go_to()
+    assert create_author.write_experiment('atl', 'ATLAS').has_error()
```
jmartinm: same as above.

tests/acceptance/test_author_creation.py
```diff
@@ -0,0 +1,174 @@
...
+        'reserach-field': 'This field is required.'
+    }
+
+    create_author.go_to()
+    assert create_author.submit_empty_form(expected_data).has_error()
+
+
+def test_submit_author(login):
+    create_author.go_to()
+    assert create_author.submit_author(input_author_data).has_error()
```
jmartinm: So in all of these `has_error()` is `True` but still they continue to reach holding pen?

jmartinm:
Besides the comments, looks good overall. @rikirenz maybe you can have a look and clarify the questions.

inspirehep/bat/pages/holding_panel_literature_list.py
jacquerie: Note that this is not an actual `SERVER_NAME`, but it's just a convoluted way of passing a constant string: https://github.com/inspirehep/inspire-next/blob/19b6cf28d3d1e59b650518134af96c0f73b8dae4/docker-compose.test.yml#L105.

So yeah, this logic shouldn't be necessary.

tests/acceptance/test_author_creation.py
jacquerie: Yeah, it's confusing. How that works is that `has_error` calls this callback closure: https://github.com/inspirehep/inspire-next/pull/1730/files#diff-ee551b72b52beb17f7613f5956a5d804R82 That closure returns `True` if, after typing `cer` in a certain input, `CERN` is suggested, so the assertion passes.

But yeah, this is badly named and designed.

tests/acceptance/test_author_creation.py
jacquerie: Hopefully https://github.com/inspirehep/inspire-next/pull/1730#discussion_r87990249 clears this confusion.

docs/tests.rst
rikirenz: This error is given because the version of firefox is new. We should update the code in order to use the gekodriver and verify that all the tests work. This in my opinion should be done in a new PR.

tests/acceptance/test_author_creation.py
rikirenz: Guys we should move the assertion inside the lambda function/method that is executed in the has_error(). In that case the has_error() it would make more sense.

It can be refactorized in order to create a more readable code. ANyway I do not thing that is "badly designed"

rikirenz: We can move also the error messages inside the tests and do directly the assertion inside the tests. If we do this it means that it will not necessary use the ArsenicResponse object. 

Let me know if you have other solution

jacquerie: What I would really like to read is a test like this:
```
When I go to the "Create Author" page
And I type "cer" in the "institution" input
Then I should see "CERN" in the suggestion box
```
A function called `has_error`, even with an assertion inside, is still confusing to read, because there's no error here!

rikirenz: Yes, I think you are right. If you wanna I can help you with the refactoring. We can write directly the assertion in the test:

```
create_author.go_to()
create_author.write_institution('cer')
assert 'CERN' in create_author.get_value_from_institution_field()
```
or
```
create_author.go_to()
assert 'CERN' in create_author.write_institution('cer')
```
or
start to use [behave](http://pythonhosted.org/behave/)

or 

if you have other ideas please let me know

tests/acceptance/test_conftest.py
```diff
@@ -22,17 +22,8 @@
```
rikirenz: Why the filename is `test_conftest` rather than `test_login` or `test_navigation_page` ?

jacquerie: Because this is in fact testing that the `login` fixture did its job, and nothing else.

inspirehep/bat/pages/holding_panel_literature_list.py
rikirenz: I have tried to remove the string but it does not work. @jmartinm What should we do with this?

jacquerie:
@kaplun this is the one to work on!

jacquerie:
BTW to get coverage from `acceptance` I don't think the current approach has ever worked. We should try http://coverage.readthedocs.io/en/latest/subprocess.html, which involves writing a `sitecustomize.py` file in the pyhon folder.

kaplun:
I don't understand why, though, since in my execution with Honcho coverage was produced.

kaplun:
We agree that this can already go into like this and will be better refactored in a second moment.

jacquerie:
We have a failure on `_submit_doi_id`: https://travis-ci.org/inspirehep/inspire-next/jobs/179262041#L1666

jmartinm:
I believe the problem here is:

```bash
$ grep '<controlfield tag="001">' demo-records-acceptance.xml                                                              
  <controlfield tag="001">902725</controlfield>
  <controlfield tag="001">1373300</controlfield>
  <controlfield tag="001">1373746</controlfield>
  <controlfield tag="001">1374216</controlfield>
  <controlfield tag="001">1407482</controlfield>
  <controlfield tag="001">1374216</controlfield>
  <controlfield tag="001">903112</controlfield>
  <controlfield tag="001">1208304</controlfield>
  <controlfield tag="001">1214548</controlfield>
  <controlfield tag="001">1303903</controlfield>
  <controlfield tag="001">1245372</controlfield>
  <controlfield tag="001">1247304</controlfield>
  <controlfield tag="001">984446</controlfield>
  <controlfield tag="001">1108541</controlfield>
```

There is a duplicated record (1374216) that is causing the conflict in ElasticSearch. @jacquerie @rikirenz

jacquerie:
Ah, good catch. How come this wasn't failing before?!

rikirenz:
> We have a failure on _submit_doi_id: https://travis-ci.org/inspirehep/inspire-next/jobs/179262041#L1666

If we are in a rush @jacquerie I think for that test we can put the `@pytest.mark.xfail(reason='link button')` .
I noticed that with link button selenium sometimes it does not work. I have not found a clean solution for that case.

jmartinm:
@jacquerie did you try to remove that rogue record from the demo records and resubmit? We can see if the build works.

docs/tests.rst
jacquerie: It should, but the documentation build does not know about the code, so it cannot point to it. Best I can do is adding a `:file:` directive, which is a little more semantical but does not point where it should.

jacquerie:
@rikirenz This is now completely broken: https://github.com/inspirehep/inspire-next/pull/1730

inspirehep/bat/pages/holding_panel_literature_list.py
jacquerie: The container that is running that code is different from the one that is serving the application (which is `web:5000`), so `Arsenic().get('/foo/bar')` can't possibly work, you will have to do `Arsenic().get('web:5000/foo/bar')`. What I was saying is that there is no reason to pass the constant string `web:5000` around as a shell variable.

inspirehep/wsgi_with_coverage.py
```diff
@@ -44,3 +48,80 @@ def save_coverage():
...
+
+del app.url_map._rules_by_endpoint['inspirehep_literature_suggest.validate']
+del app.url_map._rules_by_endpoint['inspirehep_authors_holdingpen.validate']
+del app.url_map._rules_by_endpoint['_arxiv.search']
+del app.url_map._rules_by_endpoint['_doi.search']
+
+del app.view_functions['inspirehep_literature_suggest.validate']
+del app.view_functions['inspirehep_authors_holdingpen.validate']
+del app.view_functions['_arxiv.search']
+del app.view_functions['_doi.search']
```
jacquerie: I'm reminded of this: https://www.youtube.com/watch?v=BOpm1HAWY8c#t=4m24s

inspirehep/wsgi_with_coverage.py
```diff
@@ -44,3 +48,80 @@ def save_coverage():
...
+        if request.json['arxiv_id'] in ('1001.4538', 'hep-th/9711200'):
+            return '{"messages":{"arxiv_id":{}}}'
+        if request.json['arxiv_id'] == 'hep-th.9711200':
+            return '{"messages":{"arxiv_id":{"messages":["The provided ArXiv ID is invalid - it should look similar to \'hep-th/9711200\' or \'1207.7235\'."],"state":"error"}}}'
+
+    if 'doi' in request.json:
+        if request.json['doi'] in ('10.1086/305772', 'doi:10.1086/305772'):
+            return '{"messages":{"doi":{}}}'
+        if request.json['doi'] == 'dummy:10.1086/305772':
+            return '{"messages":{"doi":{"messages":["The provided DOI is invalid - it should look similar to \'10.1086/305772\'."],"state":"error"}}}'
```
jacquerie: All these guys should become `jsonify` calls on dictionaries for readability.

inspirehep/wsgi_with_coverage.py
```diff
@@ -44,3 +48,80 @@ def save_coverage():
...
+    return literature_validate()
+
+
+@app.route('/submit/author/validate', endpoint='inspirehep_authors_holdingpen.validate', methods=['POST'])
+def mock_literature_validate():
+    """"Mock the ORCID validation"""
+    if request.json.get('orcid') == 'wrong.ORCID':
+        return '{"messages":{"orcid":{"messages":["A valid ORCID iD consists of 16 digits separated by dashes."],"state":"error"}}}'
+    elif request.json.get('orcid') == '1111-1111-1111-1111':
+        return '{"messages":{"orcid":{}}}'
```
jacquerie: Same as above.

inspirehep/wsgi_with_coverage.py
```diff
@@ -44,3 +48,80 @@ def save_coverage():
...
...ation': [], 'family': 'Davis', 'given': 'Marc'}], 'container-title': ['The Astrophysical Journal'], 'content-domain': {'crossmark-restriction': False, 'domain': []}, 'created': {'date-parts': [[2002, 7, 26]], 'date-time': '2002-07-26T18:50:42Z', 'timestamp': 1027709442000}, 'deposited': {'date-parts': [[2011, 8, 23]], 'date-time': '2011-08-23T14:08:25Z', 'timestamp': 1314108505000}, 'indexed': {'date-parts': [[2016, 9, 26]], 'date-time': '2016-09-26T15:34:12Z', 'timestamp': 1474904052331}, 'issue': '2', 'issued': {'date-parts': [[1998, 6, 20]]}, 'member': 'http://id.crossref.org/member/266', 'original-title': [], 'page': '525-553', 'prefix': 'http://id.crossref.org/prefix/10.1088', 'published-print': {'date-parts': [[1998, 6, 20]]}, 'publisher': 'IOP Publishing', 'reference-count': 83, 'score': 1.0, 'short-container-title': ['ApJ'], 'short-title': [], 'source': 'CrossRef', 'subject': ['Space and Planetary Science', 'Astronomy and Astrophysics'], 'subtitle': [], 'title': ['Maps of Dust Infrared Emission for Use in Estimation of Reddening and Cosmic Microwave Background Radiation Foregrounds'], 'type': 'journal-article', 'volume': '500', }, 'source': 'crossref', 'status': 'success'})
```
jacquerie: While these two should be extracted in a fixture.

inspirehep/wsgi_with_coverage.py
```diff
@@ -44,3 +48,80 @@ def save_coverage():
...
... excess of events is observed above the expected background, with a local significance of 5.0 standard deviations, at a mass near 125 GeV, signalling the production of a new particle. The expected significance for a standard model Higgs boson of that mass is 5.8 standard deviations. The excess is most significant in the two decay modes with the best mass resolution, gamma gamma and ZZ; a fit to these signals gives a mass of 125.3 +/- 0.4 (stat.) +/- 0.5 (syst.) GeV. The decay to two photons indicates that the new particle is a boson with spin different from one. ', 'authors': [{'author': [{'keyname': 'The CMS Collaboration'}]}], 'categories': 'hep-ex', 'comments': 'Submitted to Phys. Lett. B', 'created': '2012-07-31', 'doi': '10.1016/j.physletb.2012.08.021', 'id': '1207.7235', 'journal-ref': 'Phys. Lett. B 716 (2012) 30', 'license': 'http://arxiv.org/licenses/nonexclusive-distrib/1.0/', 'report-no': 'CMS-HIG-12-028; CERN-PH-EP-2012-220', 'request': 'http://export.arxiv.org/oai2', 'responseDate': '2016-10-17T15:22:25Z', 'title': 'Observation of a new boson at a mass of 125 GeV with the CMS experiment\n at the LHC', 'updated': '2013-01-28', }, 'source': 'arxiv', 'status': 'success'})
```
jacquerie: Same as above (extracted in a fixture).

PR #1732, opened 2016-11-11
global: inspire-schemas 2.0.0 upgrade

jmartinm:
* Updates inspire-schemas package to 2.0.0.

* Updates usage of control_number field to make sure it is always used
  as an integer.

### DIFF
--- a/inspirehep/modules/pidstore/minters.py
+++ b/inspirehep/modules/pidstore/minters.py
@@ -41,5 +41,5 @@ def inspire_recid_minter(record_uuid, data):
     if 'control_number' in data:
         args['pid_value'] = data['control_number']
     provider = InspireRecordIdProvider.create(**args)
-    data['control_number'] = str(provider.pid.pid_value)
+    data['control_number'] = provider.pid.pid_value
     return provider.pid
--- a/inspirehep/utils/bibtex.py
+++ b/inspirehep/utils/bibtex.py
@@ -259,6 +259,7 @@ class Bibtex(Export):
         key = ''
         if not self._get_citation_key():
             key = self.record['control_number']
+
         return key
 
     def _get_author(self):
--- a/inspirehep/utils/latex.py
+++ b/inspirehep/utils/latex.py
@@ -283,7 +283,7 @@ class Latex(Export):
                     pages = field.get('page_start') or field['artid']
                 try:
                     if journal and (volume != '' or pages != ''):
-                        recid = self.record['control_number', '']
+                        recid = self.record['control_number']
                         record = get_es_record('jou', recid)
                         coden = ','.join(
                             [record['coden'][0], volume, pages])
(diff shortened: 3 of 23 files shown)


inspirehep/modules/pidstore/minters.py
```diff
@@ -41,5 +41,5 @@ def inspire_recid_minter(record_uuid, data):
     if 'control_number' in data:
         args['pid_value'] = data['control_number']
     provider = InspireRecordIdProvider.create(**args)
-    data['control_number'] = str(provider.pid.pid_value)
+    data['control_number'] = provider.pid.pid_value
```
jacquerie: Are you sure that `pid.pid_value` is an `int`? I remember otherwise...

inspirehep/utils/bibtex.py
```diff
@@ -257,8 +257,10 @@ def _is_number(self, s):
     def _get_key(self):
         """Returns record key"""
         key = ''
+        # import pytest; pytest.set_trace()
```
jacquerie: Leftover.

inspirehep/utils/bibtex.py
```diff
@@ -257,8 +257,10 @@ def _is_number(self, s):
...
         if not self._get_citation_key():
             key = self.record['control_number']
+
```
jacquerie: Leftover, but I like it better like this : )

inspirehep/utils/latex.py
```diff
@@ -283,7 +283,7 @@ def _get_pubnote(self):
                     pages = field.get('page_start') or field['artid']
                 try:
                     if journal and (volume != '' or pages != ''):
-                        recid = self.record['control_number', '']
```
jacquerie: ...How could this even work? O_o

inspirehep/modules/pidstore/minters.py
jmartinm: Seems to be:

```python
In [1]: from inspirehep.modules.pidstore.providers import InspireRecordIdProvider

In [2]: import uuid

In [3]: args = dict(
   ...:         object_type='rec',
   ...:         object_uuid=uuid.uuid4(),
   ...:         pid_type='recid'
   ...:     )

In [4]:

In [4]: args['pid_value'] = 123

In [5]: provider = InspireRecordIdProvider.create(**args)

In [6]: provider.pid.pid_value
Out[6]: 123

In [7]: type(provider.pid.pid_value)
Out[7]: int
```

jacquerie: Ah-ah! With @rikirenz and @david-caro we just discovered that, if you omit line 4, the `pid_value` that is minted is actually a string.

jmartinm: which seems to be our fault :) https://github.com/inspirehep/inspire-next/pull/1744#issuecomment-260725528

PR #1736, opened 2016-11-14
Travis: add stable test suit

rikirenz:


.travis.yml
```diff
@@ -42,6 +42,21 @@ env:
...
+    - SUITE=unit_unstable
+
+matrix:
+  allow_failures:
+    - SUITE=integration_unstable
+    - SUITE=workflows_unstable
+    - SUITE=disambiguation_unstable
+    - SUITE=acceptance_unstable
+    - SUITE=pep8_unstable
+    - SUITE=unit_unstable
```
kaplun: @david-caro suggests that we take only integration for the time being.

rikirenz:
The test are failing because we did not merge the changes in the inspire-docker project (https://github.com/inspirehep/inspire-docker/pull/25)

rikirenz:
We have opened an issue on the ISBNID project (https://github.com/nekobcn/isbnid/issues/4) because the installation is having a problem with that specific package (https://travis-ci.org/inspirehep/inspire-next/jobs/183907796).

The repo is not regularly updated. Probably we need a workaround to make it works.

kaplun:
@rikirenz in case, we can always fork that package internally and ship it from within our repos.

rikirenz:
@kaplun I think is the right way to proceed. 
Do you want to set a deadline, to see if the nekobcn/isbnid#4 will be resolved?

rikirenz:
Depends on https://github.com/inveniosoftware/idutils/pull/29

rikirenz:
@kaplun @david-caro I think here we are!  :clinking_glasses:

docker-compose.deps.yml
```diff
@@ -27,7 +27,14 @@ services:
     extends:
       file: services.yml
       service: base
-    command: pip install --find-links /pip-cache -r requirements.txt --pre -e .[tests,crawler] gunicorn --exists-action i
+    command: bash -c "pip install -r ../deps/deps.txt && pip install --find-links /pip-cache -r requirements.txt --pre -e .[tests,crawler] gunicorn --exists-action i"
```
kaplun: @rikirenz is this `../deps/deps.txt` indeed created in the Docker?

rikirenz: Yes. We store all the dependencies in that file!

jacquerie:
> The only drawback is adding another element to the build matrix.

This is a significant drawback. This is making ALL builds slower in exchange for, when it happens, not having to deal immediately with a dependency-related error.

rikirenz:
I agree. It was nice understand a bit of this but we did not experience too many dependency errors in the last months.

jacquerie:
> I agree. It was nice understand a bit of this but we did not experience too many dependency errors in the last months.

Yeah, we had a nasty streak of those in a limited amount of time, then at most one of those per month...

PR #1739, opened 2016-11-14
templates: show cataloger tools to catalogers only

jmartinm:

### DIFF
--- a/tests/unit/theme/test_theme_jinja2filters.py
+++ b/tests/unit/theme/test_theme_jinja2filters.py
@@ -1227,3 +1227,29 @@ def test_weblinks_when_description_is_not_in_the_kb():
 
 def test_weblinks_when_description_is_a_falsy_value():
     assert weblinks('') == 'Link to fulltext'
+
+
+def test_is_cataloger():
+    class MockUser(object):
+        def __init__(self, user_roles):
+            self.user_roles = user_roles
+
+        @property
+        def roles(self):
+            class MockRole():
+                def __init__(self, role_name):
+                    self.role_name = role_name
+
+                @property
+                def name(self):
+                    return self.role_name
+
+            return [MockRole(name) for name in self.user_roles]
+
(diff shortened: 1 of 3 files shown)


jmartinm:
* [ ] Create test for `is_cataloger` filter.

jmartinm:
Ready for review.

tests/unit/theme/test_theme_jinja2filters.py
```diff
@@ -1227,3 +1227,29 @@ def test_weblinks_when_description_is_not_in_the_kb():
...
+
+            return [MockRole(name) for name in self.user_roles]
+
+    superadmin_user = MockUser(user_roles=['superuser'])
+    cataloger_user = MockUser(user_roles=['cataloger'])
+    no_role_user = MockUser(user_roles=[])
+
+    assert is_cataloger(superadmin_user) == True
+    assert is_cataloger(cataloger_user) == True
+    assert is_cataloger(no_role_user) == False
```
jacquerie: You can just:
```
assert is_cataloger(superadmin_user)
assert is_cataloger(cataloger_user)
assert not is_cataloger(no_role_user)
```

jacquerie:
The only problem I see with this PR is that some catalogers are already using the tool, but none of them has an account, so we should give them one before deploying!

kaplun:
In principle all catalogers have a local account on INSPIRE legacy. We should transfer their account (is legacy and Invenio 3 still using the same salting method)? Also on QA we can keep enabled local=1.

jmartinm:
Yes, most of the catalogers already have a local account, and if not we can create new ones. Note that with the direct link also it will work as usual, this is just hiding the link from the dropdown.

>is legacy and Invenio 3 still using the same salting method

Have not tried in the past porting a legacy account into Invenio3 but they seem compatible checking the user migration task (https://github.com/inveniosoftware/invenio-migrator/blob/master/invenio_migrator/tasks/users.py#L68-L86)

PR #1742, opened 2016-11-15
tasks: add control for duplicate notes

rikirenz:
During the accepting phase for an harvest article, the worker raise an error for a duplicate dictionary. This fix should check if the dictionary that is going to be added is already present and avoid the error.

### DIFF
--- a/inspirehep/modules/workflows/tasks/submission.py
+++ b/inspirehep/modules/workflows/tasks/submission.py
@@ -306,13 +306,17 @@ def send_robotupload(url=None,
 
 def add_note_entry(obj, eng):
     """Add note entry to metadata on approval."""
+    def _has_note(reference_note, notes):
+        return any(reference_note == note for note in notes)
+
     entry = {'value': '*Temporary entry*'} if obj.extra_data.get("core") \
         else {'value': '*Brief entry*'}
     if obj.data.get('public_notes') is None or \
             not isinstance(obj.data.get("public_notes"), list):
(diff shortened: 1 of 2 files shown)


inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -306,13 +306,20 @@ def _send_robotupload(obj, eng):
 
 def add_note_entry(obj, eng):
     """Add note entry to metadata on approval."""
+    def _has_note(notes, reference_note):
+        for note in notes:
+            if note == reference_note:
+                return True
+        return False
```
jacquerie:
```python
def _has_note(reference_note, notes):
    return any(reference_note == note for note in notes)
```

jacquerie:
Also missing tests. Have a look [here](https://github.com/inspirehep/inspire-next/blob/19b6cf28d3d1e59b650518134af96c0f73b8dae4/tests/unit/workflows/test_workflows_tasks_beard.py) to see how tasks can be unit tested using the appropriate test doubles.

jacquerie:
> During the accepting phase for an harvest article, the worker raise an error for a duplicate dictionary. 

Since the error comes from real life, use (a reduced version of) that article in your unit test.

rikirenz:
Ready for the review

jacquerie:
LGTM!

jacquerie:
Whoops, it was missing your signature. No biggie (we don't really use them...), but take care next time.

PR #1744, opened 2016-11-15
pidstore: fix pid type, cast from string to int

rikirenz:
cast `pidstore` value, because when we accept an article in the holdingpen the validation process is complaining about the type of this value. It requires an int but a we were passing a string.

### DIFF
--- a/inspirehep/modules/pidstore/providers.py
+++ b/inspirehep/modules/pidstore/providers.py
@@ -49,7 +49,7 @@ def _get_next_pid_from_legacy():
     url = current_app.config.get('LEGACY_PID_PROVIDER')
     next_pid = requests.get(url, headers=headers).json()
 
-    return str(next_pid)
+    return next_pid
 
 
 class InspireRecordIdProvider(BaseProvider):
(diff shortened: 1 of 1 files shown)


jmartinm:
I think the problem might be https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/pidstore/providers.py#L78 actually.

That is our `id` generator and it should be returning `int` directly. That would avoid the problem.

jmartinm:
And you could also update https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/pidstore/providers.py#L52

rikirenz:
@jmartinm I have made the changes but the test are failing. 
The line that is failing is the following:

`list(_es.create(ignore=[400]))`

Do you have an idea? I think is the problem is in the invenio-search package because it needs a string. But I am not sure.

jmartinm:
It should be ok once the new Docker image is submitted to Dockerhub.

jmartinm:
I restarted the build, and it now passes :) @rikirenz if you rebase it, we can merge.

PR #1748, opened 2016-11-16
tests: unit tests for workflows/tasks/actions

jacquerie:
- [x] `mark`
- [x] `is_marked`
- [x] `is_record_accepted`
- [x] `shall_halt_workflow`
- [x] `in_production_mode`
- [x] `add_core`
- [x] `halt_record`
- [x] `update_note`
- [x] `reject_record`
- [x] `is_record_relevant`
- [x] `is_experimental_paper`
- [x] `is_arxiv_paper`
- [x] `is_submission`
- [x] `prepare_update_payload`

### DIFF
--- a/inspirehep/modules/workflows/tasks/__init__.py
+++ b/inspirehep/modules/workflows/tasks/__init__.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this licence, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Workflows tasks."""
+
+from __future__ import absolute_import, division, print_function
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -28,6 +28,8 @@ from functools import wraps
 
 from flask import current_app
 
+from inspirehep.modules.workflows.utils import log_workflows_action
+
 from inspirehep.utils.arxiv import get_clean_arXiv_id
 from inspirehep.utils.record import get_value
 
@@ -103,8 +105,6 @@ def reject_record(message):
     """Reject record with message."""
     @wraps(reject_record)
     def _reject_record(obj, *args, **kwargs):
-        from inspirehep.modules.workflows.utils import log_workflows_action
-
         prediction_results = obj.extra_data.get("relevance_prediction")
         log_workflows_action(
             action="reject_record",
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of INSPIRE.
-# Copyright (C) 2014, 2015 CERN.
+# Copyright (C) 2014, 2015, 2016 CERN.
 #
 # INSPIRE is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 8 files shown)


kaplun:
@jacquerie great initiative. One important thing: if you see anything that is not obvious or doesn't feel right, always check with me, before writing a test that certify possibly the wrong behavior.
It's the right moment to do a thorough review of the workflow.

inspirehep/utils/arxiv.py
```diff
@@ -24,17 +24,11 @@
...
 def get_clean_arXiv_id(record):
     """Return the arXiv identifier from given record."""
-    arxiv_id = record.get("arxiv_id")
-    if not arxiv_id:
-        arxiv_eprints = record.get('arxiv_eprints', [])
-        for element in arxiv_eprints:
-            if element.get("value", ""):
-                arxiv_id = element.get("value", "")
-
+    arxiv_id = get_value(record, 'arxiv_eprints.value[-1]')
```
kaplun: Use `[0]` rather than `[-1]` as reported in #1751

tests/unit/utils/test_utils_arxiv.py
```diff
@@ -3,63 +3,80 @@
...
+
+def test_get_clean_arXiv_id_from_arxiv_eprints_selects_last():
+    record = {
+        'arxiv_eprints': [
+            {'value': 'oai:arXiv.org:0801.4782'},
+            {'value': 'oai:arXiv.org:0805.1410'},
+        ],
+    }
+
+    expected = '0805.1410'
```
kaplun: So here would be the first to be returned.

jacquerie:
> One important thing: if you see anything that is not obvious or doesn't feel right, always check with me, before writing a test that certify possibly the wrong behavior.

Naturally! An example of that was our discussion on file descriptors yesterday in chat.

jacquerie:
BTW you don't have to review this yet, because I'm still discovering things worth fixing in other parts of the code, so commits are going to be duplicated between here and other PRs that I'm spinning off of here.

jacquerie:
`is_arxiv_paper` does not handle this corner case (`arxiv_eprints` with `value`s but without `categories`):
```python
In [1]: class StubObj(object):
   ...:     def __init__(self, data):
   ...:         self.data = data
   ...:

In [2]: from inspirehep.modules.workflows.tasks.actions import is_arxiv_paper

In [3]: obj = StubObj({'arxiv_eprints': [{'value': 'oai:arXiv.org:0801.4782'}]})

In [4]: is_arxiv_paper(obj)
IndexError                                Traceback (most recent call last)
<ipython-input-4-0f3287181dbc> in <module>()
----> 1 is_arxiv_paper(obj)

/Users/jacquerie/.virtualenvs/inspire/src/inspire-next/inspirehep/modules/workflows/tasks/actions.pyc in is_arxiv_paper(obj, *args, **kwargs)
    159 def is_arxiv_paper(obj, *args, **kwargs):
    160     """Check if the record is from arXiv."""
--> 161     return bool(get_value(obj.data, "arxiv_eprints.categories", [[]])[0]) or \
    162         get_clean_arXiv_id(obj.data)
    163

IndexError: list index out of range
```

jacquerie:
`is_experimental_paper` is completely broken:
```
In [5]: from inspirehep.utils.record import get_value

In [6]: obj = StubObj({'arxiv_eprints': [{'categories': 'hep-ex'}]})

In [7]: list(get_value(obj.data, "arxiv_eprints.categories", [[]])[0])
Out[7]: ['h', 'e', 'p', '-', 'e', 'x']
```

kaplun:
Please send in also the corresponding fixes for the bugs you are finding. Also re: arXiv category, I believe there shouldn't been such case where an arXiv document comes without categories from arXiv.

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -0,0 +1,125 @@
...
+@pytest.fixture
+def another_arxiv_pdf():
+    return pkg_resources.resource_string(
+        __name__, os.path.join('fixtures', '0701031v2.pdf'))
+
+
+class StubObj(object):
+    def __init__(self, data, extra_data, files):
+        self.data = data
+        self.extra_data = extra_data
```
rikirenz: `extra_data` is always empty. Why?

jacquerie: Mh, you're right, tasks in this file don't use `extra_data` at all.

jacquerie: For now, every test file has its own small mocks/stubs/dummies. The better approach would be to write a unique mock that covers all use cases and put it in a `conftest.py` in the same folder.

jacquerie:
> Also re: arXiv category, I believe there shouldn't been such case where an arXiv document comes without categories from arXiv.

But there _could_ be, because we're not enforcing it from the schema. The fact is, the idiom `get_value(record, "foo.bar", [[]])[0]` is incorrect, and it's the common cause of both bugs (but I agree that they differ in severity). So this is going to get fixed anyway : )

kaplun:
The issue is that we could enforce it in the schema, but it's simply that given an arXiv eprint we don't know automatically its categories. Hence one might create a record with just the eprint, and resolve the categories at a later stage. However, upon harvesting we surely know them.

jacquerie:
I see. Then I'm not going to assert that in a test.

kaplun:
I mean, I think it's great to assert in the workflow test that categories *are* there. (since this would make the test to be documentation/specification for the workflow). Instead no need, IMHO, to test for the case, within the workflow, where arXiv doesn't give us categories.

jacquerie:
Minor bug: `is_record_relevant` does not match its signature, because it passes its parameters to `is_submission` which has a fixed number of parameters.

Travis: https://travis-ci.org/inspirehep/inspire-next/jobs/177641066#L1518

kaplun:
@jacquerie what do you do for:
> `is_experimental_paper` is completely broken:

Besides I understand there are no other major mis-behavior you have found right?

jacquerie:
> @jacquerie what do you do for:
>
> is_experimental_paper is completely broken:
>
> Besides I understand there are no other major mis-behavior you have found right?

Oh right! I was wrong about that. Note that I was missing `[]` around `hep-ex`, which is why I was getting that weird result.

PR #1749, opened 2016-11-16
editor: experiment suggester

harunurhan:
* Adds a suggester for accelerator_experiments which suggest experiments
and returns results with payload that is populated with $ref of the
accelerator_experiment.

I am not sure if I need to add tests for this.

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -738,6 +738,11 @@ RECORDS_REST_ENDPOINTS = dict(
                 ':json_v1_search'
             ),
         },
+        suggesters=dict(
+            experiment=dict(completion=dict(
+                field='experiment_suggest'
+            ))
+        ),
         list_route='/experiments/',
         item_route='/experiments/<pid(exp):pid_value>',
         default_media_type='application/json',
--- a/inspirehep/modules/records/mappings/records/hep.json
+++ b/inspirehep/modules/records/mappings/records/hep.json
@@ -146,9 +146,6 @@
                 "citation_count": {
                     "type": "integer"
                 },
-                "control_number": {
-                    "type": "integer"
-                },
                 "cnum": {
                     "index": "not_analyzed",
                     "type": "string"
@@ -187,6 +184,9 @@
                         }
                     },
                     "type": "object"
+                },
+                "control_number": {
+                    "type": "integer"
                 },
                 "date": {
                     "format": "yyyy||yyyy-MM||yyyy-MM-dd",
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -54,6 +54,7 @@ def enhance_record(sender, json, *args, **kwargs):
     match_valid_experiments(sender, json, *args, **kwargs)
     dates_validator(sender, json, *args, **kwargs)
     add_recids_and_validate(sender, json, *args, **kwargs)
+    populate_experiment_suggest(sender, json, *args, **kwargs)
     after_record_enhanced.send(json)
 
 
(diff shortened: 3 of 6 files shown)


inspirehep/config.py
```diff
@@ -1140,7 +1145,8 @@ def _(x):
 
 # OAuthclient
 # ===========
-orcid.REMOTE_APP['params']['request_token_params'] = {'scope': '/orcid-profile/read-limited /activities/update /orcid-bio/update'}
+orcid.REMOTE_APP['params']['request_token_params'] = {
+    'scope': '/orcid-profile/read-limited /activities/update /orcid-bio/update'}
```
jacquerie: Unrelated, will split.

inspirehep/modules/records/receivers.py
```diff
@@ -371,3 +372,17 @@ def check_if_record_is_going_to_be_deleted(sender, *args, **kwargs):
...
+    accelerator_experiments = json.get("accelerator_experiments")
+    if accelerator_experiments:
+        for accelerator_experiment in accelerator_experiments:
+            experiment = accelerator_experiment.get("experiment")
+            record = accelerator_experiment.get("record")
+            if experiment and record:
+                accelerator_experiment.update({"experiment_suggest": {
+                    "input": experiment,
+                    "payload": {"$ref": record.get("$ref")}
+                }})
```
jacquerie: Suggestion: every time you write this kind of JSON-driven logic try to see if [`get_value`](https://github.com/inspirehep/inspire-next/blob/2605d9b211d11c58763ba1ce4256f36a7790c01f/inspirehep/utils/record.py#L61-L110) can be used instead.

jacquerie: BTW this methods looks perfectly testable: you just assert that the right side effects on `json` have happened. You should be able to find examples in `tests/unit/records/test_record_receivers.py` for this.

inspirehep/modules/records/mappings/records/hep.json
```diff
@@ -527,4 +533,4 @@
         "index.percolator.map_unmapped_fields_as_string": true,
         "index.query.default_field": "global_fulltext"
     }
-}
+}
```
jacquerie: Sign that you didn't run `scripts/prettifyjson` : )

jacquerie:
> I am not sure if I need to add tests for this.

Acceptance tests for the entire suggester doing its job will probably have to wait #1730, but unit tests for the receiver can absolutely be written.

jmartinm:
Besides Jacopo's comments, LGTM.

inspirehep/config.py
harunurhan: 👍  it was my formatter

inspirehep/modules/records/receivers.py
```diff
@@ -371,3 +372,19 @@ def check_if_record_is_going_to_be_deleted(sender, *args, **kwargs):
...
+    # FIXME: Use a dedicated method when #1355 will be resolved.
+    if "experiments.json" in json.get("$schema"):
+        experiment_names = get_value(json, "experiment_names.title")
+        title_variants = force_force_list(
+            get_value(json, "title_variants.title")
+        )
+        json.update({"experiment_suggest": {
+            "input": experiment_names + title_variants,
+            "output": experiment_names[0],
+            "payload": {"$ref": get_value(json, "self.$ref")},
```
harunurhan: payload might change later today!

jacquerie: Did this already happen?

jmartinm: In the end this is good enough, we don't need extra `payload`.

@harunurhan can you rebase and we can merge? Thanks

PR #1751, opened 2016-11-17
utils: refactor arxiv utils

jacquerie:
* Simplifies `get_clean_arXiv_id`
* Removes data-only fixtures in tests
* Cleans up licenses and imports

Spin off of #1748

### DIFF
--- a/inspirehep/utils/arxiv.py
+++ b/inspirehep/utils/arxiv.py
@@ -24,17 +24,11 @@
 
 from __future__ import absolute_import, division, print_function
 
+from .record import get_value
+
 
 def get_clean_arXiv_id(record):
     """Return the arXiv identifier from given record."""
-    arxiv_id = record.get("arxiv_id")
-    if not arxiv_id:
-        arxiv_eprints = record.get('arxiv_eprints', [])
-        for element in arxiv_eprints:
-            if element.get("value", ""):
-                arxiv_id = element.get("value", "")
-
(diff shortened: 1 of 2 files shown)


jacquerie:
Ready for review.

kaplun:
LGTM however it underlines the fact that, in case of multiple arXiv eprints the code just take an arbitrary ID between the two. 

Actually if we stick with the implicit convention that within a list the first element is the preferred one, so, please change the `[-1]` into a `[0]`, in the `get_value` call.

inspirehep/utils/arxiv.py
```diff
@@ -24,17 +24,11 @@
...
 def get_clean_arXiv_id(record):
     """Return the arXiv identifier from given record."""
-    arxiv_id = record.get("arxiv_id")
-    if not arxiv_id:
-        arxiv_eprints = record.get('arxiv_eprints', [])
-        for element in arxiv_eprints:
-            if element.get("value", ""):
-                arxiv_id = element.get("value", "")
-
+    arxiv_id = get_value(record, 'arxiv_eprints.value[-1]')
```
kaplun: `[-1]` -> `[0]` as mentioned below.

jacquerie: Note that this changes behaviour, so this could mean more failed tests downstream.

kaplun: Crossifing fingers :)

PR #1752, opened 2016-11-17
setup: use pytest-httpretty

jacquerie:
* Uses `pytest-httpretty` instead of `httpretty_mock` and `httppretty_mock`.
* Fixes some licenses and imports.

Spin off of #1748


jacquerie:
Ready for review.

PR #1753, opened 2016-11-17
migrator: dump user roles

jmartinm:
* Can be applied to Invenio 2.x instance.


jmartinm:
I have tested the dumping on Labs with this PR and looks good.

PR #1757, opened 2016-11-17
tests: add unit tests for utils/helpers

jacquerie:
* Adds unit tests for methods in `inspirehep/utils/helpers.py`.

~~Depends on #1752~~
Spin off of #1748

### DIFF
--- a/tests/unit/utils/test_utils_helpers.py
+++ b/tests/unit/utils/test_utils_helpers.py
@@ -22,7 +22,138 @@
 
 from __future__ import absolute_import, division, print_function
 
-from inspirehep.utils.helpers import force_force_list
+import os
+import pkg_resources
+import tempfile
+
+import httpretty
+import pytest
+
+from inspirehep.utils.helpers import (
+    download_file,
+    download_file_to_record,
+    get_json_for_plots,
+    force_force_list,
+)
+
+
+@pytest.fixture
+def foo_bar_baz():
+    return pkg_resources.resource_string(
+        __name__, os.path.join('fixtures', 'foo-bar-baz'))
+
+
+class StubRecord(object):
+    def __init__(self, files):
+        self.files = files
+
+
+@pytest.mark.httpretty
+def test_download_file(foo_bar_baz):
+    httpretty.register_uri(
+        httpretty.GET, 'http://example.com/foo-bar-baz', body=foo_bar_baz)
+
+    filename = next(tempfile._get_candidate_names())
+
+    expected = filename
+    result = download_file(
+        'http://example.com/foo-bar-baz', output_file=filename)
+
+    assert expected == result
+
+    with open(filename, 'rb') as f:
+        assert 'foo\nbar\nbaz\n' == f.read()
+
+    os.remove(filename)
+
+
+@pytest.mark.httpretty
+def test_download_file_does_not_create_a_file_if_the_request_fails():
+    httpretty.register_uri(
+        httpretty.GET, 'http://example.com/500', body='', status=500)
+
+    filename = next(tempfile._get_candidate_names())
+
+    expected = filename
+    result = download_file(
+        'http://example.com/500', output_file=filename)
+
+    assert expected == result
+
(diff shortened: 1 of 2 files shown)


jacquerie:
Ready for review.

tests/integration/test_pidstore.py
```diff
@@ -24,11 +24,13 @@
 
 import httpretty
 import mock
+import pytest
 
 from inspirehep.modules.pidstore.providers import InspireRecordIdProvider
 
 
-def test_getting_next_recid_from_legacy(httpretty_mock, app):
+@pytest.mark.xfail
```
jacquerie: Whoops. Wrong `mark`.

tests/unit/utils/test_utils_helpers.py
```diff
@@ -20,7 +20,140 @@
...
+    filename = next(tempfile._get_candidate_names())
+
+    expected = filename
+    result = download_file(
+        'http://example.com/500', output_file=filename)
+
+    assert expected == result
+
+    with pytest.raises(IOError):
+        open(filename, 'rb')
```
jacquerie: @kaplun This documents this surprising fact: if the request fails, no file is created, but no exception is thrown. I don't think this is good behavior.

tests/unit/utils/test_utils_helpers.py
```diff
@@ -20,7 +20,140 @@
...
+
+
+@pytest.mark.httpretty
+def test_download_file_raises_if_called_without_output_file():
+    httpretty.register_uri(
+        httpretty.GET, 'http://example.com/foo-bar-baz', body=foo_bar_baz)
+
+    with pytest.raises(IOError):
+        download_file('http://example.com/foo-bar-baz')
```
jacquerie: This instead documents the weird `None` default argument, which I don't think it's a good idea either.

kaplun:
@jacquerie since the immediate objective is to deliver all this code on production, and you seem to have a found quite a source of bugs (or potential source of bugs), please send an appropriate fix as well.

In particular, raising an exception that should rather being captured upstream (e.g. in the workflow, cc @david-caro), and passing around the open socket. Or rather pre-downloading the file locally and only afterwards move-it around.

jacquerie:
These tests are blocked by the fact that I don't know what is the correct behaviour: attaching a string containing the file contents to the `files`, or an **open** file descriptor. We agreed that the former is better, but in some places of the code the latter is expected.

kaplun:
I thought we agreed that the safest thing is actually to download it locally (better on EOS) and ship around the full path to the file?

jacquerie:
> I thought we agreed that the safest thing is actually to download it locally (better on EOS) and ship around the full path to the file?

This is not at all what the users of `.files` expect, so they should be amended accordingly. To be clear, what you're proposing is that `.files` becomes a map of filename -> path on EOS, right?

jacquerie:
> To be clear, what you're proposing is that `.files` becomes a map of filename -> path on EOS, right?

This requires a modification to https://github.com/inveniosoftware-contrib/invenio-workflows-files/blob/22570da9aeb6b8029e2f88a3a5379d28a4f6b9ee/invenio_workflows_files/api.py.

david-caro:
@jacquerie is there an issue I can assign myself to?

jacquerie:
> @jacquerie is there an issue I can assign myself to?

I was just writing it!

jacquerie:
Me and @david-caro  just figured out that the treatment of files is way more busted than what these tests say. Writing the issue will have to wait until we understand a little more what's going on.

kaplun:
For this milestone there minimal viable product is:
* The FFT in the MARCXML that is sent to legacy either contain the URL of the arXiv paper
* Alternatively the FFT refers to a local file that is accessible via EOS by legacy.

Pick up what is less broken

kaplun:
> Writing the issue will have to wait until we understand a little more what's going on.

At least write a stub issue with the right milestone, so that everybody else is aware that someone is working on something that is a blocker.

PR #1745, opened 2016-11-16
tools: authorlist improvements

bittirousku:
* Better buttons.
* Better output box size.
* Always dots in initials.
* Works better with bad input data, e.g. ligatures or unnecessary whitespaces and newlines.

Still requires that there is only one affiliation per line, and affiliation IDs in author names be separated with commas.

PR #1760, opened 2016-11-18
tools: make authorlist work better with bad input

jacquerie:
Closes #1745


jacquerie:
Truth to be told, this won't fix everything. The main problem with `authorlist` is its architecture: we can't possibly work line by line, and have something that works in all cases, because there's just too many corner cases to deal with.

The right way to build this tool is to write a grammar of the things we expect to recognize, a parser that recognizes this grammar, and a walker that produces the right output given the Abstract Syntax Tree (like `invenio-query-parser` does).

PR #1761, opened 2016-11-18
theme: redirect linkedaccounts view

### LINKED ISSUE #1737: Login through ORCID redirects to /account/settings/linkedaccounts/
At the moment on qa.inspirehep.net after logging in through ORCID, a redirect is done to /account/settings/linkedaccounts/ which gives a blank page.

jmartinm:
* /account/settings/linkedaccounts is the page that invenio-oauthclient
  redirects by default after a successful login. INSPIRE does not use
  that page so the user is currently shown a blank page. This commits
  makes that page redirect to the front page which was the behaviour
  in Invenio 2.x. (closes #1737)

### DIFF
--- a/inspirehep/modules/theme/views.py
+++ b/inspirehep/modules/theme/views.py
@@ -647,6 +647,12 @@ def ajax_experiments_people():
     )
 
 
+@blueprint.route('/account/settings/linkedaccounts/', methods=['GET'])
+def linkedaccounts():
+    """."""
+    return redirect('/')
+
+
 #
 # Helpers
 #


inspirehep/modules/theme/views.py
```diff
@@ -646,6 +646,10 @@ def ajax_experiments_people():
         }
     )
 
+@blueprint.route('/account/settings/linkedaccounts/', methods=['GET'])
+def linkedaccounts():
+    """."""
```
jacquerie: Leftover.

inspirehep/modules/theme/views.py
```diff
@@ -646,6 +646,10 @@ def ajax_experiments_people():
         }
     )
 
+@blueprint.route('/account/settings/linkedaccounts/', methods=['GET'])
```
jacquerie: Sorry, shouldn't we change the url `invenio-oauthclient` redirects to, rather than adding another redirect?

inspirehep/modules/theme/views.py
jmartinm: I remember telling to myself I would write something there, did not happen apparently :P

inspirehep/modules/theme/views.py
jmartinm: Definitely.  But 💣 https://github.com/inveniosoftware/invenio-oauthclient/search?utf8=%E2%9C%93&q=invenio_oauthclient_settings.index&type=Code

PR #1764, opened 2016-11-21
Unit test extract journal info

rikirenz:
This PR should add the test for the extract_journal_info function and also fix the bug that the test found.

### DIFF
--- a/inspirehep/modules/refextract/tasks.py
+++ b/inspirehep/modules/refextract/tasks.py
@@ -64,9 +64,9 @@ def extract_journal_info(obj, eng):
                     "title"
                 )
             if "year" in extracted_publication_info:
-                pubnote["year"] = extracted_publication_info.get(
+                pubnote["year"] = int(extracted_publication_info.get(
                     "year"
-                )
+                ))
             if "page" in extracted_publication_info:
                 page_start, page_end, artid = split_page_artid(
                     extracted_publication_info.get("page"))
--- a/tests/unit/refextract/test_tasks.py
+++ b/tests/unit/refextract/test_tasks.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014, 2015, 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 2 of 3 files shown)


tests/unit/refextract/test_tasks.py
```diff
@@ -0,0 +1,55 @@
...
+    eng = DummyEng()
+
+    extract_journal_info(obj, eng)
+
+    pub_info = obj.data['publication_info'][0]
+
+    assert '082102' == pub_info.get('artid')
+    assert '55' == pub_info.get('journal_volume')
+    assert 'J. Math. Phys.' == pub_info.get('journal_title')
+    assert 2014 == pub_info.get('year')
```
jacquerie: You can just write
```python
assert obj.data['publication_info'][0] == {
    'artid': '082102',
    'journal_title': 'J. Math. Phys.',
    'journal_volume': '55',
    'year': 2014,
}
```

tests/unit/refextract/test_tasks.py
```diff
@@ -0,0 +1,55 @@
...
+    pass
+
+
+def test_extract_journal_info():
+
+    data = {
+        'publication_info': [{'pubinfo_freetext': 'J. Math. Phys. 55, 082102 (2014)'}]
+    }
+
+    obj = StubObj(data)
```
jacquerie: IMO it's ok to have
```python
obj = StubObj({
    'publication_info': [
        {'pubinfo_freetext': 'J. Math. Phys. 55, 082102 (2014)'},
    ],
})
```

tests/unit/refextract/test_tasks.py
```diff
@@ -0,0 +1,55 @@
...
+    eng = DummyEng()
+
+    extract_journal_info(obj, eng)
```
jacquerie: Here you should also test the return value (`None`) of the task, because I believe that if it's different from `None` it's used by the next task as the input (I might be wrong on this):
```python
assert extract_journal_info(obj, eng) is None
```

tests/unit/refextract/test_tasks.py
```diff
@@ -0,0 +1,55 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014, 2015, 2016 CERN.
```
jacquerie: Wrong year.

tests/unit/refextract/test_tasks.py
```diff
@@ -0,0 +1,55 @@
...
+from __future__ import absolute_import, division, print_function
+
+from inspirehep.modules.refextract.tasks import extract_journal_info
+
+
+class StubObj(object):
+    def __init__(self, data):
+        self.data = data
+        self.extra_data = {}
+        self.files = {}
```
jacquerie: Kind of weird, but ok.

inspirehep/modules/refextract/tasks.py
```diff
@@ -64,9 +64,9 @@ def extract_journal_info(obj, eng):
                     "title"
                 )
             if "year" in extracted_publication_info:
-                pubnote["year"] = extracted_publication_info.get(
+                pubnote["year"] = int(extracted_publication_info.get(
                     "year"
-                )
+                ))
```
jacquerie: I see that you steered clear from doing the refactoring this task needs : )

PR #1765, opened 2016-11-22
refextract: fix year type in extract_journal_info

jacquerie:
Closes #1764


jacquerie:
Obviously this tasks needs more unit tests and a refactoring, but for now this will do.

PR #1769, opened 2016-11-22
Fix fft in according to the article source for legacy conversion

rikirenz:
This fix should allow to export the fft field in the right format

### DIFF
--- a/inspirehep/dojson/common/base.py
+++ b/inspirehep/dojson/common/base.py
@@ -291,6 +291,7 @@ def fft(self, key, value):
         'flag': value.get('o'),
         'description': value.get('d'),
         'filename': value.get('n'),
+        'filetype': value.get('f'),
     }
 
 
@@ -304,6 +305,7 @@ def fft2marc(self, key, value):
         'o': value.get('flag'),
         'd': value.get('description'),
         'n': value.get('filename'),
+        'f': value.get('filetype'),
     }
 
 
--- a/inspirehep/modules/workflows/tasks/submission.py
+++ b/inspirehep/modules/workflows/tasks/submission.py
@@ -35,7 +35,10 @@ from retrying import retry
 from invenio_accounts.models import User
 
 from ....utils.tickets import get_instance, retry_if_connection_problems
-from .actions import in_production_mode
+from .actions import (
+    in_production_mode,
+    is_arxiv_paper,
+)
 
 
 LOGGER = logging.getLogger(__name__)
@@ -373,21 +376,27 @@ def prepare_files(obj, eng):
     if not obj.files:
         return
 
+    def _define_filename(paper, file_name):
+        if is_arxiv_paper(paper):
+            return 'arxiv:' + file_name
+
+        return file_name
+
     def _get_fft(url, name):
         return {
             'url': os.path.realpath(url),
             'docfile_type': 'INSPIRE-PUBLIC',
-            'filename': name,
+            'filename': _define_filename(obj, os.path.splitext(name)[0]),
+            'filetype': os.path.splitext(name)[1],
         }
 
     pdf_file_objs = [
         (key, obj.files[key])
-        for key in obj.files.keys
-        if key.endswith('.pdf')
+        for key in obj.files.keys if key.endswith('.pdf')
     ]
     pdf_paths = [
         _get_fft(pdf_file_obj.obj.file.uri, name)
-        for name, pdf_file_obj in pdf_file_objs
+        for name, pdf_file_obj in pdf_file_objs if pdf_file_obj
     ]
 
     obj.data['fft'] = obj.data.get('fft', []) + pdf_paths
--- a/tests/unit/workflows/test_workflows_tasks_submission.py
+++ b/tests/unit/workflows/test_workflows_tasks_submission.py
@@ -22,7 +22,13 @@
 
 from __future__ import absolute_import, division, print_function
 
-from inspirehep.modules.workflows.tasks.submission import add_note_entry
+from mock import MagicMock
+import mock
+
+from inspirehep.modules.workflows.tasks.submission import (
+    add_note_entry,
+    prepare_files,
+)
 
 
 class StubObj(object):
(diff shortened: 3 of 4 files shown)


rikirenz:
ready to be reviewed

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -392,3 +394,77 @@ def test_harvesting_arxiv_workflow_accepted(
         obj = workflow_object_class.get(obj_id)
         # It was accepted
         assert obj.status == ObjectStatus.COMPLETED
+
+        # Check the conversion for the legacy sys
+        marc_json = hep2marc.do(obj.data)
+        marcxml = legacy_export_as_marc(marc_json)
+        record = create_record(marcxml)
```
jacquerie: This is not the best record to test, because in INSPIRE we don't have it, so the things that in the assertions are on the left actually come from the current implementation.

So this is only testing that we don't regress from here, not that we create records that are the same as we have on INSPIRE...

jacquerie: Yeah, actually we can do even better that this: we store an XML fixture of the record how it is on INSPIRE, and we do a similar test (without checking at all the intermediate steps). Then the test compares the dictionary with MARC keys coming from this fixture and the dictionary with MARC keys coming through the workflow: in this way you don't have to type manually all these assertions, but can write just one.

kaplun:
@rikirenz can you rebase properly this PR?

kaplun:
This PR seems not to correspond to what the title say(!) :dancing_women:

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -373,11 +374,20 @@ def prepare_files(obj, eng):
     if not obj.files:
         return
 
+    def _define_filename(paper, file_name):
+        """Define the filename in according to the article source"""
```
jacquerie: There's no need to add docstrings to internal helpers: you're their only user, so they don't need documentation.

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -34,6 +34,7 @@
 
 from invenio_accounts.models import User
 
+from ....modules.workflows.tasks.actions import is_arxiv_paper
```
jacquerie: This is just `.actions`, so it can added to the import two lines below.

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -373,11 +374,20 @@ def prepare_files(obj, eng):
     if not obj.files:
         return
 
+    def _define_filename(paper, file_name):
+        """Define the filename in according to the article source"""
+        if is_arxiv_paper(paper):
+            return 'arxiv:' + file_name
+
+        # Default return
```
jacquerie: You don't need to write these kind of comments. It's obvious that a return returns!

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -389,7 +399,6 @@ def _get_fft(url, name):
         _get_fft(pdf_file_obj.obj.file.uri, name)
         for name, pdf_file_obj in pdf_file_objs
     ]
-
```
jacquerie: Leftover?

inspirehep/modules/workflows/tasks/submission.py
kaplun: Still isn't it nice to describe what the subfunction does, so that you don't have to reparse it with your eyes everytime?

rikirenz:
@jacquerie unit tests are coming :+1:

tests/unit/workflows/test_workflows_tasks_submission.py
```diff
@@ -35,9 +41,51 @@ class DummyEng(object):
...
 def test_add_note_entry_does_not_add_value_that_is_already_present():
     obj = StubObj({'public_notes': [{'value': '*Temporary entry*'}]}, {'core': 'something'})
     eng = DummyEng()
 
     assert add_note_entry(obj, eng) is None
     assert {'public_notes': [{'value': '*Temporary entry*'}]} == obj.data
+
+
+@mock.patch('inspirehep.modules.workflows.tasks.submission.is_arxiv_paper',
+    side_effect=fake_is_arxiv_paper)
```
jacquerie: If all you want to do is to return a value you can just `return_value=True` without having to define a mock function.

tests/unit/workflows/test_workflows_tasks_submission.py
```diff
@@ -35,9 +41,51 @@ class DummyEng(object):
...
+        'fake_record': mock_file,
+        '': mock_file,
+        'fake_record1.pdf': None,
+    })
+
+    assert prepare_files(files_collection, eng) is None
+    assert {'url': '/code/fake_uri',
+            'filetype': '.pdf',
+            'docfile_type': 'INSPIRE-PUBLIC',
+            'filename': 'arxiv:fake_record'} == files_collection.data['fft'][0]
```
jacquerie: Better (you assert more, you don't `IndexError`):
```python
assert files_collection.data['fft'] == [
    {
        ...
    }
]
```

inspirehep/dojson/common/base.py
```diff
@@ -304,6 +305,7 @@ def fft2marc(self, key, value):
         'o': value.get('flag'),
         'd': value.get('description'),
         'n': value.get('filename'),
+        'f': value.get('filetype'),
```
jacquerie: These two are modifying another thing in DoJSON and need tests as well.

tests/unit/workflows/test_workflows_tasks_submission.py
```diff
@@ -35,9 +41,51 @@ class DummyEng(object):
...
+
+        def __getitem__(self, item):
+            return self.data[item]
+
+    eng = DummyEng()
+    files_collection = MagicMock()
+    mock_file = MagicMock()
+
+    mock_file.name = MagicMock(return_value='fake_record')
+    mock_file.obj.file.uri = 'fake_uri'
```
jacquerie: I understand what you are doing with the mocks here, but I wonder if these could be pulled up in the `MockObj` definition rather than defined per test.

tests/unit/workflows/test_workflows_tasks_submission.py
```diff
@@ -35,9 +41,51 @@ class DummyEng(object):
...
+
+    files_collection.data = {}
+    files_collection.files = MockFilesCollectionIterator({
+        'fake_record.pdf': mock_file,
+        'fake_record.txt': mock_file,
+        'fake_record': mock_file,
+        '': mock_file,
+        'fake_record1.pdf': None,
+    })
```
jacquerie: An added benefit of the modification above is that these test scenarios can be split in multiple tests, which achieves better clarity.

inspirehep/dojson/common/base.py
jacquerie: Also these should be split in their own commit.

tests/unit/workflows/test_workflows_tasks_submission.py
jacquerie: BTW this is an example of Classicist vs Mockist testing: http://martinfowler.com/articles/mocksArentStubs.html#ClassicalAndMockistTesting.

tests/unit/workflows/test_workflows_tasks_submission.py
jacquerie: Yes, sorry, `StubObj`!

rikirenz: Could you show me the classicist way to simulate this behavior:
`mock_file.obj.file.uri = 'fake_uri'`?

inspirehep/modules/workflows/tasks/submission.py
david-caro: 👍 
bad code < bad code commented < 'good code' < 'good code' commented < good code that needs no comments

kaplun:
What is the status of this PR? @david-caro are you taking it into account WRT the work you are currently doing on passing over the uploaded file?

jacquerie:
> What is the status of this PR?

I'm currently merging it.

PR #1772, opened 2016-11-23
tests: unit tests for workflows/tasks/submission

rikirenz:
Addresses #1773

- [x] `submit_rt_ticket`
- [x] `create_ticket`
- [x] `reply_ticket`
- [x] `close_ticket`
- [x] `send_robotupload`
- [x] `add_note_entry`
- [x] `filter_keywords`
- [x] `prepare_keywords`
- [x] `user_pdf_get`
- [x] `prepare_files`
- [x] `remove_references`

### DIFF
--- a/tests/unit/workflows/test_workflows_tasks_submission.py
+++ b/tests/unit/workflows/test_workflows_tasks_submission.py
@@ -28,6 +28,7 @@ from inspirehep.modules.workflows.tasks.submission import (
     create_ticket,
     reply_ticket,
     close_ticket,
+    send_robotupload,
 )
 
 from six import StringIO
@@ -49,6 +50,7 @@ class MockObj(object):
     log = MockLog()
 
     def __init__(self, id_, data, extra_data):
+        self.id = id_
         self.id_user = id_
         self.data = data
         self.extra_data = extra_data
@@ -150,3 +152,41 @@ def test_close_ticket(mock_get_instance):
 
     mock_rt.edit_ticket.assert_called_with(ticket_id=1, Status='resolved')
 
+
+@pytest.mark.parametrize('extra_data_key', [None, 'something'])
+@mock.patch('inspirehep.modules.workflows.tasks.submission.in_production_mode')
+@mock.patch('inspirehep.utils.robotupload.make_robotupload_marcxml')
+@mock.patch('inspirehep.dojson.utils.legacy_export_as_marc')
+def test_send_robotupload(mock_legacy_export_as_marc,
+                          mock_make_robotupload_marcxml,
+                          mock_production_mode,
+                          extra_data_key):
+    marc_xml = '<marc_xml=data>'
+    eng = MagicMock()
+    mock_result = MagicMock()
+    marcxml_processor = MagicMock()
+    obj = MockObj(1, {}, {'ticket_id': 1})
+
+    mock_result.text = '[INFO]'
+    mock_production_mode.return_value = True
+    mock_legacy_export_as_marc.return_value = marc_xml
+    mock_make_robotupload_marcxml.return_value = mock_result
+    marcxml_processor.do.return_value = {'marc_json': 'data'}
+
+    send_robotupload_function = send_robotupload(
+        url=None,
+        marcxml_processor=marcxml_processor,
+        callback_url='callback/workflows/continue',
+        mode='insert',
+        extra_data_key=extra_data_key
+    )
+    send_robotupload_function(obj, eng)
+
+    mock_make_robotupload_marcxml.assert_called_with(
+        url=None,
+        marcxml=marc_xml,
+        callback_url='https://localhost:5000/callback/workflows/continue',
+        mode='insert',
+        nonce=1,
+        priority=5,
+    )
--- a/tests/unit/workflows/test_workflows_tasks_submission.py
+++ b/tests/unit/workflows/test_workflows_tasks_submission.py
@@ -32,6 +32,7 @@ from inspirehep.modules.workflows.tasks.submission import (
     filter_keywords,
     prepare_keywords,
     user_pdf_get,
+    prepare_files,
 )
 
 from six import StringIO
@@ -55,11 +56,12 @@ class MockObj(object):
 
     log = MockLog()
 
-    def __init__(self, id_, data, extra_data):
+    def __init__(self, id_, data, extra_data, collection_file={}):
         self.id = id_
         self.id_user = id_
         self.data = data
         self.extra_data = extra_data
+        self.files = collection_file
 
 
 def mock_context_function(user, obj):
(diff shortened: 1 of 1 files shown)


tests/integration/workflows/test_submission_workflow.py
```diff
@@ -0,0 +1,79 @@
...
+    small_app.config['PRODUCTION_MODE'] = True
+
+    assert submit_rt_ticket(obj,
+                            queue,
+                            context.get('subject'),
+                            body,
+                            'mail@ma.il',
+                            ticket_id_key)
+
+    assert mocker_rt_instance.call_count == 1
```
jacquerie: Take a look at this: https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html

rikirenz: if I remove the line: `assert mocker_rt_instance.call_count == 1`. is it good for you ? or should I do other changes?

rikirenz:
@jacquerie I just realized that I put the test in the wrong folder. This is a unit test not a integration test. I removed the line that was checking that the `get_instance()` method was triggered. BTW I did not understand your comment (about the google tests). Could you give me more explanations?

tests/unit/workflows/test_workflows_tasks_submission.py
```diff
@@ -35,9 +60,40 @@ class DummyEng(object):
...
+            side_effect=mock_get_rt_instance)
+def test_submit_rt_ticket(mocker_rt_instance):
+    """test that a ticket is sent to the RT service"""
+
+    body = ''
+    queue = ''
+    context = {}
+    obj = MockObj(1, {}, {})
+    ticket_id_key = 'ticket_id'
+    current_app.config['PRODUCTION_MODE'] = True
```
rikirenz: I have added this line to simulate the production mode. I have tried to use the `app` fixture in the `conftest.py` file but I had problems. Is it ok in this way?

jacquerie: I believe not. This is modifying the configuration for all the tests that follow, potentially introducing a dependency between tests.

jacquerie: This is the correct way to do that: https://github.com/inspirehep/inspire-next/blob/cef4089e2c381f00bcd0631e9f388db30b6bb680/tests/integration/workflows/test_arxiv_workflow.py#L271

jacquerie: And this is another way to do it, but I like it less, despite the fact that I just wrote it ._. https://github.com/inspirehep/inspire-next/blob/cef4089e2c381f00bcd0631e9f388db30b6bb680/tests/unit/workflows/test_workflows_tasks_actions.py#L164-L176

tests/unit/workflows/test_workflows_tasks_submission.py
```diff
@@ -35,9 +60,32 @@ class DummyEng(object):
...
+
+
+@mock.patch('inspirehep.modules.workflows.tasks.submission.get_instance', side_effect=mock_get_rt_instance)
+@mock.patch('inspirehep.modules.workflows.tasks.submission.in_production_mode', side_effect=mock_in_production_mode)
+def test_submit_rt_ticket(mock_production_mode, mocker_rt_instance):
+    obj = MockObj(1, {}, {})
+    current_app.config['PRODUCTION_MODE'] = True
+
+    assert submit_rt_ticket(obj, '', None, '', 'mail@ma.il', 'ticket_id')
+    assert obj.extra_data['ticket_id'] == 1
```
rikirenz: I have not found a better way to test this function. I think is useless.

tests/unit/workflows/test_workflows_tasks_submission.py
```diff
@@ -22,22 +22,55 @@
...
+            self.data = data
+            self.extra_data = extra_data
+
+    mock_file = MagicMock()
+    mock_file.create_ticket = MagicMock(return_value=1)
+    mock_rt_instance.return_value = mock_file
+    mock_production_mode.return_value = True
+
+    obj = MockObj(1, {}, {})
+    current_app.config['PRODUCTION_MODE'] = True
```
rikirenz: this line is not necessary

tests/unit/workflows/test_workflows_tasks_submission.py
```diff
@@ -22,22 +22,380 @@
...
+            self._log = StringIO()
+
+        def info(self, *args, **kwargs):
+            self._log.write(args[0])
+
+        def error(self, *args, **kwargs):
+            self.info(args[0])
+
+        def debug(self, message, format=None):
+            self.info(message)
```
jacquerie: Uhm, these three logging methods have all different signatures and implementations, despite being (essentially) the same thing....

Also, with this design one cannot see if the proper logging level happened. Maybe we should have three `StringIO` objects for the three different log levels? Maybe we should add some kind of marker (like `DEBUG:`) in front?

tests/unit/workflows/test_workflows_tasks_submission.py
```diff
@@ -22,22 +22,380 @@
...
+def mock_context_function(user, obj):
+    return {}
+
+
+@mock.patch('inspirehep.modules.workflows.tasks.submission.get_instance')
+@mock.patch('inspirehep.modules.workflows.tasks.submission.in_production_mode')
+def test_submit_rt_ticket(mock_production_mode, mock_rt_instance):
+    mock_file = MagicMock()
+    mock_file.create_ticket = MagicMock(return_value=1)
+    mock_rt_instance.return_value = mock_file
```
jacquerie: I would expect that a variable called `mock_file` is the mock of a file, but you clearly mean something different. What do you mean here?

tests/unit/workflows/test_workflows_tasks_submission.py
```diff
@@ -22,22 +22,380 @@
...
+    mock_production_mode.return_value = True
```
jacquerie: Can be pulled up in the mock declaration, or avoided entirely (Classicist vs. Mockist).

tests/unit/workflows/test_workflows_tasks_submission.py
```diff
@@ -22,22 +22,380 @@
...
+
+
+@pytest.mark.parametrize('extra_data_key', [None, 'something'])
+@mock.patch('inspirehep.modules.workflows.tasks.submission.in_production_mode')
+@mock.patch('inspirehep.utils.robotupload.make_robotupload_marcxml')
+@mock.patch('inspirehep.dojson.utils.legacy_export_as_marc')
+def test_send_robotupload(mock_legacy_export_as_marc,
+                          mock_make_robotupload_marcxml,
+                          mock_production_mode,
+                          extra_data_key):
```
jacquerie: Tip: when mock names become this unwieldy you can just use the initials of every letter: `mock_legacy_export_as_marc` -> `l_e_a_m`.

Although take note that, when that happens, maybe you're mocking too many things...

tests/unit/workflows/test_workflows_tasks_submission.py
rikirenz: It is wrong. The correct name should be `mock_ticket_obj`

tests/unit/workflows/test_workflows_tasks_submission.py
rikirenz: Do you mean: create a specific app object that simulates the production mode?

tests/unit/workflows/test_workflows_tasks_submission.py
rikirenz: :+1:

tests/unit/workflows/test_workflows_tasks_submission.py
rikirenz: :+1:

jacquerie:
Have a look at https://github.com/inspirehep/inspire-next/blob/b909f11eac21efe5b3bbe9f667337d57bb6a7423/tests/unit/workflows/test_workflows_tasks_submission.py to see how you can improve these tests.

rikirenz:
@jacquerie  Anyway after several months of tests I understood your point. When We will have time I will try to do tests in the classicist way.

tests/unit/workflows/test_workflows_tasks_submission.py
david-caro: I do prefer explicit names than the shortened ones, even if they are 'unwieldy', well, specially if they are 'unwieldy' as it's then when you are most probably going to be confused.

jacquerie:
Closed in favor of #2313

PR #1775, opened 2016-11-24
records: sync modifications in database and ElasticSearch

jmartinm:
* Adds models_committed receiver that sends to ElasticSearch the changes
  made in the RecordsMetadata table.

Signed-off-by: Javier Martin Montull <javier.martin.montull@cern.ch>

PR #1776, opened 2016-11-25
WIP records: update links for merged records

### LINKED ISSUE #484: Merging records should propagate links

jacquerie:
Depends on #1775, because we currently modify the DB, but we don't touch ES, and that is going to be handled by that receiver.

Closes #484
Closes #1660

### DIFF
--- a/inspirehep/modules/pidstore/utils.py
+++ b/inspirehep/modules/pidstore/utils.py
@@ -45,6 +45,13 @@ def get_endpoint_from_pid_type(pid_type):
     return PID_TYPE_TO_ENDPOINT[pid_type]
 
 
+def get_endpoint_from_schema(schema):
+    pid_type = get_pid_type_from_schema(schema)
+    endpoint = get_endpoint_from_pid_type(pid_type)
+
+    return endpoint
+
+
 def get_pid_type_from_endpoint(endpoint):
     """Return the ``pid_type`` corresponding to an endpoint."""
     ENDPOINT_TO_PID_TYPE = {
--- a/inspirehep/modules/pidstore/utils.py
+++ b/inspirehep/modules/pidstore/utils.py
@@ -46,10 +46,26 @@ def get_endpoint_from_pid_type(pid_type):
 
 
 def get_endpoint_from_schema(schema):
+    """Return the endpoint corresponding to a schema."""
     pid_type = get_pid_type_from_schema(schema)
     endpoint = get_endpoint_from_pid_type(pid_type)
 
     return endpoint
+
+
+def get_index_from_endpoint(endpoint):
+    """Return the index corresponding to an endpoint."""
+    ENDPOINT_TO_INDEX = {
+        'authors': 'records-authors',
+        'conferences': 'records-conferences',
+        'experiments': 'records-experiments',
+        'institutions': 'records-institutions',
+        'jobs': 'records-jobs',
+        'journals': 'records-journals',
+        'literature': 'records-hep',
+    }
+
+    return ENDPOINT_TO_INDEX[endpoint]
 
 
 def get_pid_type_from_endpoint(endpoint):
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -28,7 +28,11 @@ import six
 from flask import current_app
 
 from invenio_indexer.signals import before_record_index
-from invenio_records.signals import before_record_insert, before_record_update
+from invenio_records.signals import (
+    after_record_update,
+    before_record_insert,
+    before_record_update,
+)
 
 from inspirehep.dojson.utils import classify_field, get_recid_from_ref
 from inspirehep.modules.pidstore.utils import get_pid_type_from_schema
@@ -390,3 +394,8 @@ def populate_experiment_suggest(sender, json, *args, **kwargs):
                 'payload': {'$ref': get_value(json, 'self.$ref')},
             },
         })
+
+
+@after_record_update.connect
+def update_outdated_links(sender, *args, **kwargs):
+    """TODO."""
--- a/inspirehep/modules/records/tasks.py
+++ b/inspirehep/modules/records/tasks.py
@@ -1,0 +1,116 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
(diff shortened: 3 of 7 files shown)


tests/integration/test_record.py
```diff
@@ -347,3 +399,27 @@ def test_records_can_be_merged(app, records_not_merged_in_marcxml):
     with app.test_client() as client:
         assert client.get('/api/literature/111').status_code == 200
         assert client.get('/api/literature/222').status_code == 301
+
+
+def test_records_propagate_links_when_merged(app, records_to_be_merged):
+    before_record_update.disconnect(check_if_record_is_going_to_be_deleted)
```
jacquerie: This is why receivers are a bad idea: with just two receivers that listen for data changes on the same attributes, you run into these kind of issues, where the two receivers conflict, and you have to disable one to ensure that the other one can do its job properly.

inspirehep/modules/records/receivers.py
```diff
@@ -390,3 +394,8 @@ def populate_experiment_suggest(sender, json, *args, **kwargs):
                 'payload': {'$ref': get_value(json, 'self.$ref')},
             },
         })
+
+
+@after_record_update.connect
+def update_outdated_links(sender, *args, **kwargs):
+    """TODO."""
```
jmartinm: It's going to be part of this PR?

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,116 @@
...
+from invenio_search import current_search_client as es
+
+from inspirehep.modules.pidstore.utils import (
+    get_endpoint_from_schema,
+    get_index_from_endpoint,
+)
+
+
+@shared_task(ignore_result=True)
+def update_refs(old, new):
```
jmartinm: What is `old` and `new`? Better naming of parameters and docstring would be useful.

tests/unit/records/test_records_tasks.py
```diff
@@ -0,0 +1,151 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+from mock import patch
+
+from inspirehep.modules.records.tasks import update_links
+
+
+def test_update_links(app):
+    config = {
```
jmartinm: Why not just do:

```python
app.config['REF_UPDATER_WHITELIST'] = {
        'literature': [
            'record',
        ],
    }
```

and remove the `patch` below. Reads simpler to me.

inspirehep/modules/records/tasks.py
```diff
@@ -0,0 +1,116 @@
...
+    def _get_uuids_to_update(old):
+        def _replace_record_with_recid(path):
+            return path.replace('record', 'recid')
+
+        def _ref_to_recid(ref):
+            return int(ref.split('/')[-1])
+
+        whitelists = current_app.config['REF_UPDATER_WHITELIST']
+
+        for endpoint, whitelist in iteritems(whitelists):
```
jmartinm:
```python
six.iteritems
```

tests/integration/test_record.py
```diff
@@ -347,3 +399,27 @@ def test_records_can_be_merged(app, records_not_merged_in_marcxml):
...
+        with db.session.begin_nested():
+            record_to_be_merged = get_db_record('lit', 222)
+            record_to_be_merged['deleted'] = True
+            record_to_be_merged['new_record'] = {
+                '$ref': 'http://localhost:5000/api/literature/111'}
+            record_to_be_merged.commit()
+        db.session.commit()
+
+    with app.test_client() as client:
+        merged_record = json.loads(client.get('/api/literature/222/db').data)
```
jmartinm: Is there something trying to be tested here? Why get it through the API if it is getting retrieved from DB?

You can just do again `get_db_record('lit', 222)`

tests/integration/test_record.py
```diff
@@ -347,3 +399,27 @@ def test_records_can_be_merged(app, records_not_merged_in_marcxml):
...
+
+    expected = 'http://localhost:5000/api/experiments/333'
```
jmartinm: Might be fine, but is this want we always want?

Even though the 2 records being merged have different values in `$ref`, if the path coincides the merged records gets all the `$ref` replaced by the values in the new 'master' record?

tests/integration/test_record.py
jmartinm: I agree. Rather than changing metadata and relying on the knowledge that there is some receiver somewhere doing the magic, I would rather see and explicit call to `.merge()`

inspirehep/modules/pidstore/utils.py
```diff
@@ -45,6 +45,29 @@ def get_endpoint_from_pid_type(pid_type):
     return PID_TYPE_TO_ENDPOINT[pid_type]
 
 
+def get_endpoint_from_schema(schema):
```
jmartinm: I have a feeling this file is going out of hand with so many conversions back and forth... but don't know if there is a better solution for this case.

inspirehep/modules/records/tasks.py
jacquerie: Do you mean that you prefer to see
```python
import six

...

for k, v in six.iteritems(d):
    ...
```
rather than
```python
from six import iteritems

...

for k, v in iteritems(d):
    ...
```
?

Or maybe you didn't spot that this is `iteritems` from `six`, but you suggested me to use it?

tests/unit/records/test_records_tasks.py
jacquerie: I believe that this is permanently modifying the config for the rest of the tests, introducing a dependency on ordering.

tests/integration/test_record.py
jacquerie: Ah! We are actually testing the wrong thing!

tests/integration/test_record.py
jacquerie: I think we are doing this awkward thing because `get_db_record` would fail, since the record is now deleted. But actually, as you made me realize, we should rather be testing that record 333's `$ref` is updated, rather than this one!

inspirehep/modules/pidstore/utils.py
jacquerie: I know, this sucks. The problem is that we have so many concepts that are tightly related but are all expressed by strings with subtle differences, and conversion functions are the cleanest interface I could think of.

PR #1774, opened 2016-11-23
Travis: finally support coverage

kaplun:
Signed-off-by: Samuele Kaplun <samuele.kaplun@cern.ch>

PR #1778, opened 2016-11-27
travis: add coverage to acceptance suite

jacquerie:
Closes #1774


jacquerie:
Finally, this works. What was the problem?

Note the order of `after_success` and `after_script`: https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle. Previously, coverage was being sent to Coveralls before it was generated by sending `SIGTERM` to `web`, since the former was happening in `after_success`, while the latter was happening in `after_script`.

kaplun:
:facepunch: 

Grande!

kaplun:
@jacquerie maybe you can merge it if you don't find quickly a better reason why we create coverage ven in case of failure.

PR #1783, opened 2016-12-02
travis: bump docker version

jacquerie:
As you can see from the branch name, I started with the idea of removing [these lines](https://github.com/inspirehep/inspire-next/blob/9a9a4b4add6738d889251f8352510711a793eb9a/.travis.yml#L54-L57) and the variable above, but then I decided against it, because I believe there's value in specifying exactly which version of Docker/Docker-Compose we know are working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ env:
   global:
     - DOCKER_DATA="$HOME/docker_data"
     - DOCKER_VERSION=1.12.3-0~trusty
-    - DOCKER_COMPOSE_VERSION=1.8.0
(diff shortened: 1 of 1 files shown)


jacquerie:
@david-caro This works, but I would like your opinion on this.

david-caro:
👍 looks good to me

david-caro:
You might want to take the opportunity to bump also docker compose to 1.9.0 (latest)

jacquerie:
> You might want to take the opportunity to bump also docker compose to 1.9.0 (latest)

👍 good idea

PR #1786, opened 2016-12-06
travis: merge pep8 and unit suites

jacquerie:
@david-caro, @kaplun: like we discussed. See if you like it.

Essentially reverts #1373.


david-caro:
Looks good to me 👍

PR #1790, opened 2016-12-06
records: add Inspire record class

spirosdelviniotis:
* Adds an inspire record class between invenio record and ESrecord.
   Inheritance from invenio_records.api Record.

### DIFF
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -16,11 +16,11 @@
 # You should have received a copy of the GNU General Public License
 # along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
 #
-# In applying this licence, CERN does not waive the privileges and immunities
+# In applying this license, CERN does not waive the privileges and immunities
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-"""Record subclass to fetch records from ElasticSearch."""
+"""Inspire Records"""
 
 from __future__ import absolute_import, division, print_function
 
@@ -30,7 +30,9 @@ import arrow
 from elasticsearch.exceptions import NotFoundError
 
 from invenio_pidstore.errors import PIDDoesNotExistError
+from invenio_pidstore.models import PersistentIdentifier
 from invenio_records.api import Record
+from invenio_db import db
 
 from inspirehep.utils.record_getter import (
     RecordGetterError,
@@ -38,7 +40,39 @@ from inspirehep.utils.record_getter import (
 )
 
 
-class ESRecord(Record):
+class InspireRecord(Record):
+
+    """Record class that fetches records from DataBase."""
+
+    def merge(self, other):
+        """Redirect pidstore of current record to the other InspireRecord.
+
+        :param other: The record that self(record) is going to be redirected.
+        :type other: InspireRecord
+        """
+        pids_deleted = PersistentIdentifier.query.filter(PersistentIdentifier.object_uuid == self.id).all()
+        pid_merged = PersistentIdentifier.query.filter(PersistentIdentifier.object_uuid == other.id).one()
+        with db.session.begin_nested():
+            for pid in pids_deleted:
+                pid.redirect(pid_merged)
+                db.session.add(pid)
+
+    def delete(self):
+        """Mark as deleted all pidstores for a specific record."""
+        pids = PersistentIdentifier.query.filter(PersistentIdentifier.object_uuid == self.id).all()
+        with db.session.begin_nested():
+            for pid in pids:
+                pid.delete()
+                db.session.add(pid)
+
+        self['deleted'] = True
+        self.commit()
+
+    def _delete(self, *args, **kwargs):
+        super(InspireRecord, self).delete(*args, **kwargs)
+
+
+class ESRecord(InspireRecord):
 
     """Record class that fetches records from ElasticSearch."""
 
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -30,20 +30,14 @@ from flask_sqlalchemy import models_committed
 
 from invenio_indexer.api import RecordIndexer
 from invenio_indexer.signals import before_record_index
-from invenio_records.api import Record
 from invenio_records.models import RecordMetadata
 from invenio_records.signals import before_record_insert, before_record_update
 
 from inspirehep.dojson.utils import classify_field, get_recid_from_ref
-from inspirehep.modules.pidstore.utils import get_pid_type_from_schema
+from inspirehep.modules.records.api import InspireRecord
 from inspirehep.utils.date import create_valid_date
 from inspirehep.utils.helpers import force_force_list
-from inspirehep.utils.record import (
-    get_value,
-    merge_pidstores_of_two_merged_records,
-    soft_delete_pidstore_for_record,
-)
-from inspirehep.utils.record_getter import get_db_record
+from inspirehep.utils.record import get_value
 
 from .experiments import EXPERIMENTS_MAP
 from .signals import after_record_enhanced
@@ -56,9 +50,9 @@ def receive_after_model_commit(sender, changes):
     for model_instance, change in changes:
         if isinstance(model_instance, RecordMetadata):
             if change in ('insert', 'update'):
-                indexer.index(Record(model_instance.json, model_instance))
+                indexer.index(InspireRecord(model_instance.json, model_instance))
             else:
-                indexer.delete(Record(model_instance.json, model_instance))
+                indexer.delete(InspireRecord(model_instance.json, model_instance))
 
 
 @before_record_index.connect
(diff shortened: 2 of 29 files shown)


inspirehep/modules/records/record.py
```diff
@@ -0,0 +1,39 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+from invenio_records.api import Record
+
+
+class InspireRecord(Record):
+    """"Inspire Record inherits from Invenio Record"""
+
+    def __init__(self, data=None, model=None):
```
jmartinm: This will not be needed if you don't need to do anything extra in `__init__`

inspirehep/modules/records/record.py
```diff
@@ -0,0 +1,39 @@
...
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+""""Inspire Record inherits from Invenio Record"""
+
+from __future__ import absolute_import, division, print_function
+
+from invenio_records.api import Record
+
+
+class InspireRecord(Record):
```
jmartinm: This is a minimal proof of concept I guess, in this PR you can add the `merge` and `delete` functionality that currently is in the receivers.

inspirehep/modules/records/api.py
```diff
@@ -0,0 +1,85 @@
...
+    get_es_record_by_uuid
+)
+from inspirehep.utils.record import (
+    merge_pidstores_of_two_merged_records,
+    soft_delete_pidstore_for_record,
+)
+
+
+class InspireRecord(Record):
+    """"Inspire Record inherits from Invenio Record"""
```
jmartinm: There are four apostrophes at the beginning `""""`

inspirehep/modules/records/api.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def merge(self, new):
+        merge_pidstores_of_two_merged_records(new.id, self.id)
+
+    def delete(self, force=False):
+        if force:
+            super(InspireRecord, self).delete(force=True)
+        else:
+            self['deleted'] = True
+            soft_delete_pidstore_for_record(self.id)
```
jmartinm: I wonder since this function interacts with the session (with `pid.delete()`) if both this line and the next one should be inside:

```python
    soft_delete_pidstore_for_record(self.id)
    self.commit()
```

And then from the outside you will do:

```python
InspireRecord({}).delete()
db.session.commit()
```

inspirehep/utils/record_getter.py
```diff
@@ -109,5 +108,6 @@ def get_es_record_by_uuid(uuid):
 
 @raise_record_getter_error_and_log
 def get_db_record(pid_type, recid):
+    from inspirehep.modules.records.api import InspireRecord
```
jmartinm: Import at the top of the file

inspirehep/modules/records/api.py
spirosdelviniotis: Yes, you are right! We can put  `with db.session.begin_nested():`  and  `db.session.add(pid)`  inside the function that interacts with the session and then the user is responsible for `db.session.commit()` .
:christmas_tree: :santa:

inspirehep/utils/record_getter.py
spirosdelviniotis: That import is a workaround due to we receive an import error if I import it at the top of the file.

`Original error message: 'cannot import name InspireRecord'.
Make sure your test modules/packages have valid Python names.`

inspirehep/modules/records/api.py
```diff
@@ -0,0 +1,86 @@
...
+from invenio_records.api import Record
+
+from inspirehep.utils.record_getter import (
+    RecordGetterError,
+    get_es_record_by_uuid
+)
+from inspirehep.utils.record import (
+    merge_pidstores_of_two_merged_records,
+    soft_delete_pidstore_for_record,
+)
```
jacquerie: These two "utils" have exactly one user, which is the class below, so they are not a good fit for the `utils` folder. I'd say that their body should actually be moved to the `.delete` and `.merge`, and they should be deleted from there.

inspirehep/modules/records/api.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class InspireRecord(Record):
+
+    """Inspire Record inherits from Invenio Record"""
```
jacquerie: Python is already saying this more concisely than you by writing `class InspireRecord(Record)`. There's no need to repeat that! Docstrings (and comments in general) should say the _why_, not the _what_.

inspirehep/modules/records/api.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def merge(self, new):
```
jacquerie: Needs a docstring. Also, `new` is a little too terse of an argument, because it doesn't hint at what it is. Given what you implemented, probably `def merge(self, recid)` is a better signature, but now I'm wondering if `def merge(self, other)`, where `other` is another **record** (a different implementation), is even better.

inspirehep/modules/records/api.py
```diff
@@ -0,0 +1,86 @@
...
+        merge_pidstores_of_two_merged_records(new.id, self.id)
+
+    def delete(self, force=False):
```
jacquerie: Needs a docstring. Also, do we really need the `force` parameter? We're subclassing to get rid of the default behavior, because we need something different from this method, so what is the rationale for allowing it? Maybe @kaplun or @jmartinm know?

inspirehep/modules/records/api.py
spirosdelviniotis: You are right! I agree :)

inspirehep/modules/records/api.py
kaplun: @jacquerie the only use-case could be if we create a large bunch of records by mistake, and then we want get rid of them without wasting resources.

Alternatively you can force-delete in tearDowns within tests.

But I have no strong opinion on this.

inspirehep/utils/record_getter.py
jmartinm: Sounds like a circular dependency.

inspirehep/modules/records/api.py
```diff
@@ -16,11 +16,11 @@
 # You should have received a copy of the GNU General Public License
 # along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
 #
-# In applying this licence, CERN does not waive the privileges and immunities
+# In applying this license, CERN does not waive the privileges and immunities
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-"""Record subclass to fetch records from ElasticSearch."""
+""""Inspire Records"""
```
jmartinm: Extra apostrophe ;)

inspirehep/modules/records/api.py
```diff
@@ -30,15 +30,45 @@
...
 )
 
 
-class ESRecord(Record):
+class InspireRecord(Record):
+
+    """Record class that fetches records from DataBase."""
+
+    def merge(self, other):
+        """Merge all pidstores of deleted record to the merged record."""
```
jmartinm: To better document the `other` parameter, you can use this syntax https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/search/api.py#L50-L53

jmartinm: The docstring could read nicer with something like:

`Redirect pidstore of current record to the other InspireRecord.`

inspirehep/utils/record_getter.py
jacquerie: Yeah, in general stuff in `utils` should not depend from stuff in `modules`.

inspirehep/modules/records/api.py
```diff
@@ -30,15 +30,45 @@
...
+        pids_deleted = PersistentIdentifier.query.filter(PersistentIdentifier.object_uuid == self.id).all()
+        pids_merged = PersistentIdentifier.query.filter(PersistentIdentifier.object_uuid == other.id).one()
+        with db.session.begin_nested():
+            for pid in pids_deleted:
+                pid.redirect(pids_merged)
+                db.session.add(pid)
+
+    def delete(self):
+        """By default mark as deleted all pidstores for a specific record."""
```
jmartinm: I don't think you need the `By default` part of the sentence.

inspirehep/modules/records/receivers.py
```diff
@@ -379,15 +375,11 @@ def check_if_record_is_going_to_be_deleted(sender, *args, **kwargs):
...
-        new_recid = get_recid_from_ref(sender.get('new_record'))
-        if new_recid:
-            merged_record = get_db_record(pid_type, new_recid)
-            merge_pidstores_of_two_merged_records(merged_record.id, record.id)
-        else:
-            record = get_db_record(pid_type, control_number)
-            soft_delete_pidstore_for_record(record.id)
+    if sender.get('deleted') and new_recid:
+        merged_record = get_db_record(pid_type, new_recid)
+        record.merge(merged_record)
```
jacquerie: The point of this PR is to delete this receiver, not modifying it like this.

spirosdelviniotis: ok!

tests/integration/test_record.py
```diff
@@ -271,12 +270,12 @@ def _delete_record_from_everywhere(pid_type, record_control_number):
         object_uuid == PersistentIdentifier.object_uuid).delete()
 
     db.session.commit()
-    es.indices.refresh('records-hep')
+    # es.indices.refresh('records-hep')
```
jmartinm: If it's not needed we can remove the line?

tests/integration/test_record.py
```diff
@@ -286,7 +285,7 @@ def _delete_merged_records_from_everywhere(pid_type, merged_record_control_numbe
     db.session.delete(merged_pid)
     db.session.delete(deleted_pid)
     db.session.commit()
-    es.indices.refresh('records-hep')
+    # es.indices.refresh('records-hep')
```
jmartinm: same as above.

tests/integration/test_record.py
```diff
@@ -319,9 +317,11 @@ def test_records_can_be_merged(app, records_not_merged_in_marcxml):
         assert client.get('/api/literature/222').status_code == 200
 
     record = get_db_record('lit', 222)
+    other = get_db_record('lit', 111)
     record['deleted'] = True
     record['new_record'] = {'$ref': 'http://localhost:5000/api/record/111'}
-    record.commit()
+    # record.commit()
```
jmartinm: remove comment

spirosdelviniotis:
Just re-based with master.

jmartinm:
Why are there 6 commits on this PR? I think if you squash them all into one should be ready to merge.

spirosdelviniotis:
Squashed them all into one.

PR #1796, opened 2016-12-07
search: show correct debug output for search queries

### LINKED ISSUE #1795: ElasticSearch query debug does not reflect the query sent to ES
When `DEBUG=True` we currently log the query being sent to ES [here](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/search/query_factory.py#L76)

Problem is that `search.to_dict()` at that moment always has just the filter for the collection so the real query is not shown in the console.

I suggest moving the debug log after [this line](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/search/query.py#L47) to get correct log output.

jmartinm:
* (closes #1795).

### DIFF
--- a/tests/integration/test_search_views.py
+++ b/tests/integration/test_search_views.py
@@ -20,6 +20,9 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
+import json
+from mock import patch
+
 
 def test_search_conferences_is_there(app):
     with app.test_client() as client:
@@ -59,3 +62,41 @@ def test_search_jobs_is_there(app):
 def test_search_falls_back_to_hep(app):
     with app.test_client() as client:
         assert client.get('/search').status_code == 200
+
+
+@patch('inspirehep.modules.search.query.current_app')
+def test_search_logs(current_app_mock, app):
+    def _debug(log_output):
+        query = {
+            'query': {
+                'bool': {
+                    'filter': [{
+                        'match': {'_collections': 'Literature'}
+                    }],
+                    'must': [{
+                        'multi_match': {
+                            'query': '',
+                            'fields': [
+                                'title^3',
+                                'title.raw^10',
+                                'abstract^2',
+                                'abstract.raw^4',
+                                'author^10',
+                                'author.raw^15',
+                                'reportnumber^10',
+                                'eprint^10',
+                                'doi^10'
+                            ],
+                            'zero_terms_query': 'all',
+                        }
+                    }]
+                }
+            },
+            'from': 0,
+            'size': 10
+        }
+        assert query == json.loads(log_output)
+
+    current_app_mock.logger.debug.side_effect = _debug
+    with app.test_client() as client:
+        client.get('/api/literature/')
--- a/inspirehep/modules/search/query.py
+++ b/inspirehep/modules/search/query.py
@@ -24,6 +24,8 @@
 
 from __future__ import absolute_import, division, print_function
 
+import json
+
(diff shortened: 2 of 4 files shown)


jmartinm:
@jacquerie , what do you think about the test? Not very pretty but not sure if there is something better?

tests/integration/test_search_views.py
```diff
@@ -59,3 +61,11 @@ def test_search_jobs_is_there(app):
...
...ns": "Restricted Collection"\n                                }\n                            }\n                        ], \n                        "must": [\n                            {\n                                "match": {\n                                    "_collections": "Literature"\n                                }\n                            }\n                        ]\n                    }\n                }\n            ], \n            "must": [\n                {\n                    "multi_match": {\n                        "query": "", \n                        "fields": [\n                            "title^3", \n                            "title.raw^10", \n                            "abstract^2", \n                            "abstract.raw^4", \n                            "author^10", \n                            "author.raw^15", \n                            "reportnumber^10", \n                            "eprint^10", \n                            "doi^10"\n                        ], \n                        "zero_terms_query": "all"\n                    }\n                }\n            ]\n        }\n    }, \n    "from": 0, \n    "size": 10\n}'
```
jacquerie: Scary O_O

Can't it just be
```python
log_output = '''{
    "query": {
      ...
    }
}'''
```

jmartinm: I think it might work, will give it a try.

jacquerie:
> @jacquerie , what do you think about the test? Not very pretty but not sure if there is something better?

I think it's ok. We already had a similar test: https://github.com/inspirehep/inspire-next/blob/1694cd6cdf02091eb82d05cc7782b58818b7c341/tests/unit/records/test_records_receivers.py#L53-L63

The mock is placed at a different depth, but the idea is the same...

tests/integration/test_search_views.py
jmartinm: In any case I need to check a bit more because I added to the log output `Another Restricted Collection` to make Travis happy but I think that is a leftover from a previous test and should not be there.

jacquerie: https://github.com/inspirehep/inspire-next/blob/1694cd6cdf02091eb82d05cc7782b58818b7c341/tests/integration/test_permissions.py#L147-L207, maybe?

jmartinm: Yep, that's were I am looking. Even though models are cleaned up there is some cache in Redis about restricted collections.. so need to clean that as well.

tests/integration/test_search_views.py
```diff
@@ -20,6 +20,9 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
+import json
```
jacquerie: Unused.

tests/integration/test_search_views.py
```diff
@@ -59,3 +62,15 @@ def test_search_jobs_is_there(app):
 def test_search_falls_back_to_hep(app):
     with app.test_client() as client:
         assert client.get('/search').status_code == 200
+
+
+@patch('inspirehep.modules.search.query.current_app')
+def test_search_logs(current_app_mock, app):
+    def _debug(log_output):
+        assert all(
+            [element in log_output for element in ("filter", "fields", "size")]
```
jacquerie: I think this is always going to pass, no matter what fields you put here. What went wrong with the previous approach?

jmartinm:
@jacquerie I have changed it once more since I did not like the previous approach.

PR #1797, opened 2016-12-07
requirements: flask-celeryext github installation

### LINKED ISSUE #1641: migration: missing validation errors
For the past week or so https://sentry.cern.ch/ has not received validation errors from workers. It appears that the error was introduced in https://github.com/inspirehep/inspire-next/commit/79057cea55fa69c5444bc3f65820fd54621beac3 (reverting that commit fixes the issue), so this has to be related to configuration initialisation.

You'll probably need our `SENTRY_DSN` to diagnose this. If you want to work on this, ask me for it.

jmartinm:
* (closes #1641).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,6 +22,7 @@
 
 # Bleeding edge packages not yet released on Pypi
 -e git+https://github.com/inveniosoftware/invenio-jsonschemas.git@master#egg=invenio-jsonschemas
+-e git+https://github.com/inveniosoftware/flask-celeryext.git@master#egg=flask-celeryext
 
 # FIXME temporary branch for testing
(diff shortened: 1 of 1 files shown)


jacquerie:
Can you check if we still need the hack in https://github.com/inspirehep/inspire-next/blob/1694cd6cdf02091eb82d05cc7782b58818b7c341/inspirehep/celery.py#L35 after this?

jacquerie:
> Can you check if we still need the hack in https://github.com/inspirehep/inspire-next/blob/1694cd6cdf02091eb82d05cc7782b58818b7c341/inspirehep/celery.py#L35 after this?

Just tested that this is no longer the case. Yay for less technical debt!

PR #1809, opened 2016-12-12
authors: pass control_number to author tasks

### LINKED ISSUE #1803: Author update form: ticket do not contain the record id
After submitting an author update, the `recid` does not seem to be present in the `obj.data` so the email templates don't have the correct URL.

```
New update to http://inspirehep.net/record/ from admin@inspirehep.net.

You can edit the record here http://inspirehep.net/record//edit and apply the update from the left panel (Holding Pen).
```

Here is [ the line](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/authors/tasks.py#L134) where `recid` should be present.

And probably somewhere along [here](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/authors/views/holdingpen.py#L283) where `recid` is lost.

jmartinm:
* The emails generated doing an author workflow can access the control_number
  to properly format the URLs (closes #1803).

### DIFF
--- a/inspirehep/modules/authors/views/holdingpen.py
+++ b/inspirehep/modules/authors/views/holdingpen.py
@@ -172,8 +172,8 @@ def get_inspire_url(data):
     url = ""
     if "bai" in data and data["bai"]:
         url = "http://inspirehep.net/author/profile/" + data["bai"]
-    elif "recid" in data and data["recid"]:
-        url = "http://inspirehep.net/record/" + str(data["recid"])
+    elif "control_number" in data and data["control_number"]:
+        url = "http://inspirehep.net/record/" + str(data["control_number"])
     else:
         url = "http://inspirehep.net/hepnames"
     return url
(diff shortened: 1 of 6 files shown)


tests/unit/authors/test_authors_views.py
```diff
@@ -77,7 +77,7 @@ def test_convert_for_form_public_emails():
 
 
 def test_get_inspire_url_with_bai():
-    with_bai = Record({'bai': 'TODO'})
+    with_bai = Record({'control_number': 'TODO'})
```
jacquerie: Test name and test subject should be amended.

jacquerie: Also I don't know how did I miss that TODO when I merged these tests...

inspirehep/modules/authors/views/holdingpen.py
```diff
@@ -170,8 +170,8 @@ def get_inspire_url(data):
     url = ""
     if "bai" in data and data["bai"]:
         url = "http://inspirehep.net/author/profile/" + data["bai"]
```
jacquerie: This can be deleted in a preparatory commit (there is no `bai` key anymore).

jacquerie:
Also, `update_ticket_context` and `curation_ticket_context` should have tests.

inspirehep/modules/authors/views/holdingpen.py
jacquerie: Whoops, I confused this `data` with the one attached to `obj`.

jmartinm:
I have added some tests to this PR.

PR #1810, opened 2016-12-12
dojson: upgrade schemas to version ~4.0

rikirenz:
In this PR we are going to:

- Change the BL for the rule 650.*
- Rename the `field_categories` field to `inspire_categories`
- Change the code in according to the new structure of `inspire_categories` field

### DIFF
--- a/inspirehep/dojson/hep/model.py
+++ b/inspirehep/dojson/hep/model.py
@@ -29,7 +29,11 @@ from dojson import Overdo
 from inspirehep.utils.helpers import force_force_list
 
 from ..model import FilterOverdo, add_schema, clean_record
-from ..utils import force_single_element, get_record_ref
+from ..utils import (
+    force_single_element,
+    get_record_ref,
+    classify_field,
+)
 
 
 def add_book_info(record, blob):
@@ -53,9 +57,26 @@ def add_book_info(record, blob):
     return record
 
 
+def add_inspire_category_from_arxiv_categories(record, blob):
+    if not record.get('arxiv_eprints') or record.get('inspire_categories'):
+        return record
+
+    record.setdefault('inspire_categories', [])
+    for arxiv_category in record.get_value(record, 'arxiv_eprints.categories', default=[]):
+        inspire_category = classify_field(arxiv_category)
+        if inspire_category:
+            record['inspire_categories'].append({
+                'term': inspire_category,
+                'source': 'arxiv'
+            })
+
+    return record
+
+
 hep_filters = [
     add_schema('hep.json'),
     add_book_info,
+    add_inspire_category_from_arxiv_categories,
     clean_record,
 ]
 
--- a/inspirehep/dojson/hep/model.py
+++ b/inspirehep/dojson/hep/model.py
@@ -27,6 +27,7 @@ from __future__ import absolute_import, division, print_function
 from dojson import Overdo
 
 from inspirehep.utils.helpers import force_force_list
+from inspirehep.utils.record import get_value
 
 from ..model import FilterOverdo, add_schema, clean_record
 from ..utils import (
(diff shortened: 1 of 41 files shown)


inspirehep/dojson/hep/model.py
```diff
@@ -53,9 +57,26 @@ def add_book_info(record, blob):
     return record
 
 
+def add_inspire_category_from_arxiv_categories(record, blob):
+    if not record.get('arxiv_eprints') or record.get('inspire_categories'):
+        return record
+
+    record.setdefault('inspire_categories', [])
+    for arxiv_category in record.get('arxiv_eprints').get('categories'):
```
jacquerie: Every time you chain `get` you should be using `get_value` instead:
```python
get_value(record, 'arxiv_eprints.categories', default=[])
```

rikirenz: I will change it. Anyway `get_value` it's 5 times slower in according to the Jan's comment (https://github.com/inspirehep/inspire-next/blob/master/inspirehep/utils/record.py#L62).

jacquerie: It might be a little slower, but the access string is easily greppable, which is going to be very useful when doing codebase-wide modifications (like the one you're doing): you just have to
```
$ af field_categories.term
```
and you get all the uses.

It's also our antidote to JSON Driven Development: http://okigiveup.net/arguments-against-json-driven-development/.

rikirenz: I ve read the article. Great!

inspirehep/modules/authors/templates/authors/workflows/authorupdate.html
```diff
@@ -122,8 +122,8 @@
       {% endif %}
```
jmartinm: This file seems to be unused nowadays. I will remove it in a different PR.

jmartinm:
Checked the commit (https://github.com/inspirehep/inspire-next/pull/1810/commits/a66c07731851ec24f95c0569b0b442c8f21b6949#diff-335a381ba8a85e073da5ff087ea2ac9cR122) you mention has no tests, and the code LGTM.

tests/unit/dojson/test_dojson_common_base.py
```diff
@@ -127,153 +123,106 @@ def test_field_from_marcxml_650_with_two_a():
...
-
-
-def test_field_categories_from_65017_a_2_9_with_conference():
-    snippet = (
-        '<datafield tag="650" ind1="1" ind2="7">'
-        '  <subfield code="2">INSPIRE</subfield>'
-        '  <subfield code="9">conference</subfield>'
-        '  <subfield code="a">Accelerators</subfield>'
-        '</datafield>'
-    )  # record/1479228
```
jacquerie: Be careful: you should *never* delete tests that use real data in favor of tests that use fake data. These tests are here because they were documenting a regression; if you delete them you allow the regression to happen again.

Sometimes you have to, because the business logic changed so much that the past test is not salvageable (but it's not the case here). In this case, _explain_ why it's safe to delete the test!

tests/unit/literaturesuggest/test_literaturesuggest_tasks.py
```diff
@@ -101,15 +101,17 @@ def test_formdata_to_model_populates_collections_from_type_of_doc_thesis(u, ui):
...
     extra_data = {}
     formdata = {
         'type_of_doc': 'foo',
         'title': [
             'bar',
         ],
-        'categories': 'baz',
+        'arxiv_eprints': {
+            'categories': ['baz']
+        },
```
jacquerie: Here you are cheating: `formdata_to_model` can never have this argument, because such a structure is not part of the form data model. In fact, this is an instance of the problem I mentioned in https://github.com/inspirehep/inspire-next/pull/1885#discussion_r97901456: the DoJSON rules inside of `formdata_to_model` are conceptually wrong.

jacquerie:
@rikirenz: don't act on the above comments, they're there to explain the changes I'm going to do myself in #1885. I just need you to ACK them with a comment or something like that. If you need clarifications we can discuss them tomorrow (that is, today...).

PR #1812, opened 2016-12-12
editor: abstract source suggester

harunurhan:
* Adds a suggest for abstracts.source which is used to autocomplete
from all abstract sources in DB.

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -424,6 +424,11 @@ RECORDS_REST_ENDPOINTS = dict(
                 'inspirehep.modules.records.serializers'
                 ':cvformattext_v1_search'),
         },
+        suggesters=dict(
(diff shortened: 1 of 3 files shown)


jmartinm:
LGTM.

jacquerie:
Since this is obviously a refactor waiting to happen (do we really want to add a receiver for each autocomplete field?) this **must** have unit tests. It's the only way to guarantee that the desired behavior is documented _somewhere_, and that it will be preserved by the refactor.

harunurhan:
@jacquerie  I was about to push the unit tests, but thanks anyway. I think we will have probably more than 10 autocompletion fields and it this file will not look good. I don't what you have in mine but it would be nice to have them somewhere else grouped by type `hep`, `experiments` etc.

PR #1815, opened 2016-12-13
records: suggester enhancers refactoring

jmartinm:
* Aggregates all suggesters under one receiver.

* Separates each enhancement into its own function.

### DIFF
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -47,6 +47,10 @@ from inspirehep.utils.record_getter import get_db_record
 
 from .experiments import EXPERIMENTS_MAP
 from .signals import after_record_enhanced
+from .suggester_enhancers import (
+    populate_abstract_suggester,
+    populate_experiment_name_suggester
+)
 
 
 @models_committed.connect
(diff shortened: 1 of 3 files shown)


jmartinm:
Continuation of https://github.com/inspirehep/inspire-next/pull/1812

@jacquerie @harunurhan have a look to see if that is a better way of organizing the suggesters.

inspirehep/modules/records/receivers.py
```diff
@@ -391,36 +394,8 @@ def check_if_record_is_going_to_be_deleted(sender, *args, **kwargs):
...
-            if source:
-                abstract.update({
-                    'abstract_source_suggest': {
-                        'input': source,
-                        'output': source,
-                    },
-                })
+        populate_abstract_suggester(json)
+    elif 'experiments.json' in json.get('$schema'):
+        populate_experiment_name_suggester(json)
```
jacquerie: I don't see this growing very nicely when more suggesters are added...

jacquerie:
IMO it's a little too early to start refactoring, because we currently have two receivers with no common functionality, and in fact this refactor is just moving things around without abstracting anything.

inspirehep/modules/records/receivers.py
jmartinm: Anyway this PR was WIP to start the discussion and see if we find a better way of organizing that code, but can be closed if we don't see any improvement possible atm.

jacquerie:
I think we agreed before Christmas to wait on this refactor until we have more examples of things we want to refactor.

PR #1829, opened 2016-12-19
tests: add pytest-randomly

jacquerie:
Introduces `pytest-randomly` to randomize the order of execution of
tests, so that we discover bugs hidden by the deterministic execution
of tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -105,6 +105,7 @@ tests_require = [
     'pytest-cov>=1.8.0',
     'pytest-httpretty>=0.2.0',
     'pytest-pep8>=1.0.6',
+    'pytest-randomly>=1.1.2',
     'pytest-selenium>=1.3.1',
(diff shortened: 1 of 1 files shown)


kaplun:
:+1: although it's going to be tricky for the PR hit by such a bug. We shall explain it then to everybody how to behave in case of PR failure (e.g. if it's not a test that you expect to fail, might be due to badly clean up/test design. If it goes away after a rebuild, do not consider yourself safe, but rather open an issue with the test that failed anyway.)

jacquerie:
Interestingly this already uncovered a bug in the `workflows` test suite.

jacquerie:
It looks like `pytest-randomly` and `pytest-flake8` are incompatible. Same goes for `pytest-random-order`, but it's probably a bug in `pytest-flake8`.

PR #1833, opened 2016-12-20
Impactgraphserializer fix

jmartinm:


jmartinm:
First commit fixes a problem when `get_es_records` is called with non string recids. All across the system we treat recids (or control_number) as integers but since the column definition for `pid_value` is `String` the `in_` operator fails when integers are used.

The second commit fixes https://sentry.cern.ch/inspire-sentry/inspire-labs-qa/group/613156/

There are no good tests atm for `get_es_records` or for the impact graph serializer unfortunately.

jmartinm:
Added a test for the impactgraph change.

Not sure if we can add some useful test for the change regarding enforcing list of `str`  in `get_es_records`

tests/unit/api/test_impactgraph.py
```diff
@@ -0,0 +1,60 @@
...
+        'references': [
+            {
+                'recid': 123
+            }
+        ]
+    }
+    with app.app_context():
+        serializer.serialize(111, record)
+
+    g.assert_called_with(
```
jacquerie: I don't like this test : /

This looks like a [Change-Detector](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html) test. I know that what you actually want to test is the fact that `get_es_records` is called with an array of integers, but then that is the thing you should be testing...

jacquerie: I have an alternative idea. I'll add a couple of tests in `tests/integration/test_record.py` to assert that `get_es_records` accepts both list of integers and list of strings, and remove the type comment in the docstring.

jmartinm: Well actually this test was targeting the first commit of the PR where I moved from using the ` OR ` query building to the more efficient `get_es_records`. So it was ensuring that such function was being used to retrieve the references.

jacquerie: But it cannot ever happen that you're not calling `get_es_records`, so this test is never going to fail unless you change the code (that is, it's a Change-Detector).

jacquerie: (My point is: if the implementation of this serializer starts using `more_performant_get_es_records` then the test would start fail, even if the results are exactly the same, just obtained faster, and that is bad...)

jmartinm: Right, I am not sure if there is a better a way of asserting that references are retrieved with `mget` instead of the previous behaviour inside of the impact graph serializer.

Besides that, you can add some different tests as you propose regarding the internals of `get_es_records`, although they are not connected to what I was testing here.

PR #1842, opened 2017-01-05
records: update links for merged records

### LINKED ISSUE #484: Merging records should propagate links

jacquerie:
Closes #484

- [x] refactor `tests/integration/test_record.py`
- [x] test `update_refs` in `inspirehep/modules/records/tasks.py`

### DIFF
--- a/tests/integration/test_record.py
+++ b/tests/integration/test_record.py
@@ -28,10 +28,13 @@ from elasticsearch import RequestError
 from dojson.contrib.marc21.utils import create_record
 from invenio_db import db
 from invenio_pidstore.models import PersistentIdentifier, Redirect
+from invenio_search import current_search_client as es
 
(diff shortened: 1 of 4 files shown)


tests/integration/test_record.py
```diff
@@ -28,304 +28,263 @@
...
+    }
 
-    db.session.commit()
+    with app.app_context():
+        with db.session.begin_nested():
+            merged_uuid = record_upsert(merged_record).id
+            deleted_uuid = record_upsert(deleted_record).id
+            record_upsert(pointing_record)
+        db.session.commit()
+    es.indices.refresh('records-hep')
```
jacquerie: This is why we shouldn't search in ES to do DB modifications: there's no guarantee that they are in sync.

PR #1784, opened 2016-12-05
records: update links for merged records

spirosdelviniotis:
* Adds update references on merge functionality of Inspire record.
* Adds _redirect functionality for Inspire record.

~~Depends on #1790~~

Closes #484 
~~Closes #1660~~
~~Closes #1776~~

Signed-off-by: Spiros Delviniotis spyridon.delviniotis@cern.ch

Issue #1843, opened 2017-01-06
tests: reduce the number of application fixtures


jacquerie:
Every application fixture is a spot in which we have to specify the application configuration, which has already led to (minor) divergences between configurations. Let's kill this before it becomes a problem by having exactly one application fixture per test suite.

If a test requires a certain configuration variable to be set, that should happen in the preamble of the test, not in the fixture.

- [ ] `email_app` and `app` should be merged in `test/unit/conftest.py`
- [ ] `small_app` and `app` should be merged in `tests/integration/conftest.py`

jacquerie:
Touching the difficulty up to hard, because merging `small_app` and `app` requires to change the scope of the fixture from `function` to `session`, which implies some careful work to avoid persisting modifications from one test to the other.

jacquerie:
Commit https://github.com/inspirehep/inspire-next/commit/c97fb948a423a3ac9f09b50010863b19d85d1a13 introduced another application fixture (`workflow_app`).

PR #1844, opened 2017-01-06
tests: set celery eager mode in integration tests

jacquerie:
Spin off of #1784.


jacquerie:
The divergence between this configuration and the one of `small_app` will be fixed by fixing #1843.

PR #1853, opened 2017-01-10
records: fix impact graph serializer

jacquerie:
Closes #1833

### DIFF
--- a/tests/integration/test_serializers.py
+++ b/tests/integration/test_serializers.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


tests/integration/test_serializers.py
```diff
@@ -0,0 +1,43 @@
...
+)
+
+
+def test_impact_graph_serializer_does_not_raise_maximum_recursion_error(app):
+    serializer = ImpactGraphSerializer()
+    record = {
+        'control_number': 111,
+        'earliest_date': '1993-02-02',
+        'references': [
+            {'recid': 4328} for _ in range(sys.getrecursionlimit() + 1)
```
jacquerie: Here's how I solved our discussion in https://github.com/inspirehep/inspire-next/pull/1833#pullrequestreview-15896203: I created a test from a minimal reproduction of the bug on Sentry. I also added tests about the types that `get_es_records` accepts, so we're covered on both sides.

jacquerie: @jmartinm This is the thing I was mentioning at lunch. We could also add a similar test to the other place where this was happening, if you remember where it was...

PR #1854, opened 2017-01-10
global: add flake8

spirosdelviniotis:
* Add flake8 and plug-ins in order to discover useful code warnings.


jacquerie:
You also need to commit the configuration of `flake8` + plugins that you're using*.

* `flake8` plugins: https://pypi.python.org/pypi?%3Aaction=search&term=flake8-&submit=search

spirosdelviniotis:
In order to fix that kind of error: 
`/inspirehep/modules/records/experiments.py:158:5: F999 dictionary key 'astro' repeated with different values`

@jacquerie do you need key duplications such the below example?

```
'astro': 'Astro',
'cro': 'CRO',
'astro': 'astro',
'astro': 'Astro',
```

jacquerie:
> @jacquerie do you need key duplications such the below example?

I didn't understand the question ._.

spirosdelviniotis:
> I didn't understand the question ._.

Sorry for the misunderstanding...
Should we have only one key of any kind?

e.x.
instead of:
```
'astro': 'Astro',
'astro': 'astro',
'astro': 'Astro',
```
Should we transform the code like:
```
'astro': 'Astro'
```

If yes, then we have to delete the duplicates. Otherwise we have to add, in some way, something that ignores the specific lines of code, which is not good practise for the future.

jacquerie:
> Otherwise we have to add, in some way, something that ignores the specific lines of code, which is not good practise for the future.

You are overcomplicating this. Just delete the wrong/duplicated key/value pairs.

jacquerie:
Putting this back to WIP because we will need it after the current sprint. No action needed from you, @spirosdelviniotis.

pytest.ini
```diff
@@ -21,9 +21,15 @@
...
+addopts = --flake8 --ignore=docs --ignore=inspirehep --ignore=node_modules --cov=inspirehep --cov-report=term-missing
+flake8-max-line-length = 99
+flake8-ignore =
     tests/* ALL
+    *.py I100
     *.py E501
+    *.py F401
+    *.py F403
+    *.py F405
+    *.py E251
```
jacquerie: Why are you ignoring and correcting the same errors at the same time?

PR #1785, opened 2016-12-06
hal: rewrite module

jacquerie:
~~Closes #1567~~

~~Depends on #1856~~

PR #1856, opened 2017-01-11
theme: configure Jinja2 to remove whitespace

jacquerie:
Spin off of #1785.

Adds `lstrip_blocks` and `trim_blocks` to Jinja2's configuration, so
that most whitespace is deleted from the result.
See: http://jinja.pocoo.org/docs/dev/templates/#whitespace-control.

As @jmartinm remarked [here](https://github.com/inspirehep/inspire-next/pull/1785#discussion_r91735604), this means that all `{%-` and `-%}` are
now obsolete, and should be removed in a cleanup PR.

### DIFF
--- a/inspirehep/modules/theme/ext.py
+++ b/inspirehep/modules/theme/ext.py
@@ -63,8 +63,10 @@ class INSPIRETheme(object):
 
         app.register_blueprint(blueprint)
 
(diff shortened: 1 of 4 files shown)


jacquerie:
Well, if everything breaks it's my fault.

PR #1861, opened 2017-01-12
setup: added extra deps for web-node

david-caro:
* For now just gunicorn is needed, this simplifies a bit the
  deployment and allows to avoid custom-hidden deps out of the code.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -131,6 +131,9 @@ extras_require = {
     'migration': [
         'invenio-migrator>=1.0.0a6',
     ],
+    'web-node': [
+        'gunicorn',
+    ],
     'crawler': [
         'hepcrawl>=0.2.42',
     ],


jacquerie:
IMO that file should go through a very thorough cleanup, and this is a (very small) step in the opposite direction...

david-caro:
To properly clean it up, you should first make sure it works, no matter how clean it is if it does not work right?

jacquerie:
Sure, but why is this added as an extra? Can't it be just part of `install_requires`? We depend on it in development as well... But then, what about the redundant stuff in [this other](https://github.com/inspirehep/inspire-next/blob/8d994536f0b71f447637166e80b788e0180fe9c0/setup.py#L138-L144) extra?

david-caro:
It's not on the install_requires because for example, when you run it on a crawler node, or a task node you don't need it.

About the other stuff, as you said, it's not clean, the docker command can now be changed to use web_node, and the 'development' looks ok to me (having a dependency listed twice might be better that having an inter-dependency tree that might be less code, but harder to read, debug and maintain).

david-caro:
Dammit, I did not want to close it XD

PR #1866, opened 2017-01-17
docker tests: run a visible instance of firefox

rikirenz:
Adds some parameters in order to see the browser when the acceptance tests are running.


jacquerie:
The Travis build uses the images built by `inspire-docker`, so until https://github.com/inspirehep/inspire-docker/pull/32 is merged this won't have a `firefox` to run.

rikirenz:
Let me know if the code looks good. When everything it will be ok, I will add a commit for the documentation.

PR #1871, opened 2017-01-17
migrator: fix continuous migration

### LINKED ISSUE #1787: Test continuous migration
We have to ensure through integration testing that it works :+1:

jacquerie:
Closes #1864

Also adds integration tests for it (closes #1787).

### DIFF
--- a/tests/integration/test_migrator.py
+++ b/tests/integration/test_migrator.py
@@ -1,0 +1,166 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2016 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


tests/integration/test_migrator.py
```diff
@@ -0,0 +1,166 @@
...
+        object_uuid == PersistentIdentifier.object_uuid).delete()
+
+    db.session.commit()
+
+
+def push_to_redis(record_file):
+    record = pkg_resources.resource_string(
+        __name__, os.path.join('fixtures', record_file))
+
+    redis_url = current_app.config.get('CACHE_REDIS_URL')
```
jacquerie: This is the most notable change: since this method will be executed when an application context is available, we can use `current_app` instead of passing around the `app` fixture.

PR #1879, opened 2017-01-20
workflows: added debug logs to some functions

david-caro:
workflows: added debug logs to some functions

This way is way easier to debug worflows and asynchronous tasks that use them.

### DIFF
--- a/inspirehep/modules/workflows/utils.py
+++ b/inspirehep/modules/workflows/utils.py
@@ -21,12 +21,18 @@
 
 from __future__ import absolute_import, division, print_function
 
-import requests
 import json
+import logging
+import traceback
+from functools import wraps
 
+import requests
 from flask import current_app
 
 from .models import WorkflowsAudit
+
+
+LOGGER = logging.getLogger(__name__)
 
 
 def json_api_request(url, data, headers=None):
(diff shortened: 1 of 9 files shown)


inspirehep/modules/workflows/utils.py
```diff
@@ -80,3 +86,57 @@ def log_workflows_action(action, prediction_results,
...
+            if hasattr(obj, 'log') and hasattr(obj.log, 'debug'):
+                logfn = obj.log.debug
+            elif hasattr(current_app, 'log'):
+                logfn = current_app.log.debug
+            else:
+                logfn = LOGGER.debug
+
+            return logfn
+
+        def _proof_run(func, *args, **kwargs):
```
jmartinm: Since you have already above `func` (as the actual function that wants to be run) this parameter can be confusing.

You are actually testing here the `logfn` right?

inspirehep/modules/workflows/utils.py
```diff
@@ -80,3 +86,57 @@ def log_workflows_action(action, prediction_results,
...
+            try:
+                func(*args, **kwargs)
+            except Exception:
+                LOGGER.debug(
+                    "Error while trying to log with %s:\n%s",
+                    logfn,
```
jmartinm: Where is this defined? Is it read from the closure?

david-caro: It is yes, though it should not (was not my intention), will change.

inspirehep/modules/workflows/utils.py
david-caro: Will change the naming.

yep (if by testing you mean running even with exceptions without failing).

kaplun:
This is not a bugfix, but would really be great to have it in order to debug current and future bugs in the workflows.

inspirehep/modules/workflows/utils.py
kaplun: I am not understanding in which way to fix this. I'd like though to merge this branch...

PR #1880, opened 2017-01-20
workflows: fix continue_workflow callback

david-caro:
Refactor (and fix) callbacks to support robotupload, continue_workflow (generic), and webcoll.

### DIFF
--- a/inspirehep/modules/workflows/models.py
+++ b/inspirehep/modules/workflows/models.py
@@ -35,8 +35,16 @@ class WorkflowsAudit(db.Model):
     # Date that the action was taken
     created = db.Column(db.DateTime, default=datetime.now, nullable=False)
 
-    user_id = db.Column(db.Integer, db.ForeignKey("accounts_user.id"), nullable=True)
-    object_id = db.Column(db.Integer, db.ForeignKey("workflows_object.id"), nullable=False)
+    user_id = db.Column(
+        db.Integer,
+        db.ForeignKey("accounts_user.id"),
+        nullable=True,
+    )
+    object_id = db.Column(
+        db.Integer,
+        db.ForeignKey("workflows_object.id"),
+        nullable=False,
+    )
 
     # Score from model, action taken, recommendation from the model
     score = db.Column(db.Float, default=0, nullable=False)
@@ -50,3 +58,16 @@ class WorkflowsAudit(db.Model):
         """Save object to persistent storage."""
         with db.session.begin_nested():
             db.session.add(self)
+
+
+class WorkflowsPendingRecord(db.Model):
+
+    __tablename__ = "workflows_pending_record"
+
+    workflow_id = db.Column(
+        db.Integer,
+        db.ForeignKey("workflows_object.id", ondelete='CASCADE'),
+        primary_key=True,
+        nullable=False,
+    )
+    record_id = db.Column(db.Integer, nullable=False)
--- a/inspirehep/modules/workflows/views/callback.py
+++ b/inspirehep/modules/workflows/views/callback.py
@@ -28,11 +28,12 @@ import re
 
 from os.path import join
 from flask import Blueprint, jsonify, request, current_app
+from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
 
 from invenio_db import db
 from invenio_workflows import workflow_object_class
 
-from inspirehep.modules.cache import current_cache
+from inspirehep.modules.workflows.models import WorkflowsPendingRecord
 
 
 blueprint = Blueprint(
@@ -47,48 +48,63 @@ blueprint = Blueprint(
 def _get_base_url():
     """Return base URL for generated URLs for remote reference."""
     base_url = current_app.config.get(
-        "LEGACY_ROBOTUPLOAD_URL", "SERVER_NAME"
+        "LEGACY_ROBOTUPLOAD_URL",
+        current_app.config["SERVER_NAME"],
     )
     if not re.match('^https?://', base_url):
         base_url = 'http://{}'.format(base_url)
     return base_url
 
 
-@blueprint.route('/workflows/continue', methods=['POST'])
-def continue_workflow_callback():
-    """Handle callback to continue a workflow.
+def _continue_workflow(workflow_id, recid, result=None):
+    """Small wrapper to continue a workflow.
 
-    Expects the request data to contain a object ID in the
-    nonce field.
+    Will prepare the needed data from the record id and the result data if
+    passed.
+
+    :return: True if succeeded, False if the specified workflow id does not
+        exist.
     """
-    request_data = request.get_json()
-    id_object = request_data.get("nonce", "")
+    result = result if result is not None else {}
+    base_url = _get_base_url()
+    workflow_object = workflow_object_class.get(workflow_id)
+    if not workflow_object:
+        current_app.logger.error(
+            'No workflow object with the id %s could be found.',
+            workflow_id,
+        )
+        return False
 
-    if id_object:
-        callback_results = request_data.get("results", {})
-        workflow_object = workflow_object_class.get(id_object)
-        if workflow_object:
-            results = request_data.get("results", [])
-            for result in results:
-                status = result.get('success', False)
-                if status:
-                    recid = result.get('recid')
-                    base_url = _get_base_url()
-                    workflow_object.extra_data['url'] = join(
-                        base_url,
-                        'record',
-                        str(recid)
-                    )
-                    workflow_object.extra_data['recid'] = recid
-            # Will add the results to the engine extra_data column.
-            workflow_object.save()
-            db.session.commit()
-            workflow_object.continue_workflow(
-                delayed=True,
-                callback_results=callback_results
-            )
-            return jsonify({"result": "success"})
-    return jsonify({"result": "failed"})
+    workflow_object.extra_data['url'] = join(
+        base_url,
+        'record',
+        str(recid)
+    )
+    workflow_object.extra_data['recid'] = recid
+    workflow_object.extra_data['callback_result'] = result
+    workflow_object.save()
+
+    workflow_object.continue_workflow(delayed=True)
+    return True
+
+
+def _find_and_continue_workflow(workflow_id, recid, result=None):
+    workflow_found = _continue_workflow(
+        workflow_id=workflow_id,
+        recid=recid,
+        result=result,
+    )
+    if not workflow_found:
+        current_app.logger.warning(
+            'The workflow %s was not found.',
+            workflow_id,
+        )
+        return {
+            'status': 'failed',
+            'message': 'workflow with id %s not found.' % workflow_id,
+        }
+
+    return {'status': 'succeeded'}
 
 
 @blueprint.route('/workflows/webcoll', methods=['POST'])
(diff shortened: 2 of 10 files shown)


david-caro:
@kaplun I added error messages and such if the callback contains unexpected data.

Tested on qa

inspirehep/modules/workflows/views/callback.py
```diff
@@ -58,37 +59,87 @@ def _get_base_url():
...
-    Expects the request data to contain a object ID in the
-    nonce field.
+    Expects the request data to contain an object ID in the
+    nonce field. For a full reference of the results data structure see:
+        http://inspirehep.net/help/admin/bibupload-admin-guide#3.7
+
+    From which we only expect only one result (as only one record can be
+    managed at a time in the holding pen).
     """
+    def _send_error_message(id_object, error_message):
```
jmartinm: Why sending emails manually from here? I think it's better to use logger.{error/warning/info} which will end up in Sentry (and email will be sent anyway)

inspirehep/modules/workflows/views/callback.py
```diff
@@ -97,7 +148,13 @@ def webcoll_callback():
 
     Expects the request data to contain a list of record ids in the
     recids field.
+
+    DEPRECATED: replaced by the simpler continue_workflow_callback
```
jmartinm: The reason why one needs to wait for `webcoll` to run in legacy is because otherwise the user will receive an email saying: You can visit the record in `http://inspirehep.net/record/123` and when clicking the link the record will not be available.

Only after `webcoll` has run the record is really accessible to the user.

inspirehep/modules/workflows/views/callback.py
david-caro: Ok, just reused the same code it was being used by the `robotupload_callback`

inspirehep/modules/workflows/views/callback.py
david-caro: I know, we discussed this (sam and I) and though that simplicity is better as first step (webcoll runs each 5min, and there's a nice message saying that the record is there but will take a moment to show up)

jmartinm: Well the code was not there at the beginning, then people were confused so we had to add it.

I think it's a basic usability question here. If one receives an email saying that the record is ready and a link to it, it's expected to be there (no need to overcomplicate it saying it will be there eventually). Since it is working nowadays, I don't see a reason to go backwards.

And we know by past experience that 'it runs every 5 minutes' it's quite an illusion. There are many times things happening that slow down the task queue so this time frame is longer.

david-caro: Problem was that it's not working nowadays...

david-caro: I'm a bit confused too, when you say that "then people were confused so we had to add it" you mean that this code has been running somewhere?

jmartinm: You mean on labs.inspirehep.net ? Have not heard any complaints on that

jmartinm: It works on a daily basis on labs.inspirehep.net (https://github.com/inspirehep/inspire-next/blob/production/inspirehep/modules/workflows/views/callback.py) - see `production` branch. Since the code on legacy has not changed, maybe something is not right in the `master` branch and needs fixing.

david-caro: Ok, I'll try to fix then the current callbacks. Do you have any idea why does it use redis? (IMO the queue of pending records should be in a reliable storage, not in cache)

david-caro: btw. that change requires changing the inspire-crawler code to use the robotupload callback instead of continue_workflow right? Or that is part of another flow?

kaplun: @david-caro Ideally it could sit in a new SQL table, if you want to take the pain to port it to it :+1: I guess redis had been chosen because this was anyway a quick hack.

david-caro:
@jmartinm @kaplun still have to change the workflows to adapt to this (and config to use robotupload instead of workflow continue directly), and move from redis to db (to have persistency) but let me know if it looks ok for you the structure now.

inspirehep/modules/workflows/views/callback.py
```diff
@@ -99,26 +174,41 @@ def webcoll_callback():
...
+    results = {}
+    for recid in recids:
+        recid = str(recid)
+        if recid in results:
+            current_app.logger.warning('Received duplicated recid: %s', recid)
+            continue
+
+        if recid not in pending_records:
+            current_app.logger.warning(
+                'The record %s was not found on the pending list.',
```
jmartinm: This might not be a warning. Here you will receive all the recids that get added to production so it will be quite common that the recids are not in the `pending_records`.

inspirehep/modules/workflows/views/callback.py
```diff
@@ -99,26 +174,41 @@ def webcoll_callback():
...
             )
-    return jsonify({"result": "success"})
+            results[recid] = {
+                'status': 'failed',
+                'message': 'recid not in pending list',
+            }
+            continue
+
+        workflow_id = pending_records[recid]
+        results[recid] = _find_and_continue_workflow(
```
jmartinm: So is `/workflows/continue` still going to be used or there will be only this function left? If you send a `robotupload` and on this callback you continue the workflow, then the user will be notified before the record is really available (only available after `/workflows/webcoll`). I guess you are still missing modifying the `article` workflow so that it waits after this.

inspirehep/modules/workflows/views/callback.py
david-caro: What are the other ingestion flows?

inspirehep/modules/workflows/views/callback.py
david-caro: No, the continue callback is not used anymore in the current flows... hmm, I'll better just remove it, no need to maintain code that we don't use.

inspirehep/modules/workflows/views/callback.py
kaplun: Well records are being added directly on legacy, for quite some time, so webcoll will communicate back about new pure legacy records that labs is not aware of.

david-caro:
Please ignore the travis push check... I pushed the branch to origin instead of david-caro remote... (already deleted it).

@jmartinm U've used debug for the 'record was not found', though I'm still using warning for the duplicated entries. Should we handle those in any other way? (right now it just logs and continues without removing them or continuing any records).

Sill have to test and adapt the workflows, but let me know if this is ok.

inspirehep/modules/workflows/models.py
```diff
@@ -50,3 +58,16 @@ def save(self):
         """Save object to persistent storage."""
         with db.session.begin_nested():
             db.session.add(self)
+
+
+class WorkflowsPendingRecord(db.Model):
+
+    __tablename__ = "workflows_pending_record"
+
+    workflow_id = db.Column(db.Integer, primary_key=True, nullable=False)
```
jmartinm: Shall we make this a foreign key with CASCADE? So that if workflow is deleted the entries in this table are deletes as well.

inspirehep/modules/workflows/models.py
```diff
@@ -50,3 +58,16 @@ def save(self):
...
+    record_id = db.Column(db.Integer, nullable=False)
+
+    def save(self):
```
jmartinm: I see it is also on the model above, but I am not sure why this is needed. On models in other modules we don't have a dedicated `save` function - https://github.com/inveniosoftware/invenio-accounts/blob/master/invenio_accounts/models.py

inspirehep/modules/workflows/views/callback.py
```diff
@@ -54,41 +54,57 @@ def _get_base_url():
...
-            # Will add the results to the engine extra_data column.
-            workflow_object.save()
-            db.session.commit()
-            workflow_object.continue_workflow(
-                delayed=True,
-                callback_results=callback_results
-            )
-            return jsonify({"result": "success"})
-    return jsonify({"result": "failed"})
+    base_url = _get_base_url()
```
jmartinm: Looking into `_get_base_url()` there seems to be a problem at the beginning:

```python
base_url = current_app.config.get(
        "LEGACY_ROBOTUPLOAD_URL", "SERVER_NAME"
    )
```

The string `"SERVER_NAME"` will be used if `LEGACY_ROBOTUPLOAD_URL` is not defined. Seems like the intended behaviour was to fallback to the other config key.

inspirehep/modules/workflows/views/callback.py
```diff
@@ -54,41 +54,57 @@ def _get_base_url():
...
+            workflow_id,
+        )
+        return False
+
+    workflow_object.extra_data['url'] = join(
+        base_url,
+        'record',
+        str(recid)
+    )
+    workflow_object.extra_data['recid'] = recid
```
jmartinm: So in all cases that you call `_continue_workflow` you will have a recid available?

inspirehep/modules/workflows/views/callback.py
```diff
@@ -99,26 +115,51 @@ def webcoll_callback():
...
-            objectid = pending_records[rid]
-            workflow_object = workflow_object_class.get(objectid)
-            base_url = _get_base_url()
-            workflow_object.extra_data['url'] = join(
-                base_url, 'record', str(rid)
+    results = {}
+    for recid in recids:
+        recid = int(recid)
+        if recid in results:
+            current_app.logger.warning('Received duplicated recid: %s', recid)
```
jmartinm: I guess its fine to have a `warning` here. Not sure when a duplicate recid would be sent (could be a bug on legacy code if it happens).

inspirehep/modules/workflows/views/callback.py
```diff
@@ -99,26 +115,51 @@ def webcoll_callback():
...
-                "pending_records",
-                pending_records,
-                timeout=current_app.config["PENDING_RECORDS_CACHE_TIMEOUT"]
+            results[recid] = {
+                'status': 'failed',
+                'message': 'Recid not in pending list.',
+            }
+            continue
+
+        if not len(pending_records) > 1:
```
jmartinm: So you expect that the `WorkflowsPendingRecord` table will not have more than one recid attached to a workflow object. Maybe in the future that is not the case? One workflow could send a couple of `robotupload` and save for the same workflow id a couple of recid.

If you want to put that restriction then above you can do:

```python
pending_records = WorkflowsPendingRecord.query.filter_by(
            record_id=recid,
).one()
```

Instead of using `all()`

inspirehep/modules/workflows/models.py
david-caro: Just copied it from above, not sure why it's needed. Maybe it's a way to avoid having to run the db.session.add explicitly?

inspirehep/modules/workflows/models.py
david-caro: Can we do something similar with the record id? (if so, what should be the foreign key fqdn?)

inspirehep/modules/workflows/views/callback.py
david-caro: Good catch! Will check the config.get docs, see if there's a way to do that nicely.

inspirehep/modules/workflows/views/callback.py
david-caro: Yep, it's a required parameter, and for our current functionality, it's all we need (it's not meant to be a generic continue_workflow function).

inspirehep/modules/workflows/views/callback.py
david-caro: Currently the workflows are not able (in any case) to handle more than one record, allowing multiple will lead to some refactoring on all the code (inspire-workflows{,-files,-ui} + inspire-next).

I don't think that in the near future we are going to have any multi-record workflows, and very hopefully, that will be after migrating to invenio3 everything, and this extra table will not be needed 🤞

david-caro:
@kaplun @jmartinm This is ready to get merged, tested on qa.

david-caro:
It's not, the waits are ignored (skipped) in non-production mode.
Though it would be nice to do somehow with a mock service (to make sure that it actually goes through all the stack, instead of just calling the 'continue' function)

kaplun:
@david-caro this afternoon you were working on a way to test this. I assume this will be part of this PR?

david-caro:
Id I can get it :/, it's taking me a lot of time and I'm still really puzzled by some behaviors... (and the already known complexity to debug workflows).

kaplun:
Is there a way I can encourage you? A :beer: ?

david-caro:
I have northerner blood, you can just say 'A que no hay 🥚 🥚 ?', and I'll just do it...

david-caro:
btw. that does not take back the :beer: offering ;)

kaplun:
With pleasure :beers:

david-caro:
Okok, just added/fixed the workflow tests to include the callbacks.... way more complicated than expected, but at least now we have a starting point to test more complex flows.

There's still some 'leftovers' to be cleaned up or reused from what I guess was a previous attempt to add tests (for example, the fixtures for accepted/rejected records).

Though I think that this pr is big enough for now and that should be done later.

PR #1888, opened 2017-01-26
setup: ensure correct MANIFEST.in

kaplun:


jacquerie:

PR #1890, opened 2017-01-26
Acquisition orcid

rikirenz:
[xml_to_json] Acquisition of the ORCID in according to the schema format: 
https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/elements/acquisition_source.json#L14-L17

[json_to_marcjson] this change needs explation


rikirenz:
@kaplun https://github.com/inspirehep/inspire-next/pull/1890/commits/5fe5aaae4a0888dcd2e0e113ce6b8098f3460c41#diff-9bb8893bd35189910fd4c7730329203eR81

Here I do not know where to put the `orcid` field when we convert JSON to JSON_marc. Because when I convert from XML to JSON I take the `orcid` from the field `a`. But when I convert from JSON to JSON_marc the key value is already taken from the field `source`.

What should I do?

PR #1896, opened 2017-01-31
tests: rewrite unit/dojson/common/base

jacquerie:


kaplun:
As we were discussing IRL the other day, it would be great to also test the full pipeline: JSON->Pythonic-MARCXML->serialized MARCXML, in order to be sure that the MARCXML produced is really what we expect it to be (without issues of `bool` and `list` that are not handled properly).
Not sure if this can already start to be part of this PR or separate one.

inspirehep/dojson/utils/__init__.py
```diff
@@ -224,3 +227,8 @@ def dedupe_all_lists(obj):
         return type(obj)(new_obj)
     else:
         return obj
+
+
+def validate(instance, schema):
+    return jsonschema_validate(
+        instance, schema, resolver=LocalRefResolver('', {}))
```
rikirenz: @david-caro was telling that the one responsible for the validation of the JSON Schema should be the `inspire_schema`. Probably we should move these lines in this method https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/api.py#L32 in according to the behavior that we need. What do you think?

jacquerie: @david-caro is 100% right, this method should live in `inspire_schemas.api`. But I have a conundrum: I'm improving the tests so that I can update `inspire_schemas` more easily, but this method will come in a future version of `inspire_schemas`, so I'm deadlocked!

To break the deadlock, I contract a little technical debt here by introducing this ugly little thing, with the intention of deleting it once the schemas are upgraded.

jacquerie:
> Not sure if this can already start to be part of this PR or separate one.

Separate one. Part of what you say is already tested in https://github.com/inspirehep/inspire-next/blob/0dbfc8d8fbbf1cbef7573f76b31331fc6ed8b998/tests/unit/dojson/test_dojson_utils.py#L233-L261, though.

inspirehep/dojson/utils/__init__.py
kaplun: What about using always bleading edge `inspire-schemas`?

kaplun:
> Separate one. Part of what you say is already tested by https://github.com/inspirehep/inspire-next/blob/0dbfc8d8fbbf1cbef7573f76b31331fc6ed8b998/tests/unit/dojson/test_dojson_utils.py#L233-L261, though.

Yep part. I was thinking about something more realistic with complex output.

inspirehep/dojson/utils/__init__.py
jacquerie: The plan is to write as many tests as possible for HEP and HEPNames that also validate the produced schema. These make sense already _now_, so I want to tackle one problem at a time.

jacquerie:
> I was thinking about something more realistic with complex output.

Sure, can do. The idea is to test the key requirements (`bool` conversion, etc...) as simple unit tests, and something more complex as an integration test.

PR #1897, opened 2017-01-31
theme: add record preview endpoint

jmartinm:
* New endpoint that receives a JSON record and renders the detailed
  view. This preview is useful, e.g. for the record-editor.

### DIFF
--- a/inspirehep/modules/theme/views.py
+++ b/inspirehep/modules/theme/views.py
@@ -47,6 +47,7 @@ from invenio_search import current_search_client
 
 from inspirehep.modules.pidstore.utils import (
     get_endpoint_from_pid_type,
+    get_endpoint_from_schema,
     get_pid_type_from_endpoint,
 )
 from inspirehep.modules.records.conference_series import (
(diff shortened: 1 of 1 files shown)


inspirehep/modules/theme/views.py
```diff
@@ -658,6 +659,19 @@ def linkedaccounts():
 
 
 #
+# Record preview
+#
+
+@blueprint.route('/record/preview', methods=['POST'])
```
jacquerie: Is it a problem to make it more RESTy? Like: `/record/<int:recid/preview`. Also, the HTTP method should semantically be `GET`.

jmartinm: The thing here is that the `recid` parameter would be unused if passed, since there is no fetching being done based on that `recid`. Instead of that, the whole record is passed from the editor.

Also we might want to use this endpoint for non-existing records, doing a preview from a Holding Pen entry (which does not have recid yet) - although this might break some assumptions in our current detailed view.

> HTTP method should semantically be GET

Indeed, but since the whole JSON needs to be passed in the request from the editor, adding the whole JSON in a URL query param might get truncated.

(just for reference, this is [the line](https://github.com/inspirehep/record-editor/pull/109/files#diff-bef73574144e5725cded1a7ef2bed4c9R48) where I am using this endpoint at the moment)

jmartinm:
I will move this view to https://github.com/inveniosoftware-contrib/invenio-record-editor

https://github.com/inspirehep/inspire-next/pull/1900 is a preparatory commit.

PR #1899, opened 2017-02-01
tests: split and improve unit/dojson/hep

jacquerie:
Continues the program started in #1896: test all the things before we break all the things with the new schemas.

- [ ] Fix bug in `book_series` field when converting from MARCXML to JSON (see tests).
- [ ] Fix bug in `collaboration` field when converting from JSON to MARCXML (see tests).


jacquerie:
Splits `unit/dojson/hep` in several files patterned over the same
structure as inspirehep/dojson/hep/fields.

Also improves the tests by making them assert the validity of the
schema produced and the fact that the `hep` and `hep2marc` models are
(roughly) the inverse of each other.

Cases in which the DoJSON rule performs a non-reversible normalization
of the input are marked with a comment such as `# no roundtrip`. Tests
`xfail`ed with the reason `wrong roundtrip` are missing some data when
converting from JSON to MARCXML, while the reason `wrong conversion`
denotes that even the first conversion from MARCXML to JSON is producing
bad results. Finally, the reason `missing roundtrip` means that there
was no rule that converted from JSON to MARCXML.

tests/unit/dojson/test_dojson_hep_bd6xx.py
```diff
@@ -0,0 +1,284 @@
...
+
+    if not xml_snippet.strip().startswith('<record>'):
+        xml_snippet = '<record>%s</record>' % xml_snippet
+
+    json_data = hep.do(create_record(xml_snippet))
+    json_experiments = json_data['accelerator_experiments']
+    marc_experiments = hep2marc.do(json_data)['693']
+
+    assert marc_experiments == expected_marc
+    assert json_experiments == expected_json
```
jacquerie: This test was not updated, just migrated as-is from where it was.

PR #1903, opened 2017-02-01
tests: update/add tests for inspire-schemas v4.0

spirosdelviniotis:

### DIFF
--- a/tests/unit/dojson/test_dojson_hep_bd01x09x.py
+++ b/tests/unit/dojson/test_dojson_hep_bd01x09x.py
@@ -593,6 +593,47 @@ def test_report_numbers_from_037__a():
     assert expected == result['037']
 
 
+def test_report_numbers_from_two_037__a():
+    schema = load_schema('hep')
+    subschema = schema['properties']['report_numbers']
+
+    snippet = (
+        '<record>'
+        '  <datafield tag="037" ind1=" " ind2=" ">'
+        '    <subfield code="a">UTPT-89-27</subfield>'
+        '  </datafield>'
+        '  <datafield tag="037" ind1=" " ind2=" ">'
+        '    <subfield code="a">CALT-68-1585</subfield>'
+        '  </datafield>'
+        '</record>'
+    )  # record/26564
+
+    expected = [
+        {
+            'value': 'UTPT-89-27',
+        },
+        {
+            'value': 'CALT-68-1585',
+        },
+    ]
+    result = hep.do(create_record(snippet))
+
+    assert validate(result['report_numbers'], subschema) is None
+    assert expected == result['report_numbers']
+
+    expected = [
+        {
+            'a': 'UTPT-89-27',
+        },
+        {
+            'a': 'CALT-68-1585',
+        },
+    ]
+    result = hep2marc.do(result)
+
+    assert expected == result['037']
+
+
 def test_languages_from_041__a():
     schema = load_schema('hep')
     subschema = schema['properties']['languages']
--- a/tests/unit/dojson/test_dojson_hep_bd2xx.py
+++ b/tests/unit/dojson/test_dojson_hep_bd2xx.py
@@ -131,3 +131,68 @@ def test_titles_from_245__a_b():
     result = hep2marc.do(result)
 
     assert expected == result['245']
+
+
+def test_title_translations_from_242__a():
+    schema = load_schema('hep')
+    subschema = schema['properties']['title_translations']
+
+    snippet = (
+        '<datafield tag="242" ind1=" " ind2=" ">'
+        '  <subfield code="a">The redshift of extragalactic nebulae</subfield>'
+        '</datafield>'
+    )  # record/8352
+
+    expected = [
+        {
+            'language': 'en',
+            'title': 'The redshift of extragalactic nebulae',
+        },
+    ]
+    result = hep.do(create_record(snippet))
+
+    assert validate(result['title_translations'], subschema) is None
+    assert expected == result['title_translations']
+
+    expected = [
+        {
+            'a': 'The redshift of extragalactic nebulae',
+        },
+    ]
+    result = hep2marc.do(result)
+
+    assert expected == result['242']
+
+
+def test_title_translations_from_242__a_b():
+    schema = load_schema('hep')
+    subschema = schema['properties']['title_translations']
+
+    snippet = (
+        '<datafield tag="242" ind1=" " ind2=" ">'
+        '  <subfield code="a">Generalized Hamilton-Jacobi Formalism</subfield>'
+        '  <subfield code="b">Field Theories with Upper-Order Derivatives</subfield>'
+        '</datafield>'
+        )  # record/1501064
+
+    expected = [
+        {
+            'language': 'en',
+            'title': 'Generalized Hamilton-Jacobi Formalism',
+            'subtitle': 'Field Theories with Upper-Order Derivatives',
+        },
+    ]
+    result = hep.do(create_record(snippet))
+
+    assert validate(result['title_translations'], subschema) is None
+    assert expected == result['title_translations']
+
+    expected = [
+        {
+            'a': 'Generalized Hamilton-Jacobi Formalism',
+            'b': 'Field Theories with Upper-Order Derivatives',
+        },
+    ]
+    result = hep2marc.do(result)
+
+    assert expected == result['242']
--- a/tests/unit/dojson/test_dojson_hep_bd2xx.py
+++ b/tests/unit/dojson/test_dojson_hep_bd2xx.py
@@ -1,0 +1,133 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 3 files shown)


tests/unit/dojson/test_dojson_hep.py
```diff
@@ -1630,4 +1716,78 @@ def test_mashed_publication_info_from_773():
...
+    )  # record/1509456/export/xm
+
+    expected = {
+        'value': '978-1-78634-110-5',
+        'medium': 'print',
+    }
+    result = hep.do(create_record(snippet))
+
+    assert validate(result['isbns'], subschema) is None
+    assert expected == result['isbns'][0]
```
jacquerie:
```python
assert expected == result['isbns']
```
and above
```python
expected = [
    {
        ...
    },
]
```

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -44,9 +49,20 @@ def test_doi_but_should_be_hdl_from_0247_a():
     }]
     result = hep.do(create_record(snippet))
 
+    assert validate(result['persistent_identifiers'], subschema) is None
     assert result.get('persistent_identifiers', []) == expected
     assert result.get('dois', []) == []
```
jacquerie:
```python
assert expected == result['persistent_identifiers']
```

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -1630,4 +1716,78 @@ def test_mashed_publication_info_from_773():
...
+        {
+            'a': '978-0-19-875971-3',
+        },
+    ]
+    result = hep2marc.do(result)
+
+    assert expected == result['020']
+
+
+def test_isbns_from_020():
```
jacquerie: `test_isbns_from_020__a_b_normalizes_print`

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -1630,4 +1716,78 @@ def test_mashed_publication_info_from_773():
...
+
+def test_isbns_from_020_a_only():
+    schema = load_schema('hep')
+    subschema = schema['properties']['isbns']
+
+    snippet = (
+        '<datafield tag="020" ind1=" " ind2=" ">'
+        '  <subfield code="a">9780198759713</subfield>'
+        '</datafield>'
+    )  # record/1510325/export/xm
```
jacquerie: `/export/xm` is implied by the fact that this is MARCXML, so it can be dropped here.

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -1630,4 +1716,78 @@ def test_mashed_publication_info_from_773():
...
+            'v': '354',
+            'y': 2008
+        }
+    ]
+    result = hep2marc.do(result)
+
+    assert expected == result['773']
+
+
+def test_isbns_from_020_a_only():
```
jacquerie: `test_isbns_from_020__a`

jacquerie:
Will be merged in #1899 when the comments are addressed.

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -2040,3 +2040,37 @@ def test_title_from_245__a_b():
...
+    assert expected == result['titles']
+
+    expected = [
+        {
+            'a': 'C* algebras and differential geometry',
+        },
+    ]
+    result = hep2marc.do(result)
+
+    assert expected == result['245']
```
spirosdelviniotis: Do we need this case?
As I realized, we don't have dojson rule from json to MARCXML for 247 field.

kaplun: On production we only have 6(!) records with 247. @michamos I think this are actually outliers and they should have used 246, right? (In the TWiki it says deprecated but to me looks like the other way round).

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -1609,25 +1614,42 @@ def test_mashed_publication_info_from_773():
...
 
-    assert expected == result['publication_info'][0]
+    assert validate(result['publication_info'], subschema) is None
+    assert expected == result['publication_info']
+
+    expected = [
+        {
+            'c': ['45'],
+            'p': 'IAU Symp.',
+            '0': 1408366,
```
jacquerie: I don't think we should be producing guys like these, because they come from the `xme` format, while here we are producing the `xm` format. This is a common problem in the MARCXML -> JSON rules as they are currently written.

CC: @kaplun

tests/unit/dojson/test_dojson_hep.py
```diff
@@ -2203,3 +2203,133 @@ def test_report_numbers_from_two_037__a():
...
+        '  <subfield code="2">INSPIRE</subfield>'
+        '</datafield>'
+    )  # record/363605
+
+    expected = [
+        {
+            'keyword': 'programming: Monte Carlo',
+            'classification_scheme': 'INSPIRE',
+        },
+    ]
```
jacquerie: Watch out! Because you are not wrapping the two `<datafield>` tags with a `<record>` tag `create_record` only outputs the first.

tests/unit/dojson/test_dojson_hep.py
kaplun: @jacquerie : you are right. In general we should not export back these records IDs.

jacquerie:
These tests need to be split following what was done in #1899. @michamos, can you check them?

tests/unit/dojson/test_dojson_hep.py
michamos: Result of the discussion with @kaplun: we should write in MARCXML all IDs that exist in the XM format (those not marked XME only in https://twiki.cern.ch/twiki/bin/view/Inspire/DevelopmentRecordMarkup )

spirosdelviniotis:
Just re based !

tests/unit/dojson/test_dojson_hep_bd2xx.py
```diff
@@ -136,3 +136,37 @@ def test_titles_from_245__a_b():
     result = hep2marc.do(result)
 
     assert expected == result['245']
+
+
+def test_titles_from_245_and_247__a_9():
```
michamos: we only had 6 records in HEP with 247 content, which I all changed to 247, so this test can be removed.

tests/unit/dojson/test_dojson_hep_bd2xx.py
```diff
@@ -0,0 +1,138 @@
...
+
+    expected = [
+        {
+            'a': 'Proceedings, New Observables in Quarkonium Production',
+            'b': 'Trento, Italy',
+        },
+    ]
+    result = hep2marc.do(result)
+
+    assert expected == result['245']
```
michamos: You should add a test for the case where the two titles in `245` and `246` are different, e.g. http://inspirehep.net/record/36035. In that case, the `245` goes to the first element of `titles`, the `246`s go next.

tests/unit/dojson/test_dojson_hep_bd2xx.py
```diff
@@ -170,3 +170,68 @@ def test_titles_from_245_and_247__a_9():
     result = hep2marc.do(result)
 
     assert expected == result['245']
+
+
+def test_title_translations_from_242__a():
```
michamos: This test is wrong: if there is a `242`, `245` goes to `translated_titles` with the right language, whereas `242` goes to titles.

tests/unit/dojson/test_dojson_hep_bd2xx.py
```diff
@@ -170,3 +170,68 @@ def test_titles_from_245_and_247__a_9():
...
+        {
+            'a': 'The redshift of extragalactic nebulae',
+        },
+    ]
+    result = hep2marc.do(result)
+
+    assert expected == result['242']
+
+
+def test_title_translations_from_242__a_b():
```
michamos: Same comment.

tests/unit/dojson/test_dojson_hep_bd01x09x.py
```diff
@@ -593,6 +593,39 @@ def test_report_numbers_from_037__a():
...
+    snippet = (
+        '<datafield tag="037" ind1=" " ind2=" ">'
+        '  <subfield code="a">UTPT-89-27</subfield>'
+        '</datafield>'
+        '<datafield tag="037" ind1=" " ind2=" ">'
+        '  <subfield code="a">CALT-68-1585</subfield>'
+        '</datafield>'
+    )  # record/26564
+
+    expected = [
```
michamos: You shouldn't discard report numbers! `report_numbers` is a list for a good reason: so that we can place several values inside :wink:
You should put both in `expected` and make sure they round-trip.

tests/unit/dojson/test_dojson_hep_bd6xx.py
```diff
@@ -282,3 +284,134 @@ def test_accelerator_experiments(mock_get_record_ref, mock_get_recid_from_ref,
...
+        '  <subfield code="a">programming: Monte Carlo</subfield>'
+        '  <subfield code="2">INSPIRE</subfield>'
+        '</datafield>'
+        '<datafield tag="695" ind1=" " ind2=" ">'
+        '  <subfield code="a">electron positron: annihilation</subfield>'
+        '  <subfield code="2">INSPIRE</subfield>'
+        '</datafield>'
+    )  # record/363605
+
+    expected = [
```
michamos: again, you are discarding half of the keywords.

tests/unit/dojson/test_dojson_hep_bd2xx.py
michamos: actually, forget about this, just wrap everything with more than one `<datafield>` into a `<record>` and all titles should be preserved.

michamos:
@jacquerie @kaplun LGTM

michamos:
@jacquerie @kaplun LGTM

PR #1909, opened 2017-02-07
workflow: doctype for arXiv PDFs in FFT

kaplun:
* FIX Uses the correct FFT doctype `arXiv` when sending to legacy
  PDFs originated from arXiv. This is necessary in order to have
  legacy hide them.


kaplun:
As discussed IRL with @david-caro this is necessary in order not to disclose arXiv files attached on legacy.

kaplun:
Actually I should add a regression test.

PR #1911, opened 2017-02-08
docs: update install guide

Kjili:
Remove pip from package list as it is installed automatically within virtualenv and should not be required globally.

Add Arch Linux packages and alternatives for installing npm globally inside the virtualenv and setting up the virtualenv independently from the git repository. One use case where this comes in handy is when switching from Python 2.7 to Python 3 since you can set up two environments and switch between them as you need while working on the same code.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -7,7 +7,6 @@ System prerequisites
 
 This guide expects you to have installed in your system the following tools:
 
-* pip
 * git
 * virtualenv
 * virtualenvwrapper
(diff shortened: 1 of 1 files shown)


INSTALL.rst
```diff
@@ -23,21 +22,29 @@ CentOS
...
+will need also to add the following workarounds:
 
 .. code-block:: console
 
     $ sudo usermod -a -G $USER elasticsearch
     $ newgrp elasticsearch  # or log out and in again
     $ sudo ln -s /etc/elasticsearch /usr/share/elasticsearch/config
 
+Arch Linux
+----------
```
jacquerie: As much as I like Arch Linux, we don't have the resources to (officially) support more than one distro.

INSTALL.rst
```diff
@@ -123,6 +148,12 @@ We build assets using `npm`. Make sure you have installed it system wide.
     still setup a per-user npm modules installation that will allow you to
     install/remove modules as normal user. You can find more info `in the npm
     docs here`_.
+    
+    If you want to install npm packages directly in your virtualenv
+    just add `NPM_CONFIG_PREFIX=$VIRTUAL_ENV`
+    in the postactivate file of your virtualenv folder and you will be able to run
+    `npm install -g node-sass@3.8.0 clean-css requirejs uglify-js`
+    from inside your virtual environment.
```
jacquerie: Oh, I like this! I was actually polluting my global config with a `NPM_CONFIG_PREFIX=$HOME/.npm`.

INSTALL.rst
```diff
@@ -49,12 +56,30 @@ Create a virtual environment and clone the INSPIRE source code using `git`:
 
 .. code-block:: console
 
-    $ mkvirtualenv inspirehep
+    $ mkvirtualenv -p /usr/bin/python2.7 inspirehep
```
jacquerie: It looks like you can also do `--python=python2.7`, which looks a bit more robust.

david-caro: that will get messed up if you are already in a venv (happened to me at some point when playing with different compilations of python)

david-caro: In any case... I guess that it's safer if you don't really do strange things (like I do :S)

jacquerie: Ah, right, didn't think about that case... On the other hand in the context of the first installation that shouldn't happen.

PR #1914, opened 2017-02-13
Refactor formdata to model

rikirenz:
This PR is using the `literature_builder` class that we are developing here: https://github.com/inspirehep/inspire-schemas/pull/81

It also upgrade the schemas to the version ` inspire-schemas-19.0.0`. This version of the schema should include the builder. For this reason the `formdata_to_model` tests and other tests should fail.


inspirehep/modules/literaturesuggest/dojson/fields/literature.py
```diff
@@ -0,0 +1,337 @@
...
+@literature.over('additional_url', '^additional_url$')
+def additional_url(self, key, value):
+    field = {
+        "value": value
+    }
+    if 'urls' in self:
+        self['urls'].append(field)
+    else:
+        self['urls'] = [field]
+    raise IgnoreKey
```
rikirenz: this is wrong I am going to clean this commit

inspirehep/modules/literaturesuggest/dojson/fields/literature.py
```diff
@@ -39,271 +39,270 @@
...
-    repnums = []
-    for repnum in value:
-        repnums.append(dict(value=repnum.get('report_number')))
-    if 'report_numbers' in self:
-        self['report_numbers'].extend(repnums)
-    else:
-        self['report_numbers'] = repnums
-    raise IgnoreKey
-
+# @literature.over('abstracts', '^abstract$')
```
kaplun: Please, kill this whole folder since it's not going to be used anymore.

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,220 @@
...
+                     ('ita', 'it'),
+                     ('spa', 'es'),
+                     ('chi', 'zh'),
+                     ('por', 'pt'),
+                     ('oth', 'ot')]
+
+        if language == 'oth' and other_language:
+            return other_language
+
+        return unicode(dict(languages).get(language))
```
kaplun: Why this `unicode`?

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,220 @@
...
-    # Title source and cleanup
-    # ============================
-    try:
-        # Clean up all extra spaces in title
-        data['titles'][0]['title'] = " ".join(
-            data['titles'][0]['title'].split()
+    builder = LiteratureBuilder(source='submitter')
+
+    builder.add_authors(
+        authors=form_fields.get('authors', []),
```
kaplun: If you amend the literature builder to simply ignore None and the like, then this code can be more readable:
```python
builder.add_authors( authors=form_fields['authors'])
```
or as we were mentioning:
```python
for author in form_fields.get('authors', []):
    builder.add_author(builder.make_author(author['full_name']...)
```

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,220 @@
...
+            if arxiv_id.split('/')[0] not in arxiv_categories:
+                arxiv_categories.append(arxiv_id.split('/')[0])
+
+        builder.add_arxiv_eprint(
+            arxiv_id=form_fields.get('arxiv_id'),
+            arxiv_categories=arxiv_categories
+        )
+
+    if form_fields.get('doi'):
+        builder.add_doi(doi=form_fields.get('doi'))
```
kaplun: All these `if`s could be simplified if the builder accepted empty input.

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,220 @@
...
+    if form_fields.get('journal_title'):
+        hits = JournalsSearch().query(
+            'match', title_variants__title__lowercased=form_fields['journal_title']).execute()
+
+        if hits:
             try:
-                data['page_nr'] = int(page_end) - int(page_start) + 1
-            except (TypeError, ValueError):
+                form_fields['journal_title'] = hits[0].short_titles[0].title
+            except (AttributeError, IndexError):
```
kaplun: This should be factored out and put in some utilities. I.e. something like: `normalize_journal_title()`. @jacquerie and @jmartinm were musing some days ago about a reorganization of such utilities.

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,220 @@
...
+        date=form_fields.get('thesis_date')
+    )
+    if form_fields.get('experiment'):
+        builder.add_accelerator_experiment(experiment=form_fields.get('experiment'))
+
+    language = _language_mapping(
+        form_fields.get('language'),
+        form_fields.get('other_language')
+    )
+    if language and len(language) == 2:
```
kaplun: Why shouldn't the language being already normalized? Is it free form?

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,220 @@
...
+        builder.add_license_url(license_url=form_fields.get('license_url'))
+
+    if form_fields.get('public_notes'):
+        builder.add_public_note(form_fields.get('public_notes'))
+
+    if form_fields.get('note'):
+        builder.add_public_note(
+            form_fields.get('note'),
+            source='arXiv' if form_fields.get('categories') else 'CrossRef'
+        )
```
kaplun: I am confused here: @jmartinm are arXiv notes and Crossref notes not going into `public_notes`? Is the user able to submit some public notes? I have also seen above plenty ways of hidden notes...

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,220 @@
...
+
+    if form_fields.get('defense_data'):
+        builder.add_public_note('Presented on {0}'.format(
+            form_fields.get('defense_data')))
```
kaplun: `_data` or `_date`?

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,220 @@
...
+
+    if form_fields.get('url'):
+        builder.add_url(url=form_fields.get('url'))
+
+    if form_fields.get('additional_url'):
+        builder.add_url(url=form_fields.get('additional_url'))
+
+    if form_fields.get('report_numbers'):
+        builder.add_report_numbers(report_numbers=form_fields.get('report_numbers'))
```
kaplun: Are there more than one?

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,220 @@
...
+            id_user=obj.id_user,
+            method='orcid'
+        ).one().id
     except NoResultFound:
-        source = ''
-    if source:
-        source = source.method + ':' + source.id
-    data['acquisition_source'] = dict(
-        source=source,
+        orcid = None
```
kaplun: All this should be factored out in some utility. It shouldn't be here.

inspirehep/modules/literaturesuggest/tasks.py
michamos: Bigger problem: the form accepts a free-form language, while the schema wants ISO 639-1 two-letter codes. Right now (tested on QA), `other_language` gets put into the record. This will not validate with the newer schema. Maybe we should use something like [pycountry](https://pypi.python.org/pypi/pycountry) to extract the language? Ideally the form should autocomplete languages and transform them to ISO 639-1. What do you think @jmartinm ?

inspirehep/modules/literaturesuggest/tasks.py
rikirenz: I will implement the second option in order to reuse it in `hepcrawler`.

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,220 @@
...
-            })
+
+    if form_fields.get('extra_comments'):
+        builder.add_private_notes(form_fields.get('extra_comments'))
+
+    if form_fields.get('nonpublic_note'):
+        builder.add_private_notes(form_fields.get('nonpublic_note'))
+
+    if form_fields.get('hidden_notes'):
+        builder.add_private_notes(form_fields.get('hidden_notes'))
```
michamos: Where do all these notes come from? In the form, I only see 'additional comments' and 'references' that should become notes.

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,220 @@
...
+        journal_issue=form_fields.get('issue'),
+        journal_title=form_fields.get('journal_title'),
+        journal_volume=form_fields.get('volume'),
+        page_start=page_start,
+        page_end=page_end,
+        artid=artid
+    )
+
+    if form_fields.get('preprint_created'):
+        builder.add_imprint_date(form_fields.get('preprint_created'))
```
michamos: no, `imprint` is for books. You need to put this into `preprint_date`.

inspirehep/modules/literaturesuggest/tasks.py
michamos: @kaplun, see https://github.com/inspirehep/inspire-next/pull/1914/commits/26427d05d7885f01f6d0ead26d8783437466c2f9#r100791238

inspirehep/modules/literaturesuggest/tasks.py
michamos: Yes several report numbers

inspirehep/modules/literaturesuggest/tasks.py
kaplun: Yeah @jmartinm this should really be fixed at the form level.

inspirehep/modules/literaturesuggest/tasks.py
michamos: BTW, references are currently discarded. They only appear in the RT ticket that is created.
Could you add them as a `private_note`?

inspirehep/modules/literaturesuggest/tasks.py
kaplun: Yeah I really think the language should be already validated by this point.

inspirehep/modules/literaturesuggest/tasks.py
kaplun: Uh? But references are way structured. Actually they are not available at this stage. It's the workflow down-the-line that will run refextract and add them, so @rikirenz should not need to deal with them here.

inspirehep/modules/literaturesuggest/tasks.py
rikirenz: https://github.com/inspirehep/inspire-next/issues/1915 I have open a issue to document the problem

inspirehep/modules/literaturesuggest/tasks.py
rikirenz: Actually now there is an issue with the language. I https://github.com/inspirehep/inspire-next/issues/1915

inspirehep/modules/literaturesuggest/tasks.py
rikirenz: @kaplun could you suggest me the right module where I can put this util function?

inspirehep/modules/literaturesuggest/tasks.py
rikirenz: @kaplun could you suggest me the right module where I can put this util function?

inspirehep/modules/literaturesuggest/tasks.py
rikirenz: I am not saying that the old code was correct but it was doing, it means that in some ways the user is able to submit these data.

inspirehep/modules/literaturesuggest/tasks.py
rikirenz: @kaplun should I put references in `_private_note`?

inspirehep/modules/literaturesuggest/tasks.py
kaplun: In the chat I made a proposal (i.e. `helpers/normalizers.py`) but I am not sure if we all agree.

rikirenz: People agree! let's create this `normilizers.py`

rikirenz:
depends https://github.com/inspirehep/inspire-schemas/pull/81

inspirehep/utils/normalizers.py
```diff
@@ -20,16 +20,36 @@
...
 from __future__ import absolute_import, division, print_function
 
-from inspirehep.dojson.model import FilterOverdo, add_schema, clean_record
+from sqlalchemy.orm.exc import NoResultFound
+from invenio_oauthclient.models import UserIdentity
+
+from inspirehep.modules.search.api import JournalsSearch
+
+
+def normalized_title(journal_title):
```
jmartinm: Better to call it `normalize_title` and `normalize_orcid`

inspirehep/utils/normalizers.py
```diff
@@ -20,16 +20,36 @@
...
+        except (AttributeError, IndexError):
+            pass
+    return normalized_journal_title
 
 
-filters = [
-    add_schema('hep.json'),
-    clean_record,
-]
+def normalized_orcid(id_user):
```
jmartinm: I don't see any normalization happening here, just retrieving from DB. Is this function used from several places or you can do the query directly in the place where you need it.

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,249 @@
...
+            affiliations=force_force_list(supervisor['affiliation']),
+            supervisor=True
+        ))
+
+    builder.add_title(title=form_fields.get('title', []))
+
+    document_type = 'conference paper' if form_fields.get('conf_name') \
+        else form_fields.get('type_of_doc', [])
+
+    builder.add_if_non_empty(
```
jmartinm: This does not look very nice, for example above using `builder.add_title` and here `builder.add_if_non_empty`, why not just do:


```python
if document_type:
    builder.add_document_type(document_type)
```

Or at the top were you call `filter_empty_elements` make sure that all the empty elements are filtered out.

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,249 @@
...
         )
-        if title == title_crossref:
-            data['titles'][0]["source"] = "CrossRef"
-        else:
-            data['titles'].append({
-                'title': title_crossref,
-                'source': "CrossRef"
-            })
+
+    builder.add_if_non_empty(
```
jmartinm: As mentioned above, using everywhere `builder.add_if_non_empty` defeats a bit the purpose of having the builder, you should have methods in the builder to build these fields (and avoid passing empty values here)

inspirehep/utils/normalizers.py
jmartinm: And since it is only for journals, it should be `normalize_journal_title`

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,249 @@
...
-            try:
-                short_title = hits[0].short_titles[0].title
-                data['publication_info'][0]['journal_title'] = short_title
-            except (AttributeError, IndexError):
-                pass
-
-    # Finally, return the converted data
-    return data
+
+    builder.validate_object()
```
jmartinm: Shall we handle this gracefully? Otherwise the user will get a 500 error. If this call raises an exception, you can catch it in `literaturesuggest/views.py` and show to the user an informative error message.

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -24,205 +24,249 @@
...
+)
 
 
 def formdata_to_model(obj, formdata):
     """Manipulate form data to match literature data model."""
+    def _filter_arxiv_categories(categories, arxiv_id):
+        arxiv_categories = categories.split()
+        if len(arxiv_id.split('/')) == 2:
+            if arxiv_id.split('/')[0] not in arxiv_categories:
+                arxiv_categories.append(arxiv_id.split('/')[0])
```
michamos: * This will fail to validate if the arXiv category has been renamed, e.g. [funct-an/9707009](https://arxiv.org/abs/funct-an/9707009) has category `math.FA` now.
* The order of categories is meaningful, the primary one (the one in the arXiv ID) should always be the first.

inspirehep/modules/literaturesuggest/tasks.py
michamos: Besides, `document_type` is required and the default one is `article`.

inspirehep/modules/literaturesuggest/tasks.py
michamos: @jmartinm says that the categories come from arXiv already, so probably no need to do this in the first place

jmartinm:
I have tried this locally (with latest schema and `LiteratureBuilder`) and I see the following problems that need to be fixed:

* [ ] `_files` is in the `metadata` but it is not part of the schema.
* [ ] `acquisition_source-> submission_number` is a string and should be an integer
* [ ] `publication_info` is empty even though there was content in the form.
* [ ] `thesis_info-> degree_type` is in the metadata even though I had not selected thesis.

michamos:
> _files is in the metadata but it is not part of the schema.

is going to be fixed in inspirehep/inspire-schemas#96

inspirehep/utils/normalizers.py
jmartinm: Maybe @jacquerie has some suggestion on this?

rikirenz:
@jmartinm about the 

> `acquisition_source`-> `submission_number` is a string and should be an integer

https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/elements/acquisition_source.json#L31

are you sure?

jmartinm:
> https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/elements/acquisition_source.json#L31
are you sure?

What I meant is that it should be treated as a number (changing also the schema), but since there are too many changes happening at the same time we better leave it as it is for the moment.

So you can forget about that one.

inspirehep/utils/normalizers.py
jacquerie: Yeah, no normalization is happening here, so it does not belong in this file.

jmartinm:
Can you double check in this PR if you fill the arXiv ID field manually in the form if it ends up correctly inside of `metadata` (in the Holding Pen). Just checked and for me it was not there.

rikirenz:
> Can you double check in this PR if you fill the arXiv ID field manually in the form if it ends up correctly inside of metadata (in the Holding Pen). Just checked and for me it was not there.

@jmartinm I have tried and in the api section I am able to see the `arxiv_id`. 
screenshot from 2017-02-23 18-16-05

screenshot from 2017-02-23 18-13-47

I do not know where is this `metadata` in the holding_pen. Anyway it should be there

jmartinm:
@rikirenz If I input an arXiv ID in the form and use the import functionality I can see it as well.

What was confusing me is just filling the arXiv ID and **not doing an import**. In that case the arXiv ID does not end up in the final record.

Checking the schema I can see that in `arxiv_eprints` it is mandatory to at least have one category so if the user just inputs an arXiv ID like arXiv:1001.4538 but then does not use the import functionality we would not have the category.

This might be a rare case, but to be more complete we can improve the workflow for submissions so that in the case where an arXiv ID is there but no categories, we query arXiv to get extra information.

CC @michamos @david-caro

jmartinm:
@rikirenz I will add an issue to review if I we fix this in the UI so that it is not allowed to input an arXiv ID without using import.

david-caro:
@jmartinm I'd create a new issue for that, and fix/improve it in another pr.

jmartinm:
`publication_info` is now there :) Only thing missing is to clean up the commit logs and rebase with latest master.

inspirehep/modules/literaturesuggest/tasks.py
michamos: You forgot to remove this

jmartinm:
Follow up issue https://github.com/inspirehep/inspire-next/issues/1931

rikirenz:
> publication_info is now there :) Only thing missing is to clean up the commit logs and rebase with latest master.

Everything is clean and rebased

jacquerie:
> Everything is clean and rebased

Sorry, no: not even the unit tests were passing, and you are introducing something that makes the acceptance test time out.

kaplun:
(Approved in order not to block the PR since I am with reduced connectivity)

rikirenz:
> Sorry, no: not even the unit tests are passing, and you are introducing something that makes the acceptance tests time out.

The schema is upgraded and some tests are failing for that reason some others no (my fault). Let's discuss this in chat.

inspirehep/bat/pages/holding_panel_literature_list.py
```diff
@@ -36,7 +36,7 @@
 
 
 def go_to():
-    Arsenic().get(os.environ['SERVER_NAME'] + '/holdingpen/list/?page=1&size=10&source=submission&workflow_name=HEP&status=HALTED')
+    Arsenic().get(os.environ['SERVER_NAME'] + '/holdingpen/list/?page=1&size=10&status=HALTED')
```
jacquerie: This looks suspicious: why are you removing `workflow_name=HEP` here? Of course, the better solution is to have proper test isolation by recreating the `holdingpen` tables, but meanwhile...

PR #1920, opened 2017-02-15
literaturesuggest: use ISO 639 for language and other_language

### LINKED ISSUE #1915: Language field from form should be always 2 characters
As you can see from the schema (https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/hep.json#L500), the language should be always of 2 characters. 
But now in `literature suggest` if we select:

- Language -> Other

- And we put a specific language that is not in the form list.

screenshot from 2017-02-13 14-33-58

- Server side we are going to receive:

```
{
    language: 'oth',
    other_language 'japanese'
}
```

But we do not have a way to normalize the language in according to the standard ISO 639-1 (https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes). Actually we are doing some normalization for the fixed languages but not in a "good" way:

```
        languages = [('en', 'en'),
                     ('rus', 'ru'),
                     ('ger', 'de'),
                     ('fre', 'fr'),
                     ('ita', 'it'),
                     ('spa', 'es'),
                     ('chi', 'zh'),
                     ('por', 'pt'),
                     ('oth', 'ot')]

        if language == 'oth' and other_language:
            return other_language

        return unicode(dict(languages).get(language))

```

We can implement several solutions. I think it is necessary to discuss which is the best solution in according to:

- User Experience

- JSONSchema changes needed

jmartinm:
* Adapt to new schema that does not allow free text inside
  the language field. (closes #1915)


jmartinm:
As discussed with @StellaCh today, we have simplified the language selection to the ones currently available on INSPIRE.

screen shot 2017-02-15 at 2 32 23 pm

michamos:
I am not convinced by this hardcoded list.

* Some of those languages have only one article written in it, so it seems a bit arbitrary because the difference between 0 and 1 is not much :)
* What should the user do if the language is not in the list?

StellaCh:
We agreed with @michamos to add the option 'Other' to the list above. When 'Other' is selected, another field 'Other language' appears. This should happen very rarely.
The 'Other language' field will be a dropdown list which will contain the rest of the languages we have ISO codes for. This will solve the issue of non-identified languages and if language is 'Other' in the first dropdown, the schema will get the value from the 2nd dropdown.

jmartinm:
> We agreed with @michamos to add the option 'Other' to the list above. When 'Other' is selected, another field 'Other language' appears. This should happen very rarely.
The 'Other language' field will be a dropdown list which will contain the rest of the languages we have ISO codes for. This will solve the issue of non-identified languages and if language is 'Other' in the first dropdown, the schema will get the value from the 2nd dropdown.

Makes sense, that's what we had before this change.

But before, the Other Language was a free text box and now we can have a dropdown with other languages.

What should this other languages be? The library we use for languages includes all the ones from the ISO 639 standard (http://www-01.sil.org/iso639-3/codes.asp?order=639_2&letter=s) so the dropdown would have 200 of them. Since that will be rarely used, that ok?

StellaCh:
Yes, that sounds good! Thanks!

jmartinm:
This is the list with the languages part of the iso 639 standard - http://pastebin.com/JH2RU12c

Have a look to see if you think it's good to be shown in the Other Language dropdown.

StellaCh:
The list is ok, but can we erase all the parentheses and their content, since they don't give any added value and some of them are distracting? Concerning the option for Greek, it's not obvious that one needs to search for 'modern Greek' in order to select Greek and anyway there's no option for 'ancient Greek', so can it be changed to just 'Greek'? thanks!

jmartinm:
Branch updated and ready for review.

language

inspirehep/modules/literaturesuggest/forms.py
```diff
@@ -62,6 +64,8 @@
 BOOK_CLASS = " book-related"
 PROCEEDINGS_CLASS = " proceedings-related"
 
+FORM_LANGUAGES = ['en', 'ru', 'de', 'fr', 'it', 'es', 'zh', 'pt']
```
kaplun: Looking to: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/dojson/current_marcxml_usage/hep.txt#L1273-L1286 I think you should also add at least `jp` for Japan

inspirehep/modules/literaturesuggest/forms.py
```diff
@@ -344,26 +348,34 @@ class LiteratureForm(INSPIREForm):
...
-                 ("fre", _("French")),
-                 ("ita", _("Italian")),
-                 ("spa", _("Spanish")),
-                 ("chi", _("Chinese")),
-                 ("por", _("Portuguese")),
-                 ("oth", _("Other"))]
+    language_choices = [
+        (language, Locale('en').languages.get(language))
+        for language in FORM_LANGUAGES
+    ]
```
kaplun: Since it's anyway a short list, shouldn't they be alphabetically sorted by Label?

jmartinm: They can, except the 'Other' that its more natural to find at the end, but the rest I can do

inspirehep/modules/literaturesuggest/forms.py
jmartinm: Sure.

kaplun:
I am approving the potential changes because I am about to leave on holiday and I didn't want to block the PR :dancer:

jmartinm:
updated.

PR #1930, opened 2017-02-24
global: move from datetime.now to datetime.utcnow

### LINKED ISSUE #1928: Move datetimes from now to utcnow
To standarize the dates we produce as `now()` we should move to:

```python
In [3]: datetime.utcnow().isoformat()
Out[3]: '2017-02-23T12:58:16.285875'
```

jmartinm:
* Using datetime.utcnow across the system ensures that all times are
  consistent instead of using the local time of the server serving the
  request. (closes #1928)


david-caro:
There are a couple of datetime.now in the conf.py file, shouldn't those be changed too?

jmartinm:
Yes, there are these lines:

```python
# The short X.Y version.
version = datetime.now().strftime("v%Y%m%d")
# The full version, including alpha/beta/rc tags.
release = datetime.now().strftime("v%Y%m%d")
```

Although they are not part of the `cookiecutter` - https://github.com/inveniosoftware/cookiecutter-invenio-module/blob/master/docs/conf.py and I am not sure if those variables are used at all (cannot find the usage)

jmartinm:
I updated now the occurrences in `conf.py` as well.

PR #1933, opened 2017-02-24
Schema25 form updates

jmartinm:


inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -345,8 +345,7 @@ def prepare_keywords(obj, eng):
         # TODO: differentiate between curated and gueesed keywords
         keywords.append(
             {
-                'classification_scheme': '',
-                'keyword': keyword['label'],
+                'value': keyword['label'],
```
david-caro: there should be a 'schema' key, with probably 'INSPIRE' value (according to the jsonschema: https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/hep.json#L587

jmartinm: Yes, I think so. The schema says: `Leave empty for free form keywords`

But I guess the `magpie` keywords should be using the INSPIRE schema.

Issue #1561, opened 2016-09-13
Documentation


kaplun:
We desperately need a documentation for INSPIRE from the development point of view.
- [ ] Technologies used and pointers to tutorials (e.g. Flask, SQLAlchemy, AngularJS, Jinja2, Celery)
- [ ] INSPIRE data model, explaining Literature, Authors... and their relations
- [ ] invenio-records and how we use it
- [ ] Pidstore and how we used (e.g. `recid` concept)
- [ ] JSONSchema and how we use it
- [ ] How to search
- [ ] Patterns (e.g. memoization/caching, authorization, celery, querying the DB, modifying the DB, CLI)
- [ ] Writing tests

kaplun:
We could make it in Sphinx so that we can ensure that the examples are runnable.

bing13:
👍

jacquerie:
This is "desperately needed", and yet we have exactly 0 resources assigned on this : )

Here's a possible outline of everything we should cover:
```
1. Getting Started
  1.1 About INSPIRE
  1.2 Installing INSPIRE
2. Developers Guide
  2.1 Technologies Used
    2.1.1 Git
    2.1.2 Invenio
    2.1.3 Werkzeug/Jinja/Flask/SQLAlchemy/Celery
    2.1.4 Angular.js
  2.2 Invenio-PIDStore
  2.3 Invenio-Records
    2.3.1 JSONSchema
  2.4 Invenio-Query-Parser (?)
  2.5 Invenio-Search (?)
  2.6 Common Tasks
  2.7 Common Problems
3. Operations Guide
  3.1 Auxiliary Services
    3.1.1 BEARD
    3.1.2 GROBID
    3.1.3 Magpie
  3.2 Common Tasks
  3.3 Common Problems
4. Advanced Topics
  4.1 Developing with Docker
```
The idea is that up to 2.5 should serve as the onboarding guide for new developers, while 2.6/2.7 are the reference section you consult when you want to do something / you see an error.

kaplun:
@StellaCh, this month @ammirate (and Sebastian) arrived and indeed it appears we do not have through documentation to share with newcomers.

I wonder if, as soon as we deliver this first milestone, documentation could be the next sprint (maybe it's enough one week of collaborative work).

This would delay the plan, but might make newcomers more productive sooner.

StellaCh:
[...]

PR #1938, opened 2017-03-01
docs: documentation for Inspire

spirosdelviniotis:
* Update index.
* Reconstruction of docs.

Addresses #1561 


david-caro:
I think that some of the pages are too big and should be split on multiple ones.

About the docker section comments I added, I think they should be done in another pr, but the structure changes I think belong here.

I think we might want to split this pr in several ones, like:
* Create docs layout
* Revise getting started guide
* Revise developer guide
...

docs/about.rst
```diff
@@ -0,0 +1,36 @@
...
+
+
+===========
+1.1.1 About
+===========
+
+INSPIRE is a service designed for physicists. The aim of INSPIRE is to improve the
+communication between physicists around the world and in addition to spread their scientific
+work (papers, thesis, etc.). In addition INSPIRE service provides other features such as
+information about authors, conferences, institutions, experiments, journals and job offers.
```
spirosdelviniotis: Nice! done!

docs/developers_guide.rst
```diff
@@ -0,0 +1,115 @@
...
+
+2. Developers Guide
+===================
+
+---------------------
+2.1 Technologies Used
+---------------------
+
+2.1.1 High level overview
+-------------------------
```
spirosdelviniotis: It declares the heading hierarchy and because we have several files with different levels it needs reconstruction all the docs.

david-caro:
Split into this two:
    #1944 
    #1950 

Will move there the latest changes here

PR #1939, opened 2017-03-01
tests: add test for completed tasks in workflow

rikirenz:
Adds a control to check when an article has been accepted in the workflow process

### DIFF
--- a/inspirehep/bat/pages/holding_panel_literature_list.py
+++ b/inspirehep/bat/pages/holding_panel_literature_list.py
@@ -36,7 +36,34 @@ from ..arsenic import Arsenic, ArsenicResponse
 
 
 def go_to():
-    Arsenic().get(os.environ['SERVER_NAME'] + '/holdingpen/list/?page=1&size=10&status=HALTED&workflow_name=HEP')
+    Arsenic().get(os.environ['SERVER_NAME'] + '/holdingpen/list/?page=1&size=10&workflow_name=HEP')
+
+
+def force_load_record(xpath_command):
+    def _refresh_page():
+        go_to()
+        return force_load_record(xpath_command)
+
+    try:
+        # click the abstract link
+        WebDriverWait(Arsenic(), 10).until(
+            EC.visibility_of_element_located(
+                (By.XPATH, xpath_command)
+            )
+        ).click()
+        # take the record
+        record = WebDriverWait(Arsenic(), 10).until(
+            EC.visibility_of_element_located(
+                (By.XPATH, '//div[@class="row hp-item ng-scope"][1]')
+            )
+        ).text
+    except (ElementNotVisibleException, WebDriverException):
+        record = _refresh_page()
+
+    if 'Waiting' in record:
+        return _refresh_page()
+
+    return record
 
 
 def click_first_record():
@@ -49,8 +76,8 @@ def click_first_record():
             (By.XPATH, '(//div[@class="ng-scope"])[2]')))
 
 
-def load_submission_record(input_data):
-    def _load_submission_record():
+def load_submitted_record():
+    def _load_record():
         return (
             'Computing' in record and
             'Accelerators' in record and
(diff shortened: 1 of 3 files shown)


rikirenz:
Only the last commit should be merged.

inspirehep/bat/pages/holding_panel_literature_list.py
```diff
@@ -49,8 +49,28 @@ def click_first_record():
...
+            record = WebDriverWait(Arsenic(), 10).until(
+                EC.visibility_of_element_located(
+                    (By.XPATH, '//div[@class="row hp-item ng-scope"][1]')
+                )
+            ).text
+        except (ElementNotVisibleException, WebDriverException):
+            go_to()
+            return _force_load_record()
+
+        return record
```
rikirenz: I moved and duplicate the function `_force_load_record` because we force the reload in different ways in according to the status of an article. In addition the html code that show a `submitted` article and a `completed` article in the list is different  for this reason it requires a different xpath command

inspirehep/bat/pages/holding_panel_literature_list.py
```diff
@@ -49,8 +49,28 @@ def click_first_record():
             (By.XPATH, '(//div[@class="ng-scope"])[2]')))
 
 
-def load_submission_record(input_data):
-    def _load_submission_record():
+def load_submitted_record():
```
rikirenz: I removed the `input_data` because it was not used

tests/acceptance/conftest.py
```diff
@@ -84,7 +84,7 @@ def login(arsenic):
     top_navigation_page.log_out()
 
 
-@pytest.fixture(autouse=True, scope='module')
+@pytest.fixture(autouse=True, scope='function')
```
rikirenz: I changed the scope of this fixture because every time that we run a function test the PostgreSQL should not contain any article. Moreover when we have to do assertions about the status of an article we are assuming that only one article is present in the list and it helps also to maintain the tests more independent with each other.

tests/acceptance/test_literature_suggestion_form.py
```diff
@@ -132,16 +132,17 @@ def test_literature_create_article_journal_with_proceeding_manually(login):
...
 
     holding_panel_literature_detail.go_to()
     assert holding_panel_literature_detail.load_submitted_record(
         input_data
     ).has_error()
     assert holding_panel_literature_detail.accept_record().has_error()
 
+    holding_panel_literature_list.go_to()
+    assert holding_panel_literature_list.load_completed_record().has_error()
+
```
rikirenz: This lines are checking if the user is able to see the completed record in the UI

inspirehep/bat/pages/holding_panel_literature_list.py
```diff
@@ -60,19 +80,40 @@ def _load_submission_record():
...
+            record = WebDriverWait(Arsenic(), 10).until(
+                EC.visibility_of_element_located(
+                    (By.XPATH, '//div[@class="row hp-item ng-scope"][1]')
+                )
+            ).text
+        except (ElementNotVisibleException, WebDriverException):
+            _referesh_page()
+
+        if 'Waiting' in record:
+            return _referesh_page()
```
rikirenz: This is necessary because the workflow sometimes needs more time to set the record's status to `Completed`. For this reason if there is the Waiting keyword it means that the workflows requires more time to complete the record process.

jmartinm:
If this is useful, can it be cleaned up to just leave one commit on top of master?

jmartinm:
Can you remind what is being checked in this PR that is not in the acceptance tests we already have?

rikirenz:
This PR is checking if the user is able to see the record completed in the holdingpen.

inspirehep/bat/pages/holding_panel_literature_list.py
```diff
@@ -49,8 +49,28 @@ def click_first_record():
             (By.XPATH, '(//div[@class="ng-scope"])[2]')))
 
 
-def load_submission_record(input_data):
-    def _load_submission_record():
+def load_submitted_record():
+    def _force_load_record():
+        try:
+            # click the abstract link
```
jmartinm: What is the clicking of abstract used for?

jmartinm: What is the click on the abstract used for? You use the content inside later for assertions?

inspirehep/bat/pages/holding_panel_literature_list.py
jmartinm: _referesh_page -> _refresh_page

inspirehep/bat/pages/holding_panel_literature_list.py
```diff
@@ -60,19 +80,40 @@ def _load_submission_record():
...
-        record = WebDriverWait(Arsenic(), 10).until(
-            EC.visibility_of_element_located((By.XPATH, '//div[@class="row hp-item ng-scope"][1]'))).text
-    except (ElementNotVisibleException, WebDriverException):
+def load_completed_record():
+    def _referesh_page():
         go_to()
-        record = _force_load_record(input_data)
+        return _force_load_record()
+
+    def _force_load_record():
```
jmartinm: This code and the one above seems almost identical, it should be reused

inspirehep/bat/pages/holding_panel_literature_list.py
rikirenz: yes! Without the abstract click the record it would not have all the data that we have to check in order to make the assertions.

inspirehep/bat/pages/holding_panel_literature_list.py
```diff
@@ -60,19 +87,12 @@ def _load_submission_record():
             'Lorem ipsum dolor sit amet, consetetur sadipscing elitr.' in record
         )
 
-    record = _force_load_record(input_data)
-    return ArsenicResponse(_load_submission_record)
-
+    record = force_load_record('//div[@class="row hp-item ng-scope"][1]/div/div/div[2]/holding-pen-template-handler/div[3]/a')
```
jacquerie: This screams: "I need a debug id!"

PR #1945, opened 2017-03-02
Fix arxiv_categories in hepnames

jmartinm:


jmartinm:
Related to https://github.com/inspirehep/inspire-next/issues/1956

jmartinm:
One case missing when trying to update an author from http://inspirehep.net is for records that use INSPIRE categories (like https://inspirehep.net/record/1072429)

The update form only uses arxiv categories so the inspire categories will not be pre selected.

For HEPNAMES, can we unify to just use `arxiv_categories` ? So on dojson if we detect there is an inspire category we do the conversion to the equivalent arxiv_category. @michamos @jacquerie

jmartinm:
> For HEPNAMES, can we unify to just use arxiv_categories ? So on dojson if we detect there is an inspire category we do the conversion to the equivalent arxiv_category. @michamos @jacquerie

Talking to @michamos IRL, we decide that for the cases where inspire categories are being used, these will not be displayed in the update form. So once Travis is green this PR is ready.

PR #1946, opened 2017-03-02
tests: add doc to debug acceptance in the web node

rikirenz:
* Add documentation in order to use RemotePdb to debug acceptance tests with Docker.

### DIFF
--- a/docs/tests.rst
+++ b/docs/tests.rst
@@ -59,7 +59,7 @@ Via Docker with a graphical instance of Firefox (Linux)
   $ docker-compose -f docker-compose.test.yml run --rm acceptance
 
 
-Via Docker with a graphical instance of Firefox (macOS)
+Via Docker with a graphical instance of Firefox (MacOS)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 1. Check the first step in the `Via Docker`_ section.
@@ -109,6 +109,57 @@ Via Docker with a graphical instance of Firefox (macOS)
 .. code-block:: bash
 
   $ docker-compose -f docker-compose.test.yml run --rm acceptance
+
+How to Debug the Selenium Tests on Docker
+-----------------------------------------
+In order to debug code in the `web` container with the Docker-Selenium test suite it is necessary to use `remote-pdb`_.
+
+1. First install RemotePdb in the web container. Run the web container with the bash command attached:
+
+.. code-block:: bash
+
+  $ docker-compose run web bash
+
+Secondly install with the pip command the `remotePdb` package:
+
+.. code-block:: bash
+
+  (virtualenv) $ pip install remote-pdb
+
+Finally exit from the virtualenv and from the docker container with `ctrl + d`.
+
+2. Find a spot where you would like tracing to begin, and insert the following code:
+
+.. code-block:: python
+
+  from remote_pdb import RemotePdb
+  RemotePdb('0.0.0.0', 4444).set_trace()
+
+3. Run the acceptance tests with `docker-compose`
+
+.. code-block:: bash
+
+  $ docker-compose -f docker-compose.test.yml run --rm acceptance
+
+4. Write down the IP address of the web container because you will need it later:
+
+.. code-block:: bash
+
+  $ docker inspect inspirenext_web_1
+  {
+    [...]
+    "IPAddress": "123.456.7.890",
+    [...]
+  }
+
+5. At a certain point the tests will be blocked. It means that the remote `RemotePdb.set_trace()` has been executed.
+   In order to use the pdb console run:
+
+.. code-block:: bash
+
+  $ telnet 123.456.7.890 4444
+
+.. _`remote-pdb`: https://pypi.python.org/pypi/remote-pdb
 
 
 How to Write the Selenium Tests


docs/tests.rst
```diff
@@ -110,6 +110,57 @@ Via Docker with a graphical instance of Firefox (macOS)
 
   $ docker-compose -f docker-compose.test.yml run --rm acceptance
 
+How to Debug the Selenium Tests on Docker
+-----------------------------------------
+In order to debug code in the `web` containter with the Docker-Selenium test suite it is necessary to use `remote-pdb`_.
```
jmartinm: container

docs/tests.rst
```diff
@@ -110,6 +110,57 @@ Via Docker with a graphical instance of Firefox (macOS)
...
+
+  $ docker-compose run web bash
+
+Secondly install with the pip command the `remotePdb` package:
+
+.. code-block:: bash
+
+  (virtualenv) $ pip install remote-pdb
+
+Finally exit from the virtualenv and from the  docker container with `ctrl + d`.
```
jmartinm: extra space

docs/tests.rst
```diff
@@ -110,6 +110,57 @@ Via Docker with a graphical instance of Firefox (macOS)
...
+.. code-block:: bash
+
+  $ docker inspect inspirenext_web_1
+  {
+    [...]
+    "IPAddress": "123.456.7.890",
+    [...]
+  }
+
+5. At a certain point the tests will be blocked. It means that the remote `RemotePdb.set_trace()` has been executed.
```
jmartinm: When I click View on this file and get the preview it seems like RemotePdb.set_trace() is not getting formatted (even if you use ` ) . Same happens in other parts of the file.

rikirenz: docs

Locally in my browser it works.

jmartinm:
I think this is more complicated that it needs to be. I was now testing and for debugging I just do:

```bash
docker-compose -f docker-compose.test.yml run --rm acceptance bash
```

Inside of container:

```bash
(virtualenv)py.test --driver Firefox --host selenium --port 4444 --capability browserName firefox --html=selenium-report.html tests/acceptance/test_author_creation.py -k test_review_submission_author
```

And put anywhere where I want to stop:

```python
import pytest; pytest.set_trace()
```

This correctly stops in a pdb session, I can execute commands to selenium and continue with the tests.

rikirenz:
We have already discuss about this but I am going to write it to be clear. This procedure should help when we want to debug code in the _web_ container and not in the _acceptance_ one. Anyway probably this peace of documentation is too specific and is treating a procedure that is not very common.
What do you think @david-caro @jacquerie @kaplun ? is it useful in your opinion?

jmartinm:
> We have already discuss about this but I am going to write it to be clear. This procedure should help when we want to debug code in the web container and not in the acceptance one. Anyway probably this peace of documentation is too specific and is treating a procedure that is not very common.

I personally never encountered the need so far.

kaplun:
I am not sure I am understanding, but maybe that this documentation can be simplified in order to show how to debug the web container, regardless of the acceptance test? I.e. what if want to simply put a breakpoint on the web part, and reach it by opening myself the browser?

david-caro:
The problem is that debugging the web container, when you are running the acceptance tests is more complicated than doing it by itself (the main problem is that if you start the web container with `docker-compose run --rm web`, then when running the `acceptance` one will not see the web container you started manually and create a new one).

So the simplified version of debugging the web node is not ok for debugging it when you are running the acceptance tests.

rikirenz:
The problem is that when we run acceptance I have not found a way to debug those tests with `rdb` for this reason this doc is using `RemotePdb`.
Anyway we do not have a doc that explains how to debug celery tasks with `rdb`.
I strongly believe that is necessary have at least a doc that explain how to debug the workflow with `rdb`.
I can write it if you think is useful.

docs/tests.rst
```diff
@@ -59,7 +59,7 @@ Via Docker with a graphical instance of Firefox (Linux)
   $ docker-compose -f docker-compose.test.yml run --rm acceptance
 
 
-Via Docker with a graphical instance of Firefox (macOS)
+Via Docker with a graphical instance of Firefox (MacOS)
```
jacquerie: (`macOS` is the correct capitalization: https://www.apple.com/lae/macos/sierra/)

docs/tests.rst
```diff
@@ -110,6 +110,57 @@ Via Docker with a graphical instance of Firefox (macOS)
 
   $ docker-compose -f docker-compose.test.yml run --rm acceptance
 
+How to Debug the Selenium Tests on Docker
+-----------------------------------------
```
jacquerie: This logically belongs _after_ the section on how to write tests. Also, since we don't support running the `acceptance` natively, there's no point in specifying "on Docker".

docs/tests.rst
```diff
@@ -110,6 +110,57 @@ Via Docker with a graphical instance of Firefox (macOS)
...
+
+  $ docker-compose run web bash
+
+Secondly install with the pip command the `remotePdb` package:
+
+.. code-block:: bash
+
+  (virtualenv) $ pip install remote-pdb
+
+Finally exit from the virtualenv and from the docker container with `ctrl + d`.
```
jacquerie: All this first paragraph can be replaced with `docker-compose run --rm web pip install remote-pdb`.

docs/tests.rst
```diff
@@ -110,6 +110,57 @@ Via Docker with a graphical instance of Firefox (macOS)
...
+4. Write down the IP address of the web container because you will need it later:
+
+.. code-block:: bash
+
+  $ docker inspect inspirenext_web_1
+  {
+    [...]
+    "IPAddress": "123.456.7.890",
+    [...]
+  }
```
jacquerie: The output of `inspect` is quite massive. Instead of hunting for it with your eyes you can `grep` for it: `docker inspect inspirenext_web_1 | grep IPAddress`

docs/tests.rst
jacquerie: Single backticks and double backticks are different in Sphinx: https://mail.zope.org/pipermail/grok-dev/2008-December/006823.html

docs/tests.rst
```diff
@@ -110,6 +110,57 @@ Via Docker with a graphical instance of Firefox (macOS)
 
   $ docker-compose -f docker-compose.test.yml run --rm acceptance
 
+How to Debug the Selenium Tests on Docker
+-----------------------------------------
+In order to debug code in the `web` container with the Docker-Selenium test suite it is necessary to use `remote-pdb`_.
```
jacquerie: Yeah, but why? Better:

Unlike the other test suites, the container that is running the test code of the
``acceptance`` test suite is different from the one running the application
code. Therefore, in order to debug a test failure, we must connect remotely to
this other container. The tool to achieve this is called `remote-pdb`_.

docs/tests.rst
jacquerie: Moreover, the actual container name is `inspirenext_test-web_1` (this changed in the meantime).

PR #1944, opened 2017-03-02
docs: update getting started

spirosdelviniotis:
* Adds docker documentation

Signed-off-by: Spiros Delviniotis <spyridon.delviniotis@cern.ch>

PR #1950, opened 2017-03-02
docs: add developers guide

spirosdelviniotis:
* Adds documentations rules
* Adds frameworks overview
* Adds git
* Adds inspire record
* Adds invenio
* Adds pidstore
* Adds workflow


david-caro:
I want to do some modifications to the diagrams, can you add also the source code for them? (the xml files or something)
Also can you add a note in the docs that the diagrams where edited with draw.io? so someone else can modify them in the future. We should find also some way to have an offline copy of the app, so if the service disappears we can still edit them.

david-caro:
The tool looks really nice btw :) (and gplv3!)

david-caro:
I modified the structure and added completed some missing info, still incomplete though, but I don't want to lose the changes, and well, allow review and someone else picking it up if wanted.

docs/workflows.rst
```diff
@@ -0,0 +1,127 @@
...
+
+.. code-block:: shell
+
+    inspirehep shell
+
+
+.. code-block:: python
+
+    from invenio_workflows import workflow_object_class, ObjectStatus
+    errors = workflows_object_class.query(status=ObjectStatus.ERROR)
```
david-caro: s/workflows_object_class/workflow_object_class/

jacquerie:
Same as https://github.com/inspirehep/inspire-next/pull/1944#issuecomment-309744476.

PR #1964, opened 2017-03-03
literaturesuggest: use legacy_ICN for autocompletion

### LINKED ISSUE #1962: literaturesuggest: Affiliation autocomplete use legacy_ICN
For Autocompletion of affiliations the `legacy_ICN` field should be queried and populated in the form.

The ElasticSearch mapping (institutions.json) needs to be amended so that the content of `legacy_icn` is copied to `affautocomplete` (instead of using ICN)

jmartinm:
* Adapt to data model change in institutions and use legacy_ICN instead of
  ICN. (closes #1962)

### DIFF
--- a/inspirehep/modules/forms/static/js/forms/affiliations_typeahead.js
+++ b/inspirehep/modules/forms/static/js/forms/affiliations_typeahead.js
@@ -57,9 +57,9 @@ define([
     this.$element = $element;
 
     var suggestionTemplate = Hogan.compile(
-      '<strong>{{ ICN }}</strong><br>' +
+      '<strong>{{ legacy_ICN }}</strong><br>' +
       '<small>' +
-      '{{#new_ICN}}{{#show_future_name}}Alternative name: {{new_ICN}}<br>{{/show_future_name}}{{/new_ICN}}' +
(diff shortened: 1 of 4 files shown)


inspirehep/modules/forms/static/js/forms/affiliations_typeahead.js
```diff
@@ -80,14 +80,14 @@ define([
...
-        return data.metadata.ICN[0];
+        return data.metadata.legacy_ICN;
       },
       templates: {
         empty: function(data) {
           return 'Cannot find this affiliation in our database.';
         },
         suggestion: function(data) {
-          if (data.new_ICN != data.ICN) {
+          if (data.ICN != data.legacy_ICN) {
```
michamos: `ICN` is an array, `legacy_ICN`is a string, they will never be equal.

jmartinm:
```
```

michamos: I don't know if it needs to be an array, we didn't really work on institutions. Probably not, but for now it is (might change in the future).

jmartinm:
Blocked by pending upcoming updates to ElasticSearch mappings

PR #1976, opened 2017-03-06
holdingpen: show links to arXiv abstract and PDF

jmartinm:


jmartinm:
screen shot 2017-03-06 at 9 32 25 am

jmartinm:
This is breaking acceptance tests. Might be because some boxes are now being hidden when there is no content, so the `xpath` selector might be broken. To be checked.

inspirehep/modules/workflows/static/js/inspire_workflows_ui/templates/details.html
```diff
@@ -89,7 +89,18 @@
                 </span>
               </div>
             </div>
-
+            <div ng-if="vm.record.metadata.arxiv_eprints" class="row">
+              <div class="col-md-12">
+                <span class="font-bolder">e-Print:</span>
+                <span ng-repeat="eprint in vm.record.metadata.arxiv_eprints">
+                  <a href="http://arxiv.org/abs/{{eprint.value}}" target="_new">
+                    arXiv:{{eprint.value}}</a>&nbsp;-&nbsp;
```
michamos: can you also add something like `[{{eprint.categories[0]}}]` or whatever is the correct syntax for the first element of the categories list? this is useful information.

Issue #438, opened 2015-10-23
submission: thesis field reordering


jmartinm:
The order of the fields should be:

Degree type
Thesis Date (we leave only one date)
Institution
Supervisors

Would be good to mark in the interface Degree type and Thesis date as important to remind the user to fill them in.

Also make PhD default degree type.

jmartinm:
@gartocern this can be a good warm up task too

kaplun:
@jmartinm still needed?

jmartinm:
Just double checked and this did not get done. It will be a bootcamp task for @koursaros93

PR #1981, opened 2017-03-06
literaturesuggest: various fixes in thesis form

Koursaros93:
Removed defense date and reordered the fields as asked in issue #438
Made Phd default degree type.


jmartinm:
Please rebase this PR with latest `master` branch to see if tests are passing.

inspirehep/bat/pages/create_literature.py
```diff
@@ -131,24 +131,24 @@ def _thesis_info_population(input_data):
...
-    )
     Arsenic().find_element_by_id('institution').send_keys(
         input_data['institution']
     )
+    Arsenic().find_element_by_id('supervisors-0-name').send_keys(
+        input_data['supervisor']
+    )
+    Arsenic().find_element_by_id('supervisors-0-affiliation').send_keys(
+        input_data['supervisor-affiliation']
+    )
```
jacquerie: So, here you are reordering the selections according to the actual form order. This is not wrong, but is unrelated with what you are doing. When that happens, you should split the related changes in their own commits. For example here you have a commit that reorders things around, and one that deletes `defense_date`.

inspirehep/modules/literaturesuggest/forms.py
```diff
@@ -439,15 +439,10 @@ def _is_other_language(language):
...
-    defense_date = fields.TextField(
-        label=_('Date of Defense'),
-        description='Format: YYYY-MM-DD, YYYY-MM or YYYY.',
-        validators=[date_validator],
-        widget=defensedate_widget,
-    )
 
     degree_type = fields.SelectField(
         label=_('Degree Type'),
+        default="phd",
```
jacquerie: And this is a third commit.

tests/acceptance/test_literature_suggestion_form.py
```diff
@@ -48,12 +48,11 @@ def test_literature_create_thesis_manually(login):
...
         'report-number-1': '101',
-        'supervisor': 'Mister Yellow',
-        'supervisor-affiliation': 'CERN',
-        'thesis-date': '2001-01-01',
-        'thesis-defense': '2001-01-01',
         'degree-type': 'bachelor',
+        'thesis-date': '2001-01-01',
         'institution': 'Wisconsin U., Madison',
+        'supervisor': 'Mister Yellow',
+        'supervisor-affiliation': 'CERN',
```
jacquerie: Same comment as above: you are reordering for clarity (this is a dictionary, so the order does not matter!), so you should be precise about it.

PR #1982, opened 2017-03-06
docs: add sphinxcontrib-napoleon

spirosdelviniotis:
* Adds: configuration of sphinxcontrib-napoleon.
* Adds: models of modules in documentation.

~Depends on https://github.com/inspirehep/inspire-docker/pull/36~
~Depends on https://github.com/inspirehep/inspire-docker/pull/37~

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -40,7 +40,7 @@ services:
       service: service_base
     volumes_from:
       - static
-    command: py.test inspirehep tests/unit
+    command: ./run_tests.sh
   disambiguation:
     extends:
       service: service_base
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -6,6 +6,8 @@ SPHINXOPTS    =
 SPHINXBUILD   = sphinx-build
 PAPER         =
 BUILDDIR      = _build
+SOURCEDIR     = source
+CODEDIR       = ../inspirehep
 
 # User-friendly check for sphinx-build
 ifeq ($(shell which $(SPHINXBUILD) >/dev/null 2>&1; echo $$?), 1)
@@ -18,6 +20,7 @@ PAPEROPT_letter = -D latex_paper_size=letter
 ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
 # the i18n builder cannot share the environment and doctrees with the others
 I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
+SPHINXAPIDOC    = sphinx-apidoc -f -o $(SOURCEDIR) $(CODEDIR)
 
 .PHONY: help clean html dirhtml singlehtml pickle json htmlhelp qthelp devhelp epub latex latexpdf text man changes linkcheck doctest coverage gettext
 
@@ -48,13 +51,25 @@ help:
 	@echo "  doctest    to run all doctests embedded in the documentation (if enabled)"
 	@echo "  coverage   to run coverage check of the documentation (if enabled)"
 
+
+# function sets exit code for html target
+define report_result_html
+	if [ -s "$(1)" ]; \
+	then \
+		echo Build Failed. Check documentation_errors.log file for details.; \
+		exit 1; \
+	fi
+endef
+
+
 clean:
-	rm -rf $(BUILDDIR)/*
+	rm -rf $(BUILDDIR) $(SOURCEDIR) documentation_errors.log
 
 html:
-	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
-	@echo
-	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
+	$(SPHINXAPIDOC); \
+	rm -f source/modules.rst; \
+	$(SPHINXBUILD) -w documentation_errors.log -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	@$(call report_result_html,documentation_errors.log)
 
 dirhtml:
 	$(SPHINXBUILD) -b dirhtml $(ALLSPHINXOPTS) $(BUILDDIR)/dirhtml
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+printf "\n--------------------- Unit Tests ----------------------\n"
+py.test inspirehep tests/unit
+unit_tests_exit_code=$?
+
+printf "\n\n--------------- Creating Documentation ----------------\n"
+make -C docs html
+create_docs_exit_code=$?
+
+if [[ "$unit_tests_exit_code" != "0" ]] || [[ "$create_docs_exit_code" != "0" ]]
+then
+	exit 1
+fi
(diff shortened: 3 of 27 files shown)


docs/Makefile
```diff
@@ -49,12 +55,13 @@ help:
 	@echo "  coverage   to run coverage check of the documentation (if enabled)"
 
 clean:
-	rm -rf $(BUILDDIR)/*
+	rm -rf $(BUILDDIR) $(SOURCEDIR) documetation_errors.log
```
jmartinm: `documentation_errors`

inspirehep/modules/arxiv/__init__.py
```diff
@@ -26,4 +26,6 @@
 
 from .ext import Arxiv
 
-__all__ = ('Arxiv')
+__all__ = [
```
jmartinm: why this styling changes in this PR?

jmartinm:
Can you add instructions on what to test/what commands to run to see the documentation?

inspirehep/modules/arxiv/__init__.py
spirosdelviniotis: Basically I am receiving this error while it is `('Arxiv')` :
```
WARNING: __all__ should be a list of strings, not 'Arxiv' (in module inspirehep.modules.arxiv) -- ignoring __all__
```

jmartinm: right

```python
In [2]: ('a') == ['a']
Out[2]: False

In [3]: ('a') == 'a'
Out[3]: True
```

jmartinm: It seems like a mistake and the correct syntax to have a tuple should be `('Arxiv',)`

inspirehep/modules/records/serializers/tei_serializer.py
```diff
@@ -24,7 +24,7 @@
 
 from __future__ import absolute_import, division, print_function
 
-from inspirehep.modules.converttohal import tei
+from inspirehep.modules.hal import tei
```
david-caro: this feels strange, if this was never used, can we remove it instead?

run_tests.sh
```diff
@@ -0,0 +1,19 @@
...
+printf "\n--------------------- Unit Tests ----------------------\n"
+py.test inspirehep tests/unit
+unit_tests_exit_code=$?
+
+# Make documentation
+printf "\n\n--------------- Creating Documentation ----------------\n"
+make -C docs html
+create_docs_exit_code=$?
+
+if [ unit_tests_exit_code != 0 ]
```
david-caro:
```bash
if [[ "$unit_tests_exit_code" != "0" ]]; then
```
* `[[]]` instead of `[]` because `[]` is acually a link to the binary `/usr/bin/[`, that is an external command to bash, and undergoes all the bash expansion and spaces stuff with it's issues, while `[[]]` is a builtin command, and has more sane syntax as it does not undergo the same expansion step (it can be avoided though using proper quoting, that is sometimes tricky):
```shell
dcaro@pcrcssis001$ a='one two'

dcaro@pcrcssis001$ [ $a == 'one two' ] && echo yes || echo no
bash: [: too many arguments
no

dcaro@pcrcssis001$ [[ $a == 'one two' ]] && echo yes || echo no
yes

dcaro@pcrcssis001$ a='*'

dcaro@pcrcssis001$ [[ $a == '*' ]] && echo yes || echo no
yes

dcaro@pcrcssis001$ [ $a == '*' ] && echo yes || echo no
bash: [: too many arguments
no

dcaro@pcrcssis001$ echo $a  # when there are no quotes, bash resolves $a to * and then expands it
...
A list with all your files
...

dcaro@pcrcssis001$ echo "$a"  # with quotes, bash does not expand the * after resolving the var
*

File: docs/Makefile [hunk 5]
```diff
@@ -48,13 +51,26 @@ help:
...
 	@echo "  coverage   to run coverage check of the documentation (if enabled)"
 
+
+# function creates exit code for html target
+define report_result_html
+	if [ -s $(1) ]; \
+	then \
+		echo Build Failed. Check documentation_errors.log file for details.; \
+		exit 1; \
+	fi \
```
david-caro: This is just bash, not makefile syntax. Not that it's very important here as it's quite unlikely that we will have spaces in the `$(1)` var, but so you know, when doing variable expansions, you should be careful with the quotes:
```shell
if [[ -s "$(1)" ]]; then...
```
Specially with the `rm` commands (like the cleanup after this).

You can also remove the `\` of the `fi` line and remove the extra newline before the `endef`

run_tests.sh
```diff
@@ -0,0 +1,19 @@
...
+then
+	exit 1
+elif [ create_docs_exit_code != 0 ]
+then
+	exit 2
```
david-caro: If you want to use special exit codes for each case, you should keep in mind that there are 3 different cases, failing the docs, failing the unit, and failing both.

run_tests.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
david-caro: In my scripts I never use `/usr/bin/env bash`, for two main reasons:
* I've found that `/bin/bash` is more portable (not that we care much about that).
* `/bin/bash` is shorter
* `/bin/bash` allows you to pass default parameters no it, like -x and -e, that can be overriden when invoking the script:
```shell
$ cat myscript.sh
#!/bin/bash -exu
echo lolo
$ bash myscript.sh  # this will ignore the shebang and allow you to override any options passed to bash
$ ./myscript.sh  # this will use the options specified in the shebang
```

The only disadvantage is that if you have a custom bash installation somewhere before on your path (the only time I've seen this is when a machine has been hacked), then it will use that one instead.

run_tests.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Run unit tests
```
david-caro: No need for the comment, there's a big nice string in the next line that tells you already what this is.

run_tests.sh
spirosdelviniotis: @david-caro  Do you think we need special exit codes?

inspirehep/modules/records/serializers/tei_serializer.py
spirosdelviniotis: As I can see it's unused so I'll remove it and run tests!

docs/Makefile
spirosdelviniotis: This doesn't work for me:

```
define report_result_html
	if [[ -s "$(1)" ]]; \
	then \
		echo Build Failed. Check documentation_errors.log file for details.; \
		exit 1; \
	fi
endef
```

I'm receiving the following error:

```
/bin/sh: 1: [[: not found
```

docker-compose.test.yml
```diff
@@ -40,7 +40,7 @@ services:
       service: service_base
     volumes_from:
       - static
-    command: py.test inspirehep tests/unit
+    command: ./run_tests.sh
```
jacquerie: Why not just `py.test inspirehep tests/unit && make -C docs html`? The ASCII art in `./run-tests.sh` adds no value to the report.

spirosdelviniotis: In order to run two commands in docker-compose.yml you have to write something like this:
 `command: bash -c "py.test inspirehep tests/unit && make -C docs html"`

jacquerie: Ah, good to know. Thanks!

PR #1984, opened 2017-03-06
dojson: fix _fft

jacquerie:
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-nightly/group/619250/


david-caro:
That's my year! Extra :+1: :+1:!

PR #1989, opened 2017-03-07
literaturesuggest: validate if id already present in Hal Hidden

### LINKED ISSUE #1297: Suggestion for HAL Hidden papers should be automatically rejected
Say that a user suggests a paper that is in the HAL Hidden collection:
https://inspirehep.net/collection/HAL%20Hidden

Then:
- if the user is a French cataloguer we should say that the paper already exists
- else: if should thank for the suggestion but state that it had already been considered as pertinent...

jmartinm:
* Port HAL Hidden validator functionality. (closes #1297)

### DIFF
--- a/inspirehep/modules/forms/validators/simple_fields.py
+++ b/inspirehep/modules/forms/validators/simple_fields.py
@@ -26,8 +26,13 @@ from datetime import datetime
 from urllib import urlencode
 
 from flask import current_app
+from flask.ext.login import current_user
 from idutils import is_arxiv, is_isbn
 from wtforms.validators import ValidationError, StopValidation
+
+
+def _get_current_user_roles():
+    return [r.name for r in current_user.roles]
 
 
 def arxiv_syntax_validation(form, field):
@@ -115,7 +120,17 @@ def duplicated_doi_validator(form, field):
     if not doi:
         return
     if current_app.config.get('PRODUCTION_MODE'):
+        user_roles = _get_current_user_roles()
+
+        # First check in default collection
         inspirehep_duplicated_validator('doi:' + doi, 'DOI')
+
+        # And in Hal collection
+        try:
+            inspirehep_duplicated_validator('doi:' + doi, 'DOI', 'HAL Hidden')
+        except ValidationError:
+            if 'cataloger' in user_roles:
+                raise
 
 
 def duplicated_arxiv_id_validator(form, field):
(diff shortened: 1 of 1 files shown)


inspirehep/modules/forms/validators/simple_fields.py
```diff
@@ -115,8 +128,19 @@ def duplicated_doi_validator(form, field):
...
         return
     if current_app.config.get('PRODUCTION_MODE'):
+        user_roles = _get_current_user_roles()
+
+        # First check in default collection
         inspirehep_duplicated_validator('doi:' + doi, 'DOI')
 
+        # And in Hal collection
+        try:
+            inspirehep_duplicated_validator('doi:' + doi, 'DOI', 'HAL Hidden')
```
jacquerie: `inspirehep_duplicated_validator` and [this](https://github.com/inspirehep/inspire-next/blob/49a4197747193054894273b51732efd78dd4a25b/inspirehep/modules/workflows/tasks/matching.py#L90) duplicate the same (or very similar) logic.

jacquerie: (Which is not something that you should fix in this PR, but something we should be aware of: we shouldn't have this repetition in the codebase)

jmartinm:
The behaviour is 1:1 port of what we have on labs.inspirehep.net (as discussed with @kaplun and @annetteholtkamp https://github.com/inspirehep/inspire-next/issues/1297)

The workflow suggested sounds good, but If there is an agreement that a different functionality is needed, the corresponding Github issue should be created.

Now, for this case, if we want to deploy Invenio 3 in the following week we can either keep the same functionality (with this PR) or we can create a new issue and at a later stage the new workflow will be implemented (with an interim period where there will be no notifications about existing records in `Hal Hidden`)

kaplun:
We agree (@michamos in the room)  that this is indeed 1:1 with what was on Labs. 
I think we can have a quick win with the `if` part dedicated to regular users (non cataloguers): there we can simply ignore the duplicate error, and let the upload fail on legacy. This will then be dealt by hand by catalogers the few times this will happen. 
WDYT?

michamos:
approved so that it is not blocked.

jmartinm:
>  We agree (@michamos in the room) that this is indeed 1:1 with what was on Labs.
I think we can have a quick win with the if part dedicated to regular users (non cataloguers): there we can simply ignore the duplicate error, and let the upload fail on legacy. This will then be dealt by hand by catalogers the few times this will happen.

Alright, can do the change. Hopefully this will not happen often since a developer will have to act (checking `bibsched`, letting a cataloger know) and also in this case the user will not get an automatic email of approval/rejection and the cataloger will have to respond via RT.

inspirehep/modules/forms/validators/simple_fields.py
jmartinm: Indeed. It needs a bit of design thinking to have a reusable `match_arxiv` and `match_doi` with some extra parameter (like specifying collection) that can be reused in both places. We can leave it for another PR.

jmartinm:
Code updated. @kaplun @michamos

michamos:
:+1: for me

PR #1992, opened 2017-03-07
literaturesuggest: remove _filter_arxiv_categories

jacquerie:
Closes #1986 

Categories that are part of the arXiv identifier are not necessarily
good categories to add: they might be outdated or no longer used.

### DIFF
--- a/inspirehep/modules/literaturesuggest/tasks.py
+++ b/inspirehep/modules/literaturesuggest/tasks.py
@@ -53,14 +53,6 @@ def retrieve_orcid(id_user):
 
 def formdata_to_model(obj, formdata):
     """Manipulate form data to match literature data model."""
-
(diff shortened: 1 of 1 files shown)


jacquerie:
Whoops, in https://github.com/inspirehep/inspire-next/commit/b58c343efaaac237074c3df749401f6f0170ade6 I also removed the fix introduced in https://github.com/inspirehep/inspire-next/commit/1013b95cec26268983cb765618b52ef26c88bfb7 to ensure that `langdetect` is deterministic. It needs to be added back.

jacquerie:
> Whoops, in b58c343 I also removed the fix introduced in 1013b95 to ensure that langdetect is deterministic. It needs to be added back.

Done in https://github.com/inspirehep/inspire-next/pull/1993.

PR #1999, opened 2017-03-08
tests: add tests for the citeable field

rikirenz:

### DIFF
--- a/tests/unit/literaturesuggest/test_literaturesuggest_tasks.py
+++ b/tests/unit/literaturesuggest/test_literaturesuggest_tasks.py
@@ -22,11 +22,15 @@
 
 from __future__ import absolute_import, division, print_function
 
+import mock
+import pytest
+
 from inspirehep.modules.literaturesuggest.tasks import (
     new_ticket_context,
     reply_ticket_context,
     curation_ticket_context,
     curation_ticket_needed,
+    formdata_to_model,
 )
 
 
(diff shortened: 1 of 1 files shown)


rikirenz:
As you can see in this test we are Monkey patching a lot of function inside the `formdata_to_model`. For this reason I have added an issue here: https://github.com/inspirehep/inspire-next/issues/2000
In order to be sure that is going to be refactored. Anyway when the function is going to be refactored the test should be amended by removing the patchings that now are there.

tests/unit/literaturesuggest/test_literaturesuggest_tasks.py
```diff
@@ -47,6 +52,49 @@ def __init__(self, email):
...
+    'journal_title': 'High Energy Physics Libraries Webzine',
+    'volume': '192',
+    'year': '2011',
+}),(
+None,
+{
+    'journal_title': 'High Energy Physics Libraries Webzine',
+    'year': '2011',
+    'page_range_article_id': '2550'
+})])
```
david-caro:
```python
@pytest.mark.parametrize(
    'param1,param2',
    [
        (
            True,
            {
                'data': 'more data',
                'other data': 'more other data',
            },
         ),
        ...
    ]
)
```

rikirenz:
This tests should be located in the `inspire-schemas` project. Indeed: https://github.com/inspirehep/inspire-schemas/pull/118

PR #2001, opened 2017-03-08
holdingpen: allow setting reject reason on hep workflow

### LINKED ISSUE #1936: Holding Pen: reject reason
Currently on labs.inspirehep.net, when rejecting a user suggestion there is the option to input a reason for the rejection and this will be sent to the user in an email.

On Invenio 3 this is missing, so no reason will be sent.

jmartinm:
* Adds an extra step when rejecting a user suggestion that allows
  setting the reason. (closes #1936)

### DIFF
--- a/inspirehep/modules/theme/static/scss/holding-pen/holding-pen.scss
+++ b/inspirehep/modules/theme/static/scss/holding-pen/holding-pen.scss
@@ -611,7 +611,7 @@ $hp_table_bg: #fff;
 
   .reject {
     background-color: $alizarin;
-    color: $dark-blue;
+    color: white;
   }
 
   .accept_core {
@@ -839,4 +839,18 @@ $hp_table_bg: #fff;
 
 .green-color {
   color: $green-sea
+}
+
+#reject-reason {
+    width: 100%;
+    min-height: 200px;
+}
+
+.text-wrap {
+    white-space: pre-wrap;
+}
+
+.reason-link {
+    font-size: 0.9em;
+    color: $clouds !important;
 }
--- a/inspirehep/modules/workflows/static/js/inspire_workflows_ui/holdingpen/holdingpen.controllers.js
+++ b/inspirehep/modules/workflows/static/js/inspire_workflows_ui/holdingpen/holdingpen.controllers.js
@@ -27,6 +27,7 @@
 
     $scope.vm.selected_records = [];
     $scope.vm.selected_record_decisions = {};
+    $scope.vm.selected_record_methods = {};
 
     $scope.toggleSelection = toggleSelection;
     $scope.toggleAll = toggleAll;
@@ -34,6 +35,21 @@
     $scope.allChecked = allChecked;
     $scope.setDecision = setDecision;
     $scope.redirect = redirect;
+
+    function _add_record_method(record) {
+      method = record._source.metadata.acquisition_source.method;
+      if (method in $scope.vm.selected_record_methods) {
+        $scope.vm.selected_record_methods[method]++;
+      }
+      else {
+        $scope.vm.selected_record_methods[method] = 1;
+      }
+    }
+
+    function _remove_record_method(record) {
+      method = record._source.metadata.acquisition_source.method;
+      $scope.vm.selected_record_methods[method]--;
+    }
 
     function redirect(url) {
       $window.location = url;
@@ -47,6 +63,7 @@
       var _data_type = record._source._workflow.data_type;
 
       if (isChecked(record._id)) {
+        _remove_record_method(record);
         $scope.vm.selected_records.splice(getItemIdx(+record._id), 1);
         if (_data_type in $scope.vm.selected_record_decisions) {
           $scope.vm.selected_record_decisions[_data_type]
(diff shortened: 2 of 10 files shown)


jmartinm:
rejection

jmartinm:
One challenge here is supporting this reject reason in a nice way from the Holding Pen results page:

screen shot 2017-03-09 at 9 25 20 am

* When hitting reject in an individual item it should behave like in the detailed view. The change is not trivial since the code in the search results is a bit different than detailed page.

* When selecting many HEP workflows and on top of the page hitting Reject, do we have a single reject reason modal window and use the same for all of them?

@michamos @annetteholtkamp @StellaCh I would like to hear your opinion on how worth the effort of implementing the Reject functionality in the search results is. Do you ever use it? Or Reject should be done from detailed page? And in case you use it, what would be the ideal flow?

michamos:
I never use reject from the list view for user submissions, I always look at the detailed record, same for @annetteholtkamp I would say. Maybe @ksachs and @fschwenn use it for arXiv? in that case there is no need for a reason though.

jmartinm:
> I never use reject from the list view for user submissions, I always look at the detailed record, same for @annetteholtkamp I would say. Maybe @ksachs and @fschwenn use it for arXiv? in that case there is no need for a reason though.

Do you ever use the other actions from search results? So it would be ok to just remove the `Reject` from search results but leave the rest?

ksachs:
reject is THE button for arXiv submissions. And for arXiv I never look at the detailed page.
On the other hand we don't need/want a reject reason for arXiv, we don't send out reply emails.

jmartinm:
> reject is THE button for arXiv submissions. And for arXiv I never look at the detailed page.
On the other hand we don't need/want a reject reason for arXiv, we don't send out reply emails.

Ok perfect.

If `Reject` is not used for user submissions (since it needs more consideration and writing a reason) I could remove that button only for user submissions.

ksachs:
just thinking out loud: 
what if hitting a button for a single article in the listing pops up a new tab with the detailed view in the same state as hitting that button from the detailed page? (you know what I mean?)
The bulk processing (buttons for all selected articles) might only be available for arXiv or other sources where we don't send out emails.

jmartinm:
> what if hitting a button for a single article in the listing pops up a new tab with the detailed view in the same state as hitting that button from the detailed page? (you know what I mean?)

I think so :) kind of a shortcut that would just send you to the detailed page and press the same button (like Reject). It's actually not that simple since in that case the Reject button should be a link (like `/holdingpen/14?reject=True` then the detailed page needs to understand that URL and take the action). Not that it is not possible but not a one-line change.

michamos:
after discussing with @annetteholtkamp, we have several comments:

- For consistency, it would be good if reject is still present in the list view, but would take you to the detailed view (if not possible to pop-up the reject dialog directly).
- Why has the order been changed from Core, accept, reject as it is now on labs?
- It would be nice to have a way to select between several pre-filled reject reasons:
    * It doesn't fit
    * We already have it
    * It will be harvested automatically

If you provide a way to select between several options (and edit the resulting template), we can write something down for the other two.

jmartinm:
@michamos will look into that and let you know

jmartinm:
@michamos I have the dropdown selector mostly working now, so while I finish it you can already think about the text.

For reference, the one we use at the moment is:

```
Dear cataloger@inspirehep.net,

Thanks for suggesting "Optical Flow Fields: Dense Correspondence Fields for Highly Accurate  Large Displacement Optical Flow Estimation". We regret to inform you that we cannot include it in our database as it is outside the focus of INSPIRE. For details please check our collection policy at: https://inspirehep.net/info/hep/collection-policy.

Thanks again for collaborating with INSPIRE! We are looking forward to further suggestions you might have.
```

michamos:
Here is what I came up with. Any comments, @annetteholtkamp ?

## Rejection

Dear cataloger@inspirehep.net,

Thanks for suggesting "Optical Flow Fields: Dense Correspondence Fields for Highly Accurate  Large Displacement Optical Flow Estimation". We regret to inform you that we cannot include it in our database as it is outside the focus of INSPIRE. For details please check our collection policy at: https://inspirehep.net/info/hep/collection-policy.

Thanks again for collaborating with INSPIRE! We are looking forward to further suggestions you might have.

## Duplicate

Dear cataloger@inspirehep.net,

Thanks for suggesting "Optical Flow Fields: Dense Correspondence Fields for Highly Accurate  Large Displacement Optical Flow Estimation". We already have it in our database, you can find it here: http://inspirehep.net/record/INSERT_RECID.

Thanks again for collaborating with INSPIRE! We are looking forward to further suggestions you might have.

## Will be harvested

Dear cataloger@inspirehep.net,

Thanks for suggesting "Optical Flow Fields: Dense Correspondence Fields for Highly Accurate  Large Displacement Optical Flow Estimation". This article was published in a journal that we harvest automatically as part of our regular workflow. It is still in the queue of papers being processed and should appear in the coming weeks.

Thanks again for collaborating with INSPIRE! We are looking forward to further suggestions you might have.

jmartinm:
Pushed latest version that includes all the comments above.

* Leaves behaviour for arXiv untouched to allow for batch rejection (and no reason inside of detailed view)

* User submissions cannot be batch rejected from search results. On an individual one you can click Reject which will bring you to the detailed view.

* Added templates to user submissions Reject inside of detailed record. (with content above from @michamos )

PR #1978, opened 2017-03-06
literaturesuggest: improve conference autocomplete

jmartinm:
* Updates ElasticSearch mapping to fix conference autocompletion.
  (closes #1941)

* Amends acceptance test with a more real scenario.

Signed-off-by: Javier Martin Montull <javier.martin.montull@cern.ch>

PR #2003, opened 2017-03-09
records: regenerate mappings from schemas

jacquerie:


jmartinm:
Something specific to pay attention to in the schemas? It's just a copy/paste from our resolved inspire-schemas?

jacquerie:
> Something specific to pay attention to in the schemas? It's just a copy/paste from our resolved inspire-schemas?

So, now the tests pass, which was my first objective. Here's what I did so far:

1. Get the skeleton of the types with `domapping`
2. Fix the output so that it is a valid mapping
3. Add back only the `<thing>autocomplete` fields that we use, including the `<thing>_suggest` needed by the `record-editor`
4. Fix all the remaining issues until the tests pass

Now, before continuing, I want to sit down with you to:

5. Add back all the dates that we have
6. Add back all the analyzers that we use
7. Convert the `<thing>autocomplete`s to use the [`completion`](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-completion.html) type

Are you available on Skype tomorrow? Otherwise on Monday.

jacquerie:
As discussed on Gitter, merged in the interest of speed to unblock https://github.com/inspirehep/inspire-next/pull/1964 and https://github.com/inspirehep/inspire-next/pull/1978.

PR #2006, opened 2017-03-10
global: remove publication_info.confpaper_info

michamos:
* It was decided to remove this field from the schema, as it contains
useless information.
* Removes it from doJSON rules.
* Removes it from SEARCH_ELASTIC_KEYWORD_MAPPING in config.


jacquerie:
This is fixing a failure on Sentry, no? It would be cool if you added a test for this, taken from the Sentry report, and interlinked a bit the PR with the Sentry issue and viceversa.

michamos:
Cannot find anything related to this on sentry. Could you point me to the error?

What do you want me to test? `773` in general? or a test that asserts that `confpaper_info` is not produced (which seems a bit useless, as I grepped `confpaper_info` and it is not present any more in dojson)?

jacquerie:
> a test that asserts that `confpaper_info` is not produced

I agree that this would be a silly test, but this is not what I meant: I thought that `confpaper_info` was removed from the schema ([but it wasn't](https://github.com/inspirehep/inspire-schemas/blob/f0380968a6246b22db4cd9f4e15c047b0d963338/inspire_schemas/records/hep.json#L715-L717)), so Sentry had validation errors because of this.

michamos:
> I thought that confpaper_info was removed from the schema (but it wasn't), so Sentry had validation errors because of this.

Not yet, do you want me to remove it in the schema first, and update `inspire-schemas` to new version in this PR?

jacquerie:
> Not yet, do you want me to remove it in the schema first, and update inspire-schemas to new version in this PR?

That would be the cleanest way of doing this: you create a new major version of the schema, then you upgrade things on Inspire. Otherwise we have to merge this, upgrade the schema, and add another commit here to upgrade the schema version.

jacquerie:
You forgot [this occurrence](https://github.com/inspirehep/inspire-next/blob/fd98926452dbcb9dd27256ef205de95b4d0c013f/inspirehep/modules/records/mappings/records/hep.json#L524-L526). Other than that, LGTM, let's rebase and merge!

michamos:
Yeah, I had to respect causality, sorry about that :) Rebased now and removed the mapping.

jacquerie:
> Yeah, I had to respect causality, sorry about that :)

Damn that Einstein guy and his treacheries!

PR #2014, opened 2017-03-10
forms: prevent error message when submitting on Firefox

### LINKED ISSUE #1947: litterature suggest form: there's an error shows just after hitting submit and redirecting to the 'submission done' page
We should not show it if possible.

jmartinm:
* Firefox fires the jQuery AJAX fail handler on submit. Check the status
  code on the fail handler to only show an error when the server responds
  500. (closes #1947)

### DIFF
--- a/inspirehep/modules/forms/static/js/forms/inspire-form.js
+++ b/inspirehep/modules/forms/static/js/forms/inspire-form.js
@@ -468,12 +468,14 @@ define(function(require, exports, module) {
 
       set_status(tpl_status_saving());
 
-      $.ajax(
+      var promise = $.ajax(
         json_options({
           url: url,
           data: request_data
         })
-      ).done(function(data) {
+      );
+
+      promise.done(function(data) {
         var errors = handle_response(data);
         if (errors) {
           set_status(tpl_status_saved_with_error());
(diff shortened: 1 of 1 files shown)


inspirehep/modules/forms/static/js/forms/inspire-form.js
```diff
@@ -468,12 +468,14 @@ define(function(require, exports, module) {
 
       set_status(tpl_status_saving());
 
-      $.ajax(
+      var promise = $.ajax(
```
jacquerie: Do you really need this extra variable? I think that
```javascript
$.ajax(
  [...]
).fail(function (jqXHR, textStatus, errorThrown) {
  if (jqXHR.status == 500) {
    set_status(tpl_status_error());
  }
});
```
should work...

jmartinm: That's how it was before and while trying some things I separated it.

It is exactly the same :)

jmartinm: And I separated it because someone on the internet was wrong and was blaming the issue on Firefox to this chaining.

PR #2015, opened 2017-03-10
holdingpen: regenerate the mappings

### LINKED ISSUE #1924: Holding Pen: search does not work properly
Search inside the Holding Pen is not accurate. For example, if a submission has the title:

`Ratione explicabo Adipisci consectetur aperiam voluptates est quisquam esse numquam`

And I type `Ratione` nothing is found.

Same thing happens with author names.

jacquerie:
Adds back the `_all` field to the holdingpen mappings, so that
simple searches can be performed on those records (closes #1924).


jacquerie:
717 lines of code removed, 1 feature added : )

PR #2025, opened 2017-03-13
workflows: conditional halt in send_and_wait_robotupload

### LINKED ISSUE #2008: authors: workflow for updates should not wait for robotupload
The following commit https://github.com/inspirehep/inspire-next/commit/c97fb948a423a3ac9f09b50010863b19d85d1a13 changed the author workflow in a way that will make the workflow never complete.

The workflow needs to do:

* On update: send robotupload (in HoldingPen mode) and create a ticket.

* On new: send robotupload (in insert mode), create a ticket and wait. Once callback is received resume workflow.

So reusing `send_and_wait_robotupload` for both cases might not be possible (without modifying that function)

jmartinm:
* Renames send_and_wait_robotupload to send_robotupload.

* Halts workflow only when a callback URL is set. (closes #2008)

### DIFF
--- a/inspirehep/modules/workflows/tasks/submission.py
+++ b/inspirehep/modules/workflows/tasks/submission.py
@@ -228,16 +228,20 @@ def close_ticket(ticket_id_key="ticket_id"):
     return _close_ticket
 
 
-def send_and_wait_robotupload(
+def send_robotupload(
     url=None,
     marcxml_processor=None,
     callback_url="callback/workflows/robotupload",
     mode="insert",
     extra_data_key=None
 ):
-    """Get the MARCXML from the model and ship it."""
+    """Get the MARCXML from the model and ship it.
 
-    @wraps(send_and_wait_robotupload)
+    If callback_url is set the workflow will halt and the callback is
+    responsible for resuming it.
+    """
+
+    @wraps(send_robotupload)
     def _send_robotupload(obj, eng):
         from inspirehep.dojson.utils import legacy_export_as_marc
         from inspirehep.utils.robotupload import make_robotupload_marcxml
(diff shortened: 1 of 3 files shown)


inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -237,19 +237,21 @@ def send_and_wait_robotupload(
 ):
     """Get the MARCXML from the model and ship it."""
```
david-caro: Do you mind adding here a note about the callback_url parameter?

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -237,19 +237,21 @@ def send_and_wait_robotupload(
...
         from inspirehep.utils.robotupload import make_robotupload_marcxml
 
-        combined_callback_url = os.path.join(
-            current_app.config["SERVER_NAME"],
-            callback_url
-        )
-        if not combined_callback_url.startswith('http'):
-            combined_callback_url = "https://{0}".format(
-                combined_callback_url
+        combined_callback_url = ""
```
david-caro: I'm not sure how legacy interprets the callback_url parameter and such... maybe @kaplun can shed some light? (I tried for 5 min to find out where is the code of the batchupload on the old code... no luck).

david-caro: Another option is just try it out on qa

jmartinm: I tried yesterday with my localhost connected to inspireheptest and passing empty meant that no parameter arrived to the `robotupload` command so should be fine.

david-caro: 👍

jmartinm:
Updated.

PR #2027, opened 2017-03-13
docs: add overview to getting started

spirosdelviniotis:
* Adds overview section


INSTALL.rst
```diff
@@ -1,9 +1,30 @@
...
+
+    You should have received a copy of the GNU General Public License
+    along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+
+    In applying this licence, CERN does not waive the privileges and immunities
+    granted to it by virtue of its status as an Intergovernmental Organization
+    or submit itself to any jurisdiction.
+
+
+Native Install (CentOS - MacOS)
```
jmartinm: Not sure if it needs to be done in this PR, but adding in the title both systems might be misleading.

For example, in the first section you use `yum` which is not available in MacOS. If there is nothing specific of MacOS you can remove it from the title and later on we can add some specific section with Ubuntu or MacOS if needed.

docs/docker.rst
```diff
@@ -0,0 +1,160 @@
...
+    In applying this licence, CERN does not waive the privileges and immunities
+    granted to it by virtue of its status as an Intergovernmental Organization
+    or submit itself to any jurisdiction.
+
+
+Docker (Linux)
+==============
+
+Docker is an application that makes it simple and easy to run processes in a container,
+which are like virtual machines, more resource-friendly. For a detailed introduction to the
```
jmartinm: but more resource-friendly.

docs/docker.rst
```diff
@@ -0,0 +1,160 @@
...
+different components of a Docker container.
```
jmartinm: Was the intention to put the link as part of this sentence?

For a detailed introduction to the different components of a Docker container, you can follow this tutorial(add link here)

docs/docker.rst
```diff
@@ -0,0 +1,160 @@
...
+Useful `Tutorial
+<https://www.digitalocean.com/community/tutorials/the-docker-ecosystem-an-introduction-to-common-components>`_.
+
+
+Inspire and Docker
+##################
+
+Get the latest Docker setup.
+
+.. note:: If you are using Mac, please build a simple box with ``docker-engine`` above ``1.10`` and
```
jmartinm: @jacquerie I don't think we need this now that we have native docker?

docs/docker.rst
```diff
@@ -0,0 +1,160 @@
...
+         ``docker-compose`` above ``1.6.0``.
+
+Make sure you can run docker without ``sudo``.
+
+- ``id $USER``
+
+  If you are not in the ``docker`` group, run the following command and then restart ``docker``. If this doesn't work, just restart your machine :)
```
jmartinm: Is there a reason why restarting the machine would fix it?

docs/docker.rst
```diff
@@ -0,0 +1,160 @@
...
+ sudo pip install docker-compose
+
+- Add ``DOCKER_DATA`` env variable in your ``.bashrc`` or ``.zshrc``. In this directory you will have all the persistent data between Docker runs.
+
+.. code-block:: console
+
+ shell mkdir -p "$DOCKER_DATA"
+ export DOCKER_DATA=~/inspirehep_docker_data/
+
+By default the virtualenv and everything else will be kept on ``/tmp`` and they will be available only until the next reboot.
```
jmartinm: is this true? When dealing with virtualenv or Docker I never had to reboot the machine to get it to work.

docs/docker.rst
```diff
@@ -0,0 +1,160 @@
...
+         - cleanup the virtualenv (careful, if docker_data is set to something you care about, it will be removed):
+
+           ``sudo rm -rf "${DOCKER_DATA?DOCKER_DATA was not set, ignoring}"``
+
+Extra useful tips
+#################
+
+- Run a random shell
+
+After ``docker-compose up`` just run:
```
jmartinm: All of this don't need `docker-compose up` to work. can you double check?

docs/docker.rst
```diff
@@ -0,0 +1,160 @@
...
+
+ docker-compose restart worker
+
+- Quick and safe reindex
+
+With ``docker-compose up`` just run:
+
+.. code-block:: console
+
+ docker-compose restart worker && docker-compose run --rm web scripts/recreate_records
```
jmartinm: You can add here:

Recreate all static assets. Will download all dependencies from npm and copy all static files to `${DOCKER_DATA}/tmp/virtualenv/var/inspirehep-instance/static`

`docker-compose -f docker-compose.deps.yml run --rm assets`

docs/docker.rst
david-caro: That's because you need to refresh the groups your current user is in. That can be done by the newgrp command under, or re-logging in (with sudo, su, or well, restarting).

INSTALL.rst
david-caro: This pr is actually on top of #1944, so this commit is actually added on that one.

docs/docker.rst
david-caro: I'm not sure what you mean, but yes, `/tmp` is a tmpfs file system, that gets cleaned up when you reboot the machine, so anything that's stored there is lost on reboot.

docs/docker.rst
spirosdelviniotis: You are right!

docs/docker.rst
spirosdelviniotis: Yeah it's not need. The only reason that we want it its just for monitoring/logging purposes.

jacquerie:
This is actually superseded by #1944: there's more content there.

PR #2039, opened 2017-03-15
records: ensure date type (and format) in mappings

chris-asl:
Setting date type (instead of string) for all date-related fields.
Only special case was `hep.thesis_info.defense_date` whose values
on nightly are `PhD`. This field's type has been left as string.

### DIFF
--- a/inspirehep/modules/records/mappings/records/authors.json
+++ b/inspirehep/modules/records/mappings/records/authors.json
@@ -27,7 +27,8 @@
                 "acquisition_source": {
                     "properties": {
                         "date": {
-                            "type": "string"
+                            "type": "date",
+                            "format": "date||yyyy-MM||yyyy"
                         },
                         "email": {
                             "type": "string"
@@ -90,7 +91,8 @@
                     "type": "string"
                 },
                 "birth_date": {
-                    "type": "string"
+                    "type": "date",
+                    "format": "date||yyyy-MM||yyyy"
                 },
                 "conferences": {
                     "properties": {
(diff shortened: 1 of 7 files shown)


michamos:
The schema is authoritative, not the crap in the records. As the name and the format suggest (https://michamos.github.io/inspire-schemas/hep.html#defense-date ), the contents of `defense_date` is a date.

@jacquerie how did this get produced? searching on legacy for `502__a:**` (which is the field that doJSON populates defense date from) does not yield any result. BTW, we really should have a `format_checker` enforcing `format`s.

jacquerie:
> @jacquerie how did this get produced? searching on legacy for 502__a:** (which is the field that doJSON populates defense date from) does not yield any result.

There used to be stuff there: https://inspirehep.net/record/edit/?ln=en#state=edit&recid=897773&recrev=20160615220421, so I assume that the data was corrected in the meantime. I must have found it in the TWiki, and went with it, even if it didn't make any sense (see: https://github.com/inspirehep/inspire-next/blob/e1fd5b6b87039983d7c9abd94756bc22a7c04785/tests/unit/dojson/test_dojson_hep_bd5xx.py#L166)

jacquerie:
> BTW, we really should have a format_checker enforcing formats.

Enforcing the mappings on dates is a first step in that direction. My plan is to do a Nightly migration with these mappings and collect all the errors, then use them to inform development of DoJSON rules to normalize dates. Then, we implement format checkers at least for dates.

jacquerie:
Stalled, because the work on correcting legacy dates must start from the DB, not ES... I was wrong in asking you to do this, @chris-asl : /

jacquerie:
Will be resurrected after we've done this work for the DB.

PR #2041, opened 2017-03-15
authors: review form fixes

jmartinm:
* Several fixes to ensure that review form gets all the fields.

* Update acceptance tests.


inspirehep/modules/authors/dojson/fields/updateform.py
```diff
@@ -144,7 +144,7 @@ def bai(self, key, value):
 def inspireid(self, key, value):
     inspireid = {
         'value': value,
-        'type': 'INSPIRE'
+        'type': 'INSPIRE ID'
```
jacquerie: Same as above, `'schema': 'INSPIRE ID`.

jmartinm:
This is ready. Still more exhaustive tests would be needed but should make it more robust.

inspirehep/utils/helpers.py
```diff
@@ -115,3 +115,22 @@ def force_list(data):
...
+        dict: a dict with empty elements removed.
+    """
+    if not isinstance(data, (dict, list)):
+        return data
+    if isinstance(data, list):
+        return [v for v in (clean_empty(v) for v in data) if v or v == 0]
+    return {
+        k: v for k, v in ((k, clean_empty(v)) for k, v in data.items())
+        if v or v == 0
+    }
```
jacquerie: Isn't this duplicating https://github.com/inspirehep/inspire-next/blob/c1f136cf7131e4c930a39c21618e03238a78ccb2/inspirehep/dojson/utils/__init__.py#L197-L219 ?

jmartinm: Yes, they seem to do the same (one in place and the other one not). I can just reuse the existing one.

PR #2042, opened 2017-03-16
dojson: handle defense_date in 500__a

### LINKED ISSUE #2040: doJSON: thesis_info.defense_date is incorrectly handled
On legacy, these dates are written in `500__a` as `presented on 2017-03-12`, not `502__a`.
This may partly explain the remark in #2039.

jacquerie:
Closes #2040


michamos:
Maybe add a test for a full `thesis_info`, with info both from `500__a` and `502`?

PR #2044, opened 2017-03-16
migrator: add missing __init__.py

kaplun:


kaplun:
cc: @david-caro @jacquerie here is the missing file!

PR #2045, opened 2017-03-16
tests: acceptance without visible firefox instance

rikirenz:
I have added a service called `not_visible_acceptance`. This routine should run the acceptance tests in background. 
It is useful because sometimes we do not need to check what's going on with these tests and it allows you to do other things without be interrupted by the firefox window.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -100,6 +100,23 @@ services:
     environment:
       - SERVER_NAME=web:5000
       - DISPLAY=$DISPLAY
+  not_visible_acceptance:
+    extends:
+      service: service_base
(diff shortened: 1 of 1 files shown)


rikirenz:
if you think is not useful it is fine. But in that case we should put in the documentation how to run these tests without the firefox instance visible. IMHO during development it is very annoying.

david-caro:
We agreed to make the default the 'headless' setup, and using a custom one (maybe something like acceptance and acceptance_with_gui or similar).

Will just try it locally as it's quite useful 👍

jmartinm:
> We agreed to make the default the 'headless' setup, and using a custom one (maybe something like acceptance and acceptance_with_gui or similar).

Sounds good.

+1 for keeping both options though.

PR #2058, opened 2017-03-17
tests: use pytest-bdd for acceptance

rikirenz:
- Here there is an example of acceptance test written with **behavioral driven development**.


tests/acceptance-bdd/conftest.py
```diff
@@ -0,0 +1,99 @@
...
+def login(arsenic):
+    top_navigation_page.log_in('admin@inspirehep.net', '123456')
+    yield
+    top_navigation_page.log_out()
+
+
+@pytest.fixture(autouse=True, scope='module')
+def cleanup_workflows_tables(app):
+    with app.app_context():
+        obj_types = (
```
jmartinm: You can simplify with:

```python
WorkflowsAudit.query.delete()
WorkflowsPendingRecord.query.delete()
workflow_object_class.query().delete()
```

rikirenz: Could you explain why :dancing_men:

jmartinm: Just from the documentation - http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.delete 

it seems like you can directly call .delete() and it will remove all rows.

http://stackoverflow.com/a/16573863/890185

docker-compose.test.yml
```diff
@@ -83,7 +83,7 @@ services:
   acceptance:
     extends:
       service: test-service_base
-    command: py.test --driver Firefox --host selenium --port 4444 --capability browserName firefox --html=selenium-report.html tests/acceptance
+    command: py.test --driver Firefox --host selenium --port 4444 --capability browserName firefox --html=selenium-report.html tests/acceptance-bdd
```
rikirenz: Only to test on travis. This commit is gonna be removed.

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,392 @@
...
+    pass
+
+
+@scenario('literature_suggestion.feature', 'Accept thesis and verify the confirmation message')
+def test_verify_accepted_thesis():
+    pass
+
+
+@given('I am logged in')
+def author_user(login):
```
jmartinm: Now that we use BDD, can the sentence `I am logged in` become a reusable piece of code that does the login? Not to mix different styles, we can remove the fixture.

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+@when('I go to the holding panel list page')
+def go_to_holding_pen_literature_list():
+    Arsenic().get(os.environ['SERVER_NAME'] + '/holdingpen/list/?page=1&size=10&status=HALTED&workflow_name=HEP')
+
+
+@when(parsers.cfparse('I click on the button with id {button_id}'))
+def click_generic_button(button_id):
+    WebDriverWait(Arsenic(), 2).until(
```
jmartinm: Above you have a timeout of 10 and these ones here have 2. If I am correct this wait is implicit right? So it's the maximum waiting time before error.

You seem to be repeating a lot the code:

```
WebDriverWait(Arsenic(), 2).until(
EC.visibility_of_element_located((By.ID, button_id))
).click()
```

Would be good to have a helper with a default value for the wait and where you just pass the part inside the `until`

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+@when('I click on the accept button')
+def click_accept_button():
+    WebDriverWait(Arsenic(), 2).until(
+        EC.visibility_of_element_located((By.XPATH, '//button[@class="btn btn-warning"]'))
+    ).click()
+
+
+@when('I click on the button Skip, and fill the form manually')
```
jmartinm: This is more:

```
'I click on the button Skip, and uncollapse all panels'
```

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,392 @@
...
+            null,
+            XPathResult.FIRST_ORDERED_NODE_TYPE,
+            null
+        ).singleNodeValue.click()"""
+    )
+    Arsenic().show_title_bar()
+
+
+@when(parsers.cfparse('I select the value {input_select} in the select box with id {field_id}'))
+def select_generic_value(input_select, field_id):
```
jmartinm: This nice generic functions are now on this file `test_literature_suggestion_form.py`. Can they be moved to a file where one can find reusable `when` clauses?

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,392 @@
...
+    Select(Arsenic().find_element_by_id(field_id)).select_by_value(input_select)
+
+
+@when(parsers.cfparse('I select the value {input_select} in the subject dropdown list'))
+def select_generic_value_dropdown(input_select):
+    Arsenic().find_element_by_xpath('(//button[@type="button"])[8]').click()
```
jmartinm: Even though this function is called `select_generic_value_dropdown` this line is quite specific.

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,392 @@
...
+    assert globals()['_is_' + field + '_error_message_present']()
+
+
+@then(parsers.cfparse('I should not see the {field} error message'))
+def is_not_there_a_message_error(field):
+    assert not globals()['_is_' + field + '_error_message_present']()
+
+
+@when(parsers.cfparse('I insert <input> in the input box with id {field_id}'))
+@when(parsers.cfparse('I insert {input} in the input box with id {field_id}'))
```
jmartinm: Are both needed?

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,392 @@
...
+    assert expected_data == field.get_attribute('value')
+
+
+@then(parsers.cfparse('I should see {value_record_entry} in the record entry'))
+def check_record_detail(value_record_entry):
+    record = WebDriverWait(Arsenic(), 10).until(
+        EC.visibility_of_element_located(
+            (
+                By.XPATH,
+                '//div[@class="row hp-item ng-scope"][1]'
```
jmartinm: What is this selecting? (we can add an id to the html to have it more clear)

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,392 @@
...
+    assert value_record_entry in record
+
+
+@then(parsers.cfparse('I should see {value_record_detail} in the record detail'))
+def check_record_detail(value_record_detail):
+    record = WebDriverWait(Arsenic(), 10).until(
+        EC.visibility_of_element_located(
+            (
+                By.XPATH,
+                '(//div[@class="ng-scope"])[2]'
```
jmartinm: Same here and the selectors below, and `id` in the html will make it more readable.

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,392 @@
...
+            (
+                By.XPATH,
+                '//div[@class="row hp-item ng-scope"][1]/div/div/div[2]/holding-pen-template-handler/div[3]/a'
+            )
+        )
+    ).click()
+
+
+@then('I should see the message of confirmation')
+def click_submit_button():
```
jmartinm: wrong name of function

tests/acceptance-bdd/literature_suggestion.feature
```diff
@@ -0,0 +1,306 @@
...
+    Given I am logged in
+    When I go to the literature suggestion page
+    And I insert <input> in the input box with id arxiv_id
+    Then I should see the arxiv error message
+
+  Examples: Vertical
+    | input | hep-th.9711200 |
+
+  # DOI text input
+  Scenario: Writing the correct DOI id
```
michamos: "DOI" stands for Digital Object Identifier, so you can drop the "id" here and in the following.

tests/acceptance-bdd/literature_suggestion.feature
```diff
@@ -0,0 +1,306 @@
...
+    | input | wrong | 2016-02-30 | 2016-13 |
+
+
+
+  # URL text input
+  Scenario: Writing a correct url
+    Given I am logged in
+    When I go to the literature suggestion page
+    And I click on the button Skip, and fill the form manually
+    And I insert pdf_url_correct in the input box with id url
```
michamos: same question, why not do it as an example like for the other cases?

tests/acceptance-bdd/literature_suggestion.feature
```diff
@@ -0,0 +1,306 @@
...
+  Scenario: Writing a wrong url
+    Given I am logged in
+    When I go to the literature suggestion page
+    And I click on the button Skip, and fill the form manually
+    And I insert pdf_url_wrong in the input box with id url
+    Then I should see the url error message
+
+
+  # thesis info autocomplete for supervisor institution
+  Scenario: Use the autocomplete for the supervisor institution
```
michamos: Use -> Using for consistency with the previous tests?

tests/acceptance-bdd/literature_suggestion.feature
rikirenz: `pytest-bdd` is able to manage 2 kinds of input parameters:

- <First_type>
- {Second_type}

This is the [Gherkin](https://github.com/cucumber/cucumber/tree/master/gherkin) syntax, the variable templates are written using corner braces as _<somevalue>_ when we need to test multiple values or curly braces as _{somevalue}_ when we need to test a single value that has to be parameterized.

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,392 @@
...
+def test_import_from_doi():
+    pass
+
+
+@scenario('literature_suggestion.feature', 'Writing the wrong arXiv id')
+def test_wrong_arxiv():
+    pass
+
+
+def test_correct_arxiv():
```
rikirenz: My bad!

tests/acceptance-bdd/test_literature_suggestion_form.py
rikirenz: You are totally right! Actually yesterday I was reading again the selenium doc and I found this best practice:
http://selenium-python.readthedocs.io/page-objects.html?highlight=locator#locators

> One of the practices is to separate the locator strings from the place where they are being used. In this example, locators of the same page belong to same class.

tests/acceptance-bdd/literature_suggestion.feature
```diff
@@ -0,0 +1,306 @@
...
+    And I should see 525-553 in the input box with id page_range_article_id
+    And I should see Schlegel, David J. in the input box with id authors-0-name
+    And I should see 10.1086/305772 in the input box with id doi
+    And I should see Maps of Dust Infrared Emission for Use in Estimation of Reddening and Cosmic Microwave Background Radiation Foregrounds in the input box with id title
+
+  Examples: Vertical
+    | input | 10.1086/305772 |
+
+
+  Scenario: Submit thesis and verify record detail in the list page
```
michamos: This and the following tests should use realistic data, also containing unicode

tests/acceptance-bdd/test_literature_suggestion_form.py
rikirenz: What do you think about:  `select_value_in_literature_dropdown`

tests/acceptance-bdd/test_literature_suggestion_form.py
rikirenz: If we group the locator as the Selenium doc sad probably even though most of the locators are gonna use the XPATH syntax, they will have a specific name that can describe what they do. In order to make them more clear and at the same time avoid to change the code only for the tests.

tests/acceptance-bdd/test_literature_suggestion_form.py
rikirenz: My bad again!

tests/acceptance-bdd/test_literature_suggestion_form.py
rikirenz: yes, we need both!

tests/acceptance-bdd/test_literature_suggestion_form.py
rikirenz: you are right again! From the moment that we are going to use this framework it makes sense spend time to organize the code

tests/acceptance-bdd/test_literature_suggestion_form.py
rikirenz: I think this is not possible, I will try, but probably we still need to use the fixtures.

tests/acceptance-bdd/test_literature_suggestion_form.py
jmartinm: Since this file is all about the literature suggestion form you can drop literature from the name. Something like:

`select_value_in_subject_multiselect_widget`

can work.

tests/acceptance-bdd/literature_suggestion.feature
rikirenz: Make it work
make it right 
make it realistic 
make it fast

I will add a test with unicode in this PR

tests/acceptance-bdd/literature_suggestion.feature
kaplun: But... there are no curly brackets, here.

kaplun: Ah I see what you mean. The curly brackets are used then in the decorator.

rikirenz: You got it!

tests/acceptance-bdd/test_literature_suggestion_form.py
rikirenz: @jmartinm I have found a way!! :wink:

tests/acceptance-bdd/test_literature_suggestion_form.py
rikirenz: We need both because with that function we are able to handle 2 different statement syntax.

inspirehep/bat/steps/when.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@when(parsers.cfparse('I click on the button with id {button_id}'))
+def click_generic_button(button_id):
+    WebDriverWait(Arsenic(), 2).until(
+        EC.visibility_of_element_located((By.ID, button_id))
+    ).click()
+
+
+@when(parsers.cfparse('I click on the button link with text {text_value}'))
```
jmartinm: Is it just for links?  So no need to say `button`

tests/acceptance-bdd/literature_suggestion.feature
```diff
@@ -0,0 +1,287 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
```
jmartinm: This file is new, so just 2017

tests/acceptance-bdd/literature_suggestion.feature
```diff
@@ -0,0 +1,287 @@
...
+    | input | 10.1086/305772 |
+
+  Scenario: Writing the wrong DOI
+    Given I am logged in
+    When I go to the literature suggestion page
+    And I insert <input> in the input box with id doi
+    Then I should see the doi error message
+
+  Examples: Vertical
+    | input | dummy:10.1086/305772 | state-doi
```
jmartinm: `state-doi` is a wrong doi? We can give it a more meaningul name

tests/acceptance-bdd/literature_suggestion.feature
```diff
@@ -0,0 +1,287 @@
...
+    And I go to the holding panel list page
+    And I should see the record in the list page
+    And I click on the button link with text Show abstract
+    Then I should see Computing in the record entry
+    And I should see My Title For Test in the record entry
+    And I should see admin@inspirehep.net in the record entry
+    And I should see Mister White; Mister Brown in the record entry
+    And I should see Lorem ipsum dolor sit amet, consetetur sadipscing elitr. in the record entry
+
+  Scenario: Submit thesis and verify record detail in the detail page
```
jmartinm: `Submit thesis and verify record detail in the Holding Pen`?

tests/acceptance-bdd/literature_suggestion.feature
```diff
@@ -0,0 +1,287 @@
...
+    And I insert Lorem ipsum dolor sit amet, consetetur sadipscing elitr. in the input box with id abstract
+    And I click on the submit button
+    And I go to the holding panel list page
+    And I should see the record in the list page
+    And I click on the button link with text My Title For Test
+    And I click on the accept button
+    Then I should see the message of confirmation
+
+
+  Scenario: Submit article and verify record detail in the list page
```
jmartinm: Above you call `record detail` the detailed page of the Holding Pen, might be confusing to call it the same the entry in the search results.

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,287 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
```
jmartinm: Only 2017

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,287 @@
+# -*- coding: utf-8 -*-
```
jmartinm: For the name of the file to be consistent with the one above it can be `literature_suggestion.feature` and `test_literature_suggestion.py`

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,287 @@
...
+    Arsenic().get_element_with_locator(CL.SKIP_IMPORT_DATA, waiting_time=5).click()
+
+    WebDriverWait(Arsenic(), 10).until(
+        EC.text_to_be_present_in_element(
+            CL.LITERATURE_FORM,
+            'Type of Document',
+        )
+    )
+
+    Arsenic().get_element_with_locator(CL.CONFERENCE_TAB).click()
```
jmartinm: I would call them panel, so `CONFERENCE_PANEL` to be more accurate

tests/acceptance-bdd/test_literature_suggestion_form.py
```diff
@@ -0,0 +1,287 @@
...
+    Arsenic().get_element_with_locator(CL.DROPDOWN_BUTTON).click()
+
+
+# when statements for holdingpen pages
+@when('I go to the holding panel list page')
+def go_to_holding_pen_literature_list():
+    Arsenic().get(os.environ['SERVER_NAME'] + '/holdingpen/list/?page=1&size=10&status=HALTED&workflow_name=HEP')
+
+
+@when('I should see the record in the list page')
```
jmartinm: Is this a `when` or a `then` ?

tests/acceptance-bdd/conftest.py
```diff
@@ -0,0 +1,88 @@
...
+        yield app
+
+
+@pytest.fixture(autouse=True, scope='function')
+def arsenic(selenium, app):
+    Arsenic(selenium)
+
+
+@pytest.fixture(autouse=True, scope='function')
+def cleanup_workflows_tables(app):
```
rikirenz: This fixture is executed every time that we run a test in order to delete all the records. This helps to maintain all the tests atomic.

inspirehep/bat/locators.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class HoldingpenLiteratureListPageLocators(object):
+    RECORD_ENTRY = (By.XPATH, '//div[@class="row hp-item ng-scope"][1]')
+    SHOW_ABSTRACT_BUTTON = (By.XPATH, '//div[@class="row hp-item ng-scope"][1]/div/div/div[2]/holding-pen-template-handler/div[3]/a')
+
+
+class HoldingpenLiteratureDetailPageLocators(object):
+    PRINCIPAL_RECORD_PANEL = (By.XPATH, '(//div[@class="ng-scope"])[2]')
+    SUBMISSION_PANEL = (By.XPATH, '//p[@class="text-center ng-scope"]')
```
rikirenz: No they are respectively:

- This panel
image

- And this panel 
image

tests/acceptance-bdd/test_literature_suggestion_form.py
rikirenz: By default pytest-bdd does not allow to mix the syntax of the steps, that should remain always the following:

- Given
- Then
- When

buuuuuut actually I discovered this PR: https://github.com/pytest-dev/pytest-bdd/pull/130 where they have added a way to relax the strict Gherkin syntax. This allows to mix different steps even thought they do not follow the classic order `Given when then`. 
 
What do you think? Should we allow to use a mixed syntax? 

IMHO YES! Morover in this issue https://github.com/pytest-dev/pytest-bdd/issues/118 a guy that was writing selenium test sad that was necessary for this kind of tests use a mixed syntax.

rikirenz:
Guys I completely forget about this PR! Do you know the state of this PR? Am I missing something? Maybe there is something wrong that I missed...

kaplun:
Hi @rikirenz! I believe one issue that arose was that formally these are not end-to-end testing in the sense that are not testing UI->Python->DB but UI->Python->DB->Python->UI. I am not sure if this is actually a blocker for this PR, though. @jacquerie might want to say more.
Another potential issue is that of course these type of tests are making many assumption on the UI (which is unavoidable IMHO) therefore making more rigid the possibility of updating it in an *agile* (:dancer:) way. But I don't see a way out of this. Tests involving the UI are going to always impose some rigidity on the UI itself.

jacquerie:
The point of BDD is to express tests in the domain of the user. This PR instead is still using the language of the developer: users don't think in terms of "when I push this button and I fill this text", but in terms of "when I add a new INSPIRE category".

Another objective of this PR was to simplify the complicated architecture of the acceptance tests, and instead it is making it more complicated, as it's adding one extra level of abstraction (which, as a I said above is not abstracting at all, it's merely recasting the below level with other code).

Because of these two main reasons, I'm going to close this PR. Sorry!

PR #2061, opened 2017-03-18
docker: use different names for test containers

david-caro:
* This avoids the name collisions with the web node container (and
  index/database...).

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -23,128 +23,122 @@
 version: '2'
 
 services:
-  service_base:
+  test-service_base:
     # Overrides default inspirehep config.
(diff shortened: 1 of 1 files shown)


docker-compose.test.yml
```diff
@@ -23,128 +23,122 @@
...
-      - redis
+      - test-database
+      - test-indexer
+      - test-rabbitmq
+      - test-redis
     depends_on:
-      - worker
-  integration:
+      - test-worker
+  test-integration:
```
jacquerie: This looks like it's causing https://travis-ci.org/inspirehep/inspire-next/jobs/212426038.

PR #2063, opened 2017-03-19
forms: autocomplete without regex queries

### LINKED ISSUE #2059: Trying to autocomplete conferences with regexp symbols breaks the es search
When submitting a new literature record, when entering input in the conference, it autocompletes correctly but tries again with the retrieved string and if it has regexp chars, when trying again it breaks elasticsearch!

Example query that will generate the error:

http://localhost:5000/api/conferences/?q=conferenceautocomplete:%2Fnational.*%2F%20AND%20conferenceautocomplete:%2Facademy.*%2F%20AND%20conferenceautocomplete:%2Fof.*%2F%20AND%20conferenceautocomplete:%2Fsciences.*%2F%20AND%20conferenceautocomplete:%2Fautumn.*%2F%20AND%20conferenceautocomplete:%2Fmeeting.*%2F%20AND%20conferenceautocomplete:%2F(october.*%2F%20AND%20conferenceautocomplete:%2F1972)%3A.*%2F%20AND%20conferenceautocomplete:%2Fsymposium%3A.*%2F%20AND%20conferenceautocomplete:%2Fnew.*%2F%20AND%20conferenceautocomplete:%2Faccelerators.*%2F

jacquerie:
Closes #2053 
Closes #2059

### DIFF
--- a/inspirehep/modules/forms/static/js/forms/affiliations_typeahead.js
+++ b/inspirehep/modules/forms/static/js/forms/affiliations_typeahead.js
@@ -31,19 +31,7 @@ define([
     this.dataEngine = new Bloodhound({
       name: 'affiliations',
       remote: {
-        url: '/search?cc=Institutions&p=affautocomplete:/%QUERY.*/&of=recjson&rg=100',
-        replace: function(url, query) {
-          var query_components = query.toLowerCase().split(" ");
-          var pattern = "";
-          $.each(query_components, function(index) {
-            if (index != 0) {
-              pattern = pattern + " AND ";
-            }
-            pattern = pattern + "affautocomplete:" + "/" + encodeURIComponent(this) + ".*/";
-          })
-
-          return '/api/institutions?q=' + pattern
-        },
+        url: '/api/institutions?q=affautocomplete:%QUERY*',
         filter: function(response) {
           return $.map(response.hits.hits, function(el) { return el });
         }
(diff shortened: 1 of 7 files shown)


jmartinm:
Are you planning on changing the ES config on this PR to use the `completion suggester`? I can test the PR tomorrow, we just need to make sure that by fixing the regexp problem we don't break the previous behaviour that the code removed was doing. That is to get suggestions even if there is an space between words.

jacquerie:
> Are you planning on changing the ES config on this PR to use the `completion suggester`?

Not yet. But it would be the right thing to do.

> we just need to make sure that by fixing the regexp problem we don't break the previous behaviour that the code removed was doing. That is to get suggestions even if there is an space between words.

I'm not breaking it: for example, autocompletion for authors returns the right result "Gligorov, Vladimir Vava" when typing "Gli", "Vla" or "Vav". If you want to be 100% sure of this we can quickly double check tomorrow.

jmartinm:
> I'm not breaking it: for example, autocompletion for authors returns the right result "Gligorov, Vladimir Vava" when typing "Gli", "Vla" or "Vav". If you want to be 100% sure of this we can quickly double check tomorrow.

What about typing "Gligorov Vla" ?

jacquerie:
> What about typing "Gligorov Vla" ?

Still works.

jmartinm:
Just tried and works for me.

jmartinm:
As checked IRL, this can be simplified even further with:

`url: '/api/authors?q=authorautocomplete:%QUERY*',`

without `replace` function.

jacquerie:
I also cherry-picked here an unrelated commit from #2053 so that we merge both faster.

PR #2071, opened 2017-03-20
setup: use kaplun's invenio-migrator fork

kaplun:


kaplun:
Tested on production.

PR #2079, opened 2017-03-20
literaturesuggest: add internal_uid to acquisition_source

### LINKED ISSUE #2075: /acquisition_source/inspire_uid does not get populated by the submission
There is a `/_workflow/id_user` key in the holdingpen record, but it is not replicated in the record metadata.

jacquerie:
Closes #2075

### DIFF
--- a/tests/integration/fixtures/conference_article_expected.json
+++ b/tests/integration/fixtures/conference_article_expected.json
@@ -17,6 +17,7 @@
       "source":"submitter",
       "method":"submitter",
       "submission_number":"1",
+      "internal_uid":3,
       "orcid":"0000-0001-9412-8627",
       "email":"test_orcid_user@inspirehep.net"
    },
--- a/inspirehep/modules/literaturesuggest/tasks.py
+++ b/inspirehep/modules/literaturesuggest/tasks.py
@@ -206,6 +206,7 @@ def formdata_to_model(obj, formdata):
     builder.add_acquisition_source(
         datetime=datetime.datetime.utcnow().isoformat(),
         submission_number=obj.id,
+        internal_uid=int(obj.id_user),
         email=email,
         orcid=orcid,
         method='submitter'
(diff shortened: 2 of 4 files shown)


inspirehep/modules/workflows/workflows/article.py
```diff
@@ -76,6 +76,7 @@
 from inspirehep.modules.workflows.tasks.upload import store_record, set_schema
 from inspirehep.modules.workflows.tasks.submission import (
     add_note_entry,
+    add_internal_uid,
```
jmartinm: Why not add it directly to https://github.com/jacquerie/inspire-next/blob/3316f65c3fe9b60c68c8c59e9491a1564f431986/inspirehep/modules/literaturesuggest/tasks.py#L206? Even though `formdata_to_model` might need a refactor, for the moment it would be easier to reason if adding `acquisition_source` is in one place.

jacquerie: You're 100% right on this. Let me redo this PR...

jacquerie: There's a problem with this (even if your approach is still better): integration tests for `formdata_to_model` will need to hardcode some user id in the fixture, but it's not guaranteed to be constant across multiple runs of the tests. This is a limitation of the design of the tests of `formdata_to_model`...

tests/integration/fixtures/conference_article_expected.json
```diff
@@ -17,6 +17,7 @@
       "source":"submitter",
       "method":"submitter",
       "submission_number":"1",
+      "internal_uid":3,
```
jacquerie: This test is brittle: if the ordering of the test changes, or some other users are created during a test run, the assertion that uses this static fixture is going to fail.

But I'm taking the decision to not compromise the code quality in favor of compromising the test quality until these tests are properly refactored.

Issue #2076, opened 2017-03-20
arXiv harvest holdingpen records have no reject button


michamos:
![screenshot from 2017-03-20 18-32-27](https://cloud.githubusercontent.com/assets/2895793/24112994/9b46899e-0d9b-11e7-8f5c-53e62e9e3042.png)

https://newlabs.inspirehep.net/holdingpen/600021

jacquerie:
Fixed in #2081.

PR #2081, opened 2017-03-20
workflows: show reject method on hepcrawl workflows

david-caro:


david-caro:
Closes #2076

jmartinm:
Did the value in `acquisition_source` change? I remember when I was testing in the past with:

```
inspirehep oaiharvester harvest -u http://export.arxiv.org/oai2 -m arXiv -f 2016-06-01 -t 2016-06-04 -s 'physics:hep-lat' -a spider=arXiv -a workflow=article
```

That was the correct value and the Reject button was showing...

david-caro:
Yep, I think it changed in the schema.

PR #2097, opened 2017-03-21
workflows: populate arXiv card with arXiv papers

### LINKED ISSUE #2091: holdingpen dashboard: harvested arxiv papers don't get into the arxiv card
The source is now 'hepcrawl' instead of 'arxiv', so it has to search instead for 'acquisiton_source.source' that is arxiv.

jacquerie:
Closes #2091


jacquerie:
Tested locally that this works. This will require a reindex of the holdingpen records to work in production.

david-caro:
Tried locally several times, and with the new mapping, the literature submissions end up showing also on the arxiv card

inspirehep/config.py
```diff
@@ -1437,7 +1438,7 @@ def _(x):
             },
             "source": {
                 "terms": {
-                    "field": "metadata.acquisition_source.method",
+                    "field": "metadata.acquisition_source.source",
```
david-caro: There's a bit of confusion on what the 'source' and the 'method' is, for forms for example, the method and the source are the same, but for hepcrawl they are not (that makes sense).

Maybe we should change the method of the submissions for something like 'forms' at some point.

david-caro: @michamos what do you think?

david-caro: Currently is an open string (not enum).

michamos: but isn't `forms` a synonym for `submitter` or `user` or whatever it is we are using?

jmartinm: At the moment, for all the forms we put `submitter` as a method.

michamos: we could rename `submitter` to `forms` or something like that, but not sure if we gain anything.

david-caro: We might have at some other point sumbissions by email, or something.

The idea is to separate the source (user) from the method (form), the same way we separate them for harvests (source:arxiv,pos,wsp..., method:hepcrawl).

Or we can altogether remove one of the (method probably) and just use source.

kaplun: I think this should be moved to a broader discussion on `inspire-schema`, but I prefer not to block this PR.

PR #2111, opened 2017-03-22
Fixes to the workflows ui

david-caro:
* Now the naming of the buttons better reflects what they will do
  (though unfortunately, the names come from the workflows module).
* Show error stack only on error status.

### DIFF
--- a/inspirehep/modules/workflows/static/js/inspire_workflows_ui/templates/details.html
+++ b/inspirehep/modules/workflows/static/js/inspire_workflows_ui/templates/details.html
@@ -301,7 +301,7 @@
 
 
       <div class="col-md-12" id="errors"
-           ng-if="vm.record._extra_data._error_msg">
+         ng-if="vm.record._workflow.status === 'ERROR'">
         <holding-pen-template-handler
           template="/static/js/inspire_workflows_ui/templates/details_errors.html">
         </holding-pen-template-handler>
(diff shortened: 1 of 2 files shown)


inspirehep/modules/workflows/static/js/inspire_workflows_ui/templates/details.html
```diff
@@ -431,7 +431,7 @@
             <button class="btn btn-default"
                     ng-if="vm.record._workflow.status !== 'HALTED'"
                     ng-click="Utils.resumeWorkflow()">
-              <i class="fa fa-play" aria-hidden="true"></i> Resume
+              <i class="fa fa-play" aria-hidden="true"></i> Skip to next step
```
jmartinm: When does this happen? If record is not `HALTED` or `COMPLETED` the button will be shown, so if the workflow is ongoing and I press the button what is the effect?

And when is it useful to `Skip to next step` ?

david-caro: I don't think that this should be addressed in this pull request, but is a very good question.

About it showing, probably we should filter out also the 'running' and 'inital' states... though I've found that there's nowhere a complete statement of what those states are or should be, we (or me, whomever) should first figure out what states are there.

About it's usefulness, well, no idea what they had in mind when they put that button. Maybe if refexctract failed, and you don't care and just want to keep going or something. I don't find many other cases where it might be not just useful, but even safe.

PR #2124, opened 2017-03-23
docker: adapt to non-root user

david-caro:
* The venv dir must be created before mounting to avoid the docker
  daemon from creating it, as it does as root no matter what.


david-caro:
This requires (and is needed by...) a pr that will come to the inspire-docker repo

david-caro:
I will be merging this and the other non-root PRs today, as they are quite stable (being used locally by me, and hepcrawl).

PR #2134, opened 2017-03-24
global: bump inspire-schemas to version ~31.0

jacquerie:
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-nightly/group/619245


inspirehep/dojson/hep/rules/bd9xx.py
```diff
@@ -236,20 +236,22 @@ def references2marc(self, key, value):
     journal_artid = get_value(reference, 'publication_info.artid')
     pubnote = build_pubnote(journal_title, journal_volume, journal_pg_start,
                             journal_pg_end, journal_artid)
+    doi = get_value(reference, 'dois[0]')
+
     return {
         '0': get_recid_from_ref(value.get('record')),
         '1': get_value(reference, 'texkey'),
-        'a': get_value(reference, 'dois[0]'),
+        'a': 'doi:{}'.format(doi) if doi else doi,
```
michamos: This is wrong, you need to send back the other persistent identifiers also, see https://michamos.github.io/inspire-schemas/records/elements/reference.html#persistent-identifiers.

inspirehep/dojson/hep/rules/bd9xx.py
```diff
@@ -236,20 +236,22 @@ def references2marc(self, key, value):
...
               if a.get('role') == 'ed.'],
         'h': [a.get('full_name') for a in reference.get('authors', [])
               if a.get('role') != 'ed.'],
         'm': get_value(reference, 'misc'),
-        'o': get_value(reference, 'number'),
+        'o': get_value(reference, 'label'),
         'i': get_value(reference, 'publication_info.isbn'),
         'p': get_value(reference, 'imprint.publisher'),
-        'r': ['arXiv:' + el if el and is_arxiv(el) else el for el in repnos],
+        'r': get_value(reference, 'arxiv_eprint'),
```
michamos: This requires the addition of the `arXiv:` prefix, see https://michamos.github.io/inspire-schemas/records/elements/reference.html#arxiv-eprint.

tests/unit/dojson/test_dojson_hep_bd9xx.py
```diff
@@ -306,17 +352,15 @@ def test_references_from_999C5h_m_o_r_s_y_0():
...
             },
             'reference': {
-                'arxiv_eprints': [
-                    '1006.1263',
-                ],
+                'arxiv_eprint': '1006.1263',
                 'authors': [
-                    {'full_name': 'R. C. Myers'},
-                    {'full_name': 'A. Sinha'},
+                    {'full_name': 'Myers, R. C.'},
```
michamos: no space between initials

tests/unit/dojson/test_dojson_hep_bd9xx.py
```diff
@@ -335,20 +379,84 @@ def test_references_from_999C5h_m_o_r_s_y_0():
         {
             '0': 857215,
             'h': [
-                'R. C. Myers',
-                'A. Sinha',
+                'Myers, R. C.',
```
michamos: same

tests/unit/dojson/test_dojson_hep_bd9xx.py
```diff
@@ -335,20 +379,84 @@ def test_references_from_999C5h_m_o_r_s_y_0():
...
             ],
             'm': [
                 'Seeing a c-theorem with holography ; [hep-th]',
             ],
-            'o': 10,
-            'r': [
-                'arXiv:1006.1263',
-            ],
+            'o': '10',
+            'r': '1006.1263',
```
michamos: missing `arXiv:` prefix

tests/unit/dojson/test_dojson_hep_bd9xx.py
```diff
@@ -335,20 +379,84 @@ def test_references_from_999C5h_m_o_r_s_y_0():
...
+        {
+            'curated_relation': False,
+            'record': {
+                '$ref': 'http://localhost:5000/api/literature/780399',
+            },
+            'raw_refs': [
+                {
+                    'value': 'Int. J. Mod. Phys. A 23 (doi:10.1142/S0217751X0804055X)',
+                    'schema': 'text',
+                    'source': 'dojson'
```
michamos: no source here. `dojson` is **not** where we obtained this information from.

tests/unit/references/test_references_processors.py
```diff
@@ -0,0 +1,643 @@
...
+def test_split_refextract_handles_unicode():
+    expected = [
+        u'J. Kätlne'
+    ]
+    result = _split_refextract_authors_str('J. Kätlne et al.')
+
+    assert expected == result
+
+
+def test_is_arxiv():
```
michamos: no test?

tests/unit/references/test_references_processors.py
```diff
@@ -0,0 +1,643 @@
...
+    pass
+
+
+def test_normalize_arxiv():
```
michamos: same question

tests/unit/references/test_references_processors.py
```diff
@@ -0,0 +1,643 @@
...
+    builder = ReferenceBuilder()
+
+    builder.add_raw_reference('Phys. Rev. C 80 (doi:10.1103/PhysRevC.80.044313)')
+
+    expected = [
+        {
+            'raw_refs': [
+                {
+                    'schema': 'text',
+                    'source': 'reference_builder',
```
michamos: again this is not a source. I think `add_raw_reference` should take a `source` optional argument (or even `__init__`), like for the [`LiteratureBuilder`](https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/builders.py#L89)

tests/unit/references/test_references_processors.py
```diff
@@ -0,0 +1,643 @@
...
+
+    builder = ReferenceBuilder()
+
+    builder.add_report_number('CERN-EP-2017-047')
+
+    expected = [
+        {
+            'reference': {
+                'publication_info': {
+                    'parent_report_number': 'CERN-EP-2017-047',
```
michamos: why not [`report_number`](https://michamos.github.io/inspire-schemas/records/elements/reference.html#report-number)?

tests/unit/references/test_references_processors.py
```diff
@@ -0,0 +1,643 @@
...
+
+    builder.add_uid('hdl:10443/1646')
+
+    expected = [
+        {
+            'reference': {
+                'persistent_identifiers': [
+                    {
+                        'schema': 'HDL',
+                        'value': 'hdl:10443/1646',
```
michamos: no `hdl:` here, see https://michamos.github.io/inspire-schemas/records/hep.html#hep-json-properties-persistent-identifiers-items-properties-value

tests/unit/references/test_references_processors.py
```diff
@@ -0,0 +1,643 @@
...
+            },
+        },
+    ]
+    result = [builder.obj]
+
+    assert validate(result, subschema) is None
+    assert expected == result
+
+
+def test_add_uid_rejects_invalid_isbns():
```
michamos: what happens with a slightly less invalid ISBN, e.g. `123456789`?

inspirehep/dojson/hep/rules/bd9xx.py
ammirate: which is the criterion for returning `hdl:` or `urn:`?

michamos: if `schema` is `HDL`, you prepend `hdl:` to the `value`. If schema is `URN`, you send back the `value` (that already contains `urn:`).

inspirehep/dojson/hep/rules/bd9xx.py
michamos: BTW, also `report_number` should be put into `r`. If you search for `999C5r` in https://michamos.github.io/inspire-schemas/records/elements/reference.html#reference-json-properties-title, you will find all the fields that should end up here. You can do the same for all other fields to be sure you don't miss anything.

michamos:
Could you also take into account the non-breaking changes in 31.0.*?
From https://inspire-schemas.readthedocs.io/en/latest/_static/CHANGELOG.txt, those are:

*   MINOR 325e57cd: elements/document_type: resurrect activity report
*   MINOR cfd42dd3: hep: add year to copyright field

inspirehep/modules/references/processors.py
```diff
@@ -228,19 +228,30 @@ def add_uid(self, uid):
...
             uid_object = {}
             uid_object['schema'] = 'HDL'
             uid_object['value'] = value
             self.obj['reference']['persistent_identifiers'].append(uid_object)
+
+        elif uid.startswith('urn:'):
+            self._ensure_reference_field('persistent_identifiers', [])
+            uid_object = {}
+            uid_object['schema'] = 'URN'
+            uid_object['value'] = uid
```
michamos: you can write more concisely:
```python
uid_object = {
    'schema': 'URN',
    'value': uid
}
```

tests/unit/dojson/test_dojson_hep_bd9xx.py
```diff
@@ -403,6 +403,8 @@ def test_references_from_999C5a_h_o_s_x_y_0():
     snippet = (
         '<datafield tag="999" ind1="C" ind2="5">'
         '  <subfield code="a">doi:10.1142/S0217751X0804055X</subfield>'
+        '  <subfield code="a">hdl:10138/168995</subfield>'
```
michamos: please don't modify existing snippets, they represent pieces of existing records (as referenced in the comment following the snippet). If you need another example, please add a new test case with a real record snippet.

inspirehep/dojson/hep/rules/bd9xx.py
```diff
@@ -236,20 +236,34 @@ def references2marc(self, key, value):
     journal_artid = get_value(reference, 'publication_info.artid')
     pubnote = build_pubnote(journal_title, journal_volume, journal_pg_start,
                             journal_pg_end, journal_artid)
+
+    doi = get_value(reference, 'dois[0]')
+    a_list = ['doi:{}'.format(doi) if doi else doi]
```
michamos: instead of these two lines, you can do this to send back all dois:
`a_list = ['doi:{}'.format(doi) for doi in get_value(reference, 'dois', [])]`

inspirehep/modules/references/processors.py
```diff
@@ -180,9 +180,9 @@ def add_refextract_authors_str(self, authors_str):
         migration.
         """
         for author in _split_refextract_authors_str(authors_str):
```
ammirate: well, the function does what it says. But, since it is the only place where is used, I am going to refactor it and put the 'cleaning' code in it.

inspirehep/modules/references/processors.py
```diff
@@ -156,7 +156,6 @@ def add_raw_reference(self, raw_reference, source='reference_builder',
         self._ensure_field('raw_refs', [])
         self.obj['raw_refs'].append({
             'value': raw_reference,
-            'source': source,
```
ammirate: `source` has been replaced by `schema` in the new version, isn't it?

michamos: no, those are two different things. `source` is where the raw reference came from, `schema` in what format it is encoded.

inspirehep/dojson/hep/rules/bd9xx.py
```diff
@@ -201,7 +201,8 @@ def get_value(value):
             ('c', rb.add_collaboration),
             ('q', rb.add_title),
             ('t', rb.add_title),
-            ('h', rb.add_refextract_authors_str),
+            ('h', rb
+             .add_refextract_authors_str),
```
michamos: typo?

inspirehep/dojson/hep/rules/bd9xx.py
```diff
@@ -247,8 +248,13 @@ def references2marc(self, key, value):
...
                       if pis_schema is 'HDL' else pis_value)
 
-    a_list = filter(lambda x: x is not None, a_list)
     arxiv = get_value(reference, 'arxiv_eprint')
+    arxiv = 'arXiv:{}'.format(arxiv) if arxiv else None
+    report_n = get_value(reference, 'report_number')
+    r_list = [arxiv, report_n]
+
+    a_list = filter(lambda x: x is not None, a_list)
+    r_list = filter(lambda x: x is not None, r_list)
```
michamos: As already mentioned, this filtering for empty values is done by doJSON, no need to do it yourself.

tests/unit/dojson/test_dojson_hep_bd9xx.py
```diff
@@ -557,3 +559,67 @@ def test_references_from_999C50_h_m_o_r_y():
     result = hep2marc.do(result)
 
     assert expected == result['999C5']
+
+
+@pytest.mark.xfail
```
michamos: please provide a `reason` to the `xfail` (like `xfail(reason=Multiple report numbers are not handled)`)

inspirehep/modules/references/processors.py
```diff
@@ -90,11 +90,10 @@ def _is_arxiv(obj):
     """Small checker for arXiv report numbers."""
     # Workaround until this issue:
     # https://github.com/inveniosoftware/idutils/issues/14 is solved.
```
michamos: does this still apply?

inspirehep/modules/references/processors.py
```diff
@@ -90,11 +90,10 @@ def _is_arxiv(obj):
...
-
     arxiv_test = obj.split()
     if not arxiv_test:
         return False
-    return idutils.is_arxiv(arxiv_test[0])
+    return idutils.is_arxiv(arxiv_test[0]) or False
```
michamos: `x or False == x`, if `x` is a `bool`, what is the point of this?

tests/unit/references/test_references_processors.py
```diff
@@ -63,12 +65,26 @@ def test_split_refextract_handles_unicode():
     assert expected == result
 
 
-def test_is_arxiv():
-    pass
+def test_is_arxiv_new_identifier():
+    result = _is_arxiv('arXiv:1501.00001v1')
```
michamos: so `_is_arxiv`, confusingly, does not return a `bool`? maybe the name of the function should be changed then.

inspirehep/modules/references/processors.py
ammirate: because the match can return None. This means that the function returns a Match object if there is a match, None if there is not, and False if the split fails. 
Since the signature is `_is_arxiv` I would expect False if it is not an arxiv value, else the matched value.

michamos: :+1: I would say this is a bug of `idutils` then. Maybe you could open an issue/PR there?
Anyway, it is probably more clear to wrap return value in `bool(...)` instead of adding `or False` (with a comment to explain the reason, as it is not obvious).

tests/unit/dojson/test_dojson_hep_bd9xx.py
```diff
@@ -678,3 +678,136 @@ def test_references_from_999C50_9_r_u_h_m_o():
...
+
+    ]
+    result = hep.do(create_record(snippet))
+
+    assert validate(result['references'], subschema) is None
+    assert expected == result['references']
+
+    expected = [
+        {
+            "h": [
```
michamos: `e`, not `h`, as it is an editor.

inspirehep/modules/references/processors.py
```diff
@@ -151,7 +151,7 @@ def set_texkey(self, texkey):
 
     def add_title(self, title):
         self._ensure_reference_field('title', {})
-        self.obj['reference']['title'] = {'title': title}
+        self.obj['reference']['title'] = {'title': unicode(title)}
```
michamos: is this cast needed?

tests/unit/dojson/test_dojson_hep_bd9xx.py
```diff
@@ -736,3 +736,78 @@ def test_reference_from_999C5t_p_y_e_o():
...
+        }
+
+    ]
+
+    result = hep.do(create_record(snippet))
+
+    assert validate(result['references'], subschema) is None
+    assert expected == result['references']
+
+    expected = [
```
michamos: `c` disappeared.

inspirehep/dojson/hep/rules/bd9xx.py
```diff
@@ -236,21 +235,37 @@ def references2marc(self, key, value):
...
     pubnote = build_pubnote(journal_title, journal_volume, journal_pg_start,
                             journal_pg_end, journal_artid)
+
+    a_list = ['doi:{}'.format(doi)
+              for doi in get_value(reference, 'dois', [])]
+
+    persistent_identifiers = get_value(reference, 'persistent_identifiers',
+                                       default=[])
+    for identifier in persistent_identifiers:
+        pid = get_value(identifier, 'value')
```
jacquerie: If you do
```python
foos = get_value(record, 'foo')
for foo in foos:
    bar = get_value(record, 'bar')
    if some_condition_on(bar):
        ...
```
you're not getting any of the benefits of `get_value`, just the fact that is ~7x slower than straight dictionary access.

In this case, where you have to filter on the `bar`, it's better to use straight dictionary accesses:
```python
for foo in force_list(record.get('foo')):
    if some_condition_on(foo.get('bar')):
        ...
```

inspirehep/modules/references/processors.py
jacquerie: I wouldn't say this is a bug in `idutils`; the expected pythonic use
```python
if is_arxiv(foo):
  ...
```
is indifferent to having `False` or `None`.

👍  instead for wrapping in `bool`.

tests/unit/dojson/test_dojson_hep_bd9xx.py
jacquerie: `c` reappeared.

inspirehep/modules/references/processors.py
jacquerie: actually, :-1: for wrapping in `bool`. Python is duck typed, `False` or `None` are completely equivalent in a boolean context.

inspirehep/modules/references/processors.py
```diff
@@ -81,16 +81,26 @@ def _split_refextract_authors_str(authors_str):
...
+    res = [_improve_author(r) for r in res if all(not f(r) for f in filters)]
 
     return res
 
 
+def _improve_author(author_names):
+    res = author_names.split(' ')[-1] + ', '
+    for name in author_names.split(' ')[0:-1]:
+        res += name
+    return res
```
jacquerie: This logic is obviously too naive to really handle names. Proper treating of this problem is deferred to the time in which we seriously integrate `nameparser` as a util in INSPIRE.

jacquerie: For example: `A. Del Guerra` in https://inspirehep.net/record/1593684/export/xme.

jacquerie:
> Could you also take into account the non-breaking changes in 31.0.*?
> From https://inspire-schemas.readthedocs.io/en/latest/_static/CHANGELOG.txt, those are:
>
> MINOR 325e57cd: elements/document_type: resurrect activity report
> MINOR cfd42dd3: hep: add year to copyright field

Done, and also fixed https://sentry.cern.ch/inspire-sentry/inspire-nightly/group/619245 in the meantime.

Issue #2037, opened 2017-03-15
global: bump celery to version 4


kaplun:
As per the title.

http://docs.celeryproject.org/en/latest/faq.html#should-i-use-retry-or-acks-late
http://docs.celeryproject.org/en/latest/userguide/configuration.html?highlight=ack_late#task-reject-on-worker-lost

lnielsen:
See https://github.com/inveniosoftware/flask-celeryext/issues/22 ;-)

david-caro:
@kaplun if you are not working on this, can you move it to ready

jacquerie:
This is done in https://github.com/inspirehep/inspire-next/pull/2138, which wasn't merged as we ignore the impact it would have on the production system.

StellaCh:
Then, is it ok to close this @jacquerie ?

jacquerie:
No, we still want to do this!

PR #2138, opened 2017-03-25
global: bump celery to version 4

jacquerie:
Closes https://github.com/inspirehep/inspire-next/issues/2037.

@david-caro, I'm not going to merge this because this requires some testing on QA to ensure that this change is correct. In particular, `librabbitmq==1.6.1` needs to be uninstalled from the machines, because it's incompatible with `celery==4.0.2` (https://github.com/celery/celery/issues/3675, https://github.com/celery/librabbitmq/issues/93), and I'm fairly dissatisfied that this means that we're forced to use `pyamqp`, which ([as you wrote!](https://github.com/celery/librabbitmq/issues/84#issuecomment-237465841)) seems to be always slower than `librabbitmq`.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -22,7 +22,7 @@
 
 web: gunicorn inspirehep.wsgi -c gunicorn.cfg
 cache: redis-server
-worker: celery worker -E -A inspirehep.celery --loglevel=INFO --workdir="${VIRTUAL_ENV}" --autoreload --pidfile="${VIRTUAL_ENV}/worker.pid" --purge
+worker: celery worker -E -A inspirehep.celery --loglevel=INFO --workdir="${VIRTUAL_ENV}" --pidfile="${VIRTUAL_ENV}/worker.pid" --purge
 workermon: celery flower -A inspirehep.celery
 # mathoid: node_modules/mathoid/server.js -c mathoid.config.yaml
 indexer: elasticsearch -Dcluster.name="inspire" -Ddiscovery.zen.ping.multicast.enabled=false -Dpath.data="$VIRTUAL_ENV/var/data/elasticsearch"  -Dpath.logs="$VIRTUAL_ENV/var/log/elasticsearch"
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -30,8 +30,8 @@ services:
       service: base
     environment:
       - APP_SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://inspirehep:dbpass123@test-database:5432/inspirehep
-      - APP_BROKER_URL=amqp://guest:guest@test-rabbitmq:5672//
-      - APP_CELERY_RESULT_BACKEND=amqp://guest:guest@test-rabbitmq:5672//
+      - APP_CELERY_BROKER_URL=pyamqp://guest:guest@test-rabbitmq:5672//
+      - APP_CELERY_RESULT_BACKEND=rpc://guest:guest@test-rabbitmq:5672//
       - APP_CACHE_REDIS_URL=redis://test-redis:6379/0
       - APP_ACCOUNTS_SESSION_REDIS_URL=redis://test-redis:6379/2
       - APP_SEARCH_ELASTIC_HOSTS=test-indexer
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -69,11 +69,11 @@ SQLALCHEMY_TRACK_MODIFICATIONS = True
 
 # Celery
 # ======
-BROKER_URL = "amqp://guest:guest@localhost:5672//"
-CELERY_RESULT_BACKEND = "amqp://guest:guest@localhost:5672//"
+CELERY_BROKER_URL = "pyamqp://guest:guest@localhost:5672//"
+CELERY_RESULT_BACKEND = "rpc://guest:guest@localhost:5672//"
 CELERY_ACCEPT_CONTENT = ['json', 'msgpack', 'yaml']
 CELERY_TIMEZONE = 'Europe/Amsterdam'
-CELERY_DISABLE_RATE_LIMITS = True
+CELERY_WORKER_DISABLE_RATE_LIMITS = True
 
 # Cache
 # =====
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,8 +30,8 @@ services:
       service: base
     environment:
       - APP_SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://inspirehep:dbpass123@database:5432/inspirehep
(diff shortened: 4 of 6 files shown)


Procfile
```diff
@@ -22,7 +22,7 @@
 
 web: gunicorn inspirehep.wsgi -c gunicorn.cfg
 cache: redis-server
-worker: celery worker -E -A inspirehep.celery --loglevel=INFO --workdir="${VIRTUAL_ENV}" --autoreload --pidfile="${VIRTUAL_ENV}/worker.pid" --purge
+worker: celery worker -E -A inspirehep.celery --loglevel=INFO --workdir="${VIRTUAL_ENV}" --pidfile="${VIRTUAL_ENV}/worker.pid" --purge
```
jacquerie: `--autoreload` has been completely removed from Celery: http://docs.celeryproject.org/en/latest/whatsnew-4.0.html#features-removed-for-lack-of-funding

docker-compose.test.yml
```diff
@@ -30,8 +30,8 @@ services:
       service: base
     environment:
       - APP_SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://inspirehep:dbpass123@test-database:5432/inspirehep
-      - APP_BROKER_URL=amqp://guest:guest@test-rabbitmq:5672//
-      - APP_CELERY_RESULT_BACKEND=amqp://guest:guest@test-rabbitmq:5672//
+      - APP_CELERY_BROKER_URL=pyamqp://guest:guest@test-rabbitmq:5672//
+      - APP_CELERY_RESULT_BACKEND=rpc://guest:guest@test-rabbitmq:5672//
```
jacquerie: The `amqp` result backend has been deprecated: http://docs.celeryproject.org/en/latest/whatsnew-4.0.html#features-removed-for-lack-of-funding

inspirehep/config.py
```diff
@@ -69,11 +69,11 @@ def _(x):
...
 # Celery
 # ======
-BROKER_URL = "amqp://guest:guest@localhost:5672//"
-CELERY_RESULT_BACKEND = "amqp://guest:guest@localhost:5672//"
+CELERY_BROKER_URL = "pyamqp://guest:guest@localhost:5672//"
+CELERY_RESULT_BACKEND = "rpc://guest:guest@localhost:5672//"
 CELERY_ACCEPT_CONTENT = ['json', 'msgpack', 'yaml']
 CELERY_TIMEZONE = 'Europe/Amsterdam'
-CELERY_DISABLE_RATE_LIMITS = True
+CELERY_WORKER_DISABLE_RATE_LIMITS = True
```
jacquerie: Do we really need this? I'm going to benchmark it to see if it really improves performance, as was claimed in https://github.com/inspirehep/inspire-next/commit/95f1cfe03e6c4af6b52bab8f7e8a8d65148cee33.

inspirehep/config.py
```diff
@@ -69,11 +69,11 @@ def _(x):
 
 # Celery
 # ======
-BROKER_URL = "amqp://guest:guest@localhost:5672//"
-CELERY_RESULT_BACKEND = "amqp://guest:guest@localhost:5672//"
+CELERY_BROKER_URL = "pyamqp://guest:guest@localhost:5672//"
+CELERY_RESULT_BACKEND = "rpc://guest:guest@localhost:5672//"
 CELERY_ACCEPT_CONTENT = ['json', 'msgpack', 'yaml']
```
jacquerie: Do we really need this? I don't see why we can't just use one type of messages (of the three, `msgpack` should be the one that guarantees the smallest messages).

kaplun: AFAIK `msgpack` has some limitation on the format supported WRT `json`, and this was hindering the workflow engine.

inspirehep/config.py
kaplun: For reference on the original reason: http://docs.celeryproject.org/en/2.0-archived/userguide/tasks.html#disable-rate-limits-if-they-re-not-used

jacquerie: Just did a quick benchmark and saw that `inspirehep migrator populate -f inspirehep/demosite/data/demo-records.xml.gz` takes 5 seconds less without this line. I will try again on Nightly, which has bigger chunks.

jacquerie:
As discussed today with @david-caro, I also added a commit to use the released `Flask-CeleryExt`.

kaplun:
Should we also amend decorator's flags?

inspirehep/config.py
kaplun: But, why not moving to `msgpack` by default and see if we pass? That would give quite a boost in performance, for small tasks.

david-caro:
Currently testing on qa, cherry-picked on top of latest master

jacquerie:
> Should we also amend decorator's flags?

I'd defer the tuning of the configuration and the improvements of performance to another PR, so that we don't mistake a problem in the configuration for a problem generated by bumping Celery to version 4.

kaplun:
I meant whether flags already specified needs to be amended due to them having be renamed in Celery 4. (not sure if this has happened at all, but I seem to recall e.b. `ack_late` vs `task_ack_late`...)

jacquerie:
> I meant whether flags already specified needs to be amended due to them having be renamed in Celery 4.

Ah, good point. Here's what we have currently:
```
% ag @shared_task
inspirehep/modules/disambiguation/tasks.py
52:@shared_task(ignore_result=True)
86:@shared_task(ignore_result=True)
133:@shared_task

inspirehep/modules/migrator/tasks/access.py
31:@shared_task()

inspirehep/modules/migrator/tasks/records.py
105:@shared_task(ignore_result=True)
119:@shared_task(ignore_result=True)
149:@shared_task(ignore_result=True)
180:@shared_task(ignore_result=False, compress='zlib', acks_late=True)
203:@shared_task()

inspirehep/modules/migrator/tasks/workflows.py
99:@shared_task()
166:@shared_task()

inspirehep/modules/records/tasks.py
43:@shared_task(ignore_result=True)
```

kaplun:
Actually after a brief check in the documentation seems the arguments haven't been renamed. Just there is a `task_` prefix added when specifying global defaults in the config. So LGTM!

jacquerie:
@bavaria95 points out that we might have some performance issues: https://github.com/celery/celery/issues/3814.

kaplun:
Interesting. Also in the comparison the were using `py-amqp` in both 3 and 4, while we use `librabbitmq` on 3. So we might suffer even more...

jacquerie:
About `librabbitmq` there's an interesting comment: https://github.com/celery/celery/issues/3675#issuecomment-294129297.

jacquerie:
Stalled, will be resurrected when needed.

PR #2139, opened 2017-03-26
travis: use codecov instead of coveralls

jacquerie:


jacquerie:
Merging of the coverage reports has started working again on Coveralls (https://coveralls.io/builds/10787309), so this is no longer needed. I will also remove the hook on the repository.

Issue #2106, opened 2017-03-22
article workflow: derive inspire_categories from arxiv categories


jmartinm:
In the article workflow (user submissions, harvest,...) when there are arXiv categories, the equivalent inspire categories should be added.

We can add a new task to the [article](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/workflows/article.py#L175) workflow for this.

PR #2143, opened 2017-03-28
Add derive categories from arxiv procedure in the workflow

rikirenz:
* Adds the procedure to convert from arxiv
categories to inspire categories


rikirenz:
Closes #2106

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -143,6 +150,25 @@ def arxiv_refextract(obj, eng):
...
+    obj.data.setdefault('inspire_categories', [])
+
+    for arxiv_category in get_value(obj.data['arxiv_eprints'][0], 'categories'):
+        inspire_category = classify_field(arxiv_category)
+        if inspire_category:
+            category = {
+                'source': 'arxiv',
+                'term': inspire_category,
+            }
+            if category not in obj.data['inspire_categories']:
```
rikirenz: @michamos do you think this control is fine or is useless?

rikirenz:
@michamos Now this patch is allowing this kind of behavior:

screenshot from 2017-03-28 15-21-58

Do you think is fine?

inspirehep/modules/workflows/tasks/arxiv.py
michamos: for now it's not needed, but in the future, we will handle updates where inspire categories will already have been assigned. So probably safer to leave it.

rikirenz:
@kaplun @annetteholtkamp  this is what you were discussing. Do you think is fine?

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -0,0 +1,88 @@
...
+    schema = load_schema('hep')
+    subschema = schema['properties']['inspire_categories']
+
+    obj = StubObj(
+        {
+            'arxiv_eprints': [
+                {
+                    'categories': [
+                        'alg-geom', # with a corresponding inspire_category
+                        'q-bio', # without a corresponding inspire_category
```
michamos: that there is no 'q-bio' in the category mapping is a bug not a feature. All categories should be mapped.

rikirenz: Make sense! I have done in this way because 1 can be a bug but 3...
We have 3 categories "mapped" in this way:

| arxiv_categ | inspire_categ |
| ------------- | ------------- |
| q-bio  | None  |
| q-fin  | None  |
| stat | None |

michamos: In all three cases, the top-level category was omitted, and only the subcategories were put in the mapping. As you can see from the subcategories, all should be `Other`.

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -143,6 +150,26 @@ def arxiv_refextract(obj, eng):
...
+def arxiv_derive_inspire_categories(obj, eng):
+    """Derive inspire-categories from arxiv-categories
+
+    :param obj: Workflow Object to process
+    :param eng: Workflow Engine processing the object
+    """
+    obj.data.setdefault('inspire_categories', [])
+
+    # for arxiv_record in obj.data['arxiv_eprints']:
+    for arxiv_category in get_value(obj.data['arxiv_eprints'][0], 'categories'):
```
jacquerie: `get_value(obj.data, 'arxiv_eprints[0].categories')` ?

inspirehep/modules/workflows/tasks/arxiv.py
jacquerie: Well, can you write a failing test without this line? Yes, you can, because `inspire_categories` has `uniqueItems: true`. So you should write a test for this.

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -0,0 +1,88 @@
...
+                        'physics.gen-ph', # with a corresponding inspire_category but inspire_category is already present
```
jacquerie: You could split these cases in several tests, so that you don't have to write inline comments to explain each of them.

inspirehep/modules/workflows/tasks/arxiv.py
rikirenz: the test is already included. I am going to split them.

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -0,0 +1,225 @@
...
+    def __init__(self, data, extra_data):
+        self.data = data
+        self.extra_data = extra_data
+
+
+class DummyEng(object):
+    pass
+
+
+def test_arxiv_derive_inspire_categories():
```
rikirenz: @jacquerie do you think we should use `pytest parametrize`?

IMHO I prefer this way. But with parametrize we can write more compact tests.

jacquerie:
This change looks innocent, but modifying workflow objects is something that we don't support yet, so I recommend **not** merging this unless we have a way to deal with that on production.

rikirenz:
@jacquerie is right! It is not an easy problem to solve...  We should investigate

kaplun:
As discussed today at the technical meeting we, amending workflows is not going to be so critical as it looks:
* Literature workflows that are waiting for decision can simply be restarted from scratch (we need to make this easy though), following a code upgrade involving a workflow
* For author related workflows we could simply ask catalogers to empty the queue (and disable celery for the time being), so that there are no pending workflows
* Only issue is for user-suggestions, where restarting a workflow might trigger sending twice a ticket or email. (For that the above restart-workflow procedure could set a flag in the `obj` that can be checked by interested steps to know if we are in a situation of restart of workflow...)

kaplun:
So all in all we can still happily amend workflows for the time being.

kaplun:
Ticketized in #2153

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -143,6 +150,26 @@ def arxiv_refextract(obj, eng):
...
 
+def arxiv_derive_inspire_categories(obj, eng):
+    """Derive inspire-categories from arxiv-categories
+
+    :param obj: Workflow Object to process
+    :param eng: Workflow Engine processing the object
+    """
+    obj.data.setdefault('inspire_categories', [])
+
+    for arxiv_record in obj.data['arxiv_eprints']:
```
jmartinm: arxiv_record -> arxiv_eprint

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -0,0 +1,225 @@
...
+import pytest
+
+from inspire_schemas.utils import load_schema
+from inspirehep.dojson.utils import validate
+from inspirehep.modules.workflows.tasks.arxiv import (
+    arxiv_derive_inspire_categories
+)
+
+
+class StubObj(object):
```
jmartinm: We have this in ~6 different files. Maybe its time to add this `StubObj` into `tests/unit/helpers.py`

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -0,0 +1,225 @@
...
+    def __init__(self, data, extra_data):
+        self.data = data
+        self.extra_data = extra_data
+
+
+class DummyEng(object):
```
jmartinm: Same as above.

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -0,0 +1,225 @@
...
+        }, {
+            'source': 'user',
+            'term': 'Math and Math Physics',
+        }, {
+            'source': 'arxiv',
+            'term': 'Math and Math Physics',
+        }
+    ]
+
+    assert arxiv_derive_inspire_categories(obj, eng) is None
```
jmartinm: What is this assertion for?

rikirenz:
update: @david-caro sad that restart the workflow should be enough.

tests/unit/workflows/test_workflows_tasks_arxiv.py
rikirenz: To check that the function is not going to return anything

tests/unit/workflows/test_workflows_tasks_arxiv.py
rikirenz: @jacquerie @david-caro @kaplun I will try to make what @jmartinm is saying for this PR. Anyway it would be nice have a clear policy about how to reuse code in the tests! :smile:

kaplun:
@rikirenz I would like to merge this, but it currently conflict heavily. Could you try to rebase this PR and resubmit it?

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -143,6 +150,26 @@ def arxiv_refextract(obj, eng):
...
+def arxiv_derive_inspire_categories(obj, eng):
+    """Derive inspire-categories from arxiv-categories
+
+    :param obj: Workflow Object to process
+    :param eng: Workflow Engine processing the object
+    """
+    obj.data.setdefault('inspire_categories', [])
+
+    for arxiv_eprint in obj.data['arxiv_eprints']:
+        for arxiv_category in get_value(arxiv_eprint, 'categories'):
```
jacquerie:
```python
from itertools import chain

[...]

for arxiv_category in chain.from_iterable(get_value(obj.data, 'arxiv_eprints.categories')):
    [...]
```

That said, this is already used elsewhere, so it should become an helper in `inspirehep/utils/record.py`.

jacquerie:
Remove 134cafd027dfc9a3c746f97b4cb78c2a39ece424, it's not the right way to refactor those things.

PR #2144, opened 2017-03-28
holdingpen: ignore experiments prediction mapping

david-caro:
* This allows puttnig there anything, it's a temporary workaround
  until we formalize what should be the data structure there.

### DIFF
--- a/inspirehep/modules/workflows/mappings/holdingpen/hep.json
+++ b/inspirehep/modules/workflows/mappings/holdingpen/hep.json
@@ -11,6 +11,10 @@
                     "properties": {
                         "is-update": {
                             "type": "boolean"
+                        },
+                        "experiments_prediction": {
+                            "dynamic": "false",
+                            "type": "object"
                         }
                     },
                     "type": "object"


david-caro:
@kaplun @jacquerie this allows me to remap current workflows without failures, but we will need a way to upgrade existing workflows model... so we can start mapping by this too (or well, not sure we want to map this data anyhow).

david-caro:
btw. for new records we are not populating it as we are skipping that step.

kaplun:
Why are we skipping it? It was an interesting addition, exploiting magpie, to be used for experimental papers.

david-caro:
Because the format was not ok, it must be changed from the list of lists to the object-based one. I can write that too if you want.

kaplun:
Ah OK. It's just temporary removed. I though it has been decided to scrape it :+1: 
Yeah if it's quick, it would be great.

david-caro:
See #2145, that does not fix though the migration from the old format to the new....

jacquerie:
Closed in favor of #2145 (unless your testing on QA reveals that we have in fact records which would become incompatible on the new mapping because of this).

Issue #1773, opened 2016-11-23
tests: unit tests for workflows/tasks


jacquerie:
Just spotted that there's an opportunity to gain ~2% coverage by adding **unit tests** to the workflows tasks. Moreover, I already found [a bug](https://travis-ci.org/inspirehep/inspire-next/jobs/176430251#L1520) and some code/tests that could be refactored.

- [x] Test `workflows/tasks/actions.py` (#1748)
- [x] Test `workflows/tasks/arxiv.py` (#2307)
- [x] Test `workflows/tasks/beard.py` (#1311)
- [ ] Test `workflows/tasks/classifier.py`
- [x] Test `workflows/tasks/magpie.py` (#1309)
- [ ] Test `workflows/tasks/matching.py`
- [x] Test `workflows/tasks/refextract.py` (several PRs)
- [x] Test `workflows/tasks/submission.py` ~ (#2313)
- [x] Test `workflows/tasks/upload.py` (#2315)

jacquerie:
Essentially done, will be completed on case-by-case basis.

PR #2145, opened 2017-03-28
workflows: format experiment predictions as objects

david-caro:
* This allows to properly map them on elasticsearch.


inspirehep/modules/workflows/tasks/magpie.py
```diff
@@ -120,17 +120,23 @@ def guess_experiments(obj, eng):
...
+            limit=0.5,
+        )
+        selected_experiments = [
+            {'label': e[0], 'score': e[1]}
+            for e in selected_experiments
+        ]
         obj.extra_data["experiments_prediction"] = dict(
-            labels=labels,
-            experiments=experiments
+            experiments=selected_experiments,
```
david-caro: I've copied the `categories_prediction` for the format. I think this is not used anywhere though.

kaplun:
Github is now suggesting reviewers (based on the original code developer)! He's popping up Jan :dancer:

david-caro:
This requires though a nice way of reindexing workflow objects (or at least a non-failing way...) as this is a non-backwards compatible change.

kaplun:
We could:
* bring down the site, (fab)
* drop HP indexes, (CLI exists)
* reinitialize HP indexes, (CLI exists)
* bring up the site (fab)
* repopulate HP indexes (CLI exists).

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -195,10 +195,10 @@
     guess_categories,
     # TODO: adapt the output of guess_experiment so that it
     # can be stored in ElasticSearch (see issue #2054).
```
jacquerie: This comment can be deleted.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -338,8 +338,13 @@ def get_halted_workflow(app, record, extra_config=None):
 
     # TODO: add the experiments predictions to the workflow
     # object (see issue #2054).
```
jacquerie: This one as well.

david-caro:
> This is missing a test to ensure that we're not regressing on this bug in the future. We can use the same record that triggered it originally, if we can find it... otherwise any experimental paper will do.

I don't really understand what you want to test, the issue here was a missing feature, so all the papers were actually affected (there's no trigger). There's already an integration test that does the check for the experiment predictions, isn't that what you ask for?

jacquerie:
> There's already an integration test that does the check for the experiment predictions, isn't that what you ask for?

Clearly it wasn't enough, because it didn't catch this mistake before we deployed in front of the users. So there **must** be a useful test to write here: something was failing before, now it's no longer failing...

david-caro:
It did not catch anything because it was looking for the wrong thing (that at the time, was correct), that's why I had to change it for this pr too. There was no bug to catch either, it's a change in the format of the data that we put on ES (before it was a list of tuples of string-float, now it's a list of objects).

Just to emphasize again, there was no bug or mistake, it's a change of functionality, specifically, in the format of the prediction results.

jacquerie:
I wanted to merge this in https://github.com/inspirehep/inspire-next/pull/2277, but it looks like it doesn't work after rebasing. Can you have a look?

david-caro:
will check

david-caro:
@jacquerie btw. you know you can reuse the pr without having to open a new one right? (we have the `Allow edits from maintainers.` enabled by default)

jacquerie:
> @jacquerie btw. you know you can reuse the pr without having to open a new one right?

I know! I actually do it on purpose, so that I'm accountable for any problems I might introduce during review. If I'm happy with the code, and I just have to do cosmetic fixes (if any), I reuse the same PR.

jacquerie:
Victim of #2222.

david-caro:
Yep, locally it worked... But on the bright side, I have a working patch to do the login only once per session (was pissed by seeing the login page all the time), will push later or tomorrow morning.


--

CERN - RCS-SIS
http://inspirehep.net
On May 11, 2017 9:30 PM, Jacopo Notarstefano <notifications@github.com> wrote:

Victim of #2222<https://github.com/inspirehep/inspire-next/issues/2222>.

tests/integration/workflows/fixtures/oai_arxiv_record_with_plots.xml
```diff
@@ -24,7 +24,7 @@
             </author>
           </authors>
           <title>Spectra generated by a confined softcore Coulomb potential</title>
-          <categories>math-ph math.MP</categories>
+          <categories>math-ph math.MP astro-ph astro-ph.IM</categories>
```
jacquerie: This is what was missing to make it work, right? As I remarked elsewhere, this is not ideal, because it's "cheating" by faking a record that doesn't exists, inducing false security in our system. In fact, my plan is to complete #1773 and later tackle this test suite, so that we start trusting it more.

david-caro: I don't think that using a specially crafted record is 'cheating', actually, I think that it's the way to go, specially for acceptance/functional tests, where adding more records and test runs is a lot more costly. I don't see how this adds false security either, the test is really testing a specification, not a tautology.

I would agree though that it might be more elegant to move the set of 'experimental categories' to a configuration value, and tweak that on the test to make the record experimental instead.

jacquerie: The thing is, in the real world we don't have the luxury of crafting the records, because we have to work with whatever the users throw at us, even if invalid or wrong. If we are in the mindset that we're allowed to deal with the *nice* case, we're not going to be prepared for the ugliness that is going to hit us later.

david-caro: That is true, but it's not taking into account:

* That it does not apply to this PR, specially to this record, as it's a 'smoke' record for a smoke test, that runs just one flow with one simple, valid input.

* A crafted record does not have to be valid or 'nice', actually, I'd say that crafting a record you cam make it uglier than you'd never get.

* The ugliness that is going to hit us later, by definition, will not be caught by the records we already have.

* All existing records, are a subset of all crafted records too, so anything you can do with an existing record, you can do with a crafted one, but not otherwise (see previous two points).

In essence, you can use crafted or non-crafted records, as long as you are choosing them wisely, though with crafted records you have a lot more power on how and what to test, including the ability to create records that will test several things at the same time, that test specific issues, reducing the amount of examples and passes you have to do to test all the cases, and the time it takes to find the real records that will trigger those cases.

jacquerie:
> That it does not apply to this PR

Just for clarity: mine was a comment, not something intended to block this PR. In fact, I agree on all of your points, when you are being intellectually honest fake or real records don't really matter. The problem is that it's very easy to lapse in judgement, because "[...] you must not fool yourself – and you are the easiest person to fool". If you want a concrete example inside of the code base, look at the tests for `contact_details` in `tests/unit/dojson`.

tests/integration/workflows/fixtures/oai_arxiv_record_with_plots.xml
jacquerie: > That it does not apply to this PR

Just for clarity: mine was a comment, not something intended to block this PR. In fact, I agree on all of your points, when you are being intellectually honest fake or real records don't really matter. The problem is that it's very easy to lapse in judgement, because "[...] you must not fool yourself – and you are the easiest person to fool". If you want a concrete example inside of the code base, look at the tests for `contact_details` in `tests/unit/dojson`.

PR #2151, opened 2017-03-29
records: add custom validators

zzacharo:
Intorduce custom validators ( related to this list https://github.com/inveniosoftware-contrib/invenio-record-editor/issues/9) under `/api/editor/validate` . These validators will pass through `RECORD_EDITOR_VALIDATOR_FNS` in inspirehep/config.py to `invenio-record-editor`. The latter will aggregate all the errors and return the response which will be shown in record-editor.

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -31,6 +31,8 @@ from invenio_oauthclient.contrib import orcid
 from invenio_records_rest.facets import range_filter, terms_filter
 
 from inspirehep.modules.records.utils import get_detailed_template_from_record
+from .modules.records.validators.validator import InspireValidator
+from .modules.records.validators.validator import InspireResolver
 
 
 def _(x):
@@ -163,6 +165,10 @@ when creating records.
 
 RECORD_EDITOR_INDEX_TEMPLATE = 'inspirehep_theme/invenio_record_editor/index.html'
 RECORD_EDITOR_PREVIEW_TEMPLATE_FUNCTION = get_detailed_template_from_record
+RECORD_EDITOR_VALIDATORS = [
+    InspireValidator
+]
+RECORD_EDITOR_SCHEMA_RESOLVER = InspireResolver
 
 INSPIRE_COLLECTIONS_DEFINITION = [
     {
--- a/inspirehep/modules/migrator/tasks/records.py
+++ b/inspirehep/modules/migrator/tasks/records.py
@@ -293,6 +293,7 @@ def record_upsert(json):
         except PIDDoesNotExistError:
             record = InspireRecord.create(json, id_=None)
             # Create persistent identifier.
+           # import pytest;pytest.set_trace()
             inspire_recid_minter(str(record.id), json)
 
         if json.get('deleted'):
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -71,6 +71,10 @@ class InspireRecord(Record):
     def _delete(self, *args, **kwargs):
         super(InspireRecord, self).delete(*args, **kwargs)
 
+    def commit(self, validators=None):
+        """Override commit() to pass custom validators"""
+        super(InspireRecord, self).commit(validators)
+
 
 class ESRecord(InspireRecord):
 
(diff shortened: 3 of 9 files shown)


kaplun:
Hi @zzacharo , can you amend the commit log for this commit with something explaining what it is about? :)

inspirehep/modules/recordvalidators/validators/rules.py
```diff
@@ -0,0 +1,402 @@
...
+    except ValidationError:
+        raise
+    return {}
+
+
+def check_external_urls_if_work(record):
+    def _check_if_url_works(field, prop, value, index, errortype):
+        try:
+            resp = requests.get(value)
+            if not resp.ok:
```
jmartinm: You should be able to get more relevant error from the status code and maybe there is a message in the response. To see if there is 404, 500, 3xx,..

zzacharo:
@kaplun yes of course!

inspirehep/modules/recordvalidators/validators/rules.py
```diff
@@ -0,0 +1,402 @@
...
+                raise ConnectionError
```
jmartinm: Raising here and catching immediately below looks a bit weird, you could just


```python
if not resp.ok:
    raise ValidationError
```

inspirehep/modules/recordvalidators/validators/rules.py
```diff
@@ -0,0 +1,402 @@
...
+    except ValidationError:
+        raise
+    return {}
+
+
+def check_external_dois_if_exist(record):
+    def _check_if_doi_works(field, prop, value, index, errortype):
+        doi_url = DOI_VALIDATION_URL.format(doi=value)
+        try:
+            resp = requests.get(doi_url)
```
jmartinm: Double check this works with real values since there will be a redirect and I am not sure if `requests` library follows redirects by default.

inspirehep/modules/recordvalidators/validators/rules.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+'''
+Warnings Validator Functions
+'''
+
+
+def check_affiliations_if_authors_exist(record):
+    def _check_affiliations_for_author(author, index):
+        if 'affiliations' not in author or len(author['affiliations']) == 0:
```
kaplun: In principle we should always strip them. Not sure if this is enforced in all layers, though. But I believe we can well assume it in this type of code.

zzacharo: So it's ok to remove the second check?

jmartinm: Yes, for the moment leave it without and we see if it blows up. Then we can fix the piece of code that is adding empty lists.

jacquerie:
Mh. I have a feeling that you're reinventing the wheel here. Look at this function, and in particular the `validators` parameter: http://python-jsonschema.readthedocs.io/en/latest/creating/#creating-validators

The implementation of JSONSchema that we use allows you to pass a list of custom rules to be run at validation time, so we can use that instead of defining our own external mechanism. Moreover, the `ValidationError` class contains a lot of data that can be used to decide if something is an error or just a warning: http://python-jsonschema.readthedocs.io/en/latest/errors/#handling-validation-errors.

If you're wondering how to get the full list of errors: http://python-jsonschema.readthedocs.io/en/latest/validate/#jsonschema.IValidator.iter_errors

jmartinm:
@jacquerie That is very cool, did not think that `jsonschema` library might have such functionality.

If we made it work like that, we could centralize it so that using invenio-records API (https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L78) would call our custom validators every time.

That would ensure that no insert would go through with, for example, a duplicate DOI.

And then from the record-editor, doing a `PUT` to invenio-records API would return all the list of errors.

michamos:
:+1: We could put the basic validators which don't require awareness of the world beyond the record and the schema in `inspire-schemas`, and extend the validator with more advanced checks in inspire-next.

zzacharo:
Continues to #2306 .

PR #2155, opened 2017-03-30
Incomplete_date_forms

Glignos:
removed date widget for not being usefull for the date insertion as well as causing trouble with allowing iso date forms.

### DIFF
--- a/inspirehep/modules/literaturesuggest/forms.py
+++ b/inspirehep/modules/literaturesuggest/forms.py
@@ -130,18 +130,6 @@ def radiochoice_buttons(field, **dummy_kwargs):
     return HTMLString(u''.join(html))
 
 
-def defensedate_widget(field, **kwargs):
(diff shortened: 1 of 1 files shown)


jmartinm:
Please amend commit to follow the conventions of the project. Check `git log`

In your case it would be something like:

```
literaturesuggest: datepicker removal in thesis_date

* Removes datepicker to allow users to input dates in the formats YYYY-MM-DD,
  YYYY-MM or YYYY. (closes #2068)
```

jmartinm:
After this change `defensedate_widget` will become unused, so you can remove it from the codebase.

Can you double check that other fields with dates accept the same formats YYYY-MM-DD,
  YYYY-MM or YYYY ?

jmartinm:
See also https://github.com/inspirehep/inspire-next/pull/2157#issuecomment-290690982

Issue #1057, opened 2016-04-20
Improve PDF checking in literature suggestion form


jmartinm:
As suggested by @kaplun the following code can be added to the PDF validator:

``` python
import requests
import PyPDF2

try:
     PyPDF2.PdfFileReader(requests.get(url), strict=False)
except PyPDF2.utils.PdfReadError as err:
    raise StopValidation(err)
```

We just need to make sure that this validator runs async and probably disable it on form submission so that it does not delay the sending of the form.

kaplun:
You mean: it has run by the time the user reaches the submit button then we are good otherwise too bad?

How far is the field from the submit button? I remember it was quite close, so most of the time such check will not have time to run, correct?

jmartinm:
> How far is the field from the submit button? I remember it was quite close, so most of the time such check will not have time to run, correct?

It's actually quite at the top and not that close to the submit so might be fine. We could see if there is an ongoing request and when clicking on submit waiting for it to finish (to make sure all validators are done)

kaplun:
> We could see if there is an ongoing request and when clicking on submit waiting for it to finish (to make sure all validators are done)

:+1: That would be ideal. You can gray out the submit button and put a message next to it: `... still checking for validity of the provided PDF...`

kaplun:
cc: @jacquerie @david-caro this is a problem also when harvesting arXiv.

kaplun:
E.g. with the current code:
https://dspace.library.uu.nl/bitstream/handle/1874/18575/jacobsohn_60_relations.pdf the server is replying mimetype `image/pdf`(sic!!!) which is then considered as invalid.

PR #2156, opened 2017-03-30
literaturesuggest: prevent duplicate URLs

### LINKED ISSUE #2066: Submission should only add link to PDF if no URL provided
Currently, when a submission comes in, the PDF gets attached (optionally), but also added as an external link to the record (`urls`). This should only happen if no other URL has been provided by the user.

jmartinm:
* Do not allow a PDF to be inserted in the additional url field.

* Avoid adding two equal URLs to the record.

* Only add PDF as a link when no additional url is present. (closes #2066)

### DIFF
--- a/inspirehep/modules/forms/validators/simple_fields.py
+++ b/inspirehep/modules/forms/validators/simple_fields.py
@@ -177,6 +177,27 @@ def pdf_validator(form, field):
         raise StopValidation(message)
 
 
+def no_pdf_validator(form, field):
+    """Validate that url is not from a PDF."""
+    import requests
+
+    def get_content_type(url):
+        session = requests.Session()
+        try:
+            response = session.head(
+                url,
+                allow_redirects=True
+            )
+        except:
+            return
+        return response.headers['content-type']
+
+    message = "Please, use the field above to link to a PDF."
+
+    if field.data and get_content_type(field.data) == 'application/pdf':
+        raise StopValidation(message)
+
+
 def date_validator(form, field):
     message = ("Please, provide a valid date in the format YYYY-MM-DD, YYYY-MM"
                " or YYYY.")
--- a/inspirehep/modules/literaturesuggest/tasks.py
+++ b/inspirehep/modules/literaturesuggest/tasks.py
@@ -183,12 +183,15 @@ def formdata_to_model(obj, formdata):
         source='arXiv' if form_fields.get('categories') else 'CrossRef'
     )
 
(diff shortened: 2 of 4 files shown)


inspirehep/modules/forms/validators/simple_fields.py
```diff
@@ -177,6 +177,27 @@ def get_content_type(url):
...
+        try:
+            response = session.head(
+                url,
+                allow_redirects=True
+            )
+        except:
+            return
+        return response.headers['content-type']
+
+    message = "Please, use the field above for links to a PDF."
```
michamos: `for links` -> `to link`, we do not allow more than one in the form.

inspirehep/modules/forms/validators/simple_fields.py
```diff
@@ -177,6 +177,27 @@ def get_content_type(url):
         raise StopValidation(message)
```
michamos: don't misconfigured webservers lie about `content-type`?

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -183,12 +183,16 @@ def _is_arxiv_url(url):
...
-        builder.add_url(url=form_fields.get('additional_url'))
+    form_url = form_fields.get('url')
+    form_additional_url = form_fields.get('additional_url')
+    if form_url and not _is_arxiv_url(form_url):
+        obj.extra_data['submission_pdf'] = form_url
+        if not form_additional_url:
+            builder.add_url(url=form_url)
+
+    if form_additional_url and not _is_arxiv_url(form_additional_url):
+        if form_url != form_additional_url:
```
michamos: I don't think you need this test, now if you provide twice the same url nothing gets attached (you might want to test that).

inspirehep/modules/forms/validators/simple_fields.py
jmartinm: Yes, sometimes we get errors because of that. A more robust check is pending to be implemented https://github.com/inspirehep/inspire-next/issues/1057

inspirehep/modules/literaturesuggest/tasks.py
jmartinm: Actually this should not be possible since one field ensures that it is a PDF and the other ensures that it is not a PDF. So if validators work properly this cannot happen.

michamos: It can happen if `requests.head()` in the validator throws an exception. I still don't understand why you need this check.

Issue #2133, opened 2017-03-24
Format strings should be unicode when value can be unicode


michamos:
A typical cause of unicode errors is to have a non-unicode format string with a unicode value:
```python
In [1]: def broken_format(value):
   ...:     return 'Value: {}'.format(value)
   ...: 

In [2]: broken_format('hello')
Out[2]: 'Value: hello'

In [3]: broken_format(u'καλημέρα')
---------------------------------------------------------------------------
UnicodeEncodeError                        Traceback (most recent call last)
<ipython-input-3-f592da2d7e68> in <module>()
----> 1 broken_format(u'καλημέρα')

<ipython-input-1-6aee3f447981> in broken_format(value)
      1 def broken_format(value):
----> 2     return 'Value: {}'.format(value)

UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-7: ordinal not in range(128)

In [4]: def good_format(value):
   ...:     return u'Value: {}'.format(value)
   ...: 

In [5]: good_format(u'καλημέρα')
Out[5]: u'Value: \u03ba\u03b1\u03bb\u03b7\u03bc\u03ad\u03c1\u03b1'
```

Instead of fixing these one by one as they pop up on sentry (see #2026, #2030, #2107, #2110, #2116, #2123, #2127), it would be better to do a systematic check to solve all of them once and for all.

PR #2158, opened 2017-03-30
workflows: create & submit rt ticket handles unicode

chris-asl:
Firstly, refactors `submit_rt_ticket` to remove code for logging when not in production. 
Ensure that `create_ticket` and `submit_rt_ticket` can handle `subject` and `body` unicode strings. 

Related to https://github.com/inspirehep/inspire-next/issues/2133.

### DIFF
--- a/tests/unit/workflows/test_workflows_tasks_submission.py
+++ b/tests/unit/workflows/test_workflows_tasks_submission.py
@@ -22,11 +22,14 @@
 
 from __future__ import absolute_import, division, print_function
 
+from mock import patch
+import pytest
 from six import StringIO
 
 from inspirehep.modules.workflows.tasks.submission import (
     add_note_entry,
     prepare_files,
+    create_ticket
 )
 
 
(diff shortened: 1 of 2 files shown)


tests/unit/workflows/test_workflows_tasks_submission.py
```diff
@@ -215,3 +222,59 @@ def test_prepare_files_ignores_keys_not_ending_with_pdf():
...
+    mocked_RT_instance.create_ticket = mocked_create_ticket
+    mocked_get_instance_method.return_value = mocked_RT_instance
+
+    mockLogObj = MockLog()
+    data = {'log': mockLogObj}
+    extra_data = {'recid': 1}
+    obj = StubObj(data, extra_data, None)
+
+    assert submit_rt_ticket(obj, None, subject, body, None, "ticket_id key")\
+        is True
```
jacquerie: `assert something_true is True` can just be `assert something true`

chris-asl: You're right, just pushed a change.

Issue #2000, opened 2017-03-08
Refactor formdata_to_model in literaturesuggest. Removing dependecies


rikirenz:
Now the function `formdata_to_model` in `literaturesuggest` is doing to many things. These is the list of the operations that `formdata_to_model` should not do:

- split page artid (https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/literaturesuggest/tasks.py#L132)
- normalize journal title (https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/literaturesuggest/tasks.py#L122)
- query the user data (https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/literaturesuggest/tasks.py#L200)
- retrive orcid (https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/literaturesuggest/tasks.py#L204, https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/literaturesuggest/tasks.py#L42)

IMHO these operations should be moved in a function called `normalization_formdata`, that should be called before the `formdata_to_model`. This is how the function should be implemented:

``` Python
def normalization_formdata(formdata):
     formdata = split_page_artid_normalization(formdata)
     formdata = journal_normalization()     
     [...]
     formdata = retrive_orcid()
     return formdata
```

PR #2162, opened 2017-03-31
Remove English language + Refactor for formdata_to_model

### LINKED ISSUE #2103: workflows: 'en' language should generally not be written in the record
Currently, on the submission form, `en` is the default language, so
```json
"languages": ["en"]
```
gets written in the record. We don't want this, as the overwhelming majority of records are in English.
Instead, at some point in the submission/workflow, the following should happen:
if there is no `title_translations` or `languages` contains only `en`, remove languages.

rikirenz:
This PR wants to close #2103 #2000.


inspirehep/modules/literaturesuggest/form_normalization.py
```diff
@@ -0,0 +1,86 @@
...
+from sqlalchemy.orm.exc import NoResultFound
+
+from invenio_accounts.models import User
+from invenio_oauthclient.models import UserIdentity
+
+from inspirehep.utils.pubnote import split_page_artid
+from inspirehep.utils.normalizers import normalize_journal_title
+
+
+def normalization_formdata(workflow_object, formdata):
```
jacquerie: `s/normalization/normalize`: function names in general should be imperative (a cool read on the topic: http://steve-yegge.blogspot.ch/2006/03/execution-in-kingdom-of-nouns.html)

jacquerie: I would also argue that a `workflow_object` is consistently called `obj` in the rest of the codebase, so it's better to call it that. But I'd like @david-caro's opinion on this.

inspirehep/modules/literaturesuggest/form_normalization.py
```diff
@@ -0,0 +1,86 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2017 CERN.
```
jacquerie: You'll need to run `scripts/prettify_licenses`

inspirehep/modules/literaturesuggest/form_normalization.py
```diff
@@ -0,0 +1,86 @@
...
+
+def journal_title_normalization(data):
+    if data.get('journal_title'):
+        data['journal_title'] = normalize_journal_title(
+            data['journal_title']
+        )
+    return data
+
+
+def get_user_mail(workflow_obj, data):
```
jacquerie: `s/data/formdata/`: (You might enjoy this clickbaity article on calling variables `data`: http://archive.oreilly.com/pub/post/the_worlds_two_worst_variable.html)

inspirehep/modules/literaturesuggest/form_normalization.py
```diff
@@ -0,0 +1,86 @@
...
+            id_user=workflow_obj.id_user,
+            method='orcid'
+        ).one().id
+    except NoResultFound:
+        data['orcid'] = None
+
+    return data
+
+
+def remove_english_language(data):
```
jacquerie: It's weird that we populate this in the form, but then we remove it later. Of course this works, but can't we fix it in the form instead?

inspirehep/modules/literaturesuggest/form_normalization.py
rikirenz: IMHO I think `obj` is too much generic, but if we have a standard about that it is fine.

inspirehep/modules/literaturesuggest/form_normalization.py
rikirenz: We have already discuss about that. It should be fine.

inspirehep/modules/literaturesuggest/form_normalization.py
rikirenz: like `obj`

PR #2164, opened 2017-03-31
literaturesuggest:capitalized phd field in the correct format

Glignos:
*Edited the phd value field  to have the correct ISO(?) format

### DIFF
--- a/inspirehep/modules/literaturesuggest/forms.py
+++ b/inspirehep/modules/literaturesuggest/forms.py
@@ -637,7 +637,8 @@ class LiteratureForm(INSPIREForm):
 
         degree_type_schema = load_schema('elements/degree_type.json')
         degree_choices = [
-            (val, val.capitalize()) for val in degree_type_schema['enum']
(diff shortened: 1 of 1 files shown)


inspirehep/modules/literaturesuggest/forms.py
```diff
@@ -641,3 +641,11 @@ def __init__(self, *args, **kwargs):
         ]
         degree_choices.sort(key=lambda x: x[1])
         self.degree_type.choices = degree_choices
+        degree_choices = list(degree_choices)
```
jmartinm: There is no need for all this code actually. You can just change above:

```diff
         degree_choices = [
-            (val, val.capitalize()) for val in degree_type_schema['enum']
+            (val, val.capitalize()) if val != 'phd' else ('phd', 'PhD')
+            for val in degree_type_schema['enum']
         ]
```

jmartinm:
superseded by https://github.com/inspirehep/inspire-next/pull/2200

PR #2167, opened 2017-03-31
holdingpen: display ids for new authors detail

### LINKED ISSUE #2130: holdingpen: display BAI info
For new author additions, the BAI info is not displayed although we can find it if we open the record editor.
(requested by Margaret)

harunurhan:
* Fixes #2130

screen shot 2017-03-31 at 15 08 09

### DIFF
--- a/inspirehep/modules/workflows/static/js/inspire_workflows_ui/templates/details.html
+++ b/inspirehep/modules/workflows/static/js/inspire_workflows_ui/templates/details.html
@@ -208,15 +208,15 @@
             </div>
 
             <div ng-repeat="id in vm.record.metadata.ids" class="row">
-              <div ng-switch="id.type" class="col-md-12">
-                <span ng-switch-when="ORCID" class="font-bolder">ORCID:</span>
-                <span ng-switch-when="INSPIRE BAI" class="font-bolder">Author profile:</span>
-                <span ng-switch-when="ORCID" ng-bind="id.value"></span>
-                <span ng-switch-when="INSPIRE BAI">
-                  <a href="https://inspirehep.net/author/profile/{{id.value}}">
-                    {{id.value}}
-                  </a>
-                </span>
+              <div ng-switch="id.schema" class="col-md-12">
+                <div ng-switch-when="INSPIRE BAI">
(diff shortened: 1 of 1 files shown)


inspirehep/modules/workflows/static/js/inspire_workflows_ui/templates/details.html
```diff
@@ -208,15 +208,15 @@
...
-                    {{id.value}}
-                  </a>
-                </span>
+              <div ng-switch="id.schema" class="col-md-12">
+                <div ng-switch-when="INSPIRE BAI">
+                  <span class="font-bolder">Author profile:</span>
+                  <a href="https://inspirehep.net/author/profile/{{id.value}}">{{id.value}}</a>
+                </div>
+                <div ng-switch-when="ORCID">
+                  <span class="font-bolder">ORCID:</span>
```
kaplun: Can you also link the ORCID? to `http://orcid.org/{{id.value}}`.

harunurhan: Addressed @kaplun

kaplun:
One question though, actually also for usability experts @StellaCh @sefeg @jmartinm. Should enforce external link to be opened in a new tab with `target="_blank"`? I know formally this is not recommended. On the other hand the catalogers risk to move out from the application if they click on an external link (with respect to the holding pen) such as an ORCID or the author profile.

kaplun:
https://css-tricks.com/use-target_blank/#article-header-id-7

StellaCh:
Yes, external links should always be `target="_blank"`, otherwise users get confused and might lose track of what they're working on. 
Indeed, in cases where the user workflow is interrupted by creating new tabs, that's not recommended, but here that's not the case.

harunurhan:
@kaplun @StellaCh I made both of them to open in a new tab, but do we consider `inspirehep.net` as external or internal link?

kaplun:
@harunurhan IMHO we can consider external something as soon as it moves out of the current flow in the holding pen. Since holding pen is actually an application, everything that moves out from the flow shall be considered external link.

PR #2170, opened 2017-04-02
workflows: fix unicode issues in create_ticket

jacquerie:
Addresses #2133 
Closes #2158


jacquerie:
@chris-asl: please add here the relevant links to Sentry.

PR #2150, opened 2017-03-29
workflows: proper was_already_harvested_on_legacy

kaplun:
* Implements proper was_already_harvested_on_legacy() function to
  filter out arXiv records that are normally harvested on legacy.
  (closes #2114)

Signed-off-by: Samuele Kaplun <samuele.kaplun@cern.ch>

PR #2172, opened 2017-04-03
workflows: fix already_harvested

jacquerie:
Closes #2150 

Implements properly the `already_harvested` task so that it filters
out arXiv records that are harvested on Legacy (closes #2214).


tests/integration/workflows/fixtures/oai_arxiv_record_with_plots.xml
```diff
@@ -24,7 +24,7 @@
             </author>
           </authors>
           <title>Spectra generated by a confined softcore Coulomb potential</title>
-          <categories>math-ph math.MP hep-ex</categories>
+          <categories>math-ph math.MP</categories>
```
jacquerie: Ugly hack, will be fixed ASAP.

kaplun: ? Uh? It's not a hack. It's that the `hep-ex` category should not be ingested (as of now), but this fixture had been created for a successful ingestion (unfortunately the wrong example of record had been picked up).

kaplun:
I see you have added unit tests WRT my PR. Thanks. Anything more?

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -219,12 +208,29 @@ def article_exists(obj, eng):
...
 
 
+def is_being_harvested_on_legacy(record):
+    """Return True if the record is being harvested on Legacy.
+
+    If the record belongs to one of the CORE arXiv categories then it
+    is already being harvested on Legacy.
+    """
+    arxiv_categories = list(chain.from_iterable(
+        get_value(record, 'arxiv_eprints.categories', [])))
```
kaplun: Mmh. I am not sure this is more readable than the previous implementation... but OK.

PR #2175, opened 2017-04-03
Holdingpen ui fixes

jmartinm:


jmartinm:
In a separate PR I am going to do a trial of creating unit tests for the Angular application. That would ease certain tests rather than creating Acceptance tests.

Issue #2176, opened 2017-04-04
literature submission: wrong doc type conference paper


annetteholtkamp:
Conference papers appear with 980 "conference paper", not ConferencePaper

jmartinm:
@rikirenz can you have a look and fix this? Problem might be in `formdata_to_model` or in the dojson conversion.

jacquerie:
Problem is in DoJSON: https://github.com/inspirehep/inspire-next/blob/0dd585ef207128a75b8a35a1eda68b05b8d23bbd/inspirehep/dojson/hep/rules/bd9xx.py#L172-L176

jacquerie:
I'm going to steal this from @rikirenz because it's a quick fix for me.

PR #2179, opened 2017-04-04
dojson: fix document_type2marc

jacquerie:
Closes https://github.com/inspirehep/inspire-next/issues/2176


jacquerie:
Note that this doesn't handle `980__a:ActivityReport` because it was readded in `inspire-schemas~=31.0`, so that must happen in #2134 (or whatever @ammirate is going to push).

PR #2199, opened 2017-04-06
global: ensure unicode handling in all user-facing code

### LINKED ISSUE #2133: Format strings should be unicode when value can be unicode

chris-asl:
Closes #2133 

### DIFF
--- a/inspirehep/utils/conferences.py
+++ b/inspirehep/utils/conferences.py
@@ -51,7 +51,7 @@ def render_conferences_contributions(cnum):
 
 def conferences_in_the_same_series_from_es(seriesname):
     """Query ES for conferences in the same series."""
-    query = 'series:"{}"'.format(seriesname)
+    query = u'series:"{}"'.format(seriesname)
     return ConferencesSearch().query_from_iq(
         query
     ).params(
--- a/inspirehep/utils/experiments.py
+++ b/inspirehep/utils/experiments.py
@@ -52,7 +52,7 @@ def render_experiment_people(experiment_name):
 
 def experiment_contributions_from_es(experiment_name):
     """Query ES for conferences in the same series."""
-    query = 'accelerator_experiments.experiment:"{}"'.format(experiment_name)
+    query = u'accelerator_experiments.experiment:"{}"'.format(experiment_name)
     return LiteratureSearch().query_from_iq(
         query
     ).params(
--- a/inspirehep/modules/literaturesuggest/forms.py
+++ b/inspirehep/modules/literaturesuggest/forms.py
@@ -142,14 +142,14 @@ class CheckboxButton(object):
 
     def __call__(self, field, **kwargs):
         """Render PLUpload widget."""
-        html = '<div id="field-{0}">' \
-               '<label for="{0}">' \
-               '<input class="checkbox-ok-upload" name="{0}" type="checkbox" value="{2}">' \
-               '<strong><em>{1}</em></strong><small>&nbsp;(temporary text)</small>' \
-               '</label>' \
-               '</div>'.format(field.id,
-                               self.msg,
-                               field.default)
+        html = u'<div id="field-{0}">' \
+               u'<label for="{0}">' \
+               u'<input class="checkbox-ok-upload" name="{0}" type="checkbox" value="{2}">' \
+               u'<strong><em>{1}</em></strong><small>&nbsp;(temporary text)</small>' \
+               u'</label>' \
+               u'</div>'.format(field.id,
+                                self.msg,
+                                field.default)
         return HTMLString(html)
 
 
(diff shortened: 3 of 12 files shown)


tests/unit/utils/test_utils_pubnote.py
```diff
@@ -0,0 +1,37 @@
...
+from __future__ import absolute_import, division, print_function
+
+from inspirehep.utils.pubnote import build_pubnote
+
+
+def test_build_pubnote():
+    unicode_title = u'φοο_title'
+    unicode_volume = u'φοο_volume'
+
+    expected = u'{},{}'.format(unicode_title, unicode_volume)
```
jacquerie: Mh, read this: https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html

inspirehep/utils/conferences.py
```diff
@@ -53,7 +53,7 @@ def render_conferences_contributions(cnum):
 
 def conferences_in_the_same_series_from_es(seriesname):
     """Query ES for conferences in the same series."""
-    query = 'series:"{}"'.format(seriesname)
+    query = u'series:"{}"'.format(seriesname)
```
michamos: :+1: (sorry, just testing code review workflow with @kaplun and @jacquerie )

tests/unit/utils/test_utils_pubnote.py
chris-asl: Thanks for the link! 
I've updated the construction of `expected`.

inspirehep/utils/conferences.py
kaplun: While you are modifying this query, is there a safer way to escape `seriesname`? I assume that when it will contain a `"`, the call will blow up.

inspirehep/utils/experiments.py
```diff
@@ -54,7 +54,7 @@ def render_experiment_people(experiment_name):
 
 def experiment_contributions_from_es(experiment_name):
     """Query ES for conferences in the same series."""
-    query = 'accelerator_experiments.experiment:"{}"'.format(experiment_name)
+    query = u'accelerator_experiments.experiment:"{}"'.format(experiment_name)
```
kaplun: I see that also this part would require proper escaping of `experiment_name`.

tests/unit/utils/test_utils_pubnote.py
kaplun: Indeed I also learned something new today!

jacquerie:
> LGTM.

This is not done, though (it's still in WIP, and if you grep the codebase you'll find lots of other things to fix).

chris-asl:
(Hopefully) closes #2133

jacquerie:
> (Hopefully) closes #2133

GitHub is dumb, and understand these comments only in the first comment of the PR with a specific format: https://help.github.com/articles/closing-issues-via-commit-messages/

jmartinm:
As an alternative, we can try adding:

```
from __future__ import unicode_literals
```

on top of all files.

This needs to be tested to avoid unexpected side effects.

chris-asl:
Resurrected this PR as per @ammirate suggestion. 
Fixed some commit messages and rebased master ontop of the PR branch.

Regarding @kaplun suggestions, [here](https://github.com/inspirehep/inspire-next/pull/2199#discussion_r110407514) and [here](https://github.com/inspirehep/inspire-next/pull/2199#discussion_r110409184), I can integrate them in a another PR.

jacquerie:
@krzysztof pointed out in https://github.com/inveniosoftware/invenio/issues/3830#issue-262781413 an interesting tool to check if we actually got all these: https://github.com/cogniteev/flake8-unicode-string-literal

Spoiler alter: we didn't. But I'd say that we should add this dependency!

jacquerie:
This became stale, but most of the errors were fixed anyway. In the future we might add `unicode-string-literal` (like in https://github.com/inspirehep/inspire-dojson/commit/c8556d45eb2c16707e1237c0106565cf734d8924) to squash these errors at integration time.

Issue #2161, opened 2017-03-31
literature submission: thesis date includes time / phd lower case


annetteholtkamp:
When a submitter inserts a complete date year/month/day a time gets added:
Thesis: phd UC, Santa Barbara (2016-09-01T00:00:00) 

And the degree appears in lower case: phd

jmartinm:
Thanks. Yes, these are known issues that @Glignos is taking care of.

jacquerie:
Done in https://github.com/inspirehep/inspire-schemas/pull/138, but this is annoyingly blocked by #2134 (to get the new builder we have to bump the schema to the latest version).

jacquerie:
We actually realized that in order to fix the second issue we need to fix DoJSON: `phd` is what is written in the schema, but we want to produce back `PhD` in MARCXML.

Glignos:
@jacquerie The correct capitalization should happen also when the key is retrieved from the schema.

jacquerie:
This should be fixed. @annetteholtkamp, can you confirm?

michamos:
The PhD capitalization is fixed, not the timestamp one https://inspirehep.net/record/edit/?ln=en#state=edit&recid=1598894&recrev=20170511113857

jacquerie:
Whoops, right, because inspirehep/inspire-schemas#138 in the end landed in schema>31. We could backport it to `31.2.0` to close this quicker...

jacquerie:
We now have a schema that contains the changes in https://github.com/inspirehep/inspire-schemas/pull/124, so this should be fixed after a deploy.

jmartinm:
Fixed and deployed

PR #2200, opened 2017-04-07
literaturesuggest: capitalized phd field

jmartinm:
* Capitalized the phd value in the degree_type field. (addresses #2161)

Co-authored-by: Javier Martin Montull <javier.martin.montull@cern.ch>

### DIFF
--- a/inspirehep/modules/literaturesuggest/forms.py
+++ b/inspirehep/modules/literaturesuggest/forms.py
@@ -637,7 +637,8 @@ class LiteratureForm(INSPIREForm):
 
         degree_type_schema = load_schema('elements/degree_type.json')
         degree_choices = [
-            (val, val.capitalize()) for val in degree_type_schema['enum']
+            (val, val.capitalize()) if val != 'phd' else ('phd', 'PhD')
+            for val in degree_type_schema['enum']
         ]
         degree_choices.sort(key=lambda x: x[1])
         self.degree_type.choices = degree_choices
--- a/tests/unit/literaturesuggest/test_literaturesuggest_form.py
+++ b/tests/unit/literaturesuggest/test_literaturesuggest_form.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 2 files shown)


inspirehep/modules/literaturesuggest/forms.py
```diff
@@ -637,7 +637,8 @@ def __init__(self, *args, **kwargs):
 
         degree_type_schema = load_schema('elements/degree_type.json')
         degree_choices = [
-            (val, val.capitalize()) for val in degree_type_schema['enum']
+            (val, val.capitalize()) if val != 'phd' else ('phd', 'PhD')
+            for val in degree_type_schema['enum']
```
kaplun: This constructor starts to do too much.

These lines could be actually factored out into a small testable function:
```
def make_degree_choices():
```

The test would just check against the current expected output.

tests/unit/literaturesuggest/test_literaturesuggest_form.py
```diff
@@ -0,0 +1,34 @@
...
+from inspirehep.modules.literaturesuggest.forms import LiteratureForm
+
+
+def test_degree_choices_contains_capitalized_phd():
+    """Ensure that lowercased degree type in the schema ends up with the
+    right capitalization in the form.
+    """
+    literature_form = LiteratureForm()
+
+    assert ('phd', 'PhD') in literature_form.degree_type.choices
```
jacquerie: This is a bad test. You're just testing that an implementation detail is set in a certain way, which just acts as a checksum for the code you've just written (or rather, a redundant copy). This test is going to fail if you change the underlying implementation, but the behavior stays the same. In other words, this is a change detector test: https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html

This test should have happened at the UI level: you select "phd" from the dropdown, "PhD" comes out of the form.

jmartinm: Going back to this, since PR https://github.com/inspirehep/inspire-next/pull/2290 could use a test similar to this.

I am actually not sure about this not being a correct test. The forms are generated from Python so asserting that a certain field has certain values ensures that those values will appear in the UI.

jacquerie: But how this happens is an implementation detail, which is not a thing worth testing. If we change how the form library internally stores its data, this test is going to fail, even if everything in the UI looks correct.

BTW, this was not even what the user wanted! They wanted something solved by https://github.com/inspirehep/inspire-next/pull/2211.

jmartinm: > BTW, this was not even what the user wanted! 

❓ I think those were different issues and indeed this commit was saying addresses https://github.com/inspirehep/inspire-next/issues/2161 when it should not have.

Here there was just a display issue being addressed. Before, the dropdown was capitalizing all the options so a user would see `Phd` in the dropdown and what we wanted is to have `PhD` displayed.

So yeah, either it could be a regression test asserting on the wtforms field, or we can assert during the acceptance test of the form that the value `PhD` appears in the dropdown.

PR #2210, opened 2017-04-10
literaturesuggest: capitalized phd field

Glignos:
* Capitalized the phd value in the degree_type field. (addresses #2161)
  Also moved phd capitalizing to a seperate function to keep the constructor clean


jacquerie:
This was already done in #2200: there's no need to improve on it.

PR #2212, opened 2017-04-10
workflows: add debug logging to tasks

jacquerie:
Closes #1879 

I actually didn't test this locally yet, I just wanted to check that resuscitating this PR was easy (it was).

### DIFF
--- a/inspirehep/modules/workflows/utils.py
+++ b/inspirehep/modules/workflows/utils.py
@@ -24,12 +24,18 @@
 
 from __future__ import absolute_import, division, print_function
 
-import requests
 import json
+import logging
+import traceback
+from functools import wraps
 
+import requests
 from flask import current_app
 
 from .models import WorkflowsAudit
+
+
+LOGGER = logging.getLogger(__name__)
 
 
 def json_api_request(url, data, headers=None):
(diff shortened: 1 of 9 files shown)


inspirehep/modules/workflows/utils.py
```diff
@@ -83,3 +89,55 @@ def log_workflows_action(action, relevance_prediction,
...
+            else:
+                obj = kwargs.get('obj', kwargs.get('record'))
+            return obj
+
+        def _get_logfn(args, kwargs):
+            obj = _get_obj(args, kwargs)
+            if hasattr(obj, 'log') and hasattr(obj.log, 'debug'):
+                logfn = obj.log.debug
+            elif hasattr(current_app, 'log'):
+                logfn = current_app.log.debug
```
jacquerie: I think that this guy should be `current_app.logger` instead.

inspirehep/modules/workflows/utils.py
```diff
@@ -83,3 +89,55 @@ def log_workflows_action(action, relevance_prediction,
...
+
+            return logfn
+
+        def _proof_run(func, *args, **kwargs):
+            try:
+                func(*args, **kwargs)
+            except Exception:
+                LOGGER.debug(
+                    "Error while trying to log with %s:\n%s",
+                    logfn,
```
jacquerie: Unless I completely misunderstand what you're doing this should be `func`, no?

jacquerie: Or rather `func` in the entire function should be `logfn`, so that you shadow the outside definition?

david-caro: Magically both of them are the same... but yes, I agree that it should try to use the parameter one instead of the one in the closure. Probably will rename the parameter too, as it collides with the func from the outer decorator.

Maybe I just should move the 'helper' funcs that don't use any closures outside of the function...

PR #2213, opened 2017-04-11
forms: make pdf validators less strict

### LINKED ISSUE #1057: Improve PDF checking in literature suggestion form

jacquerie:
Makes `pdf_validator` less strict by accepting not only links that
point to resources whose reported `Content-Type` is `application/pdf`
by checking whether the first few bytes of the pointed resource are
`%PDF` (closes #1057).

### DIFF
--- a/inspirehep/modules/forms/validators/simple_fields.py
+++ b/inspirehep/modules/forms/validators/simple_fields.py
@@ -25,6 +25,7 @@ from __future__ import absolute_import, division, print_function
 from datetime import datetime
 from urllib import urlencode
 
+import requests
 from flask import current_app
 from flask.ext.login import current_user
 from idutils import is_arxiv, is_isbn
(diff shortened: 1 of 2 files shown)


inspirehep/modules/forms/validators/simple_fields.py
```diff
@@ -157,44 +158,18 @@ def duplicated_arxiv_id_validator(form, field):
...
-                url,
-                allow_redirects=True
-            )
-        except:
-            return
-        return response.headers['content-type']
-
-    message = "Please, provide an accessible direct link to a PDF document."
+    """Validate that the field contains a link to a PDF."""
+    message = 'Please, provide an accessible direct link to a PDF document.'
```
kaplun: I would go further and make this message for user firendly: "It looks like our server is not able to download this  document. Please provide an accessible direct link to a PDF document".

PR #2225, opened 2017-04-11
holdingpen: ignore experiments prediction mapping

kaplun:
* This allows putting there anything, it's a temporary workaround
  until we formalize what should be the data structure there.
  (closes #2144)


jacquerie:
Let's just do #2145, it's the right thing.

PR #2227, opened 2017-04-13
hal: rewrite module

### LINKED ISSUE #1584: HAL: fixmes
- [ ] Check the for_hal flag in the record to determine if it should be submitted
- [ ] Build a mapping between our categorization and the [HAL domains specified](https://api-preprod.archives-ouvertes.fr/ref/domain/?fl=*&wt=xml) (detailed in #1590)
- [ ] Add conference city, country, and country code fields
- [ ] Complete the helper function that specifies the [file source](https://api-preprod.archives-ouvertes.fr/ref/metadataList/?q=metaName_s:fileSource&fl=*&wt=xml) in the XML

### LINKED ISSUE #1590: HAL Domains
HAL Domains are here:
https://api-preprod.archives-ouvertes.fr/ref/domain/?fl=code_s,label_s&rows=500

We need a mapping between arXiv categories and INSPIRE categories to HAL categories.

jacquerie:
Closes #1584
Closes #1590 

Supersedes #1785


inspirehep/modules/hal/config.py
```diff
@@ -20,10 +20,46 @@
...
+}
+"""Mapping used when converting from INSPIRE categories to HAL domains."""
+
+
+#
+# Configuration used when connecting to HAL.
+#
+
+HAL_COL_IRI = 'https://api-preprod.archives-ouvertes.fr/sword/hal'
+HAL_EDIT_IRI = 'https://api-preprod.archives-ouvertes.fr/sword/'
```
kaplun: Shall we rather specify here both `preprod` and `prod` IRIs and use a runtime flag (defaulting to `PRODUCTION`) in order to switch to the correct IRIs?

A bit like: https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/contrib/orcid.py

This would simplify deployment and operations, which should no longer be aware of exact IRIs (a knowledge that is in principle known only by who implemented HAL integration).

inspirehep/modules/hal/config.py
```diff
@@ -20,10 +20,46 @@
...
+"""IRIs used by the SWORD protocol when connecting to HAL."""
+
+HAL_USER_NAME = 'hal_user_name'
+HAL_USER_PASS = 'hal_user_pass'
+"""INSPIRE credentials on HAL."""
+
+HAL_IGNORE_CERTIFICATES = False
+"""Whether to check certificates when connecting to HAL."""
```
kaplun: This is also bound to being on production?

inspirehep/modules/hal/core/tei.py
```diff
@@ -0,0 +1,135 @@
...
+    get_conference_title,
+    get_divulgation,
+    get_document_types,
+    get_doi,
+    get_domain,
+    get_journal_title,
+    get_language,
+    get_peer_reviewed,
+    get_publication_date,
+    is_published,
```
kaplun: Ideally all these should be moved into `inspire-schema` package, so that they can be naturally adjusted WRT schema changes.

kaplun: I understand this might be too much for this long-standing PR, so it might be postponed into a `technical-debt` issue.

inspirehep/modules/hal/core/tei.py
```diff
@@ -0,0 +1,135 @@
...
+    raise NotImplementedError
+
+
+def _is_comm(record):
+    document_types = get_document_types(record)
+
+    return 'conference paper' in document_types
+
+
+def _get_comm_context(record):
```
kaplun: I think it's more readable if we use human names than HAL specific abbreviations, when possible.

inspirehep/modules/hal/config.py
jacquerie: That's not a bad idea at all. Maybe not a flag, but an `APP` variable that wraps all these things together.

inspirehep/modules/hal/config.py
jacquerie: If you remember, this was needed when one of the servers in their staging environment was misconfigured, but they since fixed the bug. I chose to preserve this extra knob just to be a little more configurable, but they can be pulled in the previous `APP` variable.

inspirehep/modules/hal/core/tei.py
jacquerie: Yeah, ideally almost none of them should live in this module. Some should (like `get_divulgation` or `get_domain`), most other shouldn't.

inspirehep/modules/hal/core/tei.py
jacquerie: I disagree on this: what better names do we have for those things, if not the short ones they give them? This is what their names mean: https://api.archives-ouvertes.fr/ref/doctype/?instance_s=hal&wt=xml

kaplun: Mmh... Maybe then this naming can be referenced at the top of the module, so that it's obvious why these names have been chosen?

jacquerie: That's for sure! Documentation is woefully incomplete.

kaplun:
@jacquerie when you are back would be awesome to merge this branch.

PR #2263, opened 2017-04-21
dojson: handle keywords according to the schema

### LINKED ISSUE #2216: dojson: handling of keywords does not correspond to schema
We are missing `084`, we are populating `653` instead of `6531`, we are not handling correctly the various schemas when converting back to MARCXML.

See: https://github.com/inspirehep/inspire-schemas/blob/bdae2b7b50a68480fc0ac6452cb7508a8bad1c8c/inspire_schemas/records/hep.yml#L826-L897

jacquerie:
Closes #2216 

- [x] Reimplement `keywords` in `hep`
- [x] Reimplement `keywords2marc` in `hep2marc`
- [x] Sentry: https://sentry.cern.ch/inspire-sentry/inspire-nightly/group/821497/
- [x] Sentry: https://sentry.cern.ch/inspire-sentry/inspire-nightly/group/619244
- [x] Sentry: https://sentry.cern.ch/inspire-sentry/inspire-nightly/group/619228
- [x] Sentry: https://sentry.cern.ch/inspire-sentry/inspire-nightly/group/618716
- [x] Ensure that we don't regress on other things

First commit removes `keywords` from the `conferences` model.
Strictly speaking this needs to happen when we bump the schemas
to version ~34.0, but we might as well do it now to fix a few errors
on Sentry.

Second commit reimplements `keywords` and `keywords2marc`
in, respectively, the `hep` and `hep2marc` models.


jacquerie:
@michamos, DoJSON is currently doing something quite funky for the `source`: https://github.com/inspirehep/inspire-next/blob/12239630a763a8db96ecb6c78ebc7abec1a60c5d/inspirehep/dojson/hep/rules/bd6xx.py#L81-L93

If there's a single source, we return it.
If instead we have multiple sources, and if `conferences` is one of them, we discard all of them.
But if instead `conferences` is not among them, we return the first between `author` and `publisher` that we find.

Does this make sense? @david-caro, do you remember why you wrote it like this?

michamos:
It makes little sense to me, I think I am missing the backstory here. @david-caro ?

kaplun:
IIRC, year the reason is: DESY is adding the keywords of the conferences record into each conference contribution, in addition to whatever keyword this conference contribution might have.

kaplun:
E.g. see: https://inspirehep.net/search?ln=en&ln=en&p=6531_9%3Aconference&of=hb&action_search=Search&sf=earliestdate&so=d&rm=&rg=25&sc=0 the keywords are not those specific to the document, but does entered in the global conference.

michamos:
after talking with @kaplun, what's happening is that for some conferences, the keywords have been taken from the conference record and put in the proceedings record with `$$9conference`. We can indeed drop those for now (as there are very few records having them and we don't have keywords for conferences any longer).

jacquerie:
Ok, so we shouldn't drop the source, but we should drop the values if `conferences` is among the source.

inspirehep/dojson/hep/rules/bd6xx.py
```diff
@@ -62,133 +61,86 @@ def accelerator_experiments2marc(self, key, value):
...
-            if marc_dict.get('9') != 'magpie':
-                self['653'].append(marc_dict)
-            continue
+    keywords = self.get('keywords', [])
+    energy_ranges = self.get('energy_ranges', [])
+
+    for value in force_list(values):
+        if value.get('e'):
+            energy_ranges.append(maybe_int(value.get('e')))
+        elif value.get('a'):
```
michamos: are we sure there is never `a` and `e` in the same `695` field? maybe `if` instead of `elif` to play it safe?

inspirehep/dojson/hep/rules/bd6xx.py
```diff
@@ -62,133 +61,86 @@ def accelerator_experiments2marc(self, key, value):
...
+def keywords2marc(self, key, values):
+    """Populate the ``695`` MARC field.
+
+    Also populates the ``084`` and ``6531`` MARC fields through side effects.
+    """
+    result_695 = self.get('695', [])
+    result_084 = self.get('084', [])
+    result_6531 = self.get('6531', [])
+
+    for value in force_list(values):
```
michamos: why the `force_list`? isn't it guaranteed by the schema to be a list?

jacquerie: Right, we don't need to be safe here.

inspirehep/dojson/hep/rules/bd6xx.py
```diff
@@ -62,133 +61,86 @@ def accelerator_experiments2marc(self, key, value):
...
-@utils.ignore_value
+@hep2marc.over('695', '^energy_ranges$')
 def energy_ranges2marc(self, key, values):
-    values = force_list(values)
+    result_695 = self.get('695', [])
 
-    def _energy_to_marc_dict(energy_range):
-        return {
-            'e': energy_range,
+    for value in force_list(values):
```
jacquerie: Here as well.

inspirehep/dojson/hep/rules/bd6xx.py
jacquerie: I don't think this ever happens, but sure.

PR #2264, opened 2017-04-21
workflows: ignore %auto-ignore in arXiv XML files

### LINKED ISSUE #2195: workflows: ignore %auto-ignore in arXiv XML files
Sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs/group/820670/

XML files we receive from arXiv might contain a TeX-like comment at the top `%auto-ignore` to mean that they should not be available as separate downloads: https://arxiv.org/help/submit_tex#autoignore

Unfortunately, this makes these files invalid XML, so the step in our workflows that converts them using an XSLT fails on them.

Many thanks to @tsgit for helping diagnose this!

chris-asl:
Closes #2195


tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -152,3 +158,30 @@ def test_arxiv_derive_inspire_categories_does_nothing_with_existing_categories()
...
+    mock_eng = MockEng()
+
+    try:
+        temporary_dir = mkdtemp()
+        mock_os.path.abspath.return_value = mkdtemp()
+
+        author_list_method = arxiv_author_list()
+        author_list_method(mock_obj, mock_eng)
+    finally:
+        rmtree(temporary_dir)
```
jacquerie: You can use the [`tmpdir` fixture](https://docs.pytest.org/en/latest/tmpdir.html) for this kind of stuff.

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -152,3 +158,30 @@ def test_arxiv_derive_inspire_categories_does_nothing_with_existing_categories()
...
     assert expected == result
+
+
+@patch('inspirehep.modules.workflows.tasks.arxiv.os')
+def test_arxiv_author_list_handles_auto_ignore_from_arxiv(mock_os):
+    data = {
+        'arxiv_eprints': {
+            'value': ['1703.09986']
+        }
+    }
```
jacquerie: Note that this is incorrect: arxiv_eprints should be a list, not an object. This is why I started peppering our tests with assertions about the schema everytime we write code that relies on it. For example, a few lines above: https://github.com/inspirehep/inspire-next/blob/c14f5327de85f56da9a6565997a10fd6b3c71c43/tests/unit/workflows/test_workflows_tasks_arxiv.py#L137-L138

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -152,3 +158,30 @@ def test_arxiv_derive_inspire_categories_does_nothing_with_existing_categories()
...
+                'uri': resource_filename(__name__, 'fixtures/1703.09986.tar.gz')
+            })
+        })
+    })
+    mock_obj = MockObj(data, None, mock_arxiv_tarball)
+    mock_eng = MockEng()
+
+    try:
+        temporary_dir = mkdtemp()
+        mock_os.path.abspath.return_value = mkdtemp()
```
jacquerie: I guess the second `mkdtemp` call is a typo? You probably wanted to use the `temporary_dir` defined above...

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -152,3 +158,30 @@ def test_arxiv_derive_inspire_categories_does_nothing_with_existing_categories()
...
+def test_arxiv_author_list_handles_auto_ignore_from_arxiv(mock_os):
+    data = {
+        'arxiv_eprints': {
+            'value': ['1703.09986']
+        }
+    }
+    mock_arxiv_tarball = MockFiles({
+        '1703.09986.tar.gz': AttrDict({
+            'file': AttrDict({
+                'uri': resource_filename(__name__, 'fixtures/1703.09986.tar.gz')
```
jacquerie: Extremely small nit: `fixtures/1703.09986.tar.gz` uses the UNIX separator, which is not OS independent. So we can use `os.path.join` to use `os.path.sep` instead.

I'm actually not sure if `pkg_resources` handles this for us, but I don't have a Windows system to test this incredibly small detail.

tests/unit/workflows/test_workflows_tasks_arxiv.py
chris-asl: Yes, sorry for that..

tests/unit/workflows/test_workflows_tasks_arxiv.py
chris-asl: I see, will be using this kind of asserts when testing involves the schema.

PR #2271, opened 2017-04-22
workflows: ignore %auto-ignore in arXiv XML files

### LINKED ISSUE #2195: workflows: ignore %auto-ignore in arXiv XML files

jacquerie:
Closes #2195 
Closes #2264


jacquerie:
Just deployed this in production, and it fixed 5 out of 12 records in error state. Good job @chris-asl and @Glignos !

Issue #1646, opened 2016-10-13
global: suppress warnings during app instantiation


jacquerie:
Currently `inspirehep shell` prints the following output when `DEBUG=True`:

```
/Users/jacquerie/.virtualenvs/inspire/lib/python2.7/site-packages/flask/exthook.py:71: ExtDeprecationWarning: Importing flask.ext.cache is deprecated, use flask_cache instead.
  .format(x=modname), ExtDeprecationWarning

/Users/jacquerie/.virtualenvs/inspire/lib/python2.7/site-packages/sqlalchemy/ext/declarative/clsregistry.py:120: SAWarning: This declarative base already contains a class with the same class name and module name as flask_sqlalchemy.RecordMetadataVersion, and will be replaced in the string-lookup table.
  item.__name__

/Users/jacquerie/.virtualenvs/inspire/lib/python2.7/site-packages/sqlalchemy/orm/properties.py:194: SAWarning: On mapper Mapper|RecordMetadataVersion|records_metadata_version, primary key column 'records_metadata_version.transaction_id' is being combined with distinct primary key column 'records_metadata_version.transaction_id' in attribute 'transaction_id'.  Use explicit properties to give each column its own mapped attribute name.
  self.columns[0], self.key))

Python 2.7.12 (default, Oct 11 2016, 05:20:59)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.38)] on darwin
IPython: 5.1.0
App: inspirehep [debug]
Instance: /Users/jacquerie/.virtualenvs/inspire/bin/../var/inspirehep-instance

In [1]:
```

which gets even more annoying when you are executing several `inspirehep` commands in a row, like in `scripts/clean_assets`.
- [x] The first three `flask.ext` ones come from https://github.com/inveniosoftware/invenio-oauthclient, and will be fixed by https://github.com/inveniosoftware/invenio-oauthclient/pull/106.
- [x] I don't know what's causing the fourth one.
- [x] The next `flask.ext` one comes from https://github.com/thadeusb/flask-cache, and there's little chance of getting it fixed upstream (https://github.com/thadeusb/flask-cache/issues/151) unless we switch to https://github.com/sh4nks/flask-caching.
[...]

PR #2278, opened 2017-04-24
setup: pin Sphinx to 1.5.5

jacquerie:


jacquerie:
Sure it would have been nice to write why I did this...

jacquerie:
> Sure it would have been nice to write why I did this...

Ah, yes, the reason is that `Sphinx==1.6.2` treats warnings raised during application instantiation (the ones tracked in https://github.com/inspirehep/inspire-next/issues/1646) as errors, so the build fails.

PR #2279, opened 2017-04-24
dojson: support hidden publication_info

jacquerie:
Closes #1359

### DIFF
--- a/inspirehep/dojson/hep/rules/bd7xx.py
+++ b/inspirehep/dojson/hep/rules/bd7xx.py
@@ -107,36 +107,52 @@ def publication_info(self, key, value):
         'year': year,
         'parent_isbn': force_single_element(value.get('z')),
         'material': _get_material(value),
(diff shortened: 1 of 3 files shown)


inspirehep/modules/theme/jinja2filters.py
```diff
@@ -488,6 +488,9 @@ def publication_info(record):
     pub_infos = []
     if 'publication_info' in record:
         for pub_info in record.publication_information:
+            if pub_info.get('hidden'):
+                continue
+
```
jacquerie: This looks quite incomplete, as the `publication_info` is used for much more than just displaying it.

jacquerie: Actually, I'm considering removing it altogether: this is not the right way to interact with our schema. We need to have a smart getter that does this job for us that encapsulates the notion of "don't show hidden `publication_info`s" for us.

PR #2282, opened 2017-04-25
classifier: add core- and anti-HEP-ontology

fschwenn:
* Adds a reduced ontology with only the core HEP keywords.

* Adds an anti-HEP-ontology with keywords indicating irrelevance
  for HEP.

Signed-off-by: fschwenn <florian.schwennsen@desy.de>

### DIFF
--- a/inspirehep/taxonomies/HEPontCore.rdf
+++ b/inspirehep/taxonomies/HEPontCore.rdf
@@ -1,0 +1,16042 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- HEP ontology v.1.7 - April 2015 -->
+<!-- <Kirsten.Sachs@desy.de> <Florian.Schwennsen@desy.de> -->
+
+<rdf:RDF xmlns="http://www.w3.org/2004/02/skos/core#"
+    xmlns:dc="http://purl.org/dc/elements/1.1/"
+    xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+    xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"> 
+
+<rdf:Property rdf:about="http://cern.ch/thesauri/HEPontology.rdf#compositeOf">
+  <rdfs:comment>Expresses a composite keyword relationship to broader concepts.</rdfs:comment>
+  <rdfs:subPropertyOf rdf:resource="http://www.w3.org/2004/02/skos/core#broader"/>
+</rdf:Property>
+
+<rdf:Property rdf:about="http://cern.ch/thesauri/HEPontology.rdf#composite">
+  <rdfs:comment>Expresses the relationship to a composite keyword</rdfs:comment>
+  <rdfs:subPropertyOf rdf:resource="http://www.w3.org/2004/02/skos/core#narrower"/>
+</rdf:Property>
+
+<ConceptScheme rdf:about="http://cern.ch/thesauri/HEPontology.rdf">
+        <dc:creator rdf:resource="http://ceres.ca.gov/"/>
+        <dc:description>This is a first draft of a subject ontology/thesaurus for High Energy Physics using RDF/SKOS syntax 
+                        Developed jointly by CERN and DESY labs.</dc:description>
+        <dc:title>HEP ontology</dc:title>
+</ConceptScheme>
+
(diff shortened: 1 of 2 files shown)


kaplun:
@fschwenn do you have instead changes WRT the current full ontology? And also, how do you think of using the antihep one? As soon as there is an antiHEP keyword matched then it's a reject?

michamos:
> As soon as there is an antiHEP keyword matched then it's a reject?

I hope not, anti-HEP contains for example Fibonacci, and we have 335 core records with Fibonacci in the fulltext http://inspirehep.net/search?ln=en&ln=en&p=fulltext+fibonacci+and+tc+core&of=hb&action_search=Search&sf=earliestdate&so=d&rm=&rg=25&sc=0

kaplun:
@michamos, @fschwenn  I have digged the code and discovered that antiHEP KB is already there: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/kbs/corepar.kb
(a bit hidden given it's called `corepar.kb` rather than `antihep.kb`.)
It is used here:
https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/proxies.py#L38-L47

https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/tasks/classifier.py#L43

So the current meaning is: list all core keywords found in the paper that are not anti HEP.

Now I am puzzled: why are there keywords that are core but anti hep? What does it mean? Is it because you want to consider these as keywords that you want to have in the record but not use as criteria to evaluate coreness of a paper?

fschwenn:
The HEP ontology is up to date.
corepar.kb is the list of particle keywords (core and not-core) which in fact should not be counted for the decision whether some paper is core or not since keywords like "B" or "K" can be found in any kind of paper.
antihep.kb contains keywords which appear in articles we scan but strongly indicate that it is not relevant (like 'interplanetary' in asto-papers, 'positron emission tomography' in instrumental papers, 'Brillouin' in cond-mat papers). Appearance of such a keyword does not directly lead to rejection but together with other indicators. For the journals at DESY we use core-keywords, antiHEP keywords, #of references in INSPIRE, #of references in INSPIRE which are CORE, #of core-PACS, #of CORE-papers by the authors in combination.

kaplun:
@fschwenn coming back to this.

> For the journals at DESY we use core-keywords, antiHEP keywords, #of references in INSPIRE, #of references in INSPIRE which are CORE, #of core-PACS, #of CORE-papers by the authors in combination.

Does that mean that you display all this information and then a human manually takes the final decision?

All these metrics sounds like something we can compute and then we can train some ML to learn from the human decisions.

fschwenn:
> Does that mean that you display all this information and then a human manually takes the final decision?
Exactly.

kaplun:
What exactly do you mean by #of core-PACS? Also How do you know core authors? Given that you need first to guess who is who?

fschwenn:
I aggregated a list of mappings 'PACS->plain text' from the different edditions of PACS lists, since for selection purposes '04.60.Bc' is not very helpful but 'Phenomenology of quantum gravity' is! I also added the information whether a PACS indicates COREness like for the example '04.60.Bc' and all PACS starting with 1. If a metadata of a journal article include PACS, I (my program) translates them to plain text and counts how many are CORE.
My routine counting 'core authors' is very rude. If there are less than 20 authors on the paper, it just looks for "ea:[author as in the metadata]". It counts how many core papers each indiviual author has, how many the authors have where they all together are authors of the same paper. From these INSPIRE papers my code tries to guess whether there is fieldcode which the authors have very often. For chinese papers this is not very helpful, but for instrumentational papers the output "0,0,5;0" with fieldcode-guess=HEP-Theory tells me two authors (probably) have no CORE paper in INSPIRE and the third one probably has the same name as a theorist => paper is (probably) not interesting. In contrast "4,30,10;3" with fieldcode-guess=Instrumentation would tell me the three authors already have written 3 CORE papers as a team with the fieldcode-guess confirming that it's really them => paper is interesting. 
All in all it is not an extremely strong indicator which can not be used to immediately reject papers but I would miss it, as it sometimes draws my attention to papers which I otherwise would have missed. Also for conference proceedings this way of guessing the field code is quite efficient to distinguish 'n' from 'x' and 'p' from 'e' where it just saves a bit of my time.

PR #2283, opened 2017-04-25
workflow: fix extracting keywords from PDF

### LINKED ISSUE #2255: workflows: classifiers ignores PDF
The classification step currently looks for PDF hints in the wrong place. This is a blocker for the proper auto-reject algorithm, because current core keyword extraction is based only on abstract and title.

kaplun:
* Fixes extracting keywords from PDFs. (Closes #2255)

* Refactors retrieval of PDF unifying arXiv and submitter sources.
  Now PDFs are looked into the obj.files attribute.

### DIFF
--- a/inspirehep/modules/forms/validators/simple_fields.py
+++ b/inspirehep/modules/forms/validators/simple_fields.py
@@ -31,6 +31,8 @@ from flask.ext.login import current_user
 from idutils import is_arxiv, is_isbn
 from wtforms.validators import ValidationError, StopValidation
 
+from inspirehep.utils.url import is_pdf_link
+
 
 def _get_current_user_roles():
     return [r.name for r in current_user.roles]
@@ -62,7 +64,6 @@ def does_exist_in_inspirehep(query, collection=None):
     :param collection: collection to search in; by default searches in
         the default collection
     """
-    import requests
     from json import loads
 
     params = {
@@ -161,7 +162,7 @@ def pdf_validator(form, field):
     """Validate that the field contains a link to a PDF."""
     message = 'Please, provide an accessible direct link to a PDF document.'
 
-    if field.data and not _is_pdf_link(field.data):
+    if field.data and not is_pdf_link(field.data):
         raise StopValidation(message)
 
 
@@ -169,7 +170,7 @@ def no_pdf_validator(form, field):
     """Validate that the field does not contain a link to a PDF."""
     message = 'Please, use the field above to link to a PDF.'
 
-    if field.data and _is_pdf_link(field.data):
+    if field.data and is_pdf_link(field.data):
         raise StopValidation(message)
 
 
@@ -186,30 +187,3 @@ def date_validator(form, field):
                 break
         else:
             raise StopValidation(message)
-
-
-def _is_pdf_link(url):
-    """Return ``True`` if ``url`` points to a PDF.
-
-    Returns ``True`` if either the server replies with the correct
-    ``Content-Type`` or the first few bytes of the response are ``%PDF``.
-
-    Args:
-        url (string): a URL.
-
-    Returns:
-        bool: whether the url points to a PDF.
-
-    """
-    try:
-        response = requests.get(url, allow_redirects=True, stream=True)
-    except requests.exceptions.RequestException:
-        return False
-
-    content_type = response.headers.get('Content-Type', '')
-    correct_content_type = content_type.startswith('application/pdf')
-
-    magic_number = next(response.iter_content(4))
-    correct_magic_number = magic_number.startswith('%PDF')
-
-    return correct_content_type or correct_magic_number
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -27,12 +27,15 @@ from __future__ import absolute_import, division, print_function
 from functools import wraps
 
 from flask import current_app
+from werkzeug import secure_filename
 
-from inspirehep.modules.workflows.utils import log_workflows_action
+from inspirehep.modules.workflows.utils import log_workflows_action, get_pdf_in_workflow
 
 from inspirehep.utils.record import get_arxiv_id, get_value
+from inspirehep.utils.url import is_pdf_link
 
-from ..utils import with_debug_logging
+from ..utils import download_file_to_workflow, with_debug_logging
+from .refextract import extract_references
 
 
 def mark(key, value):
@@ -194,6 +197,24 @@ def is_submission(obj, eng):
     return False
 
 
+@with_debug_logging
+def submission_fulltext_download(obj, eng):
+    submission_pdf = obj.extra_data.get('submission_pdf')
+    if submission_pdf and is_pdf_link(submission_pdf):
+        filename = secure_filename("fulltext.pdf")
+        pdf = download_file_to_workflow(
+            workflow=obj,
+            name=filename,
+            url=submission_pdf
+        )
+        if pdf:
+            pdf['doctype'] = 'main'
+            obj.log.info("PDF provided by user from {0}".format(submission_pdf))
+            return obj.files[filename].file.uri
+        else:
+            obj.log.info("Cannot fetch PDF provided by user from {0}".format(submission_pdf))
+
+
 def prepare_update_payload(extra_data_key="update_payload"):
     @with_debug_logging
     @wraps(prepare_update_payload)
@@ -204,3 +225,24 @@ def prepare_update_payload(extra_data_key="update_payload"):
         # FIXME: Just update entire record for now
         obj.extra_data[extra_data_key] = obj.data
     return _prepare_update_payload
+
+
+@with_debug_logging
+def refextract(obj, eng):
+    """Extract references from PDF.
+
+    :param obj: Workflow Object to process
+    :param eng: Workflow Engine processing the object
+    """
+    uri = get_pdf_in_workflow(obj)
+    if uri:
+        mapped_references = extract_references(uri)
+        if mapped_references:
+            obj.data["references"] = mapped_references
+            obj.log.info("Extracted {0} references".format(
+                len(mapped_references)
+            ))
+        else:
+            obj.log.info("No references extracted")
+    else:
+        obj.log.error("Not able to download and process the PDF")
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -208,7 +208,6 @@ def submission_fulltext_download(obj, eng):
             url=submission_pdf
         )
         if pdf:
-            pdf['doctype'] = 'main'
             obj.log.info("PDF provided by user from {0}".format(submission_pdf))
             return obj.files[filename].file.uri
         else:
--- a/inspirehep/modules/workflows/tasks/arxiv.py
+++ b/inspirehep/modules/workflows/tasks/arxiv.py
@@ -74,7 +74,6 @@ def arxiv_fulltext_download(obj, eng):
             arxiv_id=arxiv_id
         )
     )
-    pdf['doctype'] = "arXiv"
     obj.log.info("PDF retrieved from arXiv for {0}".format(arxiv_id))
 
 
@@ -95,7 +94,6 @@ def arxiv_package_download(obj, eng):
         )
     )
     if tarball:
-        tarball['doctype'] = "package"
         obj.log.info("Tarball retrieved from arXiv for {0}".format(arxiv_id))
     else:
         obj.log.error("Cannot retrieve tarball from arXiv for {0}".format(arxiv_id))
@@ -129,7 +127,6 @@ def arxiv_plot_extract(obj, eng):
 
         for idx, plot in enumerate(plots):
             obj.files[plot.get('name')] = BytesIO(open(plot.get('url')))
-            obj.files[plot.get('name')]["doctype"] = "Plot"
             obj.files[plot.get('name')]["description"] = u"{0:05d} {1}".format(
                 idx, "".join(plot.get('captions', []))
             )
--- a/inspirehep/modules/workflows/tasks/arxiv.py
+++ b/inspirehep/modules/workflows/tasks/arxiv.py
@@ -37,7 +37,6 @@ from dojson.contrib.marc21.utils import create_record
 
 from inspirehep.dojson.hep import hep
 from inspirehep.dojson.utils import classify_field
-from inspirehep.utils.helpers import download_file_to_workflow
 from inspirehep.utils.record import (
     get_arxiv_categories,
     get_arxiv_id,
@@ -48,8 +47,7 @@ from plotextractor.api import process_tarball
 from plotextractor.converter import untar
 from plotextractor.errors import InvalidTarball, NoTexFilesFound
 
-from .refextract import extract_references
-from ..utils import with_debug_logging
+from ..utils import download_file_to_workflow, get_pdf_in_workflow, with_debug_logging
 
 
 REGEXP_AUTHLIST = re.compile(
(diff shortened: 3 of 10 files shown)


inspirehep/utils/helpers.py
```diff
@@ -42,23 +40,6 @@ def download_file(url, output_file=None, chunk_size=1024):
     return output_file
 
 
-def download_file_to_workflow(workflow, name, url):
```
kaplun: This really didn't belong to `utils.helpers`: I have moved it into `modules.workflows.utils` where it rather belong.

inspirehep/modules/workflows/utils.py
```diff
@@ -141,3 +146,68 @@ def _try_to_log(logfn, *args, **kwargs):
...
+        if filename.endswith('.pdf'):
+            # There's a PDF already attached
+            return obj.files[filename].file.uri
+
+    arxiv_id = get_arxiv_id(obj.data)
+    submission_pdf = obj.extra_data.get('submission_pdf')
+    if arxiv_id:
+        # It's an arXiv record, we give the PDF from arXiv
+        filename = secure_filename("{0}.pdf".format(arxiv_id))
+        if filename not in obj.files.keys:
```
david-caro: This if is not necessary, already checked at line 160

inspirehep/modules/workflows/utils.py
```diff
@@ -27,14 +27,19 @@
...
+from functools import wraps
 from flask import current_app
+from werkzeug import secure_filename
+
+from inspirehep.utils.record import get_arxiv_id
 
 from .models import WorkflowsAudit
 
 
+
```
david-caro: Extra newline

kaplun:
@david-caro do you see what is going wrong in the integration test? To me everything looks correctly mocked.

inspirehep/modules/workflows/utils.py
```diff
@@ -141,3 +145,67 @@ def _try_to_log(logfn, *args, **kwargs):
...
+@with_debug_logging
+def get_pdf_in_workflow(obj):
+    """Return the fullpath to the PDF attached to a workflow object
+
+    Return the PDF attached to a workflow object. If no PDF is already attached
+    and the workflow object is related to an arXiv record, fetches on the fly
+    the PDF from arXiv and attaches it.
+    """
+    for filename in obj.files.keys:
+        if filename.endswith('.pdf'):
```
jacquerie: This is not how you detect if a file is a PDF. This is a bad idea for at least two reasons:

1. We already experienced issues where a file has the wrong extension, causing a later exception.
2. This also represents a security vulnerability, because it trusts user input about the format of a file. Basically this code is like the clueless user that clicks on `Very_Important.exe.doc`.

inspirehep/modules/workflows/utils.py
```diff
@@ -141,3 +145,67 @@ def _try_to_log(logfn, *args, **kwargs):
...
+            url=submission_pdf
+        )
+        if pdf:
+            pdf['doctype'] = 'main'
+            obj.log.info("PDF provided by user from {0}".format(submission_pdf))
+            return obj.files[filename].file.uri
+
+    # No PDF available
+    obj.log.info("No PDF available")
+    return None
```
jacquerie: All Python functions return `None` without an explicit `return`.

inspirehep/modules/workflows/utils.py
```diff
@@ -141,3 +145,67 @@ def _try_to_log(logfn, *args, **kwargs):
...
+            name=filename,
+            url=submission_pdf
+        )
+        if pdf:
+            pdf['doctype'] = 'main'
+            obj.log.info("PDF provided by user from {0}".format(submission_pdf))
+            return obj.files[filename].file.uri
+
+    # No PDF available
+    obj.log.info("No PDF available")
```
jacquerie:
```python
# No PDF available
obj.log.info("No PDF available")
```
is just as silly as
```python
i += 1  # increment i by 1
```

inspirehep/modules/workflows/utils.py
kaplun: Good point. Just bare in mind that this is not where we download the PDF. This check here already assumes a PDF has been downloaded (thus making `get_pdf_in_workflow` idempotent).

There are two real situations: either the PDF is going to be downloaded proactively from arXiv. In that case we can trust the PDF is a PDF, either it is provided as URL by the user. Indeed this is risky and my code doesn't improve nor worsen the existing code. 
The line you mention is real to find if we indeed have already downloaded something that we think being a PDF.

I can add in a subsequent commit to ensure that what we fetch is actually a PDF. (and as @david-caro  mentioned IRL, that is not larger than a given large enough size).

inspirehep/modules/workflows/utils.py
```diff
@@ -141,3 +145,67 @@ def _try_to_log(logfn, *args, **kwargs):
...
+        filename = secure_filename("fulltext.pdf")
+        pdf = download_file_to_workflow(
+            workflow=obj,
+            name=filename,
+            url=submission_pdf
+        )
+        if pdf:
+            pdf['doctype'] = 'main'
+            obj.log.info("PDF provided by user from {0}".format(submission_pdf))
+            return obj.files[filename].file.uri
```
jacquerie: This function is doing two completely different things (fetching PDFs from arXiv and from the user submission), which is why it should rather be two functions.

inspirehep/modules/workflows/utils.py
kaplun: Yup, bad refactoring.

inspirehep/modules/workflows/utils.py
kaplun: I can split it in 2, but I will then need to wrap it, so that all the tasks that still need a PDF (plotextractor, refextractor, classifier...) don't have to know where the PDF is taken from.

inspirehep/modules/workflows/utils.py
kaplun: :+1: but now my 2 poor lines are in a corner crying because you called them `silly`. (besides it was for symmetry with the other 2 big blocks of this function).

jacquerie: I know! In fact I take issue with the other two comments as well. Comments that duplicate the code are not helpful, unless the code is not clear... but if the code is not clear, it's much better to fix the code, rather than adding a comment! In this case the code is clear, so the comments just add noise.

Tell those two lines that I'm sorry, but, you know, reviewing tends to bring the worst out of people...

inspirehep/modules/workflows/utils.py
jacquerie: Maybe then we should have another util that checks if a `WorkflowObject` has a PDF attached, instead of calling a util that does that check AND attempts to fetch the PDF if it's not there. I mean, functions should one thing well.

inspirehep/modules/workflows/utils.py
```diff
@@ -141,3 +145,54 @@ def _try_to_log(logfn, *args, **kwargs):
...
+    ``WorkflowFilesIterator``. This class inherits a custom ``__setitem__``
+    method from its parent, ``FilesIterator``, which ends up calling ``save``
+    on an ``invenio_files_rest.storage.pyfs.PyFSFileStorage`` instance
+    through ``ObjectVersion`` and ``FileObject``. This method consumes the
+    stream passed to it and saves in its place a ``FileObject`` with the
+    details of the downloaded file.
+    """
+    with closing(requests.get(url=url, stream=True)) as req:
+        if req.status_code == 200:
+            workflow.files[name] = req.raw
```
kaplun: @david-caro, @jacquerie here is the place where we risk to be overloaded by a too big requests. I don't see out of the box how we can inject here a check to stop reading the request after a configurable amount of bytes.
Any suggestion?

inspirehep/modules/forms/validators/simple_fields.py
```diff
@@ -186,30 +187,3 @@ def date_validator(form, field):
                 break
         else:
             raise StopValidation(message)
-
-
-def _is_pdf_link(url):
```
kaplun: Moving this to the more central utils, since I am going to use it also for workflows.

inspirehep/utils/url.py
```diff
@@ -38,3 +40,26 @@ def make_user_agent_string(component=""):
...
+    """
+    try:
+        response = requests.get(url, allow_redirects=True, stream=True)
+    except requests.exceptions.RequestException:
+        return False
+
+    magic_number = next(response.iter_content(4))
+    correct_magic_number = magic_number.startswith('%PDF')
+
+    return correct_magic_number
```
kaplun: I removed the check on the `content_type` because as we know servers should not be trusted.

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -180,12 +180,20 @@
...
             arxiv_refextract,
             arxiv_derive_inspire_categories,
             arxiv_author_list("authorlist2marcxml.xsl"),
         ]
     ),
+    IF(
+        is_submission,
+        [
+            submssion_fulltext_download,
+            arxiv_refextract,
```
kaplun: It now sucks that is called arxiv_refextract...

kaplun:
@jacquerie I have implemented the refactoring (minus broken code), and I have added the PDF test. If only mocking tests were passing at this point...

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -204,3 +224,24 @@ def _prepare_update_payload(obj, eng):
...
+        mapped_references = extract_references(uri)
+        if mapped_references:
+            obj.data["references"] = mapped_references
+            obj.log.info("Extracted {0} references".format(
+                len(mapped_references)
+            ))
+        else:
+            obj.log.info("No references extracted")
+    else:
+        obj.log.error("Not able to download and process the PDF")
```
jacquerie: Uhm, I guess that we had already some repetition in the code: https://github.com/inspirehep/inspire-next/blob/f173f92b7db9294d3c25de3690634a82c2d39ab2/inspirehep/modules/workflows/tasks/refextract.py#L87-L134

kaplun: What do you mean?

jacquerie: Right, scratch that, late night reviewing...

kaplun:
Everything is green!

kaplun:
I have tried hard with `git add -p` disentangling bug fixing and refactoring int his commit but it's becoming too time consuming. I am not able to obtain atomic commits that are actually correct.

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -68,90 +66,74 @@ def arxiv_fulltext_download(obj, eng):
...
-            )
+    tarball = download_file_to_workflow(
+        workflow=obj,
+        name=filename,
+        url=current_app.config['ARXIV_TARBALL_URL'].format(
+            arxiv_id=arxiv_id
         )
+    )
+    if tarball:
+        tarball['doctype'] = "package"
```
jacquerie: This key is not part of the schema: https://github.com/inspirehep/inspire-schemas/blob/1db9777634844724c522728521ab1f876b25ca8e/inspire_schemas/records/hep.yml#L99-L132

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -68,90 +66,74 @@ def arxiv_fulltext_download(obj, eng):
...
-        obj.log.error("Not able to download and process the PDF")
+            return
+        except DelegateError as err:
+            obj.log.error("Error extracting plots. Report and skip.")
+            current_app.logger.exception(err)
+            return
+
+        for idx, plot in enumerate(plots):
+            obj.files[plot.get('name')] = BytesIO(open(plot.get('url')))
+            obj.files[plot.get('name')]["doctype"] = "Plot"
```
jacquerie: Same here.

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -180,12 +181,20 @@
         is_arxiv_paper,
         [
             arxiv_fulltext_download,
+            arxiv_package_download,
```
jacquerie: Note that merging and deploying to this is going to break the workflows that are stuck on the next task (there are 5 of them).

inspirehep/modules/workflows/tasks/arxiv.py
kaplun: Is this going actually into the JSON? Isn't just staying simply in the `files` object? Note that `doctype` usage was already there before this PR.

inspirehep/modules/workflows/tasks/arxiv.py
kaplun: Yup was there before. Just touched as part of refactoring but haven't changed this behavior.

inspirehep/modules/workflows/workflows/article.py
kaplun: Since we need to kill all arxiv harvesting anyway and restarting it it's no problem, as long as there are instead no user submissions stuck.

inspirehep/modules/workflows/tasks/arxiv.py
kaplun: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/tasks/arxiv.py#L79

(just adding `tarball` for symmetry)

inspirehep/modules/workflows/tasks/arxiv.py
kaplun: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/tasks/arxiv.py#L118

inspirehep/modules/workflows/tasks/arxiv.py
jacquerie: Yes, this goes into the JSON: the `_files` property in the schema is there to back it, otherwise all records would become invalid due to `additionalProperties: False`.

inspirehep/modules/workflows/tasks/arxiv.py
jacquerie: I know that it was there before, but it's still wrong!

inspirehep/modules/workflows/tasks/arxiv.py
kaplun: So what do you propose I do? Later part in the code (already existing) uses the `doctype` key to fill in the FFT tag accordingly in order to send to legacy.

inspirehep/modules/workflows/tasks/arxiv.py
kaplun: Yeah but I can't correct everything in this PR. We can add it as an issue for later...

inspirehep/modules/workflows/tasks/arxiv.py
jacquerie: Where is this code that uses the `doctype`? I can't find it. I don't think it exists, in fact...

jacquerie:
```
% ag --python doctype

inspirehep/modules/disambiguation/config.py
32:DISAMBIGUATION_RECORD_DOCTYPE = "hep"

inspirehep/modules/disambiguation/search.py
51:    elasticsearch_type = current_app.config.get('DISAMBIGUATION_RECORD_DOCTYPE')
100:    elasticsearch_type = current_app.config.get('DISAMBIGUATION_RECORD_DOCTYPE')

inspirehep/modules/workflows/tasks/actions.py
211:            pdf['doctype'] = 'main'

inspirehep/modules/workflows/tasks/arxiv.py
77:    pdf['doctype'] = "arXiv"
98:        tarball['doctype'] = "package"
132:            obj.files[plot.get('name')]["doctype"] = "Plot"

tests/unit/forms/test_forms_validators_simple_fields.py
45:        body='<!doctype html>',
```

kaplun: E c'hai ragione! I had in my mind this:
> https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/tasks/submission.py#L403

jacquerie: I know that it sounds silly that we don't end up using the `files` property at all, but this is exactly what we were discussing the other day with @michamos...

kaplun: Yup I remember

kaplun: We will have to use it before Labs is declared master.

PR #2286, opened 2017-04-26
WIP (closes #2285)

### LINKED ISSUE #2285: testing: continuous_migration locking should be tested
9e3128682ff9dc72360173b2d0c227b908201a48 introduces protection of `continuous_migration()` via redis-lock. This should be tested so that a naïve developer doesn't later remove this functionality.

kaplun:


tests/integration/test_migrator.py
```diff
@@ -142,7 +164,41 @@ def test_continuous_migration_handles_multiple_records(app, record_1502655_and_1
     assert expected == result
 
 
+def test_continuous_migration_avoids_jamming(app, many_times_same_record):
+    with app.app_context():
```
jacquerie: You already have an application context: https://github.com/inspirehep/inspire-next/blob/319648c3d30afdbe2b34672beac6027b8858aa94/tests/integration/conftest.py#L50

tests/integration/test_migrator.py
```diff
@@ -142,7 +164,41 @@ def test_continuous_migration_handles_multiple_records(app, record_1502655_and_1
...
+        with mock.patch.dict(app.config, {'CELERY_ALWAYS_EAGER': False}):
```
jacquerie: I just realized one key thing: by the time you're running this line, the Celery app has already been instantiated, and it's too late to modify its behavior. I think that @spirosdelviniotis had a similar issue in the past, no?

jacquerie: The other thing is that you should be patching `current_app.config`, not `app.config`: I have a feeling that this was causing your woes with having to do internal imports (because you were modifying a fixture shared by other tests, not the application context on the stack).

kaplun: Ooh!

spirosdelviniotis: I had a similar issue but I didn't used mocks. The solution was to create the `app` with the `CELERY_ALWAYS_EAGER` flag.

```
app = create_app(
        DEBUG=True,
        WTF_CSRF_ENABLED=False,
        CELERY_ALWAYS_EAGER=True,
        CELERY_RESULT_BACKEND='cache',
        CELERY_CACHE_BACKEND='memory',
        CELERY_EAGER_PROPAGATES_EXCEPTIONS=True,
        TESTING=True,
    )
```

tests/integration/conftest.py
```diff
@@ -40,10 +40,11 @@ def app():
     app = create_app(
         DEBUG=True,
         WTF_CSRF_ENABLED=False,
-        CELERY_ALWAYS_EAGER=True,
+        CELERY_ALWAYS_EAGER=False,
```
jacquerie: But we don't want this...

kaplun: Yes. Right. Only my test (it was like that originally).

jacquerie: But this is the problem! We MUST do this, otherwise the Celery application has already been created with the wrong configuration, but we don't want to modify the configuration for all the other tasks. So the solution is to create another application fixture, something which I was trying to avoid (https://github.com/inspirehep/inspire-next/issues/1843).

kaplun: I don't understand: why importing locally isn't passing the correct celery app? Maybe it's too complex to answer.

PR #2289, opened 2017-04-27
editor: affilition suggester

harunurhan:
* Adds a suggester for instititution which suggestet based on
institution names and institution_acronym and returns results with
payload that is populated with necessary fields for UI.

@jmartinm  please check if the output and input make sense for the case. And I am assuming `institution_acronym` is present all the time, is it OK?

### DIFF
--- a/inspirehep/modules/records/mappings/records/institutions.json
+++ b/inspirehep/modules/records/mappings/records/institutions.json
@@ -58,9 +58,7 @@
                     "type": "string"
                 },
                 "affiliation_suggest": {
-                    "analyzer": "simple",
                     "payloads": true,
-                    "search_analyzer": "simple",
                     "type": "completion"
                 },
                 "control_number": {
--- a/inspirehep/modules/records/mappings/records/institutions.json
+++ b/inspirehep/modules/records/mappings/records/institutions.json
@@ -57,6 +57,12 @@
                 "affautocomplete": {
                     "type": "string"
                 },
+                "affiliation_suggest": {
+                    "analyzer": "simple",
+                    "payloads": true,
+                    "search_analyzer": "simple",
+                    "type": "completion"
+                },
                 "control_number": {
                     "type": "integer"
                 },
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -67,6 +67,7 @@ def enhance_record(sender, json, *args, **kwargs):
     add_recids_and_validate(sender, json, *args, **kwargs)
     populate_experiment_suggest(sender, json, *args, **kwargs)
     populate_abstract_source_suggest(sender, json, *args, **kwargs)
+    populate_affiliation_suggest(sender, json, *args, **kwargs)
     after_record_enhanced.send(json)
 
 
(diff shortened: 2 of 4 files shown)


jmartinm:
The field that should end up in the form input is `legacy_ICN`.

For input put all the ones that have `"copy_to": "affautocomplete"` https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/mappings/records/institutions.json#L14 plus all the fields in the address (https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/mappings/records/institutions.json#L31)

harunurhan:
@jmartinm  I think now it's good.

inspirehep/modules/records/mappings/records/institutions.json
```diff
@@ -57,6 +57,12 @@
                 "affautocomplete": {
                     "type": "string"
                 },
+                "affiliation_suggest": {
+                    "analyzer": "simple",
```
jacquerie: This is like this by default, there's no need to specify it explicitly.

inspirehep/modules/records/mappings/records/institutions.json
```diff
@@ -57,6 +57,12 @@
...
+                    "payloads": true,
+                    "search_analyzer": "simple",
```
jacquerie: Same as above.

tests/unit/records/test_records_receivers.py
```diff
@@ -541,7 +542,8 @@ def test_populate_experiment_suggest_populates_if_record_is_experiment():
     assert json_dict['experiment_suggest']['input'] == \
         ['foo', 'bar', 'foo_var', 'bar_var']
     assert json_dict['experiment_suggest']['output'] == 'foo'
-    assert json_dict['experiment_suggest']['payload']['$ref'] == 'http://foo/$ref'
+    assert json_dict['experiment_suggest'][
+        'payload']['$ref'] == 'http://foo/$ref'
```
jacquerie: Unrelated.

tests/unit/records/test_records_receivers.py
```diff
@@ -36,6 +36,7 @@
     references_validator,
     populate_experiment_suggest,
     populate_abstract_source_suggest,
+    populate_affiliation_suggest
```
jacquerie: Nit: missing comma.

tests/unit/records/test_records_receivers.py
```diff
@@ -598,3 +600,111 @@ def test_populate_abstract_source_suggest_does_nothing_if_record_is_not_hep():
         {'source': 'foo'},
         {'source': 'bar'},
     ]
+
+
+def test_populate_affiliation_suggest():
```
jacquerie: This test is **really** hard to follow. I will split it in several, each dealing with a different field. This is not just for readability: I will also add tests to ensure that the schema is what is asserted by the code, because the `institutions` schema is going to change soon, so it would be nice to have good errors on this.

inspirehep/modules/records/receivers.py
```diff
@@ -267,6 +268,55 @@ def populate_abstract_source_suggest(sender, json, *args, **kwargs):
...
+            'affiliation_suggest': {
+                'input': inputs,
+                'output': legacy_ICN,
+                'payload': {
+                    'institution': institution,
+                    'institution_acronym': institution_acronym,
+                    'ICN': ICN,
+                    'legacy_ICN': legacy_ICN,
+                    'department': json.get('department'),
+                    '$ref': get_value(json, 'self.$ref')
```
jacquerie: Nit: missing comma.

tests/unit/records/test_records_receivers.py
```diff
@@ -598,3 +600,111 @@ def test_populate_abstract_source_suggest_does_nothing_if_record_is_not_hep():
...
+                '$ref': 'http://foo/bar',
+            },
+        }
+    }
+
+
+def test_populate_affiliation_suggest_does_nothing_if_record_is_not_institution():
+    json_dict = {
+        '$schema': 'http://localhost:5000/schemas/records/other.json',
+        'other': 'foo'
```
jacquerie: We don't even need this key, no?

harunurhan: not really no.

tests/unit/records/test_records_receivers.py
harunurhan: do we have formatting script or something? to run before because usually when this always happen to me when I format with my text editor.

tests/unit/records/test_records_receivers.py
harunurhan: I totally agree with splitting 👍  it wasn't even easy to write and change it for me

PR #2290, opened 2017-04-27
authors: use high level categories in research_field

### LINKED ISSUE #2104: author submission: subject autocomplete
My suggestion would be to not list all arXiv subcategories, like math.AG, math.AT... but only higher-level categories. With the exception of physics.ins-det, physics.acc-phys, physics-data.an (these are the ones we take completely in HEP)

chris-asl:
Closes #2104

### DIFF
--- a/inspirehep/modules/authors/forms.py
+++ b/inspirehep/modules/authors/forms.py
@@ -449,6 +449,8 @@ class AuthorUpdateForm(INSPIREForm):
     research_field_options = [
         (val, val)
         for val in arxiv_categories_schema['enum']
+        if '.' not in val or
+        val in ('physics.ins-det', 'physics.acc-ph', 'physics.data-an')
     ]
 
     research_field = fields.SelectMultipleField(


inspirehep/modules/authors/forms.py
```diff
@@ -449,6 +449,8 @@ class AuthorUpdateForm(INSPIREForm):
     research_field_options = [
         (val, val)
         for val in arxiv_categories_schema['enum']
+        if '.' not in val or
+           val in ('physics.ins-det', 'physics.acc-ph', 'physics.data-an')
```
jmartinm: This should be aligned with the line above (not indented)

jmartinm:
To add a test for this code, I have added a comment to a previous PR that had a similar change (https://github.com/inspirehep/inspire-next/pull/2200/files#r113871190) to see what would be a good way of testing it. @jacquerie can you have a look at that?

jmartinm:
Since we agreed that this is ready. @chris-asl just rebase and we can merge it.

jacquerie:
Also please squash and fix the commit message:
```
authors: use high level categories in research_field

* Users requested that only the high level categories and some
  subcategories of physics should be presented as options for
  the `research_field` (closes #2104).

```

chris-asl:
Done! 
Will be using this commit message format henceforth.

PR #2296, opened 2017-05-01
global: set extra_data when creating a workflow

### LINKED ISSUE #2238: wrong files get attached to holding pen records
https://labs.inspirehep.net/holdingpen/619315 has an attached PDF, which is the PDF from a different record, namely https://labs.inspirehep.net/holdingpen/619303. Looks like the problem is on our side, as the submitters are different, and there is no PDF in the `formdata`.

jmartinm:
* NOTE: When creating a workflow if data or extra_data are not set
  the next workflow created will share the same data or extra_data
  instance. (closes #2238)

### DIFF
--- a/inspirehep/modules/authors/views/holdingpen.py
+++ b/inspirehep/modules/authors/views/holdingpen.py
@@ -272,6 +272,7 @@ def submitupdate():
 
     workflow_object = workflow_object_class.create(
         data={},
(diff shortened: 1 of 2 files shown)


jmartinm:
The correct solution is https://github.com/inveniosoftware-contrib/invenio-workflows/pull/38

Issue #2120, opened 2017-03-23
literature submission: journal autocomplete


annetteholtkamp:
Autocomplete inserts the full journal name, not the standard abbreviated one

annetteholtkamp:
It also only search in the abbreviated name. Entering the full name does not give any suggestions

PR #2299, opened 2017-05-02
utils/normalizers: normalize_journal_title proper search

chris-asl:
* Update journals mapping with lowercase analyzer.
* Add `lowercase_titles` field for searching the abbreviated journal
  name.

Co-Authored-By: George Lignos <glignos93@gmail.com>

Addresses #2120 (1st comment)

### DIFF
--- a/inspirehep/modules/records/mappings/records/journals.json
+++ b/inspirehep/modules/records/mappings/records/journals.json
@@ -70,7 +70,10 @@
                             "type": "string"
                         },
                         "title": {
-                            "copy_to": "journalautocomplete",
+                            "copy_to": [
+                                "journalautocomplete",
+                                "lowercase_titles"
+                            ],
                             "type": "string"
                         }
                     },
@@ -103,6 +106,10 @@
                         }
                     },
                     "type": "object"
+                },
+                "lowercase_journal_titles": {
+                    "analyzer": "lowercase_analyzer",
+                    "type": "string"
                 },
                 "peer_reviewed": {
                     "type": "boolean"
@@ -175,6 +182,7 @@
                             "type": "string"
                         },
                         "title": {
+                            "copy_to": "lowercase_titles",
                             "type": "string"
                         }
                     },
@@ -190,6 +198,17 @@
                         }
                     },
                     "type": "object"
+                }
+            }
+        }
+    },
+    "settings": {
+        "analysis": {
+            "analyzer": {
+                "lowercase_analyzer": {
+                    "filter": "lowercase",
+                    "tokenizer": "keyword",
+                    "type": "custom"
                 }
             }
         }
--- a/inspirehep/utils/normalizers.py
+++ b/inspirehep/utils/normalizers.py
@@ -22,12 +22,13 @@
 
 from __future__ import absolute_import, division, print_function
 
-from flask import current_app
+import logging
 
 from inspirehep.modules.search.api import JournalsSearch
 
 
 def normalize_journal_title(journal_title):
+    logger = logging.getLogger(__name__)
     normalized_journal_title = journal_title
     hits = JournalsSearch().query(
         'match',
@@ -38,5 +39,5 @@ def normalize_journal_title(journal_title):
         try:
             normalized_journal_title = hits[0].short_titles[0].title
         except (AttributeError, IndexError) as e:
-            current_app.logger.debug("Failed to access normalized journal title: {}".format(e))
+            logger.debug("Failed to access normalized journal title: %s", e)
     return normalized_journal_title
--- a/inspirehep/utils/normalizers.py
+++ b/inspirehep/utils/normalizers.py
@@ -22,6 +22,8 @@
 
 from __future__ import absolute_import, division, print_function
 
+from flask import current_app
+
 from inspirehep.modules.search.api import JournalsSearch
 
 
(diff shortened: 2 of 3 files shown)


tests/integration/utils/test_normalizers.py
```diff
@@ -0,0 +1,32 @@
...
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
```
jacquerie: Wrong license... where did you even find this one?!

inspirehep/utils/normalizers.py
```diff
@@ -22,19 +22,21 @@
...
+        lowercase_titles=journal_title.lower()
     ).execute()
 
     if hits:
         try:
             normalized_journal_title = hits[0].short_titles[0].title
-        except (AttributeError, IndexError):
-            pass
+        except (AttributeError, IndexError) as e:
+            current_app.logger.debug("Failed to access normalized journal title: {}".format(e))
```
jacquerie: This is not the right way to write logs. See: https://github.com/inspirehep/inspire-next/issues/1371

tests/integration/utils/test_normalizers.py
chris-asl: Sorry for that, was from the invenio-search `api` file. Will update.

inspirehep/utils/normalizers.py
chris-asl: Is there a guidelines document so that I can add this link? 
I think it would be really helpful for others.

inspirehep/modules/records/mappings/records/journals.json
```diff
@@ -175,6 +182,7 @@
                             "type": "string"
                         },
                         "title": {
+                            "copy_to": "lowercase_titles",
```
jmartinm: The field seems to be called `lowercase_journal_titles`?

inspirehep/utils/normalizers.py
```diff
@@ -22,19 +22,22 @@
 
 from __future__ import absolute_import, division, print_function
 
+import logging
```
jmartinm: Was doing some checks about this, and it is better to use:

```python
from flask import current_app

current_app.logger.debug()
```

inspirehep/utils/normalizers.py
```diff
@@ -22,19 +22,22 @@
...
 from inspirehep.modules.search.api import JournalsSearch
 
 
 def normalize_journal_title(journal_title):
+    logger = logging.getLogger(__name__)
     normalized_journal_title = journal_title
     hits = JournalsSearch().query(
         'match',
-        title_variants__title__lowercased=journal_title
+        lowercase_titles=journal_title.lower()
```
jmartinm: I am wondering if this matters, since the analyzer will do the lowercasing.

inspirehep/utils/normalizers.py
```diff
@@ -22,19 +22,22 @@
...
     ).execute()
 
     if hits:
         try:
             normalized_journal_title = hits[0].short_titles[0].title
-        except (AttributeError, IndexError):
-            pass
+        except (AttributeError, IndexError) as e:
+            logger.debug("Failed to access normalized journal title: %s", e)
```
jmartinm: Not sure how useful this message will be.

Could it be better:


```python
logger.debug("Failed to normalize journal title in document: %s", repr(hits[0]))
```

tests/integration/utils/test_normalizers.py
```diff
@@ -0,0 +1,35 @@
...
+
+from inspirehep.utils.normalizers import normalize_journal_title
+
+
+def test_normalize_journal_title(app):
+    journal_title = 'Physical Review'
+    abbreviated_journal_title = 'Phys.Rev.'
+
+    with app.test_client():
+        normalized_journal_title = normalize_journal_title(journal_title)
```
jmartinm: Why was this needed?

tests/integration/utils/test_normalizers.py
```diff
@@ -0,0 +1,35 @@
...
+from __future__ import absolute_import, division, print_function
+
+from inspirehep.utils.normalizers import normalize_journal_title
+
+
+def test_normalize_journal_title(app):
+    journal_title = 'Physical Review'
+    abbreviated_journal_title = 'Phys.Rev.'
+
+    with app.test_client():
```
jmartinm: Why was this needed?

inspirehep/modules/records/mappings/records/journals.json
chris-asl: Left there during renaming, will update now.

inspirehep/utils/normalizers.py
chris-asl: This change happened due to [this comment](https://github.com/inspirehep/inspire-next/pull/2299#pullrequestreview-35949298). 

Which of these two methods for logging, do you guys think I should use?

tests/integration/utils/test_normalizers.py
chris-asl: When we tested this with George, there was an error (regarding context) that sadly I am no able to re-produce locally now. I'll remove it.

chris-asl:
Actually, as we discussed with @jacquerie, due to the journal suggester PR, [here](https://github.com/inspirehep/inspire-next/pull/2311), I think the normalization (which comes with an extra query to ElasticSearch) isn't needed anymore. 

The integration of the journal suggester in the journals typeahead module, currently sets the value of the text field to the abbreviated title, as was requested by issue #2120. Thus, the normalized value is already there. 

Should I proceed with adding the changes from the later review or simply removing this normalizer?

jmartinm:
> Actually, as we discussed with @jacquerie, due to the journal suggester PR, here, I think the normalization (which comes with an extra query to ElasticSearch) isn't needed anymore.

We will better keep both since even though the autocomplete puts the short title in the input, a user might not select from the autocomplete (or when we import from Crossref we don't do the conversion client side).

I will take over this PR and complete it.

jmartinm:
Continued in https://github.com/inspirehep/inspire-next/pull/2392

Issue #2300, opened 2017-05-03
Literature: Search and detailed pages UI fixes


jmartinm:
After the latest data model changes, the following issues are present in the user interface:

* [x] Keywords are not shown
![screen shot 2017-05-03 at 10 11 56 am](https://cloud.githubusercontent.com/assets/580727/25652539/fdf8d3b6-2fe8-11e7-9cb4-c7fabdc975d7.png)

* [x] Plots Bootstrap carrousel is not working.
![screen shot 2017-05-03 at 10 12 23 am](https://cloud.githubusercontent.com/assets/580727/25652556/0fad3f8e-2fe9-11e7-8af3-f82d5253e6a0.png)

* [x] References data table crashes (example from record #1594850)
![screen shot 2017-05-03 at 10 13 52 am](https://cloud.githubusercontent.com/assets/580727/25652603/4536a898-2fe9-11e7-88cc-2798694b6fd6.png)

  (https://github.com/inspirehep/inspire-next/pull/2303)

* [x] Template of search results and detailed view needs to be reviewed since some elements might not be showing but Angular hides them silently.

PR #2302, opened 2017-05-03
search: prevent error when no documents found in mget

jmartinm:
* Catch ES RequestError exception to prevent the application from
  crashing when no documents are found in mget request. (addresses #2300)

* Log with warning level when this situation happens.

* Adds tests for to assert the new behaviour.

### DIFF
--- a/inspirehep/modules/search/api.py
+++ b/inspirehep/modules/search/api.py
@@ -22,9 +22,12 @@
 
 from __future__ import absolute_import, division, print_function
 
+import logging
+
 from flask import request
 from flask_security import current_user
 
+from elasticsearch import RequestError
 from elasticsearch_dsl.query import Q
 
 from invenio_search.api import DefaultFilter, RecordsSearch
@@ -37,7 +40,7 @@ from inspirehep.modules.records.permissions import (
 
 from .query_factory import inspire_query_factory
 
-
+logger = logging.getLogger(__name__)
 IQ = inspire_query_factory()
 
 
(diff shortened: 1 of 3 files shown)


tests/integration/search/test_search_api.py
```diff
@@ -0,0 +1,34 @@
...
+import mock
+
+from inspirehep.modules.search.api import LiteratureSearch
+
+
+@mock.patch('inspirehep.modules.search.api.current_app.logger.warning')
+def test_mget_warns_when_no_documents_to_get(w, app):
+    search = LiteratureSearch()
+    search.mget([])
+    w.assert_called_with("TransportError(400, u'action_request_validation_exception', u'Validation Failed: 1: no documents to get;')")
```
jmartinm: This assertion will fail if ES library changes its wording. On the other hand, it documents the current expected behaviour. @jacquerie what do you think?

inspirehep/modules/search/api.py
```diff
@@ -77,14 +78,20 @@ def mget(self, uuids, **kwargs):
...
+        try:
+            documents = current_search_client.mget(
+                index=self.Meta.index,
+                doc_type=self.Meta.doc_types,
+                body={'ids': uuids},
+                **kwargs
+            )
+            results = [document['_source'] for document in documents['docs']]
+        except RequestError as e:
+            current_app.logger.warning('{}'.format(e))
```
jacquerie: This is not the right way to do logs. See: https://github.com/inspirehep/inspire-next/issues/1371

PR #2306, opened 2017-05-03
validators: Introduce custom validators for hep records

zzacharo:
Start checking the implementation. Tests are missing.

### DIFF
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -39,6 +39,9 @@ from inspirehep.utils.record_getter import (
     get_es_record_by_uuid
 )
 
+from inspirehep.modules.records.validators.validator import InspireValidator, InspireResolver
+from jsonschema import validate
+
 
 class InspireRecord(Record):
 
@@ -70,6 +73,14 @@ class InspireRecord(Record):
 
     def _delete(self, *args, **kwargs):
         super(InspireRecord, self).delete(*args, **kwargs)
+
+    def commit(self, **kwargs):
+        """Override commit() to run custom validation before calling the parent class"""
+        validator = kwargs.pop('validator', InspireValidator)
+        if '$schema' in self and self['$schema'] is not None and validator is not None:
+            _schema = {"$ref": self['$schema']}
+            validate(self, _schema, cls=validator, resolver=InspireResolver.from_schema(_schema))
+        super(InspireRecord, self).commit()
 
 
 class ESRecord(InspireRecord):
--- a/inspirehep/modules/records/validators/__init__.py
+++ b/inspirehep/modules/records/validators/__init__.py
@@ -1,0 +1,22 @@
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
--- a/inspirehep/modules/records/validators/helpers.py
+++ b/inspirehep/modules/records/validators/helpers.py
@@ -1,0 +1,70 @@
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+
+from __future__ import absolute_import, division, print_function
+
+from jsonschema import ValidationError
+from collections import deque
+
+from invenio_db import db
+
+FIELD_REQUIRE_FIELD_VALUES_TEMPLATE = "'{field}' field requires '{required}' field to have " \
+                                      "at least one of the values {values}."
+FIELD_REQUIRE_FIELD_TEMPLATE = "'{field}' field requires '{required}' field to exist."
+FIELD_DUPLICATE_VALUES_FOUND_TEMPLATE = "'{field}' field with value '{value}' found in more than one record."
+FIELD_VALIDATION_TEMPLATE = "Field '{field}' with value '{value}' is not valid."
+
+
+def check_document_type_values_in_required_values_for_record(record, required_values):
+    document_type_values = record.get('document_type', [])
+    return len(set(document_type_values) & set(required_values)) > 0
+
+
+def check_if_field_exists_in_dict_list_x_times(field, dict_list, limit=1):
+    return len([1 for item in dict_list if field in item]) == limit
+
+
+def get_query_results(querystring, params={}):
+    res = db.session.execute(querystring, params)
+    return res
+
+
+def check_if_listfield_property_values_are_valid(record, field, prop, checker, errortype='Error'):
+    for i, item in enumerate(record.get(field, [])):
+        if prop in item:
+            for e in checker(field=field, prop=prop, value=item[prop], index=i, errortype=errortype):
+                yield e
+
+
+def query_db_for_duplicates_for_field_listitem_with_index(field, prop, value, index, errortype):
+    queryparams = dict(field=field, prop=prop, value=value)
+    querystring = "SELECT id FROM records_metadata as r, "\
+                  "json_array_elements(r.json -> :field) "\
+                  "as elem WHERE elem ->> :prop=:value"
+    results = get_query_results(querystring, queryparams)
+    if results.rowcount > 1:
+        yield ValidationError(path=deque(['{field}/{index}/{prop}'.format(field=field, index=index, prop=prop)]),
+                              message=FIELD_DUPLICATE_VALUES_FOUND_TEMPLATE.format(
+                                  field=field,
+                                  value=value
+                              ),
+                              cause=errortype)
--- a/inspirehep/modules/records/validators/rules.py
+++ b/inspirehep/modules/records/validators/rules.py
@@ -1,0 +1,150 @@
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+
+from __future__ import absolute_import, division, print_function
+
+from idutils import is_isbn
+from jsonschema import ValidationError
+import pycountry
+
+from .helpers import (
+    check_if_field_exists_in_dict_list_x_times,
+    check_document_type_values_in_required_values_for_record,
+    check_if_listfield_property_values_are_valid,
+    query_db_for_duplicates_for_field_listitem_with_index,
+    FIELD_REQUIRE_FIELD_VALUES_TEMPLATE,
+    FIELD_VALIDATION_TEMPLATE,
+    FIELD_REQUIRE_FIELD_TEMPLATE
+)
+
+
+def check_for_author_or_corporate_author_to_exist(record):
+    if all(k not in record for k in ['authors', 'corporate_author']):
+        yield ValidationError(path=[''],
+                              message='Neither an author nor a corporate author found.',
+                              cause='Error')
+
+
+def check_document_type_if_book_series_exist(record):
+    if 'book_series' in record:
+        required_values = ['book', 'proceedings', 'thesis']
+        if not check_document_type_values_in_required_values_for_record(record, required_values):
+            yield ValidationError(path=[''],
+                                  message=FIELD_REQUIRE_FIELD_VALUES_TEMPLATE.format(
+                                      field='book_series',
+                                      required='document_type',
+                                      values=required_values),
+                                  cause='Error')
+
+
+def check_document_type_if_isbns_exist(record):
+    if 'isbns' in record:
+        required_values = ['book', 'proceedings', 'thesis']
+        if not check_document_type_values_in_required_values_for_record(record, required_values):
+            yield ValidationError(path=[''],
+                                  message=FIELD_REQUIRE_FIELD_VALUES_TEMPLATE.format(
+                                      field='isbns',
+                                      required='document_type',
+                                      values=required_values),
+                                  cause='Error')
+
+
+def check_document_type_if_cnum_exist(record):
+    if check_if_field_exists_in_dict_list_x_times('cnum', record.get('publication_info', [])):
+        required_values = ['proceedings', 'conference paper']
+        if not check_document_type_values_in_required_values_for_record(record, required_values):
+            yield ValidationError(path=[''],
+                                  message=FIELD_REQUIRE_FIELD_VALUES_TEMPLATE.format(
+                                      field='cnum',
+                                      required='document_type',
+                                      values=required_values),
+                                  cause='Error')
+
+
+def check_document_type_if_thesis_info_exist(record):
+    if 'thesis_info' in record:
+        required_values = ['thesis']
+        if not check_document_type_values_in_required_values_for_record(record, required_values):
+            yield ValidationError(path=[''],
+                                  message=FIELD_REQUIRE_FIELD_VALUES_TEMPLATE.format(
+                                      field='thesis_info',
+                                      required='document_type',
+                                      values=required_values),
+                                  cause='Error')
+
+
+def check_if_isbn_is_valid(record):
+    def _check_isbn_is_valid(field, value, prop, index, errortype):
+        if not is_isbn(value):
+            yield ValidationError(path=['{field}/{index}/{prop}'.format(field=field,
+                                                                        prop=prop,
+                                                                        index=index)],
+                                  message=FIELD_VALIDATION_TEMPLATE.format(
+                                      field='isbns',
+                                      value=value
+                                  ),
+                                  cause=errortype)
+
+    for error in check_if_listfield_property_values_are_valid(record=record,
+                                                          field='isbns',
+                                                          prop='value',
+                                                          checker=_check_isbn_is_valid):
+        yield error
+
+
+def check_if_isbn_exist_in_other_records(record):
+    for error in check_if_listfield_property_values_are_valid(record=record,
+                                                          field='isbns',
+                                                          prop='value',
+                                                          checker=query_db_for_duplicates_for_field_listitem_with_index):
+        yield error
+
+
+def check_if_languages_are_valid_iso(record):
+    def _check_language(language, index):
+        try:
+            pycountry.languages.lookup(language)
+        except LookupError:
+            yield ValidationError(path=['languages/{index}'.format(index=index)],
+                                  message=FIELD_VALIDATION_TEMPLATE.format(
+                                      field='languages',
+                                      value=language
+                                  ),
+                                  cause='Error')
+    for i, item in enumerate(record.get('languages', [])):
+        for error in _check_language(item, i):
+            yield error
+
+
+def check_languages_if_title_translations_exist(record):
+    def _check_if_language_exist(translated_title, index):
+        if 'language' not in translated_title:
+            yield ValidationError(path=['title_translations/{index}'.format(index=index)],
+                                  message=FIELD_REQUIRE_FIELD_TEMPLATE.format(
+                                      field='title_translations',
+                                      required='language'
+                                  ),
+                                  cause='Error')
+
+    for i, item in enumerate(record.get('title_translations', [])):
+        for error in _check_if_language_exist(item, i):
+            yield error
--- a/inspirehep/modules/records/validators/validator.py
+++ b/inspirehep/modules/records/validators/validator.py
@@ -1,0 +1,69 @@
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+from jsonschema import Draft4Validator
+from jsonschema.validators import extend, RefResolver
+
+from inspire_schemas.utils import load_schema
+
+from .rules import (
+    check_for_author_or_corporate_author_to_exist,
+    check_if_isbn_exist_in_other_records,
+    check_if_isbn_is_valid,
+    check_document_type_if_book_series_exist,
+    check_document_type_if_isbns_exist,
+    check_document_type_if_cnum_exist,
+    check_document_type_if_thesis_info_exist,
+    check_if_languages_are_valid_iso
+)
+
+LITERATURE_VALIDATORS_LIST = [
+    check_for_author_or_corporate_author_to_exist,
+    check_if_isbn_exist_in_other_records,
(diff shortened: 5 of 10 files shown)


inspirehep/modules/records/validators/helpers.py
```diff
@@ -0,0 +1,77 @@
...
+                                      "at least one of the values {values}."
+FIELD_DUPLICATE_VALUES_FOUND_TEMPLATE = "'{field}' field with value '{value}' found in more than one record."
+GET_RESULTS_FOR_FIELD_PROPERTY_QUERYSTRING_TEMPLATE = "SELECT id FROM records_metadata as r, " \
+                                                      "json_array_elements(r.json -> '{field}') " \
+                                                      "as elem WHERE elem ->> '{prop}'='{value}'"
+FIELD_VALIDATION_TEMPLATE = "Field '{field}' with value '{value}' is not valid."
+
+
+def check_field_values_not_in_required_values_for_record(record, field, required_values):
+    document_type_values = record.get(field, [])
```
jmartinm: This is too specific of document_type. Also assumes that the field will be a list of primitives (no objects), so not much reusable.

inspirehep/modules/records/validators/helpers.py
```diff
@@ -0,0 +1,77 @@
...
+FIELD_REQUIRE_FIELD_VALUES_TEMPLATE = "'{field}' field requires '{required}' field to have " \
+                                      "at least one of the values {values}."
+FIELD_DUPLICATE_VALUES_FOUND_TEMPLATE = "'{field}' field with value '{value}' found in more than one record."
+GET_RESULTS_FOR_FIELD_PROPERTY_QUERYSTRING_TEMPLATE = "SELECT id FROM records_metadata as r, " \
+                                                      "json_array_elements(r.json -> '{field}') " \
+                                                      "as elem WHERE elem ->> '{prop}'='{value}'"
+FIELD_VALIDATION_TEMPLATE = "Field '{field}' with value '{value}' is not valid."
+
+
+def check_field_values_not_in_required_values_for_record(record, field, required_values):
```
jmartinm: To me it makes it confusing that this function looks for 'field values NOT in required values'. I would prefer to have the positive version of the function and the caller do the negation if needed.

inspirehep/modules/records/validators/rules.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+'''
+Error Validator Functions
+'''
+
+
+def check_for_author_or_corporate_author_to_exist(record):
+    if all(k not in record for k in ['authors', 'corporate_author']):
+        yield ValidationError(path=deque(['']),
```
jmartinm: It seems like `jsonschemas` will create the `deque` for you - https://github.com/Julian/jsonschema/blob/master/jsonschema/exceptions.py#L43 so no need to create it here.

inspirehep/modules/records/validators/rules.py
```diff
@@ -0,0 +1,136 @@
...
+                                                          prop='value',
+                                                          checker=_check_isbn_is_valid):
+        yield e
+
+
+def check_if_isbn_exist_in_other_records(record):
+    for e in check_if_listfield_property_values_are_valid(record=record,
+                                                          field='isbns',
+                                                          prop='value',
+                                                          checker=query_db_for_duplicates_for_field_listitem_with_index):
```
jmartinm: There is some weird break of line here

inspirehep/modules/records/validators/rules.py
```diff
@@ -0,0 +1,136 @@
...
+        yield e
+
+
+def check_languages_if_valid_iso(record):
```
jmartinm: `check_if_languages_are_valid_iso`

inspirehep/modules/records/api.py
```diff
@@ -71,6 +71,10 @@ def delete(self):
     def _delete(self, *args, **kwargs):
         super(InspireRecord, self).delete(*args, **kwargs)
 
+    def commit(self, validators=None):
+        """Override commit() to pass custom validators"""
+        super(InspireRecord, self).commit(validators)
```
zzacharo: Yes anyway I will split this functionality to another PR to be more clear.

inspirehep/modules/records/validators/helpers.py
zzacharo: Yes indeed is only for primitve type lists and I used it for document_type only. But is reusable in our code for these cases. Propbably I can change the name to be more specific and use it for document_type field only.

zzacharo:
@jmartinm I updated with the changes in a follow up commit.

zzacharo:
Commit 3a37f8d05e6ef7796a41a9d3c5365a107e5dfb56 is the commit after first review from @jmartinm and the final version. You can check also @kaplun @jacquerie .

michamos:
So in the end, it was not possible to attach the validators to the fields of the record?

zzacharo:
@michamos I couldn't because here https://github.com/Julian/jsonschema/blob/v2.3.0/jsonschema/validators.py#L87 I should have enrich the schema for every field with a pair key: value where the key would be an identifier for the custom validator and the value the callable of the validator function. Moreover, from the way `jsonschema` validates the properties each time for every field it passes the value of the field in the instance to be validated and the subschema of this field. Thus, I don't have the whole instance to make cross validation between the fields. So, this dispatcher function runs the whole bulk of validators in the whole instance as it is attached to the top level of the schema (see the link I attached for the reason again).

jmartinm:
@zzacharo note the small Flake-8 issues in Travis

inspirehep/modules/migrator/tasks/records.py
```diff
@@ -289,7 +289,7 @@ def record_upsert(json):
             pid = PersistentIdentifier.get(pid_type, control_number)
             record = InspireRecord.get_record(pid.object_uuid)
             record.update(json)
-            record.commit()
+            record.commit(validator=None)
```
jmartinm: leftover

inspirehep/modules/records/api.py
```diff
@@ -71,6 +74,15 @@ def delete(self):
     def _delete(self, *args, **kwargs):
         super(InspireRecord, self).delete(*args, **kwargs)
 
+    def commit(self, validator=InspireValidator):
+        """Override commit() to pass custom validators"""
+        # import pytest
+        # pytest.set_trace()
```
jmartinm: leftover

inspirehep/config.py
```diff
@@ -31,6 +31,8 @@
 from invenio_records_rest.facets import range_filter, terms_filter
 
 from inspirehep.modules.records.utils import get_detailed_template_from_record
+from .modules.records.validators.validator import InspireValidator
```
jmartinm: This is not used. Please review the PR and tag as `Needs Review` once it is ready

inspirehep/modules/records/api.py
```diff
@@ -71,6 +74,13 @@ def delete(self):
     def _delete(self, *args, **kwargs):
         super(InspireRecord, self).delete(*args, **kwargs)
 
+    def commit(self, validator=InspireValidator):
```
kaplun: Do we need to pass the `validator` at `commit`-time? Can it be done directly at constructor? Is there a time where we do not want to use `InspireValidator` and hence it's desirable to call `commit(validator=None)`?

inspirehep/modules/records/validators/helpers.py
```diff
@@ -0,0 +1,88 @@
...
+    '_desy_bookkeeping': ['date'],
+    'imprints': ['date'],
+    'thesis_info': ['date', 'defense_date'],
+    '_fft': ['creation_datetime'],
+    'legacy_creation_date': [],
+    'preprint_date': []
+}
+
+
+def check_field_values_not_in_required_values_for_record(record, field, required_values):
```
kaplun: I am not sure I get what this check is validating. Maybe an example in the docstring can help.

inspirehep/modules/records/api.py
zzacharo: A possible case would be on writing some tests and we don't want to have mock records that will validate through these custom rules. Otherwise, it is decided to be used as the default commit method. We can discuss about it.

inspirehep/modules/records/validators/helpers.py
zzacharo: It is a helper function that checks primitive list fields whether they have some values that are required. E.g 
```
document_type=['book','thesis']
required_values = ['book']
```

zzacharo: Also it is used only for document_type field right now so probably I can rename it in `check_document_type_values_not_in_required_values_for_record`

zzacharo:
@jmartinm @kaplun how we proceed with this?

inspirehep/modules/records/validators/__init__.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.ext
```
jacquerie: No it isn't!

inspirehep/modules/records/validators/helpers.py
```diff
@@ -0,0 +1,88 @@
...
+from collections import deque
+
+FIELD_VALUE_REQUIRE_FIELD_TEMPLATE = "'{field}' field value '{value}' requires field '{required}' to exist."
+FIELD_REQUIRE_FIELD_VALUES_TEMPLATE = "'{field}' field requires '{required}' field to have " \
+                                      "at least one of the values {values}."
+FIELD_REQUIRE_FIELD_TEMPLATE = "'{field}' field requires '{required}' field to exist."
+FIELD_DUPLICATE_VALUES_FOUND_TEMPLATE = "'{field}' field with value '{value}' found in more than one record."
+GET_RESULTS_FOR_FIELD_PROPERTY_QUERYSTRING_TEMPLATE = "SELECT id FROM records_metadata as r, " \
+                                                      "json_array_elements(r.json -> '{field}') " \
+                                                      "as elem WHERE elem ->> '{prop}'='{value}'"
```
jacquerie: Why can't this use SQLAlchemy? It supports querying inside of JSON columns.

jacquerie:
> @jmartinm @kaplun how we proceed with this?

Just by glancing at it, this PR requires extensive cleanup of docstrings, imports, licenses, filenames, and tests.

tests/integration/test_permissions.py
```diff
@@ -58,6 +58,7 @@ def _create_and_index_record(record):
 def sample_record(app):
     record = {
         '$schema': 'http://localhost:5000/schemas/records/hep.json',
+        'authors': [],
```
kaplun: Why this empty field? All throughout the code-base we are stripping away empty values (i.e. empty lists, empty objects, empty strings).

inspirehep/modules/records/validators/helpers.py
```diff
@@ -0,0 +1,88 @@
...
+FIELD_VALIDATION_TEMPLATE = "Field '{field}' with value '{value}' is not valid."
+SCHEMA_FIELDS_THAT_CAN_CONTAIN_DATE = {
+    '_desy_bookkeeping': ['date'],
+    'imprints': ['date'],
+    'thesis_info': ['date', 'defense_date'],
+    '_fft': ['creation_datetime'],
+    'legacy_creation_date': [],
+    'preprint_date': []
+}
```
jacquerie: Those are paths, no? Then specify them as paths: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/receivers.py#L313-L320.

tests/integration/test_permissions.py
jacquerie: Obviously this is because this PR adds a check that there must be either an `authors` or a `corporate_author`. But then https://github.com/inspirehep/inspire-schemas/pull/105 will break this test, so can we future-proof this?

kaplun: Yes, but then we need at least one author in the list.

jacquerie: This is exactly what I just wrote.

zzacharo: @kaplun I left it empty because the validator checks if `authors` key is exist or not so value didn't matter. I can apply a dummy one.

inspirehep/modules/records/validators/helpers.py
zzacharo: I am gonna define them as the pointed paths and use `get_value` from `records.utils` .

zzacharo:
@jacquerie I am gonna check docstrings and unused imports again. Can you point me errors in licence (date wrong?). I already deleted the line in your comment. Apart from this, filename rename is something we can discuss after if is something not cleared.

kaplun:
> Can you point me errors in licence (date wrong?).

Everywhere in the license you refer to `Invenio` instead of `INSPIRE` :wink:

inspirehep/modules/records/validators/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+# Copyright (C) 2017 CERN.
+#
+# Invenio is free software; you can redistribute it
```
kaplun: INSPIRE

inspirehep/modules/records/validators/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be
```
kaplun: INSPIRE

inspirehep/modules/records/validators/helpers.py
```diff
@@ -0,0 +1,88 @@
+#
+# This file is part of Invenio.
+# Copyright (C) 2017 CERN.
+#
+# Invenio is free software; you can redistribute it
```
kaplun: INSPIRE... etc.

inspirehep/modules/records/validators/__init__.py
zzacharo: Ok thanks @kaplun ! Wrong copy/paste :P

tests/integration/test_permissions.py
kaplun: > This is exactly what I just wrote.

Not explicitly. So at least it was clear to @zzacharo .

> I can apply a dummy one.

Yes, please do. :+1:

inspirehep/modules/records/validators/__init__.py
kaplun: ahi ahi ahi...

inspirehep/modules/records/validators/helpers.py
```diff
@@ -0,0 +1,88 @@
...
+    '_fft': ['creation_datetime'],
+    'legacy_creation_date': [],
+    'preprint_date': []
+}
+
+
+def check_field_values_not_in_required_values_for_record(record, field, required_values):
+    field_values = record.get(field, [])
+    common_list = [x for x in field_values if x in required_values]
+    return not common_list
```
jacquerie: This is an intersection of sets, so it can be written as one. See: https://github.com/inspirehep/inspire-next/blob/a718afaac615b555ad5f524e781a731cddfc015d/inspirehep/modules/workflows/tasks/matching.py#L231-L235

inspirehep/modules/records/validators/helpers.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def check_if_field_exists_in_dict_list_x_times(field, dict_list, limit=1):
+    return len([1 for item in dict_list if field in item]) == limit
+
+
+def execute_query(querystring):
+    return db.session.execute(querystring)
```
jacquerie: This helper method adds nothing, no?

inspirehep/modules/records/validators/__init__.py
jacquerie: What I can't understand is "from where?". There are no more Python files that have wrong licenses, there's a script in `scripts/prettify_licenses` that does this job automatically... and people still do the wrong thing. How?

inspirehep/modules/records/validators/helpers.py
jacquerie: Also note the refactoring opportunity.

inspirehep/modules/records/validators/rules.py
```diff
@@ -0,0 +1,156 @@
...
+    FIELD_REQUIRE_FIELD_TEMPLATE
+)
+from jsonschema import ValidationError
+from idutils import is_isbn
+import pycountry
+
+
+'''
+Error Validator Functions
+'''
```
jacquerie: Ugly. If you want to have sections you can do something like: https://github.com/inspirehep/inspire-next/blob/a718afaac615b555ad5f524e781a731cddfc015d/inspirehep/modules/theme/views.py#L86-L88.

jacquerie: That said, this file does not even have sections, so what is this lone docstring?

inspirehep/modules/records/validators/rules.py
```diff
@@ -0,0 +1,156 @@
...
+        except LookupError:
+            yield ValidationError(path=['languages/{index}'.format(index=index)],
+                                  message=FIELD_VALIDATION_TEMPLATE.format(
+                                      field='languages',
+                                      value=language
+                                  ),
+                                  cause='Error')
+    for i, item in enumerate(record.get('languages', [])):
+        for e in _check_language(item, i):
+            yield e
```
jacquerie: This is just an enum lookup in a slowly varying standard. It should be implemented by the schema.

inspirehep/modules/records/validators/helpers.py
zzacharo: just exposes the query execution rather than explicitly call it every time. Do you think I should omit it?

inspirehep/modules/records/validators/helpers.py
zzacharo: exactly! I will replace it!

inspirehep/modules/records/validators/helpers.py
jacquerie: Explicit is better than implicit, so...

inspirehep/modules/records/validators/helpers.py
zzacharo: @jacquerie do we need to check for `_desy_bookkeeping.date` and `_fft.creation_datetime` ?

zzacharo: Actually this will be moved to the PR which will introduce the rules specific for record editor on top of this PR. So, I will remove the code from here and discuss it in the future.

inspirehep/modules/records/validators/__init__.py
zzacharo: I don't know really! I fixed it anyway! PS. I didn't know about the script I will run it next time for the licences stuff!

inspirehep/modules/records/validators/helpers.py
jacquerie: >@jacquerie do we need to check for `_desy_bookkeeping.date` and `_fft.creation_datetime` ?

In your validator? I would say so, yes. Why have an invalid date?

inspirehep/modules/records/validators/rules.py
jacquerie: Although I have a feeling that we already discussed this. @jmartinm, do you remember?

inspirehep/modules/records/validators/rules.py
zzacharo: leftover from when it was all combined (errors and warnings)..

inspirehep/modules/records/validators/rules.py
zzacharo: @jacquerie What do you mean with that? Instead of using enumerate transparently every iteration to a schema array to return item and index?

jacquerie: I was referring to the entire rule: it can be implemented as an `enum` of all the possible ISO codes of languages, which will be validated by the `jsonschema` library instead that with your extension.

zzacharo: ohh ok... it's gonna be a big enum though. I think is about 2 hundred something valid iso languages...Is it ok @jmartinm @jacquerie ?

jacquerie: I don't see the problem!

zzacharo: ok if this is the case I could omit it from here and create an issue in `inspire-schemas` ?

jacquerie: I would first like confirmation from @jmartinm... I really feel that we discussed this, maybe there was a reason why it was done like this.

zzacharo:
@jacquerie regarding imports you mentioned something? In general do we follow a plan of how we organize imports? Could you share it with me?

jacquerie:
> @jacquerie regarding imports you mentioned something? In general do we follow a plan of how we organize imports? Could you share it with me?

It's a very minor thing! Typically you have:
1. Future imports
2. Vendored imports from outside Invenio
3. Vendored imports from Invenio
4. INSPIRE imports inter modula
5. INSPIRE imports intra modula

Nobody dies if you don't do this, eh, it just helps keeping things tidy.

zzacharo:
@jacquerie thanks! What is the difference between `Vendored imports from outside Invenio`
and `Vendored imports from Invenio` ?

jacquerie:
`pytest` is an import from outside Invenio, `invenio-accounts` is an import from Invenio.

zzacharo:
not-Invenio-3rd-party libraries and Invenio-related-one right? I am gonna commit the changes to different commit and check again! Is there any other issue not mentioned? For tests possibly?

inspirehep/modules/records/validators/helpers.py
zzacharo: I searched a bit how to do this and I saw the use of `astext` operator. But in our version(< 1.1.0) this operator isn't included I think. Can you confirm that or any other way that I could use?

inspirehep/modules/records/validators/rules.py
jmartinm: No reason for not having it in the schema as far as I remember.

The discussion was about what languages to present to the user in a dropdown in the user submission, so not related to this.

Then I guess we can create an issue in inspire-schemas. @michamos will have more info once he is back since I think he considered the option of adding it in the schema and maybe discarded it for some reason.

zzacharo: @jmartinm @jacquerie So should I omitted for now and create the issue?

jmartinm: > @jmartinm @jacquerie So should I omitted for now and create the issue?

We can also leave it and then remove it once (if it happens) the enum is added to the schema. No strong opinion here, @jacquerie ?

jacquerie: Better to have it than not having it... so we can see the invalid records sooner.

inspirehep/modules/records/validators/helpers.py
jacquerie: You're right that we currently can't, but you're not right on the reason why. The problem is that the `json` column in `records_metadata` is implemented as a `JSONType` from `sqlalchemy_utils`. This type doesn't support `__getitem__`, so you can't do queries that look inside of the record for certain values.

In fact, SQLAlchemy 1.0 **has** a JSON type that supports `__getitem__`: https://docs.sqlalchemy.org/en/rel_1_0/dialects/postgresql.html?highlight=json#sqlalchemy.dialects.postgresql.JSON. As you see, this guy comes from the PostgreSQL dialect, which is why it was not used upstream in `invenio_records`: they are also supporting MySQL, but at the time there was no DB-agnostic implementation, so they used the one provided by `sqlalchemy_utils`.

zzacharo: @jacquerie I see, so `astext` was supported by Postgres JsonType but our JsonType was in fact an sqlalchemy_utils type which didn't support lookup inside the data right?

tests/unit/records/test_hep_records_inspire_validators.py
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    expected = {
+        'globalErrors': [{
+            'message': 'Neither an author nor a corporate author found.',
+            'type': 'Error'
+        }]
+    }
+    result = validate(check_for_author_or_corporate_author_to_exist, sample_record)
+    assert result == expected
```
jacquerie: True, but not terribly important: the main point of having the two variables is that, when you have an assertion failure, you can invoke `py.test` with the `--pdb` flag and, when the assertion fails, you have a variable in the scope holding the expected value and one holding the actual result.

inspirehep/modules/records/validators/helpers.py
```diff
@@ -0,0 +1,72 @@
...
+def check_if_listfield_property_values_are_valid(record, field, prop, checker, errortype='Error'):
+    for i, item in enumerate(record.get(field, [])):
+        if prop in item:
+            for e in checker(field=field, prop=prop, value=item[prop], index=i, errortype=errortype):
+                yield e
+
+
+def query_db_for_duplicates_for_field_listitem_with_index(field, prop, value, index, errortype):
+    querystring = GET_RESULTS_FOR_FIELD_PROPERTY_QUERYSTRING_TEMPLATE.format(field=field, prop=prop, value=value)
+    results = get_query_results_if_exceed_limit(querystring, 1)
```
zzacharo: @jmartinm the `get_query_results_if_exceed_limit` will be used to the rules for the editor that will be applied on top of this commit after. There there is also a limit=0. I will split them to two different functions with better naming. For the template you are right I am gonna inline it.

inspirehep/modules/records/validators/helpers.py
```diff
@@ -0,0 +1,72 @@
...
+    if results:
+        yield ValidationError(path=deque(['{field}/{index}/value'.format(field=field, index=index)]),
```
zzacharo: leftover

tests/integration/test_hep_records_inspire_validators.py
```diff
@@ -0,0 +1,133 @@
...
+            {'value': '11111'}
+        ],
+        'titles': [
+            {'title': 'sample'},
+        ],
+        'self': {
+            '$ref': 'http://localhost:5000/schemas/records/hep.json',
+        }
+    }
+    dupl_sample_record = {
```
zzacharo: Because I am checking about if isbn exist in more than 1 records in database.

inspirehep/modules/records/validators/validator.py
```diff
@@ -0,0 +1,68 @@
...
+        for error in validator(instance):
+            yield error
+
+
+class InspireResolver(RefResolver):
+
+    def resolve_remote(self, uri):
+        """Resolve a uri or relative path to a schema."""
+        resolved = load_schema(uri.rsplit('.json', 1)[0])
+        if 'hep.json' in uri:
```
zzacharo: `load_schema` resolves the schema regardless if this is a `hep` schema. The check here is for applying the specific validator. In the future depending on the uri maybe we will have to enrich the resolved schema with other validators. So, I think it should be before the if statement. Otherwise, it will be quite specific although it will make sense for now.

inspirehep/modules/records/validators/helpers.py
jmartinm: Ok

zzacharo:
@jmartinm I am gonna push a new review commit to check it after the changes of your review ok?

inspirehep/modules/records/validators/validator.py
jmartinm: The thing is that you call it `inspire_validator` and above `inspire_validator_dispatcher` but in reality it is only applicable for HEP records.

So either we have this `inspire_validator` and in `inspire_validator_dispatcher` you add the `if` statements to check if it is `hep.json`, `authors.json` and so on, or you can also rename `inspire_validator` to `inspire_hep_validator`

tests/integration/test_hep_records_inspire_validators.py
jmartinm: Well ISBNs should be unique in the DB right? So if the fixture has one record with the ISBN and in the test you try to add a second one with the same ISBN, isn't that enough?

inspirehep/modules/records/validators/helpers.py
jmartinm: Now you don't use this anymore. To be deleted.

inspirehep/modules/records/validators/rules.py
michamos: I created the language enum in inspirehep/inspire-schemas#173.

zzacharo: ok so no need to check it here?

PR #2313, opened 2017-05-05
tests: unit tests for workflows/tasks/submission

jacquerie:
Addresses #1773 

- [x] `add_note_entry`
- [x] `close_ticket`
- [ ] `create_ticket`
- [x] `filter_keywords`
- [x] `prepare_files`
- [x] `prepare_keywords`
- [x] `remove_references`
- [x] `reply_ticket`
- [ ] `submit_rt_ticket`
- [x] `send_robotupload`
- [x] `wait_webcoll`

### DIFF
--- a/inspirehep/modules/workflows/tasks/submission.py
+++ b/inspirehep/modules/workflows/tasks/submission.py
@@ -34,9 +34,12 @@ from retrying import retry
 
 from invenio_accounts.models import User
 
-from ..utils import with_debug_logging
-from ....utils.tickets import get_instance, retry_if_connection_problems
+from inspirehep.dojson.utils import legacy_export_as_marc
+from inspirehep.utils.robotupload import make_robotupload_marcxml
+from inspirehep.utils.tickets import get_instance, retry_if_connection_problems
+
 from .actions import in_production_mode, is_arxiv_paper
+from ..utils import with_debug_logging
 
 
 LOGGER = logging.getLogger(__name__)
@@ -131,8 +134,6 @@ def reply_ticket(template=None,
     @with_debug_logging
     @wraps(reply_ticket)
     def _reply_ticket(obj, eng):
-        from inspirehep.utils.tickets import get_instance
-
         ticket_id = obj.extra_data.get("ticket_id", "")
         if not ticket_id:
             obj.log.error("No ticket ID found!")
@@ -194,8 +195,6 @@ def close_ticket(ticket_id_key="ticket_id"):
     @with_debug_logging
     @wraps(close_ticket)
     def _close_ticket(obj, eng):
-        from inspirehep.utils.tickets import get_instance
-
         ticket_id = obj.extra_data.get(ticket_id_key, "")
         if not ticket_id:
             obj.log.error("No ticket ID found!")
@@ -241,9 +240,6 @@ def send_robotupload(
     @with_debug_logging
     @wraps(send_robotupload)
     def _send_robotupload(obj, eng):
-        from inspirehep.dojson.utils import legacy_export_as_marc
-        from inspirehep.utils.robotupload import make_robotupload_marcxml
-
         combined_callback_url = ''
         if callback_url:
             combined_callback_url = os.path.join(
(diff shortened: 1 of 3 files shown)


jacquerie:
Bugs uncovered while writing these tests:

* Most log calls use `format` instead of what is recommended in #1371.
* Most log calls are not `unicode`-safe. Some of these bugs were already solved in #2199.
* It's unsafe to call `strip` on something that might be `None`: https://github.com/inspirehep/inspire-next/blob/5fdbd8bf4ac6bd984406719b6a3d0c91b0b4cf77/inspirehep/modules/workflows/tasks/submission.py#L158
* It's highly probable that the exception in https://github.com/inspirehep/inspire-next/blob/5fdbd8bf4ac6bd984406719b6a3d0c91b0b4cf77/inspirehep/modules/workflows/tasks/submission.py#L222-L223 is impossible to trigger.

jacquerie:
Design problems encountered while writing these tests:

* This whole thing reeks of spaghetti code: relying on some task writing a certain thing in `extra_data` so that some task at some later point will pick it up is error-prone and is guaranteed to break in the future when tasks are added/deleted/reordered.
* These function have too many responsibilities, which results in too many mocks that have to be applied to them while testing. For example [`reply_ticket`](https://github.com/inspirehep/inspire-next/blob/5fdbd8bf4ac6bd984406719b6a3d0c91b0b4cf77/inspirehep/modules/workflows/tasks/submission.py#L127-L189) contains:
  - some logic to get the `ticket_id` on which it should operate;
  - some logic to figure out if the user actually exists;
  - some logic (with fallbacks) to build the ticket body;
  - some logic to fetch an RT instance;
  - finally, some logic to reply to a ticket.

  Note that all these items also have fallback logic to log and bail out if any of these steps fail, effectively duplicating the number of responsibilities of this function.

kaplun:
> relying on some task writing a certain thing in extra_data so that some task at some later point will pick it up is error-prone and is guaranteed to break in the future when tasks are added/deleted/reordered.

Given the workflow engine we are employing, what better pattern would you propose?

jacquerie:
> Given the workflow engine we are employing, what better pattern would you propose?

Enforce dependencies in code. If task A logically requires task B to be run before it (for example: B is a machine learning algorithm, A persists its results in the record) this must be enforced in code. How? Strong conventions on naming things, ban the use of patterns like `extra_data.get(...)`.

kaplun:
> Strong conventions on naming things, ban the use of patterns like extra_data.get(...).

Actually that makes me think that the object passed around doesn't need to be really a dictionary, so we could pass an actual object with dedicated properties/methods that can then enforce the conventions.

Only issue would be serialization in the DB. But shouldn't be impossible, to just serialize as dictionary and reinstantiate back as full-fledged object.

kaplun:
Or more simply, indeed we can have setters and getters that receive the `obj` as first arguments, allowing for still keeping the `obj` as a simple dict, but hiding the complexity of writing/accessing special  *extra* keys...

david-caro:
> ban the use of patterns like extra_data.get(...)

I'm curious, why do you think that doing that (banning the usage of .get) is enforcing in code dependencies between tasks?

jacquerie:
> I'm curious, why do you think that doing that (banning the usage of .get) is enforcing in code dependencies between tasks?

To avoid failing silently. The problem is in patterns like
```python
foo = extra_data.get('foo')
if foo:
    ...
```
because this is *not* enforcing a dependency on `foo`, so if the task that populates it starts failing tomorrow we are never going to see it. Things should fail loudly and clearly.

david-caro:
That can be easily avoided with `extra_data['foo']` :)

jacquerie:
That was my point!

david-caro:
Sorry, I missed that point XS

jacquerie:
> * It's unsafe to call `strip` on something that might be `None`: https://github.com/inspirehep/inspire-next/blob/5fdbd8bf4ac6bd984406719b6a3d0c91b0b4cf77/inspirehep/modules/workflows/tasks/submission.py#L158

...and in fact: https://sentry.cern.ch/inspire-sentry/inspire-labs/group/822209/

jacquerie:
Victim of https://github.com/inspirehep/inspire-next/issues/2385.

PR #2331, opened 2017-05-11
tests: add test of workflow completion

jacquerie:
Closes #1939


jacquerie:
Merged because it provides *some* value, but long XPath expressions are brittle and difficult to maintain, and testing the state of the database from the UI misses the point of what **is** an acceptance test. Sure, we want to use the browser to simulate the users' interaction with the website, but does it even make sense to query the applications' state from the UI?

david-caro:
I agree that acceptance tests should be though as black box tests (that is, agnostic of the internals of the app). About the xpaths, aside from putting them in a nicely named variable, until we have beautiful ids on the ui, not sure much else can be done.

Issue #1987, opened 2017-03-07
Speed up acceptance tests


jmartinm:
Acceptance tests are the slowest part of our testing matrix (17min vs ~10 min for the next slowest one).

Some suggestions to make them faster:

* [ ] **Aggregate several assertions in the same test**. For example for [author form tests](https://github.com/inspirehep/inspire-next/blob/master/tests/acceptance/test_author_creation.py) we have many small tests that each have to open the browser/login/go to the form page and then they just test one input field. Would be better to aggregate all the assertions regarding the form in one test.

* [ ] **Avoid repetition**. For example for [author form tests](https://github.com/inspirehep/inspire-next/blob/master/tests/acceptance/test_author_creation.py) `test_submit_author`, `test_accept_author`, `test_reject_author`, `test_curation_author` and `test_review_submission_author` they all do exactly the same long process of (go to main page/login/go to form/fill form/go and wait in holding pen/access detail holding pen and make sure values are there). Since the input is always the same, checking once is enough.

* [ ] **Make sure tests run fast**. For some reason tests like [this one](https://github.com/inspirehep/inspire-next/blob/master/tests/acceptance/test_author_creation.py#L74) or [this one](https://github.com/inspirehep/inspire-next/blob/master/tests/acceptance/test_author_creation.py#L86) get stuck and take long to complete (see example below of `test_ORCID_format`. Other tests like the autocomplete test execute fast.
![slow-test](https://cloud.githubusercontent.com/assets/580727/23661905/de30402a-0345-11e7-941d-cd5c4793c611.gif)

jmartinm:
CC @rikirenz

rikirenz:
it would be a pleasure improve acceptance test!

jmartinm:
> It would be a pleasure improve acceptance test! 😄

Nice! Since @Koursaros93 is going to be working on front-end tasks, this would be a good way to get familiar with acceptance tests and get more training. So once he is free you can both work on it.

rikirenz:
@jacquerie Could you explain which strategies you would like to implement to reduce the time?

rikirenz:
Rather than guess where and how we should improve acceptance I have create some statistics in order to understand which are the most time consuming functions in acceptance:

|ncalls|tottime|percall|cumtime|percall| filename:lineno(function)|
| --- | --- | --- | --- | --- | --- |
|54|0|0|59.68|1.105| /code/tests/acceptance/conftest.py:94(login)|
[...]

PR #2332, opened 2017-05-11
tests.acceptance: login once per session

david-caro:
Addresses #1987

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,7 @@ nosetests.xml
 coverage.xml
 *,cover
 .hypothesis/
+tests/acceptance/login_cookies.pkl
 
 # Translations
 *.mo
--- a/tests/acceptance/conftest.py
+++ b/tests/acceptance/conftest.py
@@ -24,6 +24,8 @@
 
 from __future__ import absolute_import, division, print_function
 
+import os
+import pickle
 from time import sleep
 
 import pytest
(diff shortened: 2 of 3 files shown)


david-caro:
It seems to cut out ~40 secs...

jacquerie:
> It seems to cut out ~40 secs...

Sounds promising, but the duration of this test suite has variance which is much higher than that, so one sample does not really prove it...

david-caro:
That's where the "seems" comes from ;)


--

CERN - RCS-SIS
http://inspirehep.net
On May 11, 2017 22:31, Jacopo Notarstefano <notifications@github.com> wrote:

It seems to cut out ~40 secs...

Sounds promising, but the duration of this test suite has variance which is much higher than that, so one sample does not really prove it...

tests/acceptance/conftest.py
```diff
@@ -74,14 +76,33 @@ def app(request):
...
+    if os.path.exists(cookies_path):
+        os.unlink(cookies_path)
 
 
 @pytest.fixture
-def login(arsenic):
-    top_navigation_page.log_in('admin@inspirehep.net', '123456')
-    yield
-    top_navigation_page.log_out()
+def login(arsenic, app):
```
jacquerie: The `arsenic` fixture already depends on `app`, so this shouldn't be needed.

tests/acceptance/conftest.py
```diff
@@ -74,14 +76,33 @@ def app(request):
...
+    cookies_path = './tests/acceptance/login_cookies.pkl'
+    if not os.path.exists(cookies_path):
+        top_navigation_page.log_in('admin@inspirehep.net', '123456')
+        sleep(1)
```
jacquerie: Is this needed, or is it just to be safe?

tests/acceptance/conftest.py
```diff
@@ -74,14 +76,33 @@ def app(request):
 
 @pytest.fixture
 def arsenic(selenium, app):
-    Arsenic(selenium)
+    return Arsenic(selenium)
```
jacquerie: Oh, so this is why we have `Arsenic()` everywhere?

.gitignore
```diff
@@ -67,6 +67,7 @@ nosetests.xml
 coverage.xml
 *,cover
 .hypothesis/
+tests/acceptance/login_cookies.pkl
```
jacquerie: This is not really a coverage report, so maybe we should add an extra section for `# Application specific ignores` or something like that.

PR #2338, opened 2017-05-15
config: amend ARXIV_PDF_URL

### LINKED ISSUE #2337: Crash in holding pen after deployment
After deploying of #2283 this error appered:
https://sentry.cern.ch/inspire-sentry/inspire-labs/group/822026/

kaplun:
* Amends ARXIV_PDF_URL default to actually use export.arxiv.org
  hostname, which is the one authorizing us to fetch PDFs in a
  machine-friendly way.
  (closes #2337)

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -1368,7 +1368,7 @@ MAGPIE_API_URL = "http://magpie.inspirehep.net/api"
 
 # Harvesting and Workflows
 # ========================
-ARXIV_PDF_URL = "http://arxiv.org/pdf/{arxiv_id}"
+ARXIV_PDF_URL = "http://export.arxiv.org/pdf/{arxiv_id}"
 ARXIV_TARBALL_URL = "http://arxiv.org/e-print/{arxiv_id}"
 
 ARXIV_CATEGORIES_ALREADY_HARVESTED_ON_LEGACY = [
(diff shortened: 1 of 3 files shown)


tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -155,6 +155,28 @@ def already_harvested_on_legacy_record():
     return json_data
 
 
+@pytest.fixture
+def to_automatically_reject_record():
```
kaplun: Shit
This is from a different PR.

jacquerie:
I'm skeptical this is going to solve the issue. Fetching multiple PDFs from the naked arXiv domain doesn't appear to be an issue:
```python
>>> for i in range(10, 100):
...     response = get(url='https://arxiv.org/pdf/1703.070' + str(i) + '.pdf', stream=True)
...     print len(response.content)
```
never appears to return 0.

david-caro:
> never appears to return 0.

It might well possibly be that we are exceeding the rate at we can download pdfs

jacquerie:
> It might well possibly be that we are exceeding the rate at which we can download pdfs

In the past hour I downloaded 900 PDFs without timeouts (sorry arXiv), none failed...

inspirehep/config.py
```diff
@@ -1368,7 +1368,7 @@ def _(x):
 
 # Harvesting and Workflows
 # ========================
-ARXIV_PDF_URL = "http://arxiv.org/pdf/{arxiv_id}"
+ARXIV_PDF_URL = "http://export.arxiv.org/pdf/{arxiv_id}"
 ARXIV_TARBALL_URL = "http://arxiv.org/e-print/{arxiv_id}"
```
jacquerie: What about this guy?

jacquerie:
I'm still going to merge this because this is anyway the right thing to do, and because I'm curious to see if this actually fixes the bug tonight...

PR #2339, opened 2017-05-15
config: use export.arxiv.org instead of arxiv.org

jacquerie:
Closes #2338 

We have a verbal agreement with arXiv that we should use the
former instead of the latter to avoid throttling.


tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -270,7 +270,7 @@ def test_arxiv_plot_extract_logs_when_tarball_is_invalid(mock_process_tarball):
 
     assert arxiv_plot_extract(obj, eng) is None
 
-    expected = 'Invalid tarball https://arxiv.org/e-print/1612.00626'
+    expected = 'Invalid tarball http://export.arxiv.org/e-print/1612.00626'
```
kaplun: I am not sure if the same is for .tar.gz. It very likely is.

Issue #2328, opened 2017-05-10
invenio-classifier generates non-indexable fields


michamos:
e.g. https://labs.inspirehep.net/api/holdingpen/632571 contains `interfaces.` (note the extra `.` at the end) among `#/_extra_data/classifier_results/complete_output` which makes ElasticSearch fail (as explained by @jacquerie ).
This happens despite `interfaces.` not being present in the HEPont ontology.

@ksachs @kaplun, maybe you have some idea of how this could arise?

ksachs:
this is from author-keywords.
"Author keywords": {
          "interfaces.": [
            "interface"
          ]
        }, 

In the PDF you have: 
Keywords: Dyson model, entropic repulsion, g-measure, interfaces.
I.e. you need to strip punctuation when processing the pdf for keywords.

kaplun:
BTW With @michamos we were thinking that, since the actual problems is that ES is not able to accept objects having properties with dots, and since we can't really exclude the presence of these dots everywhere (it can be in principle in the middle of a keyword, who knows?) and since it's actually a wrong structure to have them as properties (i.e. you are never going to search for `Author keywords.interfaces=interface`), the real solution, if we really want to have this information preserved is to change the structure into:
```json
"Author keywords": [
{"author_keyword": "interfaces.",
 "matched_keywords": ["interface"]
}
]
```

Then we can also strip trailing dots. Sure this is going to be helpful.

ksachs:
If you change the schema/structure please do a complete overhaul. 
I think currently I have to search for a keyword like

    _extra_data.classifier_results.categories.background:"HEP"

kaplun:
Yeah, this is at a second stage in Elasticsearch. Indeed we should simplify these keywords... invenio-classifiers is doing so many complex things that my head is bound to explode.

kaplun:
I have just release `invenio-classifier-1.1.0`. We shall just upgrade the code base to it.

jacquerie:
> I have just release `invenio-classifier-1.1.0`. We shall just upgrade the code base to it.

Err... will this break the build in the meantime unless we pin `invenio-classifier`?

jacquerie:
> Err... will this break the build in the meantime unless we pin `invenio-classifier`?

Yup: https://travis-ci.org/inspirehep/inspire-next/jobs/233248122#L2174

Is it a quick fix on `inspire-next` or is it going to take you more than the rest of today?

kaplun:
[...]

PR #2343, opened 2017-05-17
setup: pin invenio-classifier to 1.0.1

jacquerie:
See: https://github.com/inspirehep/inspire-next/issues/2328#issuecomment-302101126


jacquerie:
Obviously temporary; the idea is that tomorrow we're going to sit together to adapt `inspire-next` to the new output of `invenio-classifier`. It would also be a nice time to tackle the relevant part of #1773.

kaplun:
Sorry missed this comment before sending my PR.

PR #2349, opened 2017-05-18
workflows: retry on error when fetching files

jacquerie:
Addresses #2337 

~~Still missing tests~~, but I'd like your feedback, @david-caro @jmartinm @kaplun.

### DIFF
--- a/inspirehep/modules/workflows/utils.py
+++ b/inspirehep/modules/workflows/utils.py
@@ -30,7 +30,9 @@ import traceback
 from contextlib import closing
 from functools import wraps
 
+import backoff
 import requests
+import urllib3
 from flask import current_app
 
 from .models import WorkflowsAudit
@@ -154,6 +156,7 @@ def get_pdf_in_workflow(obj):
     obj.log.info('No PDF available')
 
 
+@backoff.on_exception(backoff.expo, urllib3.exceptions.ProtocolError, max_tries=5)
 def download_file_to_workflow(workflow, name, url):
     """Download a file to a specified workflow.
 
@@ -164,6 +167,10 @@ def download_file_to_workflow(workflow, name, url):
     through ``ObjectVersion`` and ``FileObject``. This method consumes the
     stream passed to it and saves in its place a ``FileObject`` with the
     details of the downloaded file.
+
+    Consuming the stream might raise a ``ProtocolError`` because the server
+    might terminate the connection before sending any data. In this case we
+    retry 5 times with exponential backoff before giving up.
     """
     with closing(requests.get(url=url, stream=True)) as req:
         if req.status_code == 200:
--- a/tests/unit/workflows/test_workflows_utils.py
+++ b/tests/unit/workflows/test_workflows_utils.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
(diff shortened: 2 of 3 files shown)


inspirehep/modules/workflows/utils.py
```diff
@@ -164,8 +168,24 @@ def download_file_to_workflow(workflow, name, url):
...
+                workflow.files[name] = response.raw
+                return workflow.files[name]
+            except urllib3.exceptions.ProtocolError:
+                response.close()
+                sleep(2 ** retries)
+                response = requests.get(url, stream=True)
+
+            retries += 1
+
+    raise WorkflowsTaskException()
```
jacquerie: This needs an exception message that mentions that there was a problem downloading from `url`, but I'm not sure how to write one that Sentry is able to cluster.

jacquerie:
Failures were expected, because now I'm raising an exception that must be handled by the tasks using this helper.

inspirehep/modules/workflows/utils.py
```diff
@@ -164,8 +168,24 @@ def download_file_to_workflow(workflow, name, url):
...
+            try:
+                workflow.files[name] = response.raw
+                return workflow.files[name]
+            except urllib3.exceptions.ProtocolError:
+                response.close()
+                sleep(2 ** retries)
+                response = requests.get(url, stream=True)
+
+            retries += 1
+
```
jacquerie: Uh, I should close the first `response` here.

inspirehep/modules/workflows/utils.py
```diff
@@ -164,8 +168,24 @@ def download_file_to_workflow(workflow, name, url):
...
+    if response.ok:
+        retries = 0
+
+        while retries < MAX_RETRIES:
+            try:
+                workflow.files[name] = response.raw
+                return workflow.files[name]
+            except urllib3.exceptions.ProtocolError:
+                response.close()
+                sleep(2 ** retries)
```
jacquerie: Here I should add a random jitter to prevent a [Thundering Herd problem](https://en.wikipedia.org/wiki/Thundering_herd_problem).

jacquerie:
[`backoff`](https://pypi.python.org/pypi/backoff/1.4.2) sounds like a nice solution for this problem.

david-caro:
Look ok 👍

tests/unit/workflows/test_workflows_utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def test_download_file_to_workflow_retries_on_protocol_error():
+    with requests_mock.Mocker() as requests_mocker:
+        filename = pkg_resources.resource_filename(
+            __name__, os.path.join('fixtures', '1605.03844.pdf'))
+
+        requests_mocker.register_uri(
+            'GET', 'http://export.arxiv.org/pdf/1605.03844', [
+                {'exc': urllib3.exceptions.ProtocolError},
```
jacquerie: I previously wrote this test using `httpretty`, and I managed to hit the [same bug](https://github.com/gabrielfalcao/HTTPretty/issues/65#issuecomment-278715136) that @david-caro reported...

kaplun:
So: when the 5 retries are exhausted the client code will fail with exception, right? In this case, the whole workflow will fail at the given task and can be manually restarted later. Correct?

jacquerie:
> So: when the 5 retries are exhausted the client code will fail with exception, right? In this case, the whole workflow will fail at the given task and can be manually restarted later. Correct?

Right. You can use the test to check this: you put there 6 `ProtocolError`s and you see the last one bubble up and fail with the same error as before.

kaplun:
The next level in reliability will be to introduce into all our externally-dependent tasks this:
https://pypi.python.org/pypi/circuitbreaker/1.0.1

In this way when external service is down we can simply avoid bombarding and failing.

jacquerie:
I'd say that the next step is checking that this PR is all we need for #2337 : )

We might need to tweak the `max_retries`, or maybe I'm completely wrong...

PR #2353, opened 2017-05-18
authors: two date validation

### LINKED ISSUE #2092: author submission: you can enter invalid date ranges
For example when entering the experiment history:

screenshot from 2017-03-21 21-22-22

jmartinm:
* New js and a python validator for the author dates input. (Closes #2092)

Signed-off-by: Glignos <glignos93@gmail.com>
Co-Authored-By: Zacharias Zacharodimos <zacharias.zacharodimos@cern.ch>
Co-Authored-By: Javier Martin Montull <javier.martin.montull@cern.ch>

### DIFF
--- a/inspirehep/modules/authors/static/js/authors/authors_submission_form.js
+++ b/inspirehep/modules/authors/static/js/authors/authors_submission_form.js
@@ -1,0 +1,74 @@
+define(function (require, exports, module) {
+
+  "use strict";
+
+  var $ = require("jquery");
+  var tpl_field_message = require('hgn!js/forms/templates/field_message');
+
+  function AuthorUpdateForm() {
+
+    this.$inspireForm = $('#authorUpdateForm');
+    this.init();
+  };
+
+  AuthorUpdateForm.prototype = {
+    init: function init() {
+      var that = this;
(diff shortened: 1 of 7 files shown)


inspirehep/modules/authors/static/js/authors/authors_submission_form.js
```diff
@@ -0,0 +1,74 @@
...
+      }
+
+      var isNumber =  /^\d{4}$/.test(elemValue);
+      var isOtherNumber =  /^\d{4}$/.test(otherFieldValue);
+
+      if ( !isNumber || !isOtherNumber ) {
+        return;
+      }
+
+      if ( !compareValues ) {
```
zzacharo: I think you can just make the compare `elemValue >= otherFieldValue` here and omit the `compareValues` variable. The if...else blocks just set the elemValue, otherFieldValue.

jmartinm: oh I see, the problem is that there was a mistake in the code so I had two times the same:

```
elemValue >= otherFieldValue;
```

I have now amended it so that is correct and the comparison is different depending if you are in `start_year` or in `end_year`

Issue #2222, opened 2017-04-11
tests: investigate flakiness of test_submit_author


jacquerie:
For example: https://travis-ci.org/inspirehep/inspire-next/jobs/221069173

jacquerie:
@rikirenz I remember that we discussed this, but I don't remember what you told me about it... mind writing it down here?

rikirenz:
I do not remember about this. I am going to investigate. it should be easy.

jacquerie:
Not a `Type: Bug`, as this is not affecting users, just developers.

kaplun:
Is it just a speed issue? I.e. we are too quick in loading the page and the record is not yet there in the right state?

rikirenz:
Usually is always a time/speed issue with selenium: https://sqa.stackexchange.com/questions/14241/how-to-avoid-flakey-e2e-selenium-tests?answertab=votes#tab-top
 
From the moment that we are going to migrate all the tests in BDD. I do not think it makes a lot of sense try to figure out why that specific test is flakey. But if you think is important I will investigate.

jacquerie:
Should have been fixed by #2502.

PR #2357, opened 2017-05-19
workflows: catch the right exception

jacquerie:
Addresses #2337 

The `requests` library rewraps all `urllib3` exceptions with their own classes.


jacquerie:
Victim of #2222

kaplun:
Damn Python and it's runtime type inference.

PR #2358, opened 2017-05-19
holdingpen: sort & filter by core-related scores

### LINKED ISSUE #2113: holdingpen: sorting by core-guessing relevance is missing
@ksachs reports that it is no longer possible to sort by core guessing.

@ksachs can you detail, the exact specs of the sorting algorithm you were mentioning me?

chris-asl:
* Add sorting by core (desc), non-core (desc), rejected scores (asc).
* Add filtering by decision (core/non-core/rejected).

Closes #2113.

Co-authored-by: George Lignos <glignos93@gmail.com>

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -1457,6 +1457,7 @@ WORKFLOWS_UI_REST_FACETS = {
             "workflow_name": terms_filter('_workflow.workflow_name'),
             "is-update": terms_filter('_extra_data.is-update'),
             'subject': terms_filter('metadata.inspire_categories.term'),
+            'decision': terms_filter('_extra_data.relevance_prediction.decision'),
         },
         "aggs": {
             "status": {
@@ -1483,6 +1484,12 @@ WORKFLOWS_UI_REST_FACETS = {
                     'size': 20,
                 },
             },
+            'decision': {
+                'terms': {
+                    'field': '_extra_data.relevance_prediction.decision',
+                    'size': 20
+                }
+            }
         }
     }
 }
@@ -1501,6 +1508,32 @@ WORKFLOWS_UI_REST_SORT_OPTIONS = {
             "default_order": 'desc',
             "order": 2,
         },
+        'core': {
+            'title': 'Relevance Prediction (Desc)',
+            'fields': [
+                {
+                    '_extra_data.relevance_prediction.relevance_score': {
+                        'missing': '_last',
+                        'order': 'desc',
+                        'unmapped_type': 'float'
+                    }
+                }
+            ],
+            'order': 3,
+        },
+        'rejected': {
+            'title': 'Relevance Prediction (Asc)',
+            'fields': [
+                {
+                    '_extra_data.relevance_prediction.relevance_score': {
+                        'missing': '_last',
+                        'order': 'asc',
+                        'unmapped_type': 'float'
+                    }
+                }
+            ],
+            'order': 4,
+        }
     },
 }
 
--- a/inspirehep/modules/workflows/mappings/holdingpen/hep.json
+++ b/inspirehep/modules/workflows/mappings/holdingpen/hep.json
@@ -11,6 +11,29 @@
                     "properties": {
                         "is-update": {
                             "type": "boolean"
+                        },
+                        "relevance_prediction": {
+                            "properties": {
+                                "decision": {
+                                    "index": "not_analyzed",
+                                    "type": "string"
+                                },
+                                "scores": {
+                                    "properties": {
+                                        "CORE": {
+                                            "type": "float"
+                                        },
+                                        "Non-CORE": {
+                                            "type": "float"
+                                        },
+                                        "Rejected": {
+                                            "type": "float"
+                                        }
+                                    },
+                                    "type": "object"
+                                }
+                            },
+                            "type": "object"
                         }
                     },
                     "type": "object"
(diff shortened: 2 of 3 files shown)


inspirehep/config.py
```diff
@@ -1452,6 +1452,7 @@ def _(x):
             "workflow_name": terms_filter('_workflow.workflow_name'),
             "is-update": terms_filter('_extra_data.is-update'),
             'subject': terms_filter('metadata.inspire_categories.term'),
+            "decision": terms_filter('_extra_data.relevance_prediction.decision')
```
jacquerie: Nit: add an extra comma to the last element. If you don't do it, when you append again you get a slightly messier diff, where it looks like you deleted the previous line, while all you did was adding the extra comma.

inspirehep/config.py
```diff
@@ -1496,6 +1504,45 @@ def _(x):
             "default_order": 'desc',
             "order": 2,
         },
+        "core": {
+            "title": 'Core (Desc)',
```
jacquerie: Nit: use consistently single or double quotes. Try not to mix them (PEP8).

inspirehep/modules/workflows/mappings/holdingpen/hep.json
```diff
@@ -11,6 +11,29 @@
                     "properties": {
                         "is-update": {
                             "type": "boolean"
+                        },
+                        "relevance_prediction": {
+                            "properties": {
+                                "decision": {
+                                    "type": "string",
+                                    "index": "not_analyzed"
```
jacquerie: Every time you touch a JSON file you should use `scripts/prettify_json`.

inspirehep/config.py
```diff
@@ -1496,6 +1504,45 @@ def _(x):
...
+            "fields": [
+                {
+                    '_extra_data.relevance_prediction.scores.CORE':
+                        {
```
jacquerie: Don't add cognitive load when you don't need to add it. 3 lines above you have a list, and then you indent the objects inside. Here you indent in exactly the same way, but we're not inside of a list!

Small things like this add up. See: https://en.wikipedia.org/wiki/Stroop_effect

inspirehep/config.py
```diff
@@ -1496,6 +1504,45 @@ def _(x):
...
+                            "ignore_unmapped": True,
```
jacquerie: Deprecated long ago, and probably already removed in ES `2.4`: https://www.elastic.co/guide/en/elasticsearch/reference/1.6/search-request-sort.html#_ignoring_unmapped_fields

inspirehep/config.py
```diff
@@ -1496,6 +1504,45 @@ def _(x):
...
+                {
+                    '_extra_data.relevance_prediction.scores.Rejected':
+                        {
+                            "ignore_unmapped": True,
+                            "order": "asc"
+                        }
+                }
+            ],
+            "order": 5,
+        }
```
jacquerie: The problem here is this does not satisfy what the user expects to see. An article like https://arxiv.org/abs/hep-th/9711200, which is hugely core, has a huge CORE score, but also a big (in absolute value) Rejected score. If we implement this feature like this then all hugely core articles will come before all records that were actually rejected...

inspirehep/config.py
```diff
@@ -1478,6 +1479,13 @@ def _(x):
                     'size': 20,
                 },
             },
+            'decision': {
+                'terms': {
+                    'field': '_extra_data.relevance_prediction.decision',
+                    'size': 20
+
```
chris-asl: Nit: extra space.

chris-asl:
We need to change some design and implementation details to actually get sorting by the 3 scores.
What is currently available as data for sorting is: 
```
"scores": {
    "CORE",
    "Non-CORE",
    "Rejected"
}
```
and `relevance_score` which is targeted for sorting ([here](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/tasks/beard.py#L91))
```
relevance_score := "scores.CORE" + 10, if "decision" == "CORE"
                | "scores.Rejected" - 10, if "decision" == "Rejected"
                | "scores.Non-Core", if "decision" == "Non-CORE"
``` 
This provides a linear metric, like so: 
```
Rejected          Non-Core                Core
|---------------------|---------------------|
```

As you can see, sorting only using this metric, by three values cannot be done.  

Thus, we can sort by `Core` or `Rejected` by using `relevance_score` and introduce another field, e.g. `irrelevance_score`, whose values are: 
```
irrelavance_score := "scores.Non-Core" + 10, if "decision" == "Non-CORE",
                  | 0, otherwise
```

What do you think? @jacquerie @jmartinm

david-caro:
This might be alleviated by being able to filter by decision, so if you want to get the 'non-core' only and sort by relevance, then you use the 'Non-core' facet + sorting with this criteria, and will get you the non-core only sorted by relevance.

Though I'm not sure if that's what we want or not 😬

jmartinm:
As @david-caro says, I think the filter by `decision` plus sorting by relevance will fulfill the needs from https://github.com/inspirehep/inspire-next/issues/2113

```
best records first, i.e.
CORE: descending (the larger the better)
NON-CORE: descending (the larger the better)
Reject: ascending (the smaller the better)
```

jmartinm:
Small thing, can you amend the commit to put (closes #2113) inline. Check `git log` for examples.

And then rebase with latest master branch.

chris-asl:
Done :+1:

jmartinm:
@chris-asl @Glignos if you can rebase it with latest master, ping me and I will merge it

inspirehep/modules/workflows/mappings/holdingpen/hep.json
```diff
@@ -11,6 +11,29 @@
...
+                                        },
+                                        "Non-CORE": {
+                                            "type": "float"
+                                        },
+                                        "Rejected": {
+                                            "type": "float"
+                                        }
+                                    },
+                                    "type": "object"
+                                }
```
jacquerie: You're adding a mapping for this score, but you're sorting on another!

chris-asl: @jacquerie Was going to fix this. I guess you're continuing it at #2406 ?

PR #2359, opened 2017-05-19
literaturesuggest: experiments autocomplete sorting

### LINKED ISSUE #2102: author submission: experiment autocomplete
The experiments should be listed alphabetically

Glignos:
* Add a sorting function in the experiments autocomplete
  closes #2102

Signed-off-by: Glignos <glignos93@gmail.com>

### DIFF
--- a/inspirehep/modules/forms/static/js/forms/experiments_typeahead.js
+++ b/inspirehep/modules/forms/static/js/forms/experiments_typeahead.js
@@ -32,12 +32,14 @@ define([
       remote: {
         url: '/api/experiments?q=experimentautocomplete:%QUERY*',
         filter: function(response) {
-          return $.map(response.hits.hits, function(el) { return el });
-        }
-      },
+          return $.map(response.hits.hits, function(el) { return el }).sort(function(x, y) {
+          return x.metadata.experiment_names[0].title.localeCompare(y.metadata.experiment_names[0].title);
+        });
+      }
+    },
       datumTokenizer: function() {},
       queryTokenizer: Bloodhound.tokenizers.whitespace,
-      limit: 100,
+      limit: 3,
     });
 
     this.dataEngine.initialize();


inspirehep/modules/forms/static/js/forms/experiments_typeahead.js
```diff
@@ -32,12 +32,14 @@ define([
       remote: {
         url: '/api/experiments?q=experimentautocomplete:%QUERY*',
         filter: function(response) {
-          return $.map(response.hits.hits, function(el) { return el });
-        }
-      },
+          return $.map(response.hits.hits, function(el) { return el }).sort(function(x, y) {
+          return x.metadata.experiment_names[0].title.localeCompare(y.metadata.experiment_names[0].title);
```
jacquerie: This line is logically nested to the previous line, so it should also be nested in the code. Otherwise one sees two consecutive `return` statements and thinks that there's a bug.

inspirehep/modules/forms/static/js/forms/experiments_typeahead.js
```diff
@@ -32,12 +32,14 @@ define([
       remote: {
         url: '/api/experiments?q=experimentautocomplete:%QUERY*',
         filter: function(response) {
-          return $.map(response.hits.hits, function(el) { return el });
-        }
-      },
+          return $.map(response.hits.hits, function(el) { return el }).sort(function(x, y) {
```
jacquerie: `map`ping the identity function on `response.hits.hits` seems completely pointless...

jacquerie: Ah, this was introduced in https://github.com/inspirehep/inspire-next/pull/1602, but there's absolutely no point in doing it...

inspirehep/modules/forms/static/js/forms/experiments_typeahead.js
jacquerie: Also, this line is doing three things, which can just as easily happen on three lines:
```javascript
var x_title = x.metadata.experiment_names[0].title,
    y_title = y.metadata.experiment_names[0].title;

return x_title.localeCompare(y_title);
```

PR #2361, opened 2017-05-19
arxiv: re-download PDF if splash page

### LINKED ISSUE #2360: We should handle arXiv splash pages
See: https://sentry.cern.ch/inspire-sentry/inspire-labs/group/822222/

The actual data fetched was of type `.html.gz` (instead of PDF) and contained:
```html
[...]
<h1>PDF for <a href="http://arxiv.org/abs/1705.06383">1705.06383</a></h1>
<h2>We are now attempting to automatically create some PDF from the article's source.</h2>
<p>...this may take a little time.</p>

<p>For convenience, your browser has been asked to automatically
<a href="/pdf/1705.06383">reload this URL</a> in 10 seconds.</p>

<p>With some very old browsers you may need to manually reload. If you do 
this, you are <i>strongly</i> advised to wait the recommended 10 seconds: 
multiple rapid-fire requests could result in an errant trigger of our
<a href="http://arxiv.org/help/robots">rampaging robot</a> alert.</p>

<h2>What's happening?</h2>
[...]
```

The download from arXiv should check for this situation and sleep and retry few times more.

cc: @jacquerie

kaplun:
* Re-downloads a PDF from arXiv in case the served file is not really
  a PDF (most likely it's the usual arXiv splash page).
  (closes #2360)


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -71,7 +74,15 @@ def arxiv_fulltext_download(obj, eng):
...
     if pdf:
-        obj.log.info('PDF retrieved from arXiv for %s', arxiv_id)
+        if pdf.get_version().mimetype == 'application/pdf':
+            obj.log.info('PDF retrieved from arXiv for %s', arxiv_id)
+        else:
+            # We need to delete the failed PDF
+            del obj['files'][filename]
+            pdf.delete()
+            obj.log.warning('PDF retrieved from arXiv for %s', arxiv_id)
+            raise RuntimeError("PDF from arXiv for {arxiv_id} isn't yet ready".format(
```
jacquerie: Instead of generating a synthetic `RuntimeError` I think you can use [`on_predicate`](https://github.com/litl/backoff#backoffon_predicate).

jacquerie: BTW I think that this kind of exception message will generate one distinct unclustered message on Sentry, which is less than ideal.

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -71,7 +74,15 @@ def arxiv_fulltext_download(obj, eng):
...
 
     if pdf:
-        obj.log.info('PDF retrieved from arXiv for %s', arxiv_id)
+        if pdf.get_version().mimetype == 'application/pdf':
+            obj.log.info('PDF retrieved from arXiv for %s', arxiv_id)
+        else:
+            # We need to delete the failed PDF
+            del obj['files'][filename]
+            pdf.delete()
+            obj.log.warning('PDF retrieved from arXiv for %s', arxiv_id)
```
jmartinm: The message on this warning is the same as above, is that correct?

inspirehep/modules/workflows/tasks/arxiv.py
kaplun: If I use the predicate though I would need to clean up `obj['files'][filename]` at the beginning of every call.

kaplun: > BTW I think that this kind of exception message will generate one distinct unclustered message on Sentry per failure.

You mean after the 5th time when it's no longer captured, right?

inspirehep/modules/workflows/tasks/arxiv.py
kaplun: Yeah.

inspirehep/modules/workflows/tasks/arxiv.py
jacquerie: No, every single time.

kaplun: Mmh. @jacquerie using the predicate wouldn't make it any more readable because:
* I would have to always delete potential previously downloaded PDFs
* I wouldn't have to check if the last PDF is good or not. It would simply not repeat the download and leave the dangling broken data.

kaplun: > BTW I think that this kind of exception message will generate one distinct unclustered message on Sentry per failure.

But it's not possible to aggregate per exception, can I? I can aggregate log messages, but what about exceptions?

kaplun: Sentry should actually be able to aggregate per exception, regardless of the value, when there is a stacktrace: https://docs.sentry.io/learn/rollups/#grouping-by-stacktrace

kaplun: Given all the above, I believe the code is better like this with an exception. Maybe a better one than Runtime?

kaplun:
Yeah. As usual this is quite tricky to test. I can do it in unit test, for as much worth as they value.

kaplun:
My only issue with testing this, is that I will have to mock precisely the tricky aspect on invenio-files. Hence I won't really test much. If I screwed my interpretation of invenio-files, tests will still pass. (Now the code I wrote was thanks to introspection during live session, so hopefully I have not screwed my interpretation of invenio-files).

tests/unit/helpers/mocks.py
```diff
@@ -54,8 +54,9 @@ def __init__(self, *args, **kwargs):
 
 
 class MockFiles(object):
-    def __init__(self, data):
+    def __init__(self, data, mimetype='application/pdf'):
```
kaplun: This additional parameter is hack to personalize the behavior of the whole hierarchy.

jacquerie: I don't  like this. It's a responsibility of a `MockFileObject` to know its own default `mimetype`, not the class above it...

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -71,7 +74,15 @@ def arxiv_fulltext_download(obj, eng):
     )
 
     if pdf:
-        obj.log.info('PDF retrieved from arXiv for %s', arxiv_id)
+        if pdf.get_version().mimetype == 'application/pdf':
+            obj.log.info('PDF retrieved from arXiv for %s', arxiv_id)
+        else:
+            # We need to delete the failed PDF
+            del obj.files[filename]
+            pdf.delete()
```
jacquerie: Why do you need to do both? What happens if you only do the second?

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -71,7 +74,15 @@ def arxiv_fulltext_download(obj, eng):
     )
 
     if pdf:
-        obj.log.info('PDF retrieved from arXiv for %s', arxiv_id)
+        if pdf.get_version().mimetype == 'application/pdf':
+            obj.log.info('PDF retrieved from arXiv for %s', arxiv_id)
+        else:
+            # We need to delete the failed PDF
```
jacquerie: Yeah, but why?

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -71,7 +74,15 @@ def arxiv_fulltext_download(obj, eng):
...
+            del obj.files[filename]
+            pdf.delete()
+            warning_message = "PDF from arXiv for %s isn't yet ready"
+            obj.log.warning(warning_message, arxiv_id)
+            raise RuntimeError(warning_message % arxiv_id)
```
jacquerie: I think it's time to introduce the `WorkflowsTaskError` I was thinking of introducing in my last PR. It's certainly more expressive than a RuntimeError. BTW, does `backoff` respect inheritance?

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -28,6 +28,8 @@
 import re
 from functools import wraps
 
+import backoff
+
```
jacquerie: Nit: no space here.

inspirehep/modules/workflows/tasks/arxiv.py
kaplun: I'd say better safe then sorry. My head was exploding in trying to understand the what would happen in plainly overriding. Doing both is anyway not wrong.

tests/unit/helpers/mocks.py
kaplun: Yeah, but `MockFileObject` are instantiated indirectly down the road... how can I pass them this information otherwise?

Why it has to be complex to test :cry: ?

inspirehep/modules/workflows/tasks/arxiv.py
kaplun: What inheritance are you referring to?

jacquerie: I mean, if you define a `backoff` on `RuntimeError`, will it also work on an error that inherits from `RuntimeError`?

jacquerie: Yeah, it does (just tested it locally).

kaplun:
@jacquerie I think I have addressed all your requests

kaplun:
(sorry missing file in the commit)

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -82,6 +83,37 @@ def test_arxiv_fulltext_download_logs_on_success():
 
 
 @pytest.mark.httpretty
+def test_arxiv_fulltext_download_not_ready():
```
jacquerie: Let's make it a real sentence: `test_arxiv_fulltext_download_retries_when_fulltext_is_not_ready`

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -82,6 +83,37 @@ def test_arxiv_fulltext_download_logs_on_success():
...
+    httpretty.register_uri(
+        httpretty.GET, 'http://export.arxiv.org/pdf/1605.03844',
+        body=pkg_resources.resource_string(
+            __name__, os.path.join('fixtures', 'notreadyblob.gz')))
```
jacquerie: In order to prove the point of _retrying_ we can in fact provide two responses: the first one fails, the second one works.

kaplun:
@jacquerie we agree you are taking over this PR, right? :)

jacquerie:
> we agree you are taking over this PR, right? :)

Yes! I was pointing out stuff you (or anyone who reads my emails) might find interesting.

PR #2363, opened 2017-05-22
holdingpen: wider search bar

jmartinm:


jacquerie:
Nit: signature was missing from the commit message.

PR #2365, opened 2017-05-22
theme: remove back to INSPIRE toast message

jmartinm:
* Since the message is not helpful at the moment and gets in the way
  when performing other tasks.


jacquerie:
Had to submit #2367 after rebasing (and retouching the commit message) because I can't push to your repository.

PR #2372, opened 2017-05-23
workflows: fix arxiv_plot_extract

### LINKED ISSUE #2371: We are storing str(file) rather than files.
I just discovered on production this:
```
$ cat /opt/inspire/var/data/workflows/files/02/63/bec7-4e2b-45c0-a501-36bb7b64c573/data
<open file '/opt/inspire/var/data/workflows/files/98/f2/2326-2f4d-4c96-bea6-7be3b4bb5ff4/data_files/timing.png', mode 'r' at 0x152e3390>
```
Somehow `str(open_file)` is saved into `data` rather than `open_file`.

kaplun:
* Do actually save plots rather than their name.
  (closes #2371)

### DIFF
--- a/inspirehep/modules/workflows/tasks/arxiv.py
+++ b/inspirehep/modules/workflows/tasks/arxiv.py
@@ -122,7 +122,8 @@ def arxiv_plot_extract(obj, eng):
             return
 
         for idx, plot in enumerate(plots):
-            obj.files[plot.get('name')] = BytesIO(open(plot.get('url')))
+            with open(plot['url']) as plot_file:
+                obj.files[plot.get('name')] = plot_file
             obj.files[plot.get('name')]['description'] = u'{0:05d} {1}'.format(
                 idx, ''.join(plot.get('captions', []))
             )


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -122,7 +122,7 @@ def arxiv_plot_extract(obj, eng):
             return
 
         for idx, plot in enumerate(plots):
-            obj.files[plot.get('name')] = BytesIO(open(plot.get('url')))
```
kaplun: BytesIO transform a string into a stream. Here we already have a stream!

kaplun:
This bugfix is trivial. I can't make a test right now (have to leave). 
I guess a good old regression test is what is needed. But I already expect it to be 100 times larger than the characters I have corrected.

jacquerie:
> But I already expect it to be 100 times larger than the characters I have corrected.

So what? Ensuring that the code we write is correct is hard, so we don't do it?

inspirehep/modules/workflows/tasks/arxiv.py
jacquerie: ...and we leak a file descriptor.

david-caro: Hmm... the only thing I see makes sense to use a bytesio would be:
```python
obj.files[plot.get('name')] = BytesIO(open(plot.get('url'), 'rb').read())
```

Maybe it's one of those python2<->python3 thingies...

btw. I think that you should not use 'get' here, just old `['mykey']` so it  does not silently fail

david-caro: True! Remember to wrap it around `with open('bla') as bla_fd:` or similar to avoid it :)

kaplun: @jacquerie why you say so? Only on jython. The file is closed as soon as it gets out of scope, AFAIK

kaplun:
> So what? Ensuring that the code is correct is hard, so we don't do it?

It's just that I feel I am spending an incredible amount of time testing, and close to zero in comparison implementing. I am really feeling at the end of the day that the ratio shouldn't be this one, otherwise we will reach having a really complete service only in many years.

kaplun:
But of course I will write the test. As it seems that indeed up to now no plots where actually extracted...

jacquerie:
> It's just that I feel I am spending an incredible amount of time testing, and close to zero in comparison implementing.

The code that you just fixed was written in the golden age in which no testing was ever done. Look how amazing it turned out to be!

PR #2373, opened 2017-05-23
workflows: retry arxiv pdf download

### LINKED ISSUE #2360: We should handle arXiv splash pages

jacquerie:
Closes #2360 
Closes #2361 

Testing this is a little more complicated than expected, so I'm going to go forward on it and add some technical debt for later.

### DIFF
--- a/tests/unit/helpers/mocks.py
+++ b/tests/unit/helpers/mocks.py
@@ -67,11 +67,40 @@ class MockFiles(object):
         return item in self.data
 
     def __setitem__(self, key, value):
-        self.data[key] = {'key': key}
+        self.data[key] = MockFileObject(key=key)
+
+    def __delitem__(self, key):
+        del self.data[key]
 
     @property
     def keys(self):
         return self.data.keys()
+
+
+class MockFileObject(object):
+    def __init__(self, key):
+        self.obj = {'key': key}
+
+    def __eq__(self, other):
+        return self.obj['key'] == other.obj['key']
+
+    def __getitem__(self, key):
+        return self.obj[key]
+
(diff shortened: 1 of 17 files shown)


tests/unit/helpers/mocks.py
```diff
@@ -67,13 +67,42 @@ def __contains__(self, item):
...
+        pass
+
+    def get_version(self):
+        return MockObjectVersion()
+
+
+class MockObjectVersion(object):
+    @property
+    def mimetype(self):
+        return 'application/pdf'
```
kaplun: What happened to the test for when arXiv reply the splash page?

kaplun:
Ah. I read now your comment on the PR. OK...

PR #2374, opened 2017-05-23
workflows: fix arxiv_plot_extract

### LINKED ISSUE #2371: We are storing str(file) rather than files.

jacquerie:
Closes #2371 
Closes #2372 

Makes so that actual files are attached as plots, and not the
string representations of their file descriptors.

### DIFF
--- a/inspirehep/modules/workflows/tasks/arxiv.py
+++ b/inspirehep/modules/workflows/tasks/arxiv.py
@@ -31,7 +31,6 @@ from functools import wraps
 import backoff
 from flask import current_app
 from lxml.etree import XMLSyntaxError
-from six import BytesIO
 from wand.exceptions import DelegateError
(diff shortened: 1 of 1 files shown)


kaplun:
... but after the discussion on `.get()` you merged the version with `.get('url')` without tests, rather than `['url']` :confused:

jacquerie:
You don't change behavior where you have no tests.

PR #2375, opened 2017-05-26
workflows: support storage of original source

### LINKED ISSUE #1666: JSON Store
In order to support the needs of the three-ways merging of JSON-Merger we need to implement a JSON-store.

Such store would need to store the last JSON version of a record as it came from a given source. E.g. one entry for the JSON from arXiv, one for the JSON from e.g. Elsevier. The stored JSON is the one right before it's being merged into the record.

Such store could be implemented as:

object_uuid, data, source, version/timestamp.

API could be:

``` python
def store_json(object_uuid, json_data, source):
    pass

def get_last_json(object_uuid, source):
    pass

def garbage_collect():
    remove any previous version of a given source
```

kaplun:
* Implements easy API functions:
  store_original_record_for_source() and
  retrieve_original_record_for_source() allowing to preserve and
  later retrieve the original JSON source corresponding to a given
  publisher. (closes #1666)

### DIFF
--- a/inspirehep/modules/workflows/models.py
+++ b/inspirehep/modules/workflows/models.py
@@ -26,6 +26,9 @@ from __future__ import absolute_import, division, print_function
 
 from datetime import datetime
 
+from sqlalchemy.dialects import postgresql
+from sqlalchemy_utils.types import JSONType, UUIDType
+
 from invenio_db import db
 
 
@@ -74,3 +77,26 @@ class WorkflowsPendingRecord(db.Model):
         nullable=False,
     )
     record_id = db.Column(db.Integer, nullable=False)
+
+
+class WorkflowsRecordSources(db.Model):
+
+    __tablename__ = "workflows_record_sources"
+    __table_args__ = (
+        db.PrimaryKeyConstraint('record_id', 'source'),
+    )
+    source = db.Column(db.Text, default="", nullable=False)
+    record_id = db.Column(
+        UUIDType,
+        db.ForeignKey("records_metadata.id", ondelete='CASCADE'),
+        primary_key=True,
+        nullable=False,
+    )
+    json = db.Column(
+        JSONType().with_variant(
+            postgresql.JSON(none_as_null=True),
+            'postgresql',
+        ),
+        default=lambda: dict(),
+        nullable=True
+    )
--- a/inspirehep/modules/workflows/tasks/merging.py
+++ b/inspirehep/modules/workflows/tasks/merging.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Tasks related to record merging."""
+
+from __future__ import absolute_import, division, print_function
+
+from invenio_pidstore.resolver import PersistentIdentifier
+
+from inspirehep.utils.record import get_source, get_recid
+
+from ..utils import with_debug_logging, retrieve_root_json as _retrieve_root_json
+
+@with_debug_logging
+def retrieve_root_json(obj, eng):
+    """Retrieve the root JSON.
+
+    Retrieves the root JSON corresponding to the current matched record and the
+    provided source.
+    The root version is stored in `obj.extra_data['root_json']`.
+    """
+    source = get_source(obj.metadata)
+    recid = get_recid(obj.metadata)
+    record_uuid = PersistentIdentifier.get('literature', recid).object_uuid
+    obj.extra_data['root_json'] = _retrieve_root_json(record_uuid, source)
--- a/inspirehep/modules/workflows/utils.py
+++ b/inspirehep/modules/workflows/utils.py
@@ -34,8 +34,9 @@ import backoff
 import requests
 import urllib3
 from flask import current_app
+from invenio_db import db
 
-from .models import WorkflowsAudit
+from .models import WorkflowsAudit, WorkflowsRecordSources
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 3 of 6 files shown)


tests/integration/workflows/test_utils.py
```diff
@@ -0,0 +1,62 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Tests for Workflows utils."""
```
jacquerie: We don't add docstrings to tests because we rely on naming conventions: in fact, this file should be called `tests/integration/workflows/test_workflow_utils.py`.

inspirehep/modules/workflows/utils.py
```diff
@@ -176,3 +177,41 @@ def download_file_to_workflow(workflow, name, url):
...
+    Given a previously matched ``record_id`` (i.e. its UUID), the ``source``
+    information (e.g. `arXiv` or `Elsevier`) returns the original record if
+    existing or `None` otherwise.
+    """
+    entry = WorkflowsRecordSources.query.filter(
+        WorkflowsRecordSources.record_id == record_id,
+        WorkflowsRecordSources.source == source
+    ).first()
+    if entry:
+        return entry.json
```
jacquerie: It really looks like to me that you're reimplementing `one_or_none` here.

tests/integration/workflows/test_utils.py
```diff
@@ -0,0 +1,62 @@
...
+import pytest
+
+from invenio_db import db
+
+from inspirehep.modules.workflows.utils import store_original_record_for_source, retrieve_original_record_for_source
+
+
+@pytest.fixture()
+def dummy_record(small_app):
+    from invenio_records.api import Record
```
jacquerie: Why is this import internal? If it doesn't need to be, it should be pulled up with the others. Moreover, I think we use everywhere `InspireRecord` because its `delete` plays nicer with the tests.

inspirehep/modules/workflows/utils.py
```diff
@@ -176,3 +177,41 @@ def download_file_to_workflow(workflow, name, url):
         if req.status_code == 200:
             workflow.files[name] = req.raw
             return workflow.files[name]
+
+
+def store_original_record_for_source(record_id, source, data):
+    """Store the original record for a given source.
+
+    Given a previously matched ``record_id`` (i.e. its UUID), the ``source``
```
jacquerie: If the `record_id` is in fact a `uuid` it should be called `record_uuid`, no? Otherwise I would expect to be able to put a `control_number` there.

inspirehep/modules/workflows/utils.py
```diff
@@ -176,3 +177,41 @@ def download_file_to_workflow(workflow, name, url):
         if req.status_code == 200:
             workflow.files[name] = req.raw
             return workflow.files[name]
+
+
+def store_original_record_for_source(record_id, source, data):
```
jacquerie: The API (this function and the next) sounds a bit awkward: names are long, but do not explain themselves very well. That said, good names come from their users, and since this PR does not include uses it's hard to tell which better names we could give them.

tests/integration/workflows/test_utils.py
kaplun: OK. I add it because I cut & paste from a file in the same folder which had it.

tests/integration/workflows/test_utils.py
kaplun: > Why is this import internal?

Was thinking to import globally the functions that are under tests, while import locally the rest of the auxiliary functions. But indeed no other good reason.

> because its `delete` plays nicer with the tests.

Mmh. `InspireRecord.delete()` is not good for tests because it's a soft delete, so tables are not well cleaned afterwards. On the other hand I see can user `InspireRecord._delete()` which is implementing Invenio's one, which rightfully can implement a `force`-delete, and apply all the ondelete-cascade triggers.

inspirehep/modules/workflows/utils.py
kaplun: You are right. It's a bit sad that the original APIs called it `record_uuid` in the first place, because this is mixing up type with semantic. But too bad.

inspirehep/modules/workflows/utils.py
kaplun: Yup, usage soon to come in additional commits.

Possible better name could be:
* `store(/put/persist?)_root_json()`
* `retrieve(/get/recover?)_root_json()`

WDYT?

inspirehep/utils/record.py
```diff
@@ -164,3 +164,13 @@ def is_complete(publication_info):
...
+
+
+def get_source(record):
+    """Return the first available source of a record."""
+    return get_value(record, 'acquisition_source[0].source')
+
+
+def get_recid(record):
+    """Return the record ID."""
+    return get_value(record, 'self_recid')
```
kaplun: `self_recid` is only in ES. We shall use `control_number`.

inspirehep/modules/workflows/tasks/merging.py
```diff
@@ -0,0 +1,44 @@
...
+@with_debug_logging
+def retrieve_root_json(obj, eng):
+    """Retrieve the root JSON.
+
+    Retrieves the root JSON corresponding to the current matched record and the
+    provided source.
+    The root version is stored in `obj.extra_data['root_json']`.
+    """
+    source = get_source(obj.metadata)
+    recid = get_recid(obj.metadata)
```
kaplun: This is `.data`

inspirehep/modules/workflows/models.py
```diff
@@ -74,3 +77,26 @@ class WorkflowsPendingRecord(db.Model):
...
+class WorkflowsRecordSources(db.Model):
+
+    __tablename__ = "workflows_record_sources"
+    __table_args__ = (
+        db.PrimaryKeyConstraint('record_id', 'source'),
+    )
+    source = db.Column(db.Text, default="", nullable=False)
+    record_id = db.Column(
+        UUIDType,
+        db.ForeignKey("records_metadata.id", ondelete='CASCADE'),
```
jacquerie: This only allows for a 1-N relationship between records and workflow records. This is not the right relationship, because we might ingest things that will not become records, but are workflow records (we ingested them, but we also rejected them). This needs instead to be a M-N relationship with a connecting table.

See: https://en.wikipedia.org/wiki/One-to-many_(data_model) vs. https://en.wikipedia.org/wiki/Many-to-many_(data_model).

jacquerie: CC: @rikirenz

kaplun: I don't see why this requires an N-M model in the case of non ingested records. Even in that case only one workflow record (identified how?) will match to N sources.

But IMHO to wishing to merge rejected record introduces an unnecessary complexity. If they were rejected the first time, they are surely NON-CORE, therefore, if this time we are accepting them, we can simply take the current version. (this was not discussed but I believe such simplification will make our goals more realistic). We can always have rejected records to be instantiated as stub record, thus be stored as regular one...

jacquerie: As I said, because with this DB schema you cannot model the (typical) case in which you harvest a record, but you decide not to accept it in INSPIRE.

david-caro: This is not a relation between wokflow records and records, it's between records and sources no? Records being recids, merging both, workflow and non-workflow, that's why we reserve a record id at the start of the workflow no?

kaplun: @jacquerie even in that case it's a 1 to N relation but with a different table. And it's not going to be typical. The typical is accepted paper and publisher or arxiv updates come. 

@david-caro correct. It's unrelated with workflows.

jacquerie: You want @rikirenz and @ammirate to use this to store the distinct sources that contribute to a record. Problem: when you first ingested this record, before it was accepted by a curator, which record were you pointing at with this column?

inspirehep/modules/workflows/models.py
```diff
@@ -74,3 +77,26 @@ class WorkflowsPendingRecord(db.Model):
...
+        primary_key=True,
```
jacquerie: If `record_id` is a primary key for this table then you can have at most one source per...

inspirehep/modules/workflows/models.py
kaplun: I think you should not check this PR anymore, since it's superseded by #2423 where they have implemented all this business logic.

There is no record in this table at the beginning. This table is populated the first time only after the record is actually accepted by the curator and added into the system (thus receiving a UUID).

inspirehep/modules/workflows/models.py
kaplun: Fully correct! This flag is wrong. @rikirenz can you correct this into your PR?

kaplun:
Note this is superseded by #2423 .

PR #2376, opened 2017-05-26
build: pin invenio-workflows to the major version

david-caro:


david-caro:
This allows us to upgrade invenio-workflows when there's no major version change (for example, to get https://github.com/inveniosoftware-contrib/invenio-workflows/pull/41)

jacquerie:
By reading [PEP 440](https://www.python.org/dev/peps/pep-0440/#compatible-release), it looks like this constraint can be more succintly expressed as `invenio-workflows~=6.0.5`.

PR #2377, opened 2017-05-29
setup: pin requests and docutils

jacquerie:
This build broke because `requests==2.15.0` was a bad version: https://travis-ci.org/inspirehep/inspire-next/builds/236661038

The [current build](https://travis-ci.org/inspirehep/inspire-next/builds/236879618) instead breaks because `requests~=2.16.5` has changed significantly how exception handling happens in `requests`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -126,6 +126,7 @@ extras_require = {
         'mock',
         'Sphinx<1.6',
         'sphinxcontrib-napoleon>=0.6.1',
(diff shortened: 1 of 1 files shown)


jacquerie:
...great. `docutils==0.14rc1` was also broken: https://travis-ci.org/inspirehep/inspire-next/jobs/237030100#L1789.

PR #2399, opened 2017-06-01
holding-pen: fix #2383

### LINKED ISSUE #2383: HoldingPen: remove duplicates from categories filter
The [categories filter](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/static/js/inspire_workflows_ui/holdingpen/holdingpen.filters.js#L55-L81) should remove duplicates.

Otherwise, for example creating a submission with the arXiv ID `hep-th/9711200` will trigger the following error in the Holding Pen results page:

```
inspire-workflows.details.659ac3b8.js:12811 Error: [ngRepeat:dupes] Duplicates in a repeater are not allowed. Use 'track by' expression to specify unique keys. Repeater: category in record._source.metadata | categories, Duplicate key: string:Theory-HEP, Duplicate value: Theory-HEP
```

harunurhan:
fix ng-repeat duplicate key error

## Description

categories filter returns unique array now to avoid ng-repeat
duplicate key error.

## Related Issue

#2383 

## Checklist:

- [x ] I linked the related issue(s) in the corresponding commit logs.
- [x ] I wrote [good commit log messages](https://github.com/torvalds/subsurface-for-dirk/blob/5f15ad5a86ada3c5e574041a5f9d85235322dabb/README#L92-L119).
- [x ] My code follows the code style of this project.
- [x ] All new and existing tests passed.

### DIFF
--- a/inspirehep/modules/workflows/static/js/inspire_workflows_ui/holdingpen/holdingpen.filters.js
+++ b/inspirehep/modules/workflows/static/js/inspire_workflows_ui/holdingpen/holdingpen.filters.js
@@ -78,7 +78,10 @@
         }
       }
 
-      return categories;
+      var uniqueCategories = categories.filter(function (value, index, self) {
+        return self.indexOf(value) === index;
+      });
+      return uniqueCategories;
     }
   }
 


harunurhan:
I couldn't test it since couldn't reproduce the issue on my local. If anyone knows an id to submit to holdingpen that reproduces this issue, please let me know.

jacquerie:
> If anyone knows an id to submit to holdingpen that reproduces this issue, please let me know.

Try `hep-th/9711200`!

jacquerie:
BTW watch out, it looks that you're adding some external modules as Git submodules on top of your change...

inspirehep/modules/workflows/static/js/inspire_workflows_ui/holdingpen/holdingpen.filters.js
```diff
@@ -78,7 +78,10 @@
         }
       }
 
-      return categories;
+      var uniqueCategories = categories.filter(function (value, index, self) {
+        return self.indexOf(value) === index;
+      });
```
jacquerie: This looks O(n^2)... Shouldn't it be possible to do in O(n log n)?

jacquerie: Well, not that it matters, since we're dealing with something like 2/3 elements anyway...

harunurhan: yes could be done better, but as you said, array is like to be small, max 5-6 I think. So didn't want to complicate the code for a little performance improvement :)

Issue #2385, opened 2017-05-30
travis: abnormally flaky acceptance suite


jacquerie:
After merging https://github.com/inspirehep/inspire-next/pull/2124 and https://github.com/inspirehep/inspire-docker/pull/38 we started seeing the following timeout:

* https://travis-ci.org/inspirehep/inspire-next/jobs/236285528#L1210
* https://travis-ci.org/inspirehep/inspire-next/jobs/236434695#L1220
* https://travis-ci.org/inspirehep/inspire-next/jobs/236879622#L1231

This timeout also happened in the original merge build of #2124, but it worked after a retry.

We've had 16 builds since that build, and 4 have failed because of this error, for an empirical flakiness rate of 25%, which is way too high.

jacquerie:
Another instance: https://travis-ci.org/inspirehep/inspire-next/jobs/237478388#L1218

jacquerie:
In the immortal words of DJ Khaled, "and another one": https://travis-ci.org/inspirehep/inspire-next/jobs/237407705#L1242

rikirenz:
@jacquerie I notice this as well... Honestly I do not have an idea about how to solve this issue.
Did you experience this also locally? Do you have an idea how to debug this? 

I have checked the issues in https://github.com/SeleniumHQ/docker-selenium/issues but there are not issues related.

PR #2410, opened 2017-06-07
dojson: make publication_info2marc handle unicode

jacquerie:
## Related Issue
https://sentry.cern.ch/inspire-sentry/inspire-labs/group/822408/

## Checklist:


jacquerie:
We should probably account for `\u2013` in `artid` as a special case, because it is a separator that people might type instead of `-`. Same goes for `\u2010`.

jacquerie:
Victim of #2385.

PR #2411, opened 2017-06-07
workflows: decompress gzipped pdf downloads

### LINKED ISSUE #2408: PDF with compressed content-encoding are not decompressed
## Expected Behavior
If servers serve us with content-encoding: gzip, we should on the fly decode it.

## Current Behavior
This is currently not happening. So in: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/utils.py#L177 we are serving gzipped data.

jacquerie:
## Related Issue
Closes #2408

## Checklist:


kaplun:
Thanks! I was trying to solve it as well using `iter_content()`. I missed the existence of the flag you used.

PR #2420, opened 2017-06-09
setup: properly pin dependencies

kaplun:


kaplun:
@jacquerie, fetching `/deps/deps.txt` from the latest docker image, I have wrote a little script based on `requirements-parser` to update our `setup.py`. This could be the base for future updates.

setup.py
```diff
@@ -32,117 +32,115 @@
...
-    'pep257>=0.7.0',
-    'pytest-cache>=1.0',
-    'pytest-cov>=1.8.0',
-    'pytest-flake8>=0.8.1',
-    'pytest-httpretty>=0.2.0',
-    'pytest-selenium>=1.3.1',
-    'pytest>=2.8.0',
-    'mock>=1.3.0',
-    'requests_mock',
+    'check-manifest>=0.35,~=0',
```
jacquerie: From https://www.python.org/dev/peps/pep-0440/#compatible-release:
> This operator MUST NOT be used with a single segment version number such as `~=1`.

setup.py
```diff
@@ -32,117 +32,115 @@
...
-    'dojson>=1.3.0',
-    'Flask>=0.11.1',
-    'Flask-Breadcrumbs>=0.3.0',
-    'Flask-Caching>=1.0.1',
-    'Flask-Script>=2.0.5',
-    'flask-shell-ipython>=0.2.2',
-    'fs<2.0',  # TODO: remove once invenio-files-rest#130 is fixed
-    'jsmin',
-    'pytest-runner>=2.7.0',
+    'amqp>=1.4.9,~=1.4',
```
jacquerie: Oh, and I don't believe that a constraint like `~=1.4` actually picks up version `1.5`, for example. I think you have to write `~=1.0`.

kaplun: I don't think we want version `1.5`, right?  Only 1.4.10 onwards...

setup.py
kaplun: Oh.

setup.py
david-caro: `~=1.4` does actually pick up `1.5` too (just tested), though `~=0.4.0` does not.

david-caro:
This is still valuable to some point no? We should discuss a bit though, as I don't have a clear idea of what to pin, to which versions, and for which purpose (qa/dev/prod...). Or if the stricter pins should go here or somewhere else (say a docker image with just the app, or something external to the code repo).

kaplun:
Right we should have gentle pin (not aggressive ones) on Docker and use the docker-generated `deps.txt`on deployment, that guarantees that these deps have passed all tests.

PR #2426, opened 2017-06-12
docker: pin postgres to 9.4.5

jacquerie:
## Description
Currently on production we are using this particular version, offered
by CERN's DB on Demand service. To ensure that we don't introduce code
that depends on future features of PostgreSQL, which would break on
production, this commit pins its Docker image to that specific version.

## Checklist:

### DIFF
--- a/services.yml
+++ b/services.yml
@@ -35,7 +35,7 @@ services:
     environment:
       - ES_HEAP_SIZE=2g
   database:
-    image: postgres
+    image: postgres:9.4.5
     environment:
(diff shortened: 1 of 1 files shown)


jacquerie:
CC: @jmartinm

jacquerie:
Similarly, we probably want to constrain `rabbitmq` and `redis` with what's on production. @david-caro, can you help us here?

david-caro:
Currently we have rabbitmq-server-3.3.5 and redis-3.2.3

PR #2437, opened 2017-06-14
xrootd: first commit

### LINKED ISSUE #1086: Move to EOS
Here's the recipe, courtesy of @lnielsen:
- [ ] install `xrootd-python` bindings (see below)
- [ ] install `XRootDPyFS`
- [ ] point `invenio-files-rest` to `root://eospublic.cern.ch//eos/<project>/`
- [ ] run `k5reauth -x -f -k /etc/<keytab> <user>@CERN.CH` on all the machines

Notes: 
- `k5reauth` as a service is being added to the puppet Invenio module: when a machine starts it will start `k5reauth`.
- a service account is assumed.
- To build `Xrootd` for a `virtualenv`: see https://gitlab.cern.ch/ai/it-puppet-hostgroup-zenodo/blob/qa/code/files/zenodo/tasks.py#L81
- To make sure `xrootdpyfs` is loaded: https://github.com/zenodo/zenodo/blob/master/zenodo/modules/xrootd/ext.py

kaplun:
Copies from Zenodo the XRootD integration. Unfortunately this is not testable on Travis. However by default, mocking XRootD is equivalent to having it disabled. So the real test will happen hopefully on QA.

Closes #1086

Currently we are using Fuse-mounted EOS on production and this has many stability and performance issues. Hopefully by switching to XRootD (which is the
native interface of EOS) will allow INSPIRE to use EOS following its original design, i.e. a blob-store, rather than a POSIX filesystem. This should, hopefully improve performance issue (if not also stability issues), at least within the workflow files.

* Copies from Zenodo the integration with XRootD, to be used as
  storage, instead of Fuse-mounted EOS.
  (closes #1086)

### DIFF
--- a/inspirehep/modules/xrootd/__init__.py
+++ b/inspirehep/modules/xrootd/__init__.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""XRootD integration of INSPIRE."""
+
+from __future__ import absolute_import, division, print_function
+
+from .ext import INSPIREXRootD
--- a/inspirehep/modules/xrootd/ext.py
+++ b/inspirehep/modules/xrootd/ext.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""INSPIRE XRootD extension."""
+
+from __future__ import absolute_import, division, print_function
+
+from pkg_resources import DistributionNotFound, get_distribution
+
+try:
+    # Import XRootDPyFS if available so that its
+    # opener gets registered on PyFilesystem.
+    get_distribution('xrootdpyfs')
+    import xrootdpyfs
+    XROOTD_ENABLED = True
+except DistributionNotFound:
+    XROOTD_ENABLED = False
+    xrootdpyfs = None
+
+
+class INSPIREXRootD(object):
+
+    """INSPIRE XRootD extension."""
+
+    def __init__(self, app=None):
+        """Extension initialization."""
+        if app:
+            self.init_app(app)
+
+    def init_app(self, app):
+        """Extension registration and configuration."""
+        app.config['XROOTD_ENABLED'] = XROOTD_ENABLED
+        if XROOTD_ENABLED:
+            # Overwrite reported checksum from CERN EOS due to XRootD 3.3.6.
+            app.config['XROOTD_CHECKSUM_ALGO'] = 'md5'
+            app.config['FILES_REST_STORAGE_FACTORY'] = 'invenio_xrootd:eos_storage_factory'
+        app.extensions['inspire-xrootd'] = self
--- a/setup.py
+++ b/setup.py
@@ -152,7 +152,11 @@ extras_require = {
         'kwalitee',
         'honcho',
         'gunicorn',
-    ]
+    ],
+    'xrootd': [
+        'invenio-xrootd>=1.0.0a4',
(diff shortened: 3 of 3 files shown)


inspirehep/modules/xrootd/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+# along with INSPIRE; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+#
+# NOTE: this code has been copied from Zenodo
+# https://github.com/zenodo
```
kaplun: Not sure if this is the best way to phrase it.

lnielsen:
Note that if you serve large file you may need to use uwsgi instead of gunicorn

inspirehep/modules/xrootd/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+
+"""XRootD extension."""
+
+from __future__ import absolute_import, print_function
+
+from .ext import INSPIREXRootD
+
+__all__ = ('INSPIREXRootD', )
```
jacquerie: See: https://github.com/inspirehep/inspire-next/commit/485319e4071c6ca90b310c8579f52ec3c5d9100d

kaplun:
> Note that if you serve large file you may need to use uwsgi instead of gunicorn

In principle we are serving back only PDF. Actually we are not yet serving anything :D invenio-files* is still used only in the backoffice for the holding pen, so it's more just to attach PDFs to records.

But I guess you suggest uwsgi so that we don't serve files via Python but via C :+1:

inspirehep/modules/xrootd/__init__.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2017 CERN.
```
jacquerie: Wrong year.

inspirehep/modules/xrootd/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+#
+# INSPIRE is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
```
jacquerie: Wrong license.

inspirehep/modules/xrootd/__init__.py
jacquerie: IMO this code is trivial enough to not warrant specific attribution... it's just an extension declaration, a pattern that is not specific of XRootD or Zenodo.

inspirehep/modules/xrootd/ext.py
```diff
@@ -0,0 +1,62 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
```
jacquerie: When using `scripts/prettify_licenses` you still have to be careful about this...

kaplun: Oh! First time I am using it :+1:

jacquerie:
Something happened, and `modules/xrootd/__init__.py` is now an empty file...

kaplun:
> Something happened, and modules/xrootd/__init__.py is now an empty file...

It's on purpose. It's not needed anymore, because anyway the extension is referenced explicitly from `setup.py` as in `inspirehep.modules.xrootd.ext:INSPIREXRootD`.

Since there was no longer Python code in it, I also removed the copyright/license.

kaplun:
Nevermind. An empty `__init__.py` is not valid according to flake8, due to missing import for `print_function` etc.

lnielsen:
> But I guess you suggest uwsgi so that we don't serve files via Python but via C 👍

Actually no, it's still through Python, but gunicorn have some really bad read/write loops that reads 8k from eos, then sends 8k to the client etc.....this gives really bad throughput from EOS where chunksizes of ~2MB gives much better performance.

Long-term we're trying to outsource the file serving to EOS (probably hidden behind a proxy) but that's not yet in place.

jacquerie:
> the extension is referenced explicitly from `setup.py` as in `inspirehep.modules.xrootd.ext:INSPIREXRootD`

Note however how yours is the only extension that does this...

inspirehep/modules/xrootd/ext.py
david-caro: Is that script idempotent? Can we add it as part of a check or something?

jacquerie: The script is idempotent, but what is produced is not guaranteed to be correct, as in the example above. If you read its code it's just a simple text substitution that takes the first 21 lines that look like comments and replaces those with the right license. This was enough for the typical case, in which a license was wrong because it had the wrong year or referred to the wrong project.

It's easy to make it more robust (better algorithm: you remove lines until you see something that is not a comment, and then you add the right license back), but I'd rather not fail builds because of a wrong license or wrongly formatted files: it's just frustrating for the developer (Trust me, I've been there with Invenio 2). I'd like to add this as a smudge filter, so that all this is both correct and invisible to the developer.

kaplun:
I have implemented all suggestions.

jacquerie:
ACK, I have some small fixes to make.

inspirehep/modules/xrootd/ext.py
```diff
@@ -0,0 +1,56 @@
...
+        """Extension initialization."""
+        if app:
+            self.init_app(app)
+
+    def init_app(self, app):
+        """Extension registration and configuration."""
+        app.config['XROOTD_ENABLED'] = XROOTD_ENABLED
+        if XROOTD_ENABLED:
+            # Overwrite reported checksum from CERN EOS due to XRootD 3.3.6.
+            app.config['XROOTD_CHECKSUM_ALGO'] = 'md5'
```
lnielsen: FYI: this should be needed anymore once eospublic is upgraded wednesday 21st june to xrootd 4

inspirehep/modules/xrootd/ext.py
```diff
@@ -0,0 +1,56 @@
...
+    # opener gets registered on PyFilesystem.
+    get_distribution('xrootdpyfs')
+    import xrootdpyfs
+    XROOTD_ENABLED = True
+except DistributionNotFound:
+    XROOTD_ENABLED = False
+    xrootdpyfs = None
+
+
+class INSPIREXRootD(object):
```
lnielsen: Probably we should just add this into Invenio-XRootD seeing that it's the same on Zenodo and Inspire.

kaplun: Yeah wasn't sure if it was worth a new package. Indeed Invenio-XRootD seems the natural candidate now that you said it.

PR #2441, opened 2017-06-14
Include CDS and HAL collections in duplicate validators

### LINKED ISSUE #2424: Check for duplicates in CDS Hidden collection
In the same way that we currently look for duplicates in the HAL Hidden collection, we should also look for duplicates in the CDS Hidden collection.

https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/forms/validators/simple_fields.py#L132

Glignos:
## Description
Also searches arXiv identifiers and DOIs in the CDS Hidden collection
when looking for duplicates.

## Related Issue
Closes #2424

## Checklist:

### DIFF
--- a/inspirehep/modules/forms/validators/simple_fields.py
+++ b/inspirehep/modules/forms/validators/simple_fields.py
@@ -56,11 +56,11 @@ def isbn_syntax_validation(form, field):
         raise StopValidation(message)
 
 
-def does_exist_in_inspirehep(query, collection=None):
+def does_exist_in_inspirehep(query, collections=None):
     """Check if there exist an item in the db which satisfies query.
 
     :param query: http query to check
-    :param collection: collection to search in; by default searches in
+    :param collections: collections to search in; by default searches in
         the default collection
     """
     import requests
@@ -71,8 +71,11 @@ def does_exist_in_inspirehep(query, collection=None):
         'of': 'id'
     }
 
-    if collection:
-        params['cc'] = collection
+    if collections:
+        if len(collections) == 1:
+            params['cc'] = collections[0]
+        else:
+            params['c'] = collections
 
     json_reply = requests.get(
         "http://inspirehep.net/search?", params=params).text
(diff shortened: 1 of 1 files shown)


inspirehep/modules/forms/validators/simple_fields.py
```diff
@@ -71,8 +71,12 @@ def does_exist_in_inspirehep(query, collection=None):
         'of': 'id'
     }
 
-    if collection:
-        params['cc'] = collection
+    if len(kwargs) > 1:
+        collections = []
+        for key, value in kwargs.iteritems():
+            if 'collection' in key:
+                collections.append(value)
```
jacquerie: To avoid this somewhat awkward logic (which leads to you having to have parameters called `collection1` and `collection2`) can't you just have `collection` be a list, and iterate over it?

inspirehep/modules/forms/validators/simple_fields.py
```diff
@@ -89,15 +93,18 @@ def does_exist_in_inspirehep(query, collection=None):
...
     """
-    if does_exist_in_inspirehep(inspire_query, collection):
+    if does_exist_in_inspirehep(inspire_query, **kwargs):
         url = "http://inspirehep.net/search?" + urlencode({'p': inspire_query})
-        if collection:
-            url += '&' + urlencode({'cc': collection})
+        if len(kwargs) > 1:
+            for key, value in kwargs.iteritems():
+                if 'collection' in key:
+                    url += '&' + urlencode({'cc': value})
```
jacquerie: Same as above.

inspirehep/modules/forms/validators/simple_fields.py
```diff
@@ -129,7 +136,7 @@ def duplicated_doi_validator(form, field):
 
         # And in Hal collection
```
jacquerie: Comments here and below also need to be updated.

inspirehep/modules/forms/validators/simple_fields.py
jmartinm: +1 that would be better

PR #2443, opened 2017-06-15
setup: harmonize the major version pins

david-caro:
* This uses the format `~=[major].0,>=[major[.feat[.minor]]]` to all
  the major version pins, as we were using the bad
  `~=[major].[feat].[minor]` in some places, actually pinning to the
  feature version.

Closes #2420

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ install_requires = [
     'invenio-oaiharvester==1.0.0a2',
     'invenio-utils==0.2.0',  # Not fully Invenio 3 ready
     'invenio>=3.0.0a1,<3.1.0',
-    'inspire-crawler~=0.2.7',
+    'inspire-crawler~=0.0,>=0.2.7',
     'inspire-schemas~=33.0,>=33.0.2',
     'dojson>=1.3.0',
     'Flask>=0.11.1',
@@ -95,14 +95,14 @@ install_requires = [
     'elasticsearch<3.0.0',
     'Flask-Login<0.4.0',
     'invenio-workflows~=6.0,>=6.0.5',
-    'invenio-workflows-files~=0.0.4',
-    'invenio-workflows-ui~=1.0.31',
+    'invenio-workflows-files~=0.0,>=0.0.4',
+    'invenio-workflows-ui~=1.0,>=1.0.31',
     'elasticsearch-dsl<2.2.0',
     'pycountry>=17.1.8',
     'Flask_CeleryExt>=0.3.0',
     'python-redis-lock~=3.2',
-    'backoff~=1.4.2',
-    'requests~=2.15.1',
+    'backoff~=1.0,>=1.4.2',
+    'requests~=2.0,>=2.15.1',
 ]
 
 tests_require = [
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -53,7 +53,7 @@
     'invenio-assets>=1.0.0b2',
     'invenio-base>=1.0.0a11',
     'invenio-celery>=1.0.0a4',
-    'invenio-classifier~=1.1.0',
+    'invenio-classifier~=1.0,>=1.1.0',
```
jacquerie: `~1.0` is what I wanted as well, but @kaplun preferred the stricter `~1.1`... In a way, it depends if the project is following SemVer or not.

david-caro: yep, if the project does not follow semantic versioning it's useless.
So this should be pinned to <1.2?

I found that docutils should be pinned to <0.14 too, or with our current sphinx version, it fails like https://github.com/coala/coala/issues/4280

setup.py
```diff
@@ -74,7 +74,7 @@
     'invenio-oaiharvester==1.0.0a2',
     'invenio-utils==0.2.0',  # Not fully Invenio 3 ready
     'invenio>=3.0.0a1,<3.1.0',
-    'inspire-crawler~=0.2.7',
+    'inspire-crawler~=0.0,>=0.2.7',
```
kaplun: `~=0.0` and `~=0.2` are equivalent in this context, but I believe the latter is more meaningful.

setup.py
```diff
@@ -95,14 +95,14 @@
     'elasticsearch<3.0.0',
     'Flask-Login<0.4.0',
     'invenio-workflows~=6.0,>=6.0.5',
-    'invenio-workflows-files~=0.0.4',
-    'invenio-workflows-ui~=1.0.31',
+    'invenio-workflows-files~=0.0,>=0.0.4',
```
kaplun: Ditto.

jacquerie: ?

This is `0.0.4`...

kaplun: Sorry :dancer:

setup.py
```diff
@@ -95,14 +95,14 @@
...
+    'invenio-workflows-ui~=1.0,>=1.0.31',
     'elasticsearch-dsl<2.2.0',
     'pycountry>=17.1.8',
     'Flask_CeleryExt>=0.3.0',
     'python-redis-lock~=3.2',
-    'backoff~=1.4.2',
-    'requests~=2.15.1',
+    'backoff~=1.0,>=1.4.2',
+    'requests~=2.0,>=2.15.1',
```
kaplun: Then *ditto* here.

setup.py
david-caro: yes, but if you pair the first like this: `~=0.0,>=0.2` it's even more meaningful and less prone to errors of pinning to the feature version, say for example that you have to change it and pin it to a minimum version of 0.3.5, with the single solution it's easy to end up with `~=0.3.5`, and that will not do what you want, as it will pin to the feat version, so you will never get 0.4, but with this solution you will most probably end up with `~=0.0,>=0.3.5`, that does exactly what you want. The idea of using the 0.0 instead of 0.2/0.3 is to make clear that the first part of the version restriction is only to pin the major version.

david-caro: ... that was way more text than I expected ...

jacquerie: Yeah, I agree with @david-caro: let's uniformize (cit.) to the pattern that we're already using.

PR #2446, opened 2017-06-16
workflows: fix refextract task 

zzacharo:
## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/refextract.py
+++ b/inspirehep/modules/workflows/tasks/refextract.py
@@ -24,18 +24,14 @@
 
 from __future__ import absolute_import, division, print_function
 
-import json
-
 from timeout_decorator import timeout
 
 from refextract import extract_journal_reference, extract_references_from_file
 
-from inspirehep.utils.helpers import maybe_int
-from inspirehep.utils.record import get_value
+from inspirehep.modules.references.processors import ReferenceBuilder
+from inspirehep.utils.helpers import force_list, maybe_int
 from inspirehep.utils.pubnote import split_page_artid
-
-# FIXME get journal mappings for refextract
-# from inspirehep.utils.knowledge import get_mappings_from_kbname
+from inspirehep.utils.record import get_value
 
 
 def extract_journal_info(obj, eng):
@@ -90,48 +86,34 @@ def extract_journal_info(obj, eng):
 def extract_references(filepath):
     """Extract references from PDF and return in INSPIRE format."""
     references = extract_references_from_file(
-        filepath,
-        reference_format=u"{title},{volume},{page}",
-    )
-    mapped_references = []
-    if references.get('references'):
-        for ref in references.get('references'):
-            reference = {
-                'curated_relation': False,
-                'raw_refs': [
-                    {
-                        'position': '',
-                        'schema': '',
-                        'source': '',
-                        'value': json.dumps(ref),
-                    }
-                ],
-                'record': {
-                    '$ref': ''
-                },
-                'reference': {
-                    'authors': ref.get('author'),
-                    'book_series': ref.get('book_series'),
-                    'collaboration': ref.get('collaboration'),
-                    'dois': ref.get('dois'),
-                    'imprint': ref.get('imprint'),
-                    'misc': ref.get('misc'),
-                    'number': ref.get('linemarker'),
-                    'persistent_identifiers': ref.get(
-                        'persistent_identifiers'
-                    ),
-                    'publication_info': {
-                        'year': ref.get('year'),
-                        'journal_title': ref.get('journal_title'),
-                        'journal_volume': ref.get('journal_volume'),
-                        'page_start': ref.get('journal_page'),
-                        'artid': ref.get('journal_reference'),
-                    },
-                    'texkey': ref.get('texkey'),
-                    'titles': ref.get('titles'),
-                    'urls': ref.get('urls'),
-                },
-            }
-            mapped_references.append(reference)
+        filepath, reference_format=u'{title},{volume},{page}')
 
-    return mapped_references
+    result = []
+
+    for reference in force_list(references.get('references')):
+        rb = ReferenceBuilder()
+        mapping = [
+            ('author', rb.add_refextract_authors_str),
+            ('collaboration', rb.add_collaboration),
+            ('doi', rb.add_uid),
+            ('hdl', rb.add_uid),
+            ('isbn', rb.add_uid),
+            ('journal_reference', rb.set_pubnote),
+            ('linemarker', rb.set_label),
+            ('misc', rb.add_misc),
+            ('publisher', rb.set_publisher),
+            ('reportnumber', rb.add_report_number),
+            ('texkey', rb.set_texkey),
+            ('title', rb.add_title),
+            ('url', rb.add_url),
+            ('year', rb.set_year),
+        ]
+
+        for field, method in mapping:
+            for el in force_list(reference.get(field)):
+                if el:
+                    method(el)
+
+        result.append(rb.obj)
+
+    return result
--- a/tests/unit/workflows/test_workflows_tasks_refextract.py
+++ b/tests/unit/workflows/test_workflows_tasks_refextract.py
@@ -94,7 +94,12 @@ def test_extract_journal_info_handles_year_an_empty_string():
 
 
 def test_extract_references_handles_unicode():
+    schema = load_schema('hep')
+    subschema = schema['properties']['references']
+
     filename = pkg_resources.resource_filename(
         __name__, os.path.join('fixtures', '1704.00452.pdf'))
 
-    extract_references(filename)
+    result = extract_references(filename)
+
+    assert validate(result, subschema) is None


zzacharo:
We can start reviewing this. There are things that are in purpose not refactored yet as we need to discuss if anything is missing or not taken in consideration.

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -86,49 +87,45 @@ def extract_journal_info(obj, eng):
 
 def extract_references(filepath):
     """Extract references from PDF and return in INSPIRE format."""
+
+    def field_builder(builder):
+        def _build_field(value_l):
+            for value in value_l:
+                builder(value)
+        return _build_field
```
jacquerie: This high order function is the partial application of `map` to `builder`. It should be possible to rewrite it in terms of `functools.partial` and `map`.

tests/unit/workflows/test_workflows_tasks_refextract.py
```diff
@@ -24,15 +24,16 @@
...
 from inspire_schemas.utils import load_schema
+
 from inspirehep.dojson.utils import validate
 from inspirehep.modules.workflows.tasks.refextract import (
     extract_journal_info,
     extract_references,
 )
-
-from mocks import MockEng, MockObj
+from inspirehep.utils.record import get_value
```
jacquerie: As for the imports, the only thing that should change here is that you should add the last one, and everything else was ok before.

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -86,49 +87,45 @@ def extract_journal_info(obj, eng):
...
+                ('publisher', field_builder(rb.set_publisher)),
+                ('journal_reference', field_builder(rb.set_pubnote)),
+                ('journal_year', field_builder(rb.set_year)),
+                ('year', field_builder(rb.set_year)),
+                ('url', field_builder(rb.add_url)),
+            ]
+
+            for field, method in mapping:
+                if field in ref:
+                    method(get_value(ref, '{}'.format(field)))
```
jacquerie: `'{}'.format(field)` when `field` is a string is just `field`!

inspirehep/modules/workflows/tasks/refextract.py
jacquerie: Perhaps you don't even need it...

inspirehep/modules/workflows/tasks/refextract.py
zzacharo: oups yep forgot to refactor after change something!

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -86,49 +87,45 @@ def extract_journal_info(obj, eng):
...
-                    },
-                    'texkey': ref.get('texkey'),
-                    'titles': ref.get('titles'),
-                    'urls': ref.get('urls'),
-                },
-            }
-            mapped_references.append(reference)
+
+    # TODO: these keys are returned but we don't need them .
+    # excluded_keys = ['linemarker', 'journal_volume', 'journal_title', 'journal_page']
```
michamos: That's not true. `linemarker` should go to [`label`](https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/elements/reference.yml#L94). The others are redundant with `journal_reference` indeed.

zzacharo: ok perfect that's why I left the comment because I didn't know what `linemarker` was. I will add it!

inspirehep/modules/workflows/tasks/refextract.py
zzacharo: Do you have a more elegant solution based on this mapping that I use? I am trying to find an elegant way of doing this.

michamos:
Other keys that are extracted and not added, but should be (from https://github.com/inspirehep/refextract/blob/master/refextract/references/config.py#L57):
- `isbn`
- `hdl`

zzacharo:
@michamos Can you point me to a pdf that these fields should be extracted?

michamos:
http://arxiv.org/pdf/1705.00994.pdf: [10] has a HDL, [30] has an ISBN.

zzacharo:
@michamos thanks! Any other field missing because from https://github.com/inspirehep/refextract/blob/master/refextract/references/config.py#L57 there are also some other fields that I don't include?

michamos:
The two that are missing are `recid` and `urldesc`. The first one is not needed as it is not `refextract`s resonsibility to associate records to references but the matcher, and the second one seems not to work, so you can skip it.

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -86,49 +86,43 @@ def extract_journal_info(obj, eng):
 
 def extract_references(filepath):
     """Extract references from PDF and return in INSPIRE format."""
+
+    def field_builder(builder, value_l):
+        for value in value_l:
```
jmartinm: so refextract always returns lists?

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -86,49 +86,43 @@ def extract_journal_info(obj, eng):
...
-                    'urls': ref.get('urls'),
-                },
-            }
-            mapped_references.append(reference)
+
+    if references:
+        for ref in references:
+            rb = ReferenceBuilder()
+            mapping = [
+                ('raw_ref', partial(field_builder, rb.add_raw_reference)),
```
jmartinm: Would something like `map(ref.get('raw_ref'), rb.add_raw_reference)` also work?

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -86,49 +86,43 @@ def extract_journal_info(obj, eng):
...
+                ('collaboration', partial(field_builder, rb.add_collaboration)),
+                ('doi', partial(field_builder, rb.add_uid)),
+                ('texkey', partial(field_builder, rb.set_texkey)),
+                ('reportnumber', partial(field_builder, rb.add_report_number)),
+                ('title', partial(field_builder, rb.add_title)),
+                ('misc', partial(field_builder, rb.add_misc)),
+                ('publisher', partial(field_builder, rb.set_publisher)),
+                ('journal_reference', partial(field_builder, rb.set_pubnote)),
+                ('journal_year', partial(field_builder, rb.set_year)),
+                ('year', partial(field_builder, rb.set_year)),
```
jmartinm: so both `year` and `journal_year` are the same?

inspirehep/modules/workflows/tasks/refextract.py
zzacharo: it seems so.

inspirehep/modules/workflows/tasks/refextract.py
zzacharo: which is the `map` function? But as I understand from map it will.

inspirehep/modules/workflows/tasks/refextract.py
zzacharo: As far as I observed I saw that the same value is produced in these fields. @michamos is this correct?

michamos: `year` seems a bit more generic, as book references have a `year` but no `journal_year` (coming from a hardcoded knowledge base).

zzacharo: So, setting the value of both fields to references `year` field is incorrect?

michamos: It's useless because the second call to `set_year` will overwrite the value set by the first call.

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -86,49 +86,43 @@ def extract_journal_info(obj, eng):
 
 def extract_references(filepath):
     """Extract references from PDF and return in INSPIRE format."""
+
+    def field_builder(builder, value_l):
+        for value in value_l:
+            builder(value)
```
jacquerie: Now this higher order function is just `map`: you get a function and a list, and you apply the function to each element of the list.

inspirehep/modules/workflows/tasks/refextract.py
jacquerie: In fact, this means that you have to add a `set_journal_year` call to the reference builder.

tests/unit/workflows/test_workflows_tasks_refextract.py
```diff
@@ -93,8 +94,11 @@ def test_extract_journal_info_handles_year_an_empty_string():
...
-def test_extract_references_handles_unicode():
+def test_extract_references_outputs_valid_data():
     filename = pkg_resources.resource_filename(
         __name__, os.path.join('fixtures', '1704.00452.pdf'))
 
-    extract_references(filename)
+    extracted_references = extract_references(filename)
+    ref_schema = get_value(load_schema('hep'), 'properties.references')
+
+    validate(extracted_references, ref_schema)
```
jacquerie: You're testing more than before, but you could be testing even more: as @michamos suggested (perhaps only IRL?), the fact that the first reference is correctly extracted.

michamos: Not only IRL: https://github.com/inspirehep/inspire-next/pull/2446#pullrequestreview-44556610

inspirehep/modules/workflows/tasks/refextract.py
michamos: another option is to amend the builder: if the reference already contains a `publication_info/journal_title`, use `publication_info/year`, otherwise `imprint/date`.

inspirehep/modules/workflows/tasks/refextract.py
zzacharo: Should I refactor this on my own or will you do it?

tests/unit/workflows/test_workflows_tasks_refextract.py
zzacharo: So, check the response of the first reference extracted by the given pdf is the reference record that we expect to be exported by the builder?

inspirehep/modules/workflows/tasks/refextract.py
zzacharo: I can do this! But this should be done in different PR as I will extend `ReferenceBuilder` ? Or is fine in this PR?

jacquerie:
> It might also be useful to have a test that takes, say, the first reference of a PDF and checks that all relevant fields of `references` in the record are populated.

I tried this, but the output of `pdftext` is too inconsistent to provide an acceptable test.

jacquerie:
Blocked by the fact that https://github.com/inspirehep/inspire-docker/pull/47 is not actually fixing the build of `inspire-docker`, so the build of `inspire-next` is also broken...

PR #2447, opened 2017-06-16
workflows: add indexes for on delete cascade

kaplun:

Adds indexes to columns which are subject of a foreign key with
on delete cascade.


kaplun:
This is because:
> Lesson learnt today with @jmartinm on DBs: if you have a foreign key with `on cascade delete`, the referred table should have an index on the corresponding column. Otherwise deleting rows from the referring table will take forever.

PR #2423, opened 2017-06-12
Add Merger to the Workflow

rikirenz:
Add Merger to the Workflow

## Description
This PR wants to add a step to the workflow in order to use the `inspire-json-merger` project to merge articles. 

## Related Issue
This is the related issue: #2324 and these are the google docs that we produced: 
- JSON Merger -  Proposed solution:
https://docs.google.com/document/d/1DG2FXo0W9w1c05rnr7YTviWwzOvYk2KfeblXPWwxNZw/edit?usp=sharing
- JSON Merger - Root retrieval:
https://docs.google.com/document/d/1BDG14AsY5FEtGXX10uR6U8CNTJnZuLXldgm6K4uOrCM/edit?usp=sharing
- Spreadsheet with all the rules: https://docs.google.com/spreadsheets/d/1IKgKNgp0uStX07Uv7FECpkGX5J6xdZ57kErhS4Ra7BQ/edit?usp=sharing

(Closes #2322, #2375, #2294, #2527)
(Addresses #2696 still missing the editor part)

## Motivation and Context
This change is required by curators in order to preserve the highest number of changes for a given article

## Checklist:

- [x] I have all the information that I need (if not, move to `RFC` and look for it).
- [x] I linked the related issue(s) in the corresponding commit logs.
- [x] I wrote [good commit log messages](https://github.com/torvalds/subsurface-for-dirk/blob/5f15ad5a86ada3c5e574041a5f9d85235322dabb/README#L92-L119).
- [x] My code follows the code style of this project.
- [x] I've added any new docs if API/utils methods were added.
- [x] I have updated the existing documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #2448, opened 2017-06-16
Refactor workflows test

david-caro:
Use `conftest.py` for the reusable fixtures and add a `helpers`
module with all the functions that are reusable.

Trying to add more features to the callbacks module in the workflows
found out really hard to add new tests for it.

### DIFF
--- a/tests/integration/workflows/test_arxiv_workflow.py
+++ b/tests/integration/workflows/test_arxiv_workflow.py
@@ -44,49 +44,8 @@ from invenio_workflows import (
     workflow_object_class,
 )
 
-from inspire_dojson.hep import hep
-from inspirehep.factory import create_app
+from inspirehep.dojson.hep import hep
 from inspirehep.modules.converter.xslt import convert
-from inspirehep.modules.workflows.models import (
-    WorkflowsAudit,
-    WorkflowsPendingRecord,
-)
-
-
-@pytest.fixture(autouse=True)
-def cleanup_workflows_tables(small_app):
-    with small_app.app_context():
-        obj_types = (
-                WorkflowsAudit.query.all(),
-                WorkflowsPendingRecord.query.all(),
-                workflow_object_class.query(),
-        )
-        for obj_type in obj_types:
-            for obj in obj_type:
-                obj.delete()
-
-        db.session.commit()
-
-
-@pytest.fixture
-def workflow_app():
-    app = create_app(
-        BEARD_API_URL="http://example.com/beard",
-        DEBUG=True,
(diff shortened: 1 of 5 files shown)


jacquerie:
Victim of #2385

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -44,339 +37,19 @@
...
-    assert "approved" not in obj.extra_data
-
-    return workflow_uuid, eng, obj
+from mocks import (
+    fake_download_file,
+    fake_beard_api_request,
+    fake_magpie_api_request,
+    fake_beard_api_block_request,
+    fake_refextract_extract_references_from_file,
+)
```
rikirenz: Are you sure that this is going to work? How you can import things here if there are not inside a module

david-caro: I added them to the path in conftest.py (by tweaking the sys.path variable), I agree though that for me it's nicer to have them inside inspirehep module, something like `from inspirehep.testlib.mocks import ...`.
Though I'm following the lead of the unit tests, that already do it (no strong opinion, so don't want to start a debate about it).

rikirenz: I agree :+1:

david-caro:
Will wait until #2423 is merged, as that one also refactors a bit the current tests code.

rikirenz:
I would  like to work on this PR. @david-caro could you list all the improvements that you wanted to do?
Plus I have seen this error https://travis-ci.org/inspirehep/inspire-next/jobs/243758622 I am not able to understand this `calls` module

rikirenz:
@david-caro it seems that from the moment that you created the PR I cannot assign the review to you. Anyway it would be nice have your opinion :smile:

tests/integration/workflows/helpers/apps.py
```diff
@@ -0,0 +1,48 @@
...
+        PRODUCTION_MODE=True,
+        LEGACY_ROBOTUPLOAD_URL=(
+            'http://localhost:1234'
+        ),
+        MAGPIE_API_URL="http://example.com/magpie",
+        WTF_CSRF_ENABLED=False,
+    )
+
+    with app.app_context():
+        yield app
```
david-caro: This being a fixture makes no sense having it here, as it will not be loaded as such by pytest. Hmm... we should not make it a function anyhow, as we want to be able to use the `session` option in the fixture so it's loaded only once per run right? (it's quite slow...)

Issue #2428, opened 2017-06-12
global: bump inspire-schemas to version ~35.0


jacquerie:
See: https://github.com/inspirehep/inspire-schemas/pull/170

PR #2449, opened 2017-06-16
global: bump inspire-schemas to version ~35.0

jacquerie:
## Related Issue
Closes https://github.com/inspirehep/inspire-next/issues/2428

## Checklist:

### DIFF
--- a/inspirehep/dojson/experiments/rules.py
+++ b/inspirehep/dojson/experiments/rules.py
@@ -57,12 +57,12 @@ def dates(self, key, values):
 def experiment(self, key, values):
     """Populate the ``experiment`` key.
 
-    Also populates the ``accelerator`` and the ``insitution`` key
+    Also populates the ``accelerator`` and the ``institutions`` key
     through side effects.
     """
     experiment = self.get('experiment', {})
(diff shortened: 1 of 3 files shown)


jacquerie:
Since `inspire-schemas==34.x.y` (I don't know `x` and `y`) it requires `six` as a setup dependency. In order to have it installed in the Docker image we need to have a successful build and upload to DockerHub, which is currently broken with the following failure: https://travis-ci.org/inspirehep/inspire-docker/builds/243212256

https://github.com/inspirehep/inspire-next/pull/2450 should fix this failure, so tomorrow a good image will be uploaded and this PR will be unblocked.

jacquerie:
The problem here is that `inspire-schemas` has long been non-installable through `pip` in a pristine environment since the `six` dependency was added to its `setup.py`, which happened a while ago. We were not seeing this failure because I was building binary wheels for the previous versions.

I pushed a wheel of `inspire-schemas==35.0.1` to unblock this PR, and I will probably do the same for the next two schema version, but I believe that this should be solved in `inspire-schemas`, @david-caro...

PR #2457, opened 2017-06-20
tests: get rid of es waits by adding health checks

david-caro:
* This requires getting docker-compose > 1.10
* This makes the dependent containers wait for the dependencies to be
  in a healthy state before starting.
* As we are sure now that ES is already healthy before running the
  tests, there's no point on doing a sleep to wait for it to be up.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -49,29 +49,48 @@ services:
       - static
     ports:
       - "5000:5000"
-    links:
-      - database
-      - indexer
-      - rabbitmq
-      - redis
+    depends_on:
+      database:
+        condition: service_healthy
+      indexer:
+        condition: service_healthy
+      rabbitmq:
+        condition: service_healthy
+      redis:
+        condition: service_healthy
     environment:
       - APP_DEBUG=True
       - APP_ASSETS_DEBUG=True
+
   worker:
     extends:
       service: service_base
     command: celery worker -E -A inspirehep.celery --loglevel=INFO --purge
     volumes_from:
       - static
-    links:
-      - database
-      - indexer
-      - rabbitmq
-      - redis
+    depends_on:
+      database:
+        condition: service_healthy
+      indexer:
+        condition: service_healthy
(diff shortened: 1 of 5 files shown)


david-caro:
@rikirenz @ammirate with this you don't have to add again the `sleep(10)` to the tests, though you might have to upgrade your docker-compose.

jacquerie:
I have a feeling that one of the least frequent sources of flakiness was caused by ES/some other service not being ready in time for the tests... this PR should have fixed that :+1:

docker-compose.yml
```diff
@@ -81,14 +100,38 @@ services:
...
       service: database
     volumes:
       - "${DOCKER_DATA}/tmp/inspirehep_db/data:/var/lib/postgresql/data/pgdata"
+    healthcheck:
+      timeout: 5s
+      interval: 5s
+      retries: 5
+      test:
+        - "CMD-SHELL"
+        - "export PGPASSWORD=dbpass123 && psql --quiet --no-align --tuples-only --host 127.0.0.1 --username inspirehep --dbname inspirehep -c 'SELECT 1'"
```
jacquerie: I wonder if there's anything simpler than this... on my local system I have a `pg_isready` command.

jacquerie: Yeah, it was added in 9.3, so we have it as well here: https://www.postgresql.org/docs/9.3/static/app-pg-isready.html

jacquerie:
> I think that one of the (least frequent) sources of flakiness was ES/some other service not being ready in time for the tests... this PR should have fixed that :+1:

Well, ironically the same kind of build failure that I thought was happening for this reason happened on the merge build of this very PR. So no, this wasn't improved, but this solution is definitely better than the `sleep` that was there before.

docker-compose.yml
david-caro: I got that command from the docker image `healthcheck/postgres` on dockerhub, did not spent much time on it.

PR #2458, opened 2017-06-20
travis: use the newer 'edge' runners

david-caro:


david-caro:
**DON'T MERGE** this is just to test if we have any issues with the new images, the new images will be automatically upgraded by travis without any change needed on our side.

jacquerie:
Ouch: https://travis-ci.org/inspirehep/inspire-next/jobs/245090635#L645

The `travis` user might no longer have uid 1000...

david-caro:
Being the sudo images, everything should be running as root no? Will try some stuff...

jacquerie:
> Being the sudo images, everything should be running as root no? Will try some stuff...

That's possible, although they could be using a user which is equivalent to root, like `mihai` on Nightly. In fact, I was getting the same error that disabled Nightly for a few days before I repaired the permissions...

david-caro:
user 2000! xd twice as good!

david-caro:
I think there's something strange going on on travis, it seems to be running on a different commit than the last here, as it's not exporting the new var.

jacquerie:
> I think there's something strange going on on travis, it seems to be running on a different commit than the last here, as it's not exporting the new var.

Mh, doesn't appear to be the case anymore. Your invocation of `id` is definitely there: https://travis-ci.org/inspirehep/inspire-next/jobs/245120873#L625

david-caro:
This requires this other pr https://github.com/inspirehep/inspire-docker/pull/47 to get fixed.

Issue #2462, opened 2017-06-21
Phonetic block generation produces non-compatible records 


rikirenz:
In the module `inspirehep/modules/authors/receivers.py` if this line is executed: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/authors/receivers.py#L94 we produce a non-compatible record.

## Expected Behavior
It should put an empty value instead of `None`

## Current Behavior
It puts an `None`

## Steps to Reproduce (for bugs)
You should try to execute this code, and the error it will appear:
``` Python
from __future__ import absolute_import, division, print_function

import httpretty
import mock
import pytest
from flask import current_app

from jsonschema import validate as jsonschema_validate

from inspire_schemas.utils import load_schema
from inspire_schemas.utils import LocalRefResolver

from inspirehep.modules.authors import receivers

@pytest.mark.httpretty
def test_phonetic_block_generation_broken():
    extra_config = {
        "BEARD_API_URL": "http://example.com/beard",
    }

    with mock.patch.dict(current_app.config, extra_config):
        httpretty.register_uri(
            httpretty.POST,
            "{base_url}/text/phonetic_blocks".format(
                base_url=current_app.config.get('BEARD_API_URL')),
            content_type="application/json",
            body='{"phonetic_blocks": {}}',
            status=200)

        json_dict = {
            "authors": [{
                "full_name": "** NOT VALID **"
            }]
        }

        receivers.assign_phonetic_block(json_dict)

        assert json_dict['authors'][0]['signature_block'] is None

        def validate(instance, schema):
            return jsonschema_validate(
                instance, schema, resolver=LocalRefResolver('', {}))

        schema = load_schema('hep')
        subschema = schema['properties']['authors']['items']['properties']['signature_block']

        assert validate(
            json_dict['authors'][0]['signature_block'],
            subschema
        ) is None

``` 

## Context
I was working on this PR:https://github.com/inspirehep/inspire-next/pull/2423 and at certain point a received a validation error related to the code that I linked

jacquerie:
Ah, makes sense. The schema is expecting a string but we try to put a `null` there. We can fix this now, before #2423.

PR #2464, opened 2017-06-21
authors: fix the signature_block assignment 

rikirenz:
## Description        
The function `assign_phonetic_block` is handling the case missing phonetic blocks putting `None`. This kind of behavior is not schema compatible

## Related Issue
Closes https://github.com/inspirehep/inspire-next/issues/2462


michamos:
cannot you just leave out`signature_block` in that case? at some point in the future (inspirehep/inspire-schemas#105) empty strings won't be allowed any more.

rikirenz:
of course I can... Some tests are going to fail but we can fix them in this PR

Issue #2466, opened 2017-06-22
workfrows.callbacks: Add negative tests


david-caro:
Currently we only test positive results from legacy, we should test also a couple errors just in case.

PR #2467, opened 2017-06-22
workflows.callbacks: handle robotupload errors

### LINKED ISSUE #2083: Crash when double-sending callback
Apparently there is a way in which legacy sends 2 feedbacks to labs (e.g. first about a failure, and then about a success) and the second fails for DB integrity error:

https://sentry.cern.ch/inspire-sentry/inspire-labs/group/820620/

```python
Stacktrace (most recent call last):

  File "flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "flask/app.py", line 1517, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "inspirehep/modules/workflows/views/callback.py", line 203, in robotupload_callback
    db.session.commit()
  File "sqlalchemy/orm/scoping.py", line 157, in do
    return getattr(self.registry(), name)(*args, **kwargs)
  File "sqlalchemy/orm/session.py", line 864, in commit
    self.transaction.commit()
  File "sqlalchemy/orm/session.py", line 455, in commit
    self._prepare_impl()
  File "sqlalchemy/orm/session.py", line 435, in _prepare_impl
    self.session.flush()
  File "sqlalchemy/orm/session.py", line 2080, in flush
    self._flush(objects)
  File "sqlalchemy/orm/session.py", line 2198, in _flush
    transaction.rollback(_capture_exception=True)
  File "sqlalchemy/util/langhelpers.py", line 60, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
  File "sqlalchemy/orm/session.py", line 2162, in _flush
    flush_context.execute()
  File "sqlalchemy/orm/unitofwork.py", line 373, in execute
    rec.execute(self)
  File "sqlalchemy/orm/unitofwork.py", line 532, in execute
    uow
  File "sqlalchemy/orm/persistence.py", line 178, in save_obj
    mapper, table, insert)
  File "sqlalchemy/orm/persistence.py", line 771, in _emit_insert_statements
    execute(statement, multiparams)
  File "sqlalchemy/engine/base.py", line 914, in execute
    return meth(self, multiparams, params)
  File "sqlalchemy/sql/elements.py", line 323, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "sqlalchemy/engine/base.py", line 1010, in _execute_clauseelement
    compiled_sql, distilled_params
  File "sqlalchemy/engine/base.py", line 1146, in _execute_context
    context)
  File "sqlalchemy/engine/base.py", line 1341, in _handle_dbapi_exception
    exc_info
  File "sqlalchemy/util/compat.py", line 203, in raise_from_cause
    reraise(type(exception), exception, tb=exc_tb, cause=cause)
  File "sqlalchemy/engine/base.py", line 1139, in _execute_context
    context)
  File "sqlalchemy/engine/default.py", line 450, in do_execute
    cursor.execute(statement, parameters)
```

This is NOT a full blocker to deployment...

### LINKED ISSUE #2397: Webcoll feedback from legacy seems to be ignored
## Expected Behavior
Holding pen entry: https://labs.inspirehep.net/api/holdingpen/649811 should be in status completed.

## Current Behavior
Instead it is showing `"_message": "Waiting for webcoll.",` however webcoll on legacy has already communicated to Labs the record is there.

david-caro:
Now the errors from the callback results from legacy are properly
handled (at least, more usefully).

*NOTE*: The tests will come after the workflow tests refactor.
        See #2466

When we receive an error from legacy, now the workflow is set to
'ERROR' status and the callback error message is added to the
`_error_message` so it gets shown in the holdingpen.

Closes #2083 and closes #2397.

Now it will set the workflow in error state and put the message of the
robotupload error on it so it can be viewed from the ui and restarted
if needed.
Also avoids using the recid `-1` completely, leaving the pending
records table coherent.

### DIFF
--- a/inspirehep/modules/workflows/views/callback.py
+++ b/inspirehep/modules/workflows/views/callback.py
@@ -31,10 +31,10 @@ from flask import Blueprint, jsonify, request, current_app
 from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
 
 from invenio_db import db
-from invenio_workflows import workflow_object_class
+from invenio_workflows import workflow_object_class, ObjectStatus
+from invenio_workflows.errors import WorkflowsMissingObject
 
 from inspirehep.modules.workflows.models import WorkflowsPendingRecord
-
 
 blueprint = Blueprint(
     'inspire_workflows',
@@ -67,8 +67,9 @@ def _continue_workflow(workflow_id, recid, result=None):
     """
     result = result if result is not None else {}
     base_url = _get_base_url()
-    workflow_object = workflow_object_class.get(workflow_id)
-    if not workflow_object:
+    try:
+        workflow_object = workflow_object_class.get(workflow_id)
+    except WorkflowsMissingObject:
         current_app.logger.error(
             'No workflow object with the id %s could be found.',
             workflow_id,
@@ -83,8 +84,9 @@ def _continue_workflow(workflow_id, recid, result=None):
     workflow_object.extra_data['recid'] = recid
     workflow_object.extra_data['callback_result'] = result
     workflow_object.save()
-
+    db.session.commit()
     workflow_object.continue_workflow(delayed=True)
+
     return True
 
 
@@ -100,11 +102,39 @@ def _find_and_continue_workflow(workflow_id, recid, result=None):
             workflow_id,
         )
         return {
-            'status': 'failed',
+            'success': False,
             'message': 'workflow with id %s not found.' % workflow_id,
         }
 
-    return {'status': 'succeeded'}
+    return {
+        'success': True,
+        'message': 'workflow with id %s continued.' % workflow_id,
+    }
+
+
+def _put_workflow_in_error_state(workflow_id, error_message, result):
+    try:
+        workflow_object = workflow_object_class.get(workflow_id)
+    except WorkflowsMissingObject:
+        current_app.logger.error(
+            'No workflow object with the id %s could be found.',
+            workflow_id,
+        )
+        return {
+            'success': False,
+            'message': 'workflow with id %s not found.' % workflow_id,
+        }
+
+    workflow_object.status = ObjectStatus.ERROR
+    workflow_object.extra_data['callback_result'] = result
+    workflow_object.extra_data['_error_msg'] = error_message
+    workflow_object.save()
+    db.session.commit()
+
+    return {
+        'success': True,
+        'message': 'workflow %s updated with error.' % workflow_id,
+    }
 
 
 @blueprint.route('/workflows/webcoll', methods=['POST'])
@@ -113,6 +143,16 @@ def webcoll_callback():
 
     Expects the request data to contain a list of record ids in the
     recids field.
+
+    Example:
+        An example of callback::
+
+            $ curl \\
+                http://web:5000/callback/workflows/webcoll \\
+                -H "Host: localhost:5000" \\
+                -F 'recids=1234'
+
+
     """
     recids = dict(request.form).get('recids', [])
     response = {}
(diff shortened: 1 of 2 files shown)


inspirehep/modules/workflows/views/callback.py
```diff
@@ -104,7 +103,32 @@ def _find_and_continue_workflow(workflow_id, recid, result=None):
...
-    return {'status': 'succeeded'}
+    return {
+        'status': 'succeeded',
+        'message': 'workflow with id %s continued.' % workflow_id,
+    }
+
+
+def _put_workflow_in_error_state(workflow_id, error_message, result):
+    workflow_object = workflow_object_class.get(workflow_id)
+    if not workflow_object:
```
jmartinm: I am not sure about this, if I try in the shell, if `get` fails an exception is thrown:

```
In [19]: workflow_object_class.get(12323)
WorkflowsMissingObject                    Traceback (most recent call last)
<ipython-input-19-45b961fd0548> in <module>()
----> 1 workflow_object_class.get(12323)

/opt/inspire/lib/python2.7/site-packages/invenio_workflows/api.pyc in get(cls, id_)
    154             except NoResultFound:
    155                 raise WorkflowsMissingObject("No object for for id {0}".format(
--> 156                     id_
    157                 ))
    158             return cls(model)

WorkflowsMissingObject: No object for for id 12323
```

inspirehep/modules/workflows/views/callback.py
```diff
@@ -104,7 +103,32 @@ def _find_and_continue_workflow(workflow_id, recid, result=None):
...
+        )
+        return {
+            'status': 'failed',
+            'message': 'workflow with id %s not found.' % workflow_id,
+        }
+
+    workflow_object.status = ObjectStatus.ERROR
+    workflow_object.extra_data['callback_result'] = result
+    workflow_object.extra_data['_error_message'] = error_message
+    workflow_object.save()
```
jmartinm: In all other occurrences of `workflow_object.save()` it is always followed by `db.session.commit()` (except on this file).

I would say it is needed since `save()` just adds to the session.

inspirehep/modules/workflows/views/callback.py
```diff
@@ -104,7 +103,32 @@ def _find_and_continue_workflow(workflow_id, recid, result=None):
...
+    return {
+        'status': 'succeeded',
```
jmartinm: Is it a success?

inspirehep/modules/workflows/views/callback.py
```diff
@@ -170,55 +263,32 @@ def robotupload_callback():
...
     workflow_id = request_data.get('nonce', '')
     results = request_data.get('results', [])
-    response = {}
+    responses = {}
     for result in results:
         recid = int(result.get('recid'))
 
-        if recid in response:
+        if recid in responses:
+            # this should never happen
```
jmartinm: These comments are always freaky

inspirehep/modules/workflows/views/callback.py
david-caro: done

inspirehep/modules/workflows/views/callback.py
david-caro: there's one place where it does not: https://github.com/inveniosoftware-contrib/invenio-workflows-ui/blob/master/invenio_workflows_ui/api.py#L197

But I think that might be a bug...
Done

inspirehep/modules/workflows/views/callback.py
david-caro: Putting the workflow in error state was a success at this point, this function does not know about the 'higher' intention, just about putting a workflow in error state. It fails when it's unable to find the workflow to put in error state.

inspirehep/modules/workflows/views/callback.py
david-caro: +💯 
But it's true... though I don't trust much the callback data, that's why it's still there the if with a custom log statement.

inspirehep/modules/workflows/views/callback.py
kaplun: I think @jmartinm referred to the spelling: `succeeded` Vs. `success` :+1:

inspirehep/modules/workflows/views/callback.py
```diff
@@ -153,12 +179,79 @@ def webcoll_callback():
...
         db.session.delete(pending_record)
         db.session.commit()
-        if response[recid].get('status') == 'failed':
-            continue
 
     return jsonify(response)
 
 
+def _robotupload_has_error(result):
+    recid = int(result.get('recid'))
```
kaplun: This will raise exception in case there's no `recid`.

inspirehep/modules/workflows/views/callback.py
```diff
@@ -153,12 +179,79 @@ def webcoll_callback():
...
+    if already_pending_ones:
+        current_app.logger.warning(
+            'The record %s was already found on the pending list.',
+            recid
+        )
+        response = {
+            'status': 'failed',
+            'message': 'Recid %s already in pending list.' % recid,
+        }
+        return response
```
kaplun: Is it a critical error to the point of making the workflow to fail? Isn't this triggered whenever legacy sends twice the callback? Can't we simply ignore this?

inspirehep/modules/workflows/views/callback.py
david-caro: It should always have a recid according to the interface definition, if not, then it should fail yep.

inspirehep/modules/workflows/views/callback.py
david-caro: This is triggered whenever we get a robotupload for a record that is already pending to get the webcall, I don't think we should ignore it as that possibly means that the first try got messed up somewhere and the second failed to detect that it was already there, maybe we should define more in detail all the possible cases.

david-caro:
Btw. I changed the format of the responses to mimic the way legacy tells us if it succeded or not (using the `"success": True` field instead of status.
I hope it's more readable now...

david-caro:
@kaplun is there anything missing here?

kaplun:
I am not sure Let's merge and see if it works! :dancer:

jacquerie:
> It is difficult to know just by looking at the code and with no tests if the right thing is going to happen though

I agree 100%... I hope that you're correct, @david-caro ! I can't say I understand this code at all.

I'm going to deploy this tomorrow morning, so that @annetteholtkamp will give us feedback ASAP and for the whole day. If by the end of the day we're not confident that this can handle the night's harvest, we can just revert.

jacquerie:
This appears to have handled perfectly well tonight's harvest.

PR #2468, opened 2017-06-22
travis: restore the permissions to the correct uid

david-caro:
Currently we are leaving the code directory as owned by id 1000,
though the travis runners use uid 2000. That prevents coverage from
reading the coverage file.

* This should fix the current coverage issues where the report does
  not show up.


jacquerie:
Ah, right, yesterday I merged https://github.com/inspirehep/inspire-next/pull/2446 without coverage thinking that some problem happened in the connection to Coveralls, didn't think about this...

Issue #2478, opened 2017-06-24
global: bump inspire-schemas to version ~40.0


jacquerie:
See: https://github.com/inspirehep/inspire-schemas/pull/154

PR #2479, opened 2017-06-24
global: bump inspire-schemas to version ~40.0

jacquerie:
## Related Issue
Closes https://github.com/inspirehep/inspire-next/issues/2478

## Checklist:


tests/unit/dojson/test_dojson_institutions.py
```diff
@@ -31,77 +31,6 @@
...
-        '   <subfield code="d">32.540776</subfield>'
-        '   <subfield code="d">15.561010</subfield>'
-        '</datafield>'
-    )  # record/1442294
-
-    result = institutions.do(create_record(snippet))
-
-    assert 'location' not in result
-
-
```
jacquerie: This is wrong... these should become tests on `address`.

inspirehep/modules/records/receivers.py
```diff
@@ -240,55 +239,6 @@ def populate_abstract_source_suggest(sender, json, *args, **kwargs):
...
-                    'ICN': ICN,
-                    'department': json.get('department'),
-                    'institution': institution,
-                    'institution_acronym': institution_acronym,
-                    'legacy_ICN': legacy_ICN,
-                },
-            },
-        })
-
-
```
jacquerie: I should also remove the corresponding mapping.

jacquerie:
Tests pass, but this is not yet done!

tests/unit/dojson/test_dojson_institutions.py
jacquerie: In the end I just wrote a simpler test that doesn't test all of the above because I don't believe that we're going to regress on this.

PR #2486, opened 2017-06-26
migrator: avoid indefinite locking of continuous migration

jmartinm:
* Adds expiration time to the redis lock to avoid a situation where
  the process crashes and the lock will remain.

* Sets auto_renewal=True so that while a process has the lock acquired
  and is executing code, the lock will not expire.

## Checklist:


jmartinm:
The reason why the context manager was not used is because it is then not possible to execute `blocking=False`

PR #2494, opened 2017-06-27
docs: add psql documentation

rikirenz:
* Adds some useful tips in order to query the psql database

Signed-off-by: Riccardo Candido <riccardo.candido@gmail.com>


rikirenz:
Let me know if you think is useful :wink: @jacquerie

docs/postgresql.rst
```diff
@@ -0,0 +1,94 @@
...
+.. code-block:: bash
+
+    (virtualenv)[inspirehep_web]$ psql -h inspirenext_database_1 inspirehep inspirehep
+    Password for user inspirehep: dbpass123
+
+.. code-block:: sql
+
+    psql (9.2.18, server 9.4.5)
+    WARNING: psql version 9.2, server version 9.4.
+             Some psql features might not work.
```
jacquerie: You can avoid this by running directly `docker-compose exec database psql -U inspirehep`, instead of following the steps above. Note "`exec`" instead of "`run`", as you want to reuse the same container that you have already running.

docs/postgresql.rst
```diff
@@ -0,0 +1,94 @@
...
+
+
+5. Search a record with the pid
+--------------------------------
+
+Given the pid of a record you can obtain the record running this query:
+
+.. code-block:: sql
+
+    select * from pidstore_pid, records_metadata where records_metadata.id = pidstore_pid.object_uuid where pidstore_pid.id = YOUR_PID_ID;
```
jacquerie: Do you know about `get_db_record` ?

docs/postgresql.rst
```diff
@@ -0,0 +1,94 @@
...
+Now you have an interactive console to query the inspire SQL database
+
+
+4. Search a record with the uuid
+--------------------------------
+
+A list of useful commands:
+
+- ``\h`` this lists all the sql commands that you can run
+- ``\d`` this lists all the relations
```
jacquerie: FYI, this actually lists a superset of the relations, also containing the id sequences. The right command is `\dt`.

docs/postgresql.rst
rikirenz: Yes I know. This is just an example to query a table. The is that with sql you can query the data in the way that you prefer, for instance: using the like operator.

This documentation wants to show how to have access to the db. For sure you can obtain the same result using the functions that people developed in inspire-next but sometimes when you develop it is necessary to have more flexibility. This is the sense of having this peace of documentation.

jacquerie: > sometimes when you develop it is necessary to have more flexibility

Absolutely. These are all excellent considerations, which should be part of the documentation you just wrote!

docs/postgresql.rst
rikirenz: You are partially right! :smile: 
The command that you are suggesting is assuming that you have already a running container so you need anyway to run this command line at least: 
``` bash 
docker-compose run --rm web
```
and after you can run the command that you are suggesting, that is:
``` bash
docker-compose exec database psql -U inspirehep
```
So at the end you have to run 2 command lines in both cases plus there isn't any assumption about the fact that the containers are already up.
The way that you are suggesting anyway allows you to access the database without password and I do not like this because from the moment that we can document the password of our database we should do it.

docs/postgresql.rst
rikirenz: Actually in order to be very precise it shows also views. So I will update the description because I think that is valuable.

docs/postgresql.rst
```diff
@@ -0,0 +1,94 @@
...
+             Some psql features might not work.
+    Type "help" for help.
+
+    inspirehep=#
+
+
+Now you have an interactive console to query the inspire SQL database
+
+
+4. Search a record with the uuid
```
jmartinm: Note that you have this section duplicated.

rikirenz: My bad! :cry:

docs/postgresql.rst
rikirenz: I changed my idea. You are right there is no reason to show everything! I will replace `\d` with `\dt`

david-caro:
Btw. I just used this docs... and they were useful xd
👍

docs/postgresql.rst
jacquerie: Providing a list of steps is how you would explain something without caring that they _understand_ what they are doing: "If you want X to happen, do A, B, and C in this order". Documentation instead should help people understand what they are doing and why. It might miss some steps: these act as "exercises to the reader" and check their understanding of the core concepts. For example, if you run `docker-compose exec database psql -U inspirehep` without a running `database` container you get a pretty clear `ERROR: No container found for database_1`. At some point you have to assume the intelligence of the reader, otherwise all documentation would begin with "1. Turn your computer on." : )

jacquerie: > The way that you are suggesting anyway allows you to access the database without password and I do not like this because from the moment that we can document the password of our database we should do it.

I didn't understand this instead. You can access the DB without password because you're attaching yourself to a running container that has already authenticated against the database using the (cleartext) password contained in its access string. Having to type again the password does not increase security.

rikirenz: Of course this make sense. But still, give the password allows to open a connection to the database also in a situation that is not the one that we are assuming in this document. In this case in fact you have access to a container that is already authenticated with the database.
But in real life can happen that you do not have this so I think the information of the password is useful. Anyway I will refactor a bit in order to put everything we said and will submit the pr again

rikirenz:
I updated the doc in according with the last comment. So I changed from `WIP` to `ready_for_review`

docs/postgresql.rst
```diff
@@ -0,0 +1,94 @@
...
+    or submit itself to any jurisdiction.
+
+
+How to Connect to the PostgreSQL Database
+=========================================
+
+1. About
+--------
+
+This document specifies how to connect and query the inspire PostgreSQL database
```
kaplun: Can you detail a bit more? What is the context? Docker?

docs/postgresql.rst
```diff
@@ -0,0 +1,94 @@
...
+
+
+2. Run the web container
+------------------------
+
+The first step is run the web container
```
kaplun: Why? What happens when I am doing this?

ammirate:
Going to merge this!

Issue #2495, opened 2017-06-27
refextract: Improve normaized extracted references


zzacharo:
Create improved normalized references in order to increase the possibility of the reference to match a record.

Here you can consume the design document:
https://docs.google.com/document/d/1Y1SLnYc9vMdssstHbr4356t5Ww-HbHtQo8koRuEHgUc/edit#

PR #2496, opened 2017-06-27
WIP refextract: Improve normalized extracted references

zzacharo:
Adds a celery beat task to run periodically and build the journal-kb file.

Related issue: https://github.com/inspirehep/inspire-next/issues/2495

### DIFF
--- a/inspirehep/modules/refextract/tasks.py
+++ b/inspirehep/modules/refextract/tasks.py
@@ -1,0 +1,61 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Refextract tasks."""
+
+from __future__ import absolute_import, division, print_function
+
+import contextlib
+
+from celery import shared_task
+import codecs
+from flask import current_app
+
+from invenio_db import db
+
+
+@contextlib.contextmanager
+def file_open_for_editing(fpath):
+    fh = codecs.open(fpath, encoding='utf-8', mode='w')
+
+    yield fh
+
+    fh.close()
+
+
+@shared_task()
+def create_journal_kb_file():
+    querystring = "SELECT short_titles, title_variants " \
+                  "FROM records_metadata as r, " \
+                  "json_array_elements(r.json -> 'short_titles') as short_titles, " \
+                  "json_array_elements(r.json -> 'title_variants') as title_variants"
+
+    results = db.session.execute(querystring)
+
+    if results.rowcount > 1:
+        with file_open_for_editing(current_app.config['REFEXTRACT_JOURNAL_KB_PATH']) as fh:
+
+            for row in results:
+                alt_title = row['title_variants'].get('title')
+                short_title = row['short_titles'].get('title')
+                fh.write(u'{}---{}'.format(alt_title, short_title))
+                fh.write('\n')
--- a/inspirehep/modules/workflows/tasks/refextract.py
+++ b/inspirehep/modules/workflows/tasks/refextract.py
@@ -83,10 +83,13 @@ def extract_journal_info(obj, eng):
 
 
 @timeout(5 * 60)
-def extract_references(filepath):
+def extract_references(filepath, custom_kbs_file=None):
     """Extract references from PDF and return in INSPIRE format."""
     references = extract_references_from_file(
-        filepath, reference_format=u'{title},{volume},{page}')
+        filepath,
+        override_kbs_files=custom_kbs_file,
+        reference_format=u'{title},{volume},{page}'
+    )
 
     result = []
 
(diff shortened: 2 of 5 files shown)


zzacharo:
Is ready to check the task code. We should configure the schedule of the task (how frequently should be triggered). Also think about the path to the file that will write the data. You can test the output if you define a local path in `refextract/tasks.py` on `create_journal_kb_file` as default value.

jacquerie:
Just had a quick look and it seems ok, but I can't see the part in which you configure `refextract` to use the files you generate. How do you do that?

zzacharo:
I don't do yet :P. So, in order to do that I think I will pass it to the workflow but I don't know the shared path that I will store the kb file. The path that my task will store the data will be the one that refextract will take.

jacquerie:
> I don't do yet :P

Ok, but this should be part of this PR, otherwise you won't be achieving what the title says!

zzacharo:
Of course it's just pending now.

inspirehep/modules/refextract/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+def file_open_for_editing(fpath):
+    fh = codecs.open(fpath, encoding='utf-8', mode='w')
+
+    yield fh
+
+    fh.close()
+
+
+@shared_task()
+def create_journal_kb_file(path):
```
kaplun: You can postpone configuring this into `config.py`. So when we deploy we can set a path that is sharable across nodes...

inspirehep/modules/refextract/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+    querystring = "SELECT short_titles, title_variants " \
```
michamos: you need to sort the rows on the length of `title_variants` from longest to shortest as in https://github.com/inspirehep/refextract/blob/master/refextract/references/kbs/journal-titles.kb, in order to prevent wrong partial matches.

inspirehep/modules/refextract/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+
+    results = db.session.execute(querystring)
+
+    if results.rowcount > 1:
+        with file_open_for_editing(path) as fh:
+
+            for row in results:
+                alt_title = row['title_variants'].get('title')
+                short_title = row['short_titles'].get('title')
+                fh.write(alt_title + '---' + short_title)
```
michamos: maybe use `format()` here?

jacquerie:
> Of course it's just pending now.

Okok, but since you were asking for my review I thought I was missing something...

inspirehep/modules/refextract/tasks.py
zzacharo: This is done here anyway I think https://github.com/inspirehep/refextract/blob/master/refextract/references/kbs.py#L615 .

kaplun: Yeah see: https://github.com/inspirehep/invenio/blob/prod/modules/docextract/lib/refextract_kbs.py#L718

zzacharo:
@jacquerie @kaplun @michamos I updated this PR with a complete solution. The task is configured to run every hour.

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -28,6 +28,7 @@
 
 from refextract import extract_journal_reference, extract_references_from_file
 
+from inspirehep.config import REFEXTRACT_JOURNAL_KB_PATH
```
jacquerie: You should never do `from inspirehep.config import ...`: this prevents this path from being overridable on production. When you have to import from the config you need to have an application context and access `current_app.config`

zzacharo: oh ok, I didn't know that! And also I tested the query with all the journals from here http://inspirehep.net/dumps/inspire-dump.html and it takes around 0,33 sec.

jacquerie:
> The task is configured to run every hour.

Did you benchmark how much it takes to run on a full dump of institutions? If you didn't, migrate one from http://inspirehep.net/dumps/inspire-dump.html.

zzacharo:
@jacquerie Do we need `institutions` ? Is building only journals, but I can test it.

inspirehep/modules/workflows/tasks/refextract.py
jacquerie: > And also I tested the query with all the journals from here http://inspirehep.net/dumps/inspire-dump.html and it takes around 0,33 sec.

Nice, so we can easily cover the use case presented by @annetteholtkamp.

jacquerie:
> @jacquerie Do we need institutions ?

Err, no, I was working on https://github.com/inspirehep/inspire-next/pull/2479 (which is about Institutions) and got them mixed up...

zzacharo:
@jacquerie I created a follow-up commit for continuing review.

inspirehep/modules/workflows/tasks/refextract.py
zzacharo: Every hour is fine for the task to run? Or we want different schedule?

zzacharo:
> For this to work, the time that one needs to wait between 2. and 3. should be as short as possible.
@michamos Problably we need to add an option of triggering the journal-kb build on demand.

michamos:
:+1: that would be even better. Since it's so fast compared to the actual reference extraction, 3. could rebuild the journal knowledge base first.

inspirehep/modules/workflows/tasks/refextract.py
michamos: you don't need this `import` anymore.

kaplun: > And also I tested the query with all the journals from here http://inspirehep.net/dumps/inspire-dump.html and it takes around 0,33 sec.

I now wonder though how it would go down in real condition, i.e. with another 1M records adding noise.

zzacharo: The bottleneck is how many journals will be added. I think now is tested with 3500 journals more or less but the time is quite small anyway I think.

zzacharo:
@jacquerie @kaplun Regarding this https://travis-ci.org/inspirehep/inspire-next/jobs/247895938#L2025 should I create a dummy file on conftest and configure tests to use that?

inspirehep/modules/refextract/tasks.py
```diff
@@ -0,0 +1,64 @@
...
+import contextlib
+
+from celery import shared_task
+import codecs
+from flask import current_app
+
+from invenio_db import db
+
+
+from inspirehep.config import REFEXTRACT_JOURNAL_KB_PATH
```
kaplun: This should be imported from `current_app.config` otherwise we can't override it on production.

kaplun: This is a left over

zzacharo: I am fixing and squash all the commits to merge it.

PR #2510, opened 2017-07-05
matching: backoff when legacy is down

### LINKED ISSUE #2481: Legacy matcher fails badly if legacy is down
If Legacy is down, the legacy matcher step in the workflow fails badly with a JSON error. It should rather retry for a while at least, and then fail with a meaningful error.

See: https://sentry.cern.ch/inspire-sentry/inspire-labs/group/821845/

kaplun:
Protect legacy matching with the backoff decorator.

(Closes #2481)

### DIFF
--- a/inspirehep/modules/workflows/tasks/matching.py
+++ b/inspirehep/modules/workflows/tasks/matching.py
@@ -28,12 +28,14 @@ import datetime
 import os
 import re
 import traceback
+import backoff
 from functools import wraps
 from itertools import chain
 
 import requests
 import six
 from flask import current_app
+from simplejson import JSONDecodeError
 
 from inspirehep.utils.datefilter import date_older_than
 from inspirehep.utils.record import (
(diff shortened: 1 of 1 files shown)


inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -46,6 +48,7 @@
 
 
 @with_debug_logging
+@backoff.on_exception(backoff.expo, Exception, max_tries=7)
```
kaplun: Setting this to 7 to reach almost one minute (legacy sometimes throttle clients for 10 seconds...)

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -28,12 +28,14 @@
...
 import re
 import traceback
+import backoff
 from functools import wraps
 from itertools import chain
 
 import requests
 import six
 from flask import current_app
+from simplejson import JSONDecodeError
```
kaplun: This is what is raised internally by `requests.get().json()`

PR #2511, opened 2017-07-05
setup: advertize celery tasks in entry_points

kaplun:
Explicitly advertize celery tasks in corresponding entry points so that they are all correctly registered for Celery.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -283,6 +283,12 @@ setup(
             'inspire_workflows_audit = inspirehep.modules.workflows.models',
             'inspire_disambiguation = inspirehep.modules.disambiguation.models',
         ],
+        'invenio_celery.tasks': [
+            'inspire_refextract = inspirehep.modules.refextract.tasks',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -283,6 +283,12 @@
             'inspire_workflows_audit = inspirehep.modules.workflows.models',
             'inspire_disambiguation = inspirehep.modules.disambiguation.models',
         ],
+        'invenio_celery.tasks': [
```
jmartinm: We should prefix them with `inspire_` to avoid collisions with upstream Invenio

setup.py
```diff
@@ -283,6 +283,12 @@
...
+            'refextract = inspirehep.modules.refextract.tasks',
+            'authors = inspirehep.modules.authors.tasks',
+            'disambiguation = inspirehep.modules.disambiguation.tasks',
+            'records = inspirehep.modules.records.tasks',
```
jmartinm: For example, `records` is already in use:

```
records = invenio_migrator.tasks.records
```

PR #2513, opened 2017-07-05
workflow: preserve references

kaplun:
- [x] *I have removed tests that covered the removed feature.*

* Now that refextract is using fresh journal information there's
  no more need to drop references before sending them to legacy.

### DIFF
--- a/inspirehep/modules/workflows/tasks/submission.py
+++ b/inspirehep/modules/workflows/tasks/submission.py
@@ -405,10 +405,3 @@ def prepare_files(obj, eng):
         obj.data['_fft'] = obj.data.get('_fft', []) + result
         obj.log.info('Non-user PDF files added to FFT.')
         obj.log.debug('Added PDF files: {}'.format(result))
-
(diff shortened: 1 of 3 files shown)


michamos:
If I understand correctly, this means that there are no record ids in the references, and that we let bibrank dynamically match records until we have a reference matcher on labs. Is that OK?

kaplun:
So... I believe it is, because:
* IDs would be added nevertheless when exporting back records to labs via prodsync.
* For the final user, there's no practical difference between IDs guessed by bibrank or IDs guessed by refextract.

david-caro:
Don't we have any tests that check if the references are extracted?

jacquerie:
> Don't we have any tests that check if the references are extracted?

We do: https://github.com/inspirehep/inspire-next/blob/7fad81f08139c9b5f726a361e52128e577f28bfe/tests/unit/workflows/test_workflows_tasks_refextract.py#L96-L105

PR #2517, opened 2017-07-06
Acceptance: remove has_error functions from all the tests  

rikirenz:
## Description
This PR wants to remove the `has_error` function that we used in the acceptance test.

## Motivation and Context
The has_error function was confusing because was not used in the proper way. With this change in the `ArsenicResponse` class we are able to decide which kind of assertion we want to test plus the naming with this change is going to be more meaningful

## Checklist:

### DIFF
--- a/inspirehep/bat/arsenic.py
+++ b/inspirehep/bat/arsenic.py
@@ -72,7 +72,24 @@ class Arsenic(object):
 
 
 class ArsenicResponse(object):
-    has_error = None
+    statement = None
+    negative_statement = None
 
-    def __init__(self, has_error):
-        self.has_error = has_error
+    def __init__(self, statement, negative_statement=None):
+        """Init method.
+
+        :param statement: sets the function that will
+        be called to verify that a given event occured
+        :type source: function
+        :param negative_statement: sets the function that will
+        be called to verify that a given event did not occur
+        :type record: function
+        """
+        self.statement = statement
+        self.negative_statement = negative_statement
+
+    def assert_no_errors(self):
+        self.statement()
+
+    def assert_errors(self):
+        self.negative_statement()
--- a/inspirehep/bat/pages/create_author.py
+++ b/inspirehep/bat/pages/create_author.py
@@ -43,7 +43,7 @@ def go_to():
 
 def write_institution(institution, expected_data):
     def _write_institution():
-        return expected_data in Arsenic().write_in_autocomplete_field(
+        assert expected_data in Arsenic().write_in_autocomplete_field(
             'institution_history-0-name', institution)
 
     return ArsenicResponse(_write_institution)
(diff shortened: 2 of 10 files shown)


rikirenz:
@kaplun This meta-programming IMHO is necessary in order to avoid that the tests use directly selenium. In this way with the `ArsenicRepsonse` the test does not have to care about the selenium framework, as result we have: 
- The test is able to self document itself because is more readable
- We can reuse more code

😄

inspirehep/bat/arsenic.py
```diff
@@ -72,7 +72,15 @@ def __getattr__(self, item):
 
 
 class ArsenicResponse(object):
-    has_error = None
+    statement = None
+    negative_statement = None
 
-    def __init__(self, has_error):
-        self.has_error = has_error
+    def __init__(self, statement, negative_statement=None):
```
david-caro: We should add here some docs on how to use this, and what the params mean.

rikirenz: Added :fist_oncoming:

jacquerie:
This PR is trading one antipattern for two antipatterns. We didn't need `ArsenicResponse` and the `has_error` mechanism in the first place, since they wrap what would have been, if written semantically, `assert` statements.

david-caro:
@jacquerie please, provide an alternative, I still think that this PR improves the current status of the code by moving the assert to the point where it actually does the check if the page was properly loaded, and helping then with the debugging of the issue (currently, you end up in a context where you are just checking if `assert True` or `assert False`, not really helping to debug why).

jacquerie:
Yes, it's bad, but this PR goes deeper in the bad direction: instead of simply writing assertions like `assert 'this text' in this_field` these tests had the complicated mechanism of `has_error` and closures and what not, and this PR makes that even more complicated by adding another similar mechanism. Really, the only alternative I see here is to slowly rewrite these tests one by one.

david-caro:
The problem there is that the assertion (or assertions) that need to be made depend on the page you are in, and that information is in the domain of the bat pages. What the `ArsenicResponse` does is to define an interface that all the pages have to implement to unify that `assert` at the test level.

I don't think that the idea of such interface is wrong, I think that it's actually good (abstracting the tests from the individual nuances of each page, you don't really know if having the string `alert_wrn_smthng` is actually an error or a success, also all the xpath and similar things get "hidden" under a function improving the readability of the whole test). Just that the current implementation is lacking expressiveness (though I think that this PR also improves that point too, instead of the ambiguous `has_error` it now has `assert_errors` and `assert_no_errors` that actually match what they do).

About the passing of closures, it can be done also with subclassing, though it will increase the boilerplate needed (having to add a new `ArsenicResponse` subclass to each page).

PR #2525, opened 2017-07-11
arxiv: Fix wrong mimetype for arxiv pdfs

### LINKED ISSUE #2523: Wrong mime type leads to refextract fail
## Current Behavior

`refextract` fails when tries to parse file content as it expects a pdf and gets mimetype HTML.
The problem is that here https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/tasks/arxiv.py#L75 the check for `pdf` passes but the content is `html` and so `refextract` fails when tries here https://github.com/inspirehep/refextract/blob/master/refextract/references/engine.py#L1395 with `magic` library to detect the content.

See sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs/group/824570/events/150921036/

zzacharo:
Downloaded pdf files from arxiv, sometimes were wrong recognized by
`invenio-files-rest` as mimetype 'application/pdf' but the content
was html. Therefore, `refextract` was correctly raising an exception.
For this reason, the check of mimetype was made more strict before
downloading the file.

(Closes #2523)

## Related Issue
#2523 

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/arxiv.py
+++ b/inspirehep/modules/workflows/tasks/arxiv.py
@@ -40,6 +40,7 @@ from inspire_dojson.hep import hep
 from inspire_dojson.utils.arxiv import classify_field
 from inspirehep.modules.converter import convert
 from inspirehep.utils.record import get_arxiv_categories, get_arxiv_id
+from inspirehep.utils.url import is_pdf_link
 
 from plotextractor.api import process_tarball
 from plotextractor.converter import untar
@@ -66,20 +67,19 @@ def arxiv_fulltext_download(obj, eng):
     """
     arxiv_id = get_arxiv_id(obj.data)
     filename = secure_filename('{0}.pdf'.format(arxiv_id))
+    url = current_app.config['ARXIV_PDF_URL'].format(arxiv_id=arxiv_id)
+
+    if not is_pdf_link(url):
+        raise DownloadError("{url} is not serving a PDF file.".format(url=url))
+
     pdf = download_file_to_workflow(
         workflow=obj,
         name=filename,
-        url=current_app.config['ARXIV_PDF_URL'].format(arxiv_id=arxiv_id),
+        url=url,
     )
 
-    if pdf and pdf.get_version().mimetype == 'application/pdf':
+    if pdf:
         obj.log.info('PDF retrieved from arXiv for %s', arxiv_id)
-    elif pdf:
-        # We need to delete the failed download as otherwise it would
-        # create a new version instead of replacing the previous one.
-        del obj.files[filename]
-        pdf.delete()
-        raise DownloadError()
     else:
         obj.log.error('Cannot retrieve PDF from arXiv for %s', arxiv_id)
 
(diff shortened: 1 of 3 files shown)


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -66,20 +67,19 @@ def arxiv_fulltext_download(obj, eng):
     """
     arxiv_id = get_arxiv_id(obj.data)
     filename = secure_filename('{0}.pdf'.format(arxiv_id))
+    url = current_app.config['ARXIV_PDF_URL'].format(arxiv_id=arxiv_id)
+
+    if not is_pdf_link(url):
+        raise DownloadError()
```
kaplun: Can you pass a message in `DownloadError()`?
E.g.
```python
raise DownloadError("{url} is not serving a PDF file.".format(url=url))
```

or similar.

zzacharo: ok.

zzacharo:
@kaplun check!

kaplun:
@zzacharo it doesn't pass tests... can you have a look?

zzacharo:
@kaplun I am checking it!

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -81,6 +82,7 @@ def test_arxiv_fulltext_download_logs_on_success():
     assert expected == result
 
 
+@pytest.mark.xfail(raises=DownloadError)
```
kaplun: Wait... aren't you simply disabling the test in this way?

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -106,12 +108,7 @@ def test_arxiv_fulltext_download_logs_on_error():
     obj = MockObj(data, extra_data, files=files)
     eng = MockEng()
 
-    assert arxiv_fulltext_download(obj, eng) is None
-
-    expected = 'Cannot retrieve PDF from arXiv for 1605.03814'
-    result = obj.log._error.getvalue()
-
-    assert expected == result
+    arxiv_fulltext_download(obj, eng)
```
kaplun: Here you can assert that the exception is raised, if this what you wanted to assert.

zzacharo: Yes I did it like this but it's simpler to to mark it as xfailed expecting the same excetpion, isn't it? I can do it other way if you want. I don't know the best practice about this.

tests/unit/workflows/test_workflows_tasks_arxiv.py
zzacharo: No it says that I expect to fail with an exception = DownloadError. It is like asserting that there is an exception.

tests/unit/workflows/test_workflows_tasks_arxiv.py
kaplun: Nope. If you mark it as xfailed it means that either the test is buggy, or the tested code is buggy. Either way it's something that needs to be fixed:
https://docs.pytest.org/en/latest/skipping.html#xfail-mark-test-functions-as-expected-to-fail

Here the fact that `DownloadError` is raised is not unexpected, but it's the feature that you want to test.

zzacharo: Ok yes I will change it.

zzacharo:
@kaplun check it again!

kaplun:
OK! P.s. when you repush, can you always also rebase on top of latest master? This allows to merge it straight away if all good.

PR #2530, opened 2017-07-13
authors: phonetic_blocks via native beard

kaplun:

* Uses native beard instance in order to compute author phonetic
  blocks, thus avoiding relying on sometimes flaky external service
  for a very small functionality.
  (addresses #1282)

### DIFF
--- a/inspirehep/modules/authors/receivers.py
+++ b/inspirehep/modules/authors/receivers.py
@@ -22,12 +22,7 @@
 
 from __future__ import absolute_import, division, print_function
 
-import json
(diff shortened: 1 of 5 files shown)


kaplun:
Finally I've decided to do it: I have dropped the requirement to use http://beard.inspirehep.net and simply added `beard` as a dependency. Anyway, we were already having scikit-learn and the whole beast included.
In this way we can be much faster and more stable (since we don't have to rely on an external microservice to just compute a string transformation).

PR #2544, opened 2017-07-19
Rt api

harunurhan:

### DIFF
--- a/inspirehep/modules/editor/views.py
+++ b/inspirehep/modules/editor/views.py
@@ -57,7 +57,8 @@ def create_rt_ticket():
         return jsonify(
             success=True,
             data={
-                'ticket_id': str(ticket_id)
+                'id': str(ticket_id),
+                'link': tickets.get_rt_link_for_ticket(ticket_id)
             }
         )
     else:
--- a/inspirehep/modules/editor/views.py
+++ b/inspirehep/modules/editor/views.py
@@ -1,0 +1,106 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""INSPIRE editor blueprint."""
+
+from __future__ import absolute_import, division, print_function
+
+from flask import (
+    abort,
+    Blueprint,
+    jsonify,
+    request,
+)
+from flask_login import current_user, login_required
+
+from ...utils import tickets
+
+
+blueprint = Blueprint('inspirehep_editor',
+                      __name__,
+                      url_prefix='/editor',)
+
+
+@blueprint.route('/rt/tickets/create', methods=['POST'])
+# TODO: @login_required
+def create_rt_ticket():
+    """View to create an rt ticket"""
+    json = request.json
+    ticket_id = tickets.create_ticket(json['queue'],
+                                      current_user.email,
+                                      json['description'],
+                                      json['subject'],
+                                      int(json['recid']),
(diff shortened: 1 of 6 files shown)


inspirehep/modules/editor/views.py
```diff
@@ -0,0 +1,105 @@
...
+    request,
+)
+from flask_login import current_user, login_required
+
+from ...utils import tickets
+
+
+blueprint = Blueprint('inspirehep_editor',
+                      __name__,
+                      url_prefix='/editor',)
```
jmartinm: No need of last comma

inspirehep/modules/editor/views.py
```diff
@@ -0,0 +1,105 @@
...
+def create_rt_ticket():
+    """View to create an rt ticket"""
+    json = request.json
+    ticket_id = tickets.create_ticket(json['queue'],
+                                      current_user.email,
+                                      json['description'],
+                                      json['subject'],
+                                      json['recid'],
+                                      Owner=json['owner'])
+    if ticket_id == -1:
```
jmartinm: So when it succeeds the ticket `id == -1` ? 😮

inspirehep/modules/editor/views.py
```diff
@@ -0,0 +1,105 @@
...
+        return jsonify(
+            success=True,
+            data={
+                'ticket_id': ticket_id
+            }
+        )
+    else:
+        abort(500)
```
jmartinm: Will it be possible for the client to understand this response? Or better to return `jsonify` with `success=False` and some explanation message.

inspirehep/modules/editor/views.py
```diff
@@ -0,0 +1,105 @@
...
+                      __name__,
+                      url_prefix='/editor',)
+
+
+@blueprint.route('/rt/tickets/create', methods=['POST'])
+# TODO: @login_required
+def create_rt_ticket():
+    """View to create an rt ticket"""
+    json = request.json
+    ticket_id = tickets.create_ticket(json['queue'],
```
jmartinm: Are you sure `json` will always have all the fields? Otherwise you should protect it with `json.get()`

inspirehep/modules/editor/views.py
harunurhan: hahah, good catch :)

inspirehep/modules/editor/views.py
harunurhan: well, I don't know what to put because it wasn't just successful for some internal reason. That's why I return `Internal Server Error`

inspirehep/modules/editor/views.py
```diff
@@ -0,0 +1,109 @@
...
+
+@blueprint.route('/rt/tickets/<ticket_id>/resolve', methods=['GET'])
+@editor_manage_tickets_permission.require(http_exception=403)
+def resolve_rt_ticket(ticket_id):
+    """View to resolve an rt ticket"""
+    tickets.resolve_ticket(ticket_id)
+    return jsonify(success=True)
+
+
+@blueprint.route('/rt/tickets/<recid>', methods=['GET'])
```
jacquerie: Note something funny: this route overrides the one above on `/rt/tickets/create`, because you haven't specified a type for `recid`, so it uses the string `"create"` as a `recid`... : )

inspirehep/modules/editor/views.py
```diff
@@ -0,0 +1,109 @@
...
+@editor_manage_tickets_permission.require(http_exception=403)
+def get_tickets_for_record(recid):
+    """View to get rt ticket belongs to given record"""
+    tickets_for_record = tickets.get_tickets_by_recid(recid)
+    simplified_tickets = map(_simplify_ticket_response, tickets_for_record)
```
jacquerie: Minor nit, but it's more Pythonic to use a list comprehension instead of `map` : https://www.artima.com/weblogs/viewpost.jsp?thread=98196

inspirehep/modules/editor/views.py
```diff
@@ -0,0 +1,109 @@
...
+blueprint = Blueprint('inspirehep_editor',
+                      __name__,
+                      url_prefix='/editor',)
+
+
+@blueprint.route('/rt/tickets/create', methods=['POST'])
+@editor_manage_tickets_permission.require(http_exception=403)
+def create_rt_ticket():
+    """View to create an rt ticket"""
+    json = request.json
```
jacquerie: It's discouraged to shadow names that are part of the standard library (like `json`) because it might lead to hard-to-diagnose bugs.

setup.py
```diff
@@ -216,7 +216,8 @@
             'view_restricted_collection'
             ' = inspirehep.modules.records.permissions:'
             'action_view_restricted_collection',
-            'admin_holdingpen_authors = inspirehep.modules.authors.permissions:action_admin_holdingpen_authors'
+            'admin_holdingpen_authors = inspirehep.modules.authors.permissions:action_admin_holdingpen_authors',
+            'editor_manage_tickets = inspirehep.modules.editor.permissions:action_editor_manage_tickets'
```
jacquerie: Nit: this example shows why putting trailing commas everywhere is better, because it leads to simpler diffs (unlike what happened here).

PR #2580, opened 2017-07-31
config: configure invenio-access cache

jmartinm:
* Cache will save permission information to avoid extra DB queries.
  Invalidation is handled by invenio-access.

## Checklist:


jmartinm:
Realized we were missing this config while fixing https://github.com/inveniosoftware/invenio-access/pull/127 in invenio-access. Tested that it works as expected locally.

lnielsen:
Note that the latest released version on PyPI do not have an important fix for the cache which makes it cache wrong permissions when using parameters for an action.

Also, just a heads up, we're about to do some refactoring on Invenio-Access to make it work consistently, logical and simple……we will try to do it without breaking existing code

jmartinm:
Closing this to avoid using the cache while `invenio-access` is in buggy state

PR #2590, opened 2017-08-03
utils: is_pdf_link sanity check

### LINKED ISSUE #2589: literature submission: pdf validator bug
Inputting in the form the PDF url http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.502.4449&rep=rep1&type=pdf crashes the validator.

See https://sentry.cern.ch/inspire-sentry/inspire-labs/group/864577/

Reason is that response is empty so `next` call raises `StopIteration`

jmartinm:
* Avoid raising Stopiteration when response is empty. (closes #2589)

## Checklist:

### DIFF
--- a/inspirehep/utils/url.py
+++ b/inspirehep/utils/url.py
@@ -58,7 +58,10 @@ def is_pdf_link(url):
     except requests.exceptions.RequestException:
         return False
 
-    magic_number = next(response.iter_lines(1))
+    try:
+        magic_number = next(response.iter_lines(1))
+    except StopIteration:
+        return False
     correct_magic_number = magic_number.startswith('%PDF')
 
     return correct_magic_number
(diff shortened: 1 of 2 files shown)


inspirehep/utils/url.py
```diff
@@ -58,7 +58,10 @@ def is_pdf_link(url):
     except requests.exceptions.RequestException:
         return False
 
-    magic_number = next(response.iter_lines(1))
+    try:
+        magic_number = next(response.iter_lines(1))
+    except StopIteration:
+        return False
```
jacquerie: I think there is a better way to do this... I remember discussing this with @kaplun. Let me search for it.

jacquerie: Nope, you cannot use `response.raw.read(4)` because you don't know in advance how many gzipped bytes you will have to read before decoding enough to have the first 4 bytes.

So I just added a test for your change and we're good.

PR #2613, opened 2017-08-11
refextract: Include `journal_titles` normalization in journal-kb

### LINKED ISSUE #2612: Refextract should also use the journal_titles for normalization
## Expected Behavior

When presented with a reference containing the full journal title of the journal, refextract should normalize it to the short title. For this to happen [`create_journal_kb_file`](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/refextract/tasks.py#L46) should also look at `journal_titles` instead of only `title_variants`.

zzacharo:

## Related Issue

Closes #2612 

## Checklist:

### DIFF
--- a/inspirehep/modules/refextract/tasks.py
+++ b/inspirehep/modules/refextract/tasks.py
@@ -44,10 +44,13 @@ def file_open_for_editing(fpath):
 
 @shared_task()
 def create_journal_kb_file():
-    querystring = "SELECT short_titles, title_variants " \
-                  "FROM records_metadata as r, " \
-                  "json_array_elements(r.json -> 'short_titles') as short_titles, " \
-                  "json_array_elements(r.json -> 'title_variants') as title_variants"
+    querystring = "SELECT short_title, long_title FROM records_metadata as r, " \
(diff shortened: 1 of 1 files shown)


jacquerie:
> Actually this is incomplete because the requirements were [...]

If only there was a way to encode requirements in such a way that they are automatically verified... : )

zzacharo:
@michamos https://github.com/inspirehep/refextract/blob/master/refextract/references/tag.py#L133 here I think it does some normalization in the reference line reagrding punctuation and spaces. Is all caps needed for the matching? For short_titles I can add them in the kb creation.

jacquerie:
> If only there was a way to encode requirements in such a way that they are automatically verified... : )

You can use this as a pattern: https://github.com/jacquerie/inspire-next/commit/a60d040115b5f9428385261b8fc52d86b4302c39

jacquerie:
My poor heart couldn't stand those files at 0% coverage, so I merged a commit like the one above that incorporated @kaplun's comments on it.

michamos:
@jacquerie this should be revived now that we have modified the schema.

PR #2618, opened 2017-08-14
tests: remove year validation tests

jacquerie:
## Description:
Commit 1e571aa is already testing this logic and is much faster.

## Related Issue:
Addresses https://github.com/inspirehep/inspire-next/issues/1987.

## Checklist:

### DIFF
--- a/inspirehep/bat/pages/create_author.py
+++ b/inspirehep/bat/pages/create_author.py
@@ -65,23 +65,6 @@ def write_advisor(advisor, expected_data):
     return ArsenicResponse(_write_advisor)
 
 
-def write_year(input_id, error_message_id, year):
-    def _write_year():
-        return 'is not a valid year' in message_err
-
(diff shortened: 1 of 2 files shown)


david-caro:
Mmm... they are not testing the same thing, for example, this test would have catch the current issue where Annete was not seeing the error message in the ui when entering an incorrect date, while the other test would not be able to catch that type of issue.
Imo we should keep validating the form visually, at least, a positive and a negative entry on every input, as the goal is not to validate that it correctly detects any valid or invalid value, but that when that happens, you get feedback on the ui.

jacquerie:
It's true that they don't test the UI, but I argue that we don't need that part. If the validator function is doing its job, we just need to test once that some JavaScript function will manipulate the DOM in the right way, and not for every single field.

BTW, the bug you mention is already gone, @iulianav fixed it : )

PR #2619, opened 2017-08-14
tests: split acceptance in two

jacquerie:
## Description
Temporarily removes the disambiguation module and its associated test suite
as it has become stale and unmaintained. Moreover, this module depends
crucially on the Authors schema, which is not yet ready and won't be for a while.

Splits the `acceptance` test suite in `acceptance-authors` and `acceptance-literature`
in order to reduce the build time to roughly 10 minutes.

## Related Issue
Addresses https://github.com/inspirehep/inspire-next/issues/1987.

## Checklist:


jacquerie:
The improvement in testing time is modest, and probably not worth the extra hassle of having to fish again `disambiguation` from the depths of `git`'s history when its time will come.

PR #2572, opened 2017-07-27
global: handle files between records and workflows

zzacharo:
## Description:
Adds the ``files`` property to ``InspireRecord`` by inheriting from the
``Record`` of ``invenio_records_files``. This allows us to attach plots
and documents to records (closes #2551 and closes #2552).

Also modifies those workflows tasks that deal with records so that they
use builder methods to populate the ``documents`` and ``figures`` keys,
and those will then be converted by DoJSON to the ``FFT`` fields that
Legacy understands.

## Related Issues:
Closes #2551 and closes #2552.

## Checklist:
- [x] I have all the information that I need (if not, move to `RFC` and look for it).
- [x] I linked the related issue(s) in the corresponding commit logs.
- [x] I wrote [good commit log messages](https://github.com/torvalds/subsurface-for-dirk/blob/5f15ad5a86ada3c5e574041a5f9d85235322dabb/README#L92-L119).
- [ ] My code follows the code style of this project.
- [ ] I've added any new docs if API/utils methods were added.
- [ ] I have updated the existing documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #2623, opened 2017-08-15
WIP migration initial commit

zzacharo:
I create it in order if someone wants to take it as long as I am away to be able to. It needs first dojson modification.

### DIFF
--- a/inspirehep/modules/migrator/tasks.py
+++ b/inspirehep/modules/migrator/tasks.py
@@ -59,7 +59,7 @@ from inspirehep.modules.pidstore.minters import inspire_recid_minter
 from inspirehep.modules.pidstore.utils import get_pid_type_from_schema
 from inspirehep.modules.records.api import InspireRecord
 from inspirehep.modules.records.receivers import index_after_commit
-from inspirehep.modules.records.utils import file_resolving
+from inspirehep.modules.records.utils import open_url_or_path, FailedToOpenUrlPath
 
 from .models import InspireProdRecords
 
@@ -290,20 +290,15 @@ def create_record(record):
 
 
 def attach_documents_and_figures(record):
-    control_number = record.get('control_number', record.get('recid'))
     for attachment in chain(record.get('documents', []), record.get('figures', [])):
-        file_url = attachment.get('url')
-        if control_number:
-            filename = '{}_{}'.format(control_number, attachment.get('key'))
-        else:
-            filename = attachment.get('key')
-        with file_resolving(file_url) as fp:
-            record.files[filename] = fp
-            record_file_obj = record.files[filename]
-            attachment['url'] = '/api/files/{bucket}/{key}'.format(
-                bucket=record_file_obj.bucket_id,
-                key=record_file_obj.key
-            )
+        try:
+            with open_url_or_path(attachment.get('url')) as fp:
+                try:
+                    record.attach_file(attachment['key'], fp)
+                except ValueError as e:
+                    logger.error(e)
+        except FailedToOpenUrlPath as e:
+            logger.error(e)
 
 
 def record_insert_or_replace(json):
--- a/inspirehep/modules/migrator/tasks.py
+++ b/inspirehep/modules/migrator/tasks.py
@@ -59,6 +59,7 @@ from inspirehep.modules.pidstore.minters import inspire_recid_minter
 from inspirehep.modules.pidstore.utils import get_pid_type_from_schema
 from inspirehep.modules.records.api import InspireRecord
 from inspirehep.modules.records.receivers import index_after_commit
+from inspirehep.modules.records.utils import file_resolving
 
 from .models import InspireProdRecords
 
@@ -288,9 +289,27 @@ def create_record(record):
     return json
 
 
+def attach_documents_and_figures(record):
+    control_number = record.get('control_number', record.get('recid'))
+    for attachment in chain(record.get('documents', []), record.get('figures', [])):
+        file_url = attachment.get('url')
+        if control_number:
+            filename = '{}_{}'.format(control_number, attachment.get('key'))
+        else:
+            filename = attachment.get('key')
+        with file_resolving(file_url) as fp:
+            record.files[filename] = fp
+            record_file_obj = record.files[filename]
+            attachment['url'] = '/api/files/{bucket}/{key}'.format(
+                bucket=record_file_obj.bucket_id,
+                key=record_file_obj.key
+            )
+
+
 def record_insert_or_replace(json):
     """Insert or replace a record."""
     control_number = json.get('control_number', json.get('recid'))
+
     if control_number:
         pid_type = get_pid_type_from_schema(json['$schema'])
         try:
@@ -303,6 +322,8 @@ def record_insert_or_replace(json):
             record = InspireRecord.create(json, id_=None)
             # Create persistent identifier.
             inspire_recid_minter(str(record.id), json)
+
+        attach_documents_and_figures(record)
 
         if json.get('deleted'):
             new_recid = get_recid_from_ref(json.get('new_record'))
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -25,6 +25,7 @@
 from __future__ import absolute_import, division, print_function
 
 from datetime import datetime
+import itertools
 
 import arrow
 from elasticsearch.exceptions import NotFoundError
@@ -89,6 +90,50 @@ class InspireRecord(Record):
             storage_class=storage_class
         )
         return bucket
+
+    def attach_file(self, file_name, file):
+        """Attach the file ``file`` to the record's files properties,
+        updating the record's metadata..
+        The file is attached only if there is a key in ``documents`` or
+        ``figures`` which key match ``file_name``.
+        Then. the name of the file is changed adding the record's
+        ``control_number``as prefix. This change is reflected in
+        the record's metadata too.
+        Args:
+            file_name(string): the name of the file to attach. It must be a
+                key of ``documents`` or ``figures``.
+            file(file): the file to attach to the file.
+        Return:
+            (string): the new file's key.
+        Raise:
+            ValueError in case ``file_name`` is not in record's documents and
+                figures.
+        Example:
+            >>with open(my_file) as stream:
+                record.attach_file('NewFile', stream)
+        """
+
+        control_number = self.get('control_number')
+        normalized_file_name = u'{}_{}'.format(control_number, file_name)
+
+        for attachment in itertools.chain(
+                self.get('documents', []),
+                self.get('figures', [])
+        ):
+            if attachment['key'] == file_name:
+                #
+                self.files[normalized_file_name] = file
+                attachment['key'] = normalized_file_name
+                attachment['url'] = '/api/files/{bucket}/{key}'.format(
+                    bucket=self.files[normalized_file_name].bucket_id,
+                    key=normalized_file_name
+                )
+                return normalized_file_name
+        raise ValueError(
+            "The key {} does not belong to any document or figure".format(
+                file_name
+            )
+        )
 
 
 class ESRecord(InspireRecord):
--- a/inspirehep/modules/records/utils.py
+++ b/inspirehep/modules/records/utils.py
@@ -24,12 +24,19 @@
 
 from __future__ import absolute_import, division, print_function
 
+import logging
+import requests
+from contextlib import closing, contextmanager
+from six.moves.urllib.parse import urlparse
 from flask import current_app
 
 from inspirehep.modules.pidstore.utils import (
     get_endpoint_from_pid_type,
     get_pid_type_from_schema
 )
+
+
+LOGGER = logging.getLogger(__name__)
 
 
 def get_endpoint_from_record(record):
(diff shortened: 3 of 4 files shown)


zzacharo:
@jacquerie @michamos comments on this? It needs tests of course.

zzacharo:
it depends on #2572 to be merged first.

zzacharo:
@jacquerie @michamos can you review this?

inspirehep/modules/migrator/tasks.py
```diff
@@ -288,9 +289,23 @@ def create_record(record):
     return json
 
 
+def update_documents_and_figures_for_record(record):
+    for attachment in chain(record.get('documents', []), record.get('figures', [])):
+        file_url = attachment.get('url')
+        filename = attachment.get('key')
+        with file_resolving(file_url) as fp:
+            record.files[filename] = fp
```
michamos: so you didn't implement what we discussed yesterday about mangling the `key` to include the `recid` in it?

zzacharo: Yes I didn't include it. I will do it now!

tests/integration/test_records.py
```diff
@@ -367,3 +369,54 @@ def test_records_files_attached_correctly(app):
...
+def test_records_files_are_populated_correctly_for_local_paths(app):
+    record_json = {
+        '$schema': 'http://localhost:5000/schemas/records/hep.json',
+        'document_type': [
+            'article',
+        ],
+        'titles': [
+            {'title': 'foo'},
+        ],
+        '_collections': ['Literature'],
```
jacquerie: It doesn't look like the code requires the above data to run. Fixtures should be as minimal as possible, to avoid confusing the reader with irrelevant details.

inspirehep/modules/migrator/tasks.py
```diff
@@ -288,9 +289,23 @@ def create_record(record):
...
 
 
+def update_documents_and_figures_for_record(record):
+    for attachment in chain(record.get('documents', []), record.get('figures', [])):
+        file_url = attachment.get('url')
+        filename = '{}_{}'.format(record.get('recid') + attachment.get('key'))
+        with file_resolving(file_url) as fp:
+            record.files[filename] = fp
+            record_file_obj = record.files[filename]
+            attachment['url'] = '/files/{bucket}/{key}'.format(
```
jmartinm: Shouldn't this url start with /api ?

inspirehep/modules/migrator/tasks.py
```diff
@@ -288,9 +289,23 @@ def create_record(record):
     return json
 
 
+def update_documents_and_figures_for_record(record):
+    for attachment in chain(record.get('documents', []), record.get('figures', [])):
+        file_url = attachment.get('url')
+        filename = '{}_{}'.format(record.get('recid') + attachment.get('key'))
```
jacquerie: Something's off here! And probably an assertion is missing from the tests.

inspirehep/modules/records/utils.py
```diff
@@ -44,3 +47,16 @@ def get_detailed_template_from_record(record):
...
+@contextmanager
+def file_resolving(fpath):
+    known_schemes = ['http', 'https']
+    if urlparse(fpath).scheme in known_schemes:
+        with closing(requests.get(url=fpath, stream=True)) as req:
+            if req.status_code == 200:
+                req.raw.decode_content = True
+                yield req.raw
+    else:
+        with open(fpath, mode='r') as fh:
```
jmartinm: Here I would add a try: except: for `IOError` in case the path does not exist

inspirehep/modules/migrator/tasks.py
```diff
@@ -288,9 +289,23 @@ def create_record(record):
     return json
 
 
+def update_documents_and_figures_for_record(record):
```
jacquerie: What about `attach_document_and_figures` ?

tests/integration/test_records.py
```diff
@@ -367,3 +369,54 @@ def test_records_files_attached_correctly(app):
...
+                {'title': 'foo'},
+            ],
+            '_collections': ['Literature'],
+            'documents': [{
+                'key': 'Fulltext.pdf',
+                'url': 'http://www.mdpi.com/2218-1997/3/1/24/pdf',
+            }]  # DESY harvest
+        }
+
+        record = InspireRecord.create(record_json, id_=None)
```
jmartinm: Do you need `id_=None` here and not above?

inspirehep/modules/records/utils.py
```diff
@@ -24,6 +24,9 @@
 
 from __future__ import absolute_import, division, print_function
 
+import requests
+from contextlib import closing, contextmanager
+from urlparse import urlparse
```
jacquerie: `urlparse` should come from `six` for future Python 3 compatibility. See: https://pythonhosted.org/six/#module-six.moves

tests/integration/test_records.py
zzacharo: no this a leftover.

inspirehep/modules/records/utils.py
zzacharo: Do we need to silently continue if something is wrong with the path? @jacquerie

inspirehep/modules/migrator/tasks.py
zzacharo: This is already under api application cause is provided from invenio-files-rest...

inspirehep/modules/migrator/tasks.py
zzacharo: We discussed to have the recid attached to the filename for making filenames unique. This is called inside `record_insert_or_replace` which actually ensures that there is a recid. For the tests I refactor the assertions and the snippets.

zzacharo: oups now I saw it :P I am fixing it!

tests/integration/test_records.py
zzacharo: these are required from schema. Otherwise, it fails validation!

zzacharo:
@michamos I had included `/api` prefix in my previous PR so I updated now. Check again @jacquerie and @michamos and I will talk with david to check it on qa before we merge it.

zzacharo:
@jacquerie I added the try...except block check it again! I want just to check the `control_number` that I am prepending on filename if it's eventually needed.

inspirehep/modules/migrator/tasks.py
```diff
@@ -288,9 +289,27 @@ def create_record(record):
     return json
 
 
+def attach_documents_and_figures(record):
```
david-caro: this function is probably (or can be adapted) to be the same as the one in the other pr: https://github.com/inspirehep/inspire-next/pull/2760/files#diff-3cf20a9f11cd40bad5e31856ef439299R94

We should put that function in the records module and reuse it from there, probably renamed.

david-caro: That means though basing one PR in the other... so as this one is shorter and probably will get merged first, I'd use this one as the base.

inspirehep/modules/migrator/tasks.py
```diff
@@ -288,9 +289,27 @@ def create_record(record):
...
+    control_number = record.get('control_number', record.get('recid'))
```
david-caro: If you are expecting the record to have at least one of those (control_number or recid) you should assert it's not none, if you are requiring it to have a recid, that can be done by just using `record['recid']` instead of using `get` there, so it fails if it does not have a recid.

inspirehep/modules/migrator/tasks.py
```diff
@@ -288,9 +289,27 @@ def create_record(record):
...
+    for attachment in chain(record.get('documents', []), record.get('figures', [])):
+        file_url = attachment.get('url')
```
david-caro: same here, if you are expecting a key to exist in the dict you are using, instead of `get` use `[]` to fail quick if it's not there.

inspirehep/modules/records/utils.py
```diff
@@ -44,3 +51,19 @@ def get_detailed_template_from_record(record):
     """Return the detailed template corresponding to the given record."""
     endpoint = get_endpoint_from_record(record)
     return current_app.config['RECORDS_UI_ENDPOINTS'][endpoint]['template']
+
+
+@contextmanager
+def file_resolving(fpath):
```
david-caro: I think this should be renamed to `resolve_file_path` or similar, an infinitive is a bad name for a function.

david-caro: Probably `open_url_or_path` is better, as it's actually returning a 'file like object' as the `open` builtin does.

inspirehep/modules/records/utils.py
```diff
@@ -44,3 +51,19 @@ def get_detailed_template_from_record(record):
...
+        with closing(requests.get(url=fpath, stream=True)) as req:
+            if req.status_code == 200:
+                req.raw.decode_content = True
+                yield req.raw
+    else:
+        try:
+            with open(fpath, mode='r') as fh:
+                yield fh
+        except (OSError, IOError) as e:
+            LOGGER.error(e)
```
david-caro: it should probably not swallow the exception here

inspirehep/modules/records/utils.py
```diff
@@ -44,3 +51,19 @@ def get_detailed_template_from_record(record):
...
+
+
+@contextmanager
+def file_resolving(fpath):
+    known_schemes = ['http', 'https']
+    if urlparse(fpath).scheme in known_schemes:
+        with closing(requests.get(url=fpath, stream=True)) as req:
+            if req.status_code == 200:
+                req.raw.decode_content = True
+                yield req.raw
```
david-caro: and it should probably raise an error if the status code is not 2XX

tests/integration/test_records.py
```diff
@@ -351,19 +353,78 @@ def test_get_es_records_accepts_lists_of_strings(app):
 
 
 def test_records_files_attached_correctly(app):
-    json = {
+    record_json = {
```
david-caro: better name 👍

tests/integration/test_records.py
```diff
@@ -351,19 +353,78 @@ def test_get_es_records_accepts_lists_of_strings(app):
...
-    record = InspireRecord.create(json)
+    record = InspireRecord.create(record_json)
     record.files['fulltext.pdf'] = StringIO.StringIO()
     record.commit()
 
     assert 'fulltext.pdf' in record.files
+
+
+@patch('inspirehep.modules.records.utils.open', mock_open())
+def test_records_files_are_populated_correctly_for_local_paths(app):
```
david-caro: this probably should be renamed to `test_attach_documents_and_figures_....`, it's kind of a convention to start the (specially unit) tests with `test` followed by the name of the function or class it tests.

inspirehep/modules/records/utils.py
```diff
@@ -54,16 +61,22 @@ def get_detailed_template_from_record(record):
...
-        with closing(requests.get(url=fpath, stream=True)) as req:
-            if req.status_code == 200:
-                req.raw.decode_content = True
-                yield req.raw
-    else:
-        try:
+    try:
+        if urlparse(fpath).scheme in known_schemes:
+            with closing(requests.get(url=fpath, stream=True)) as resp:
+                if resp.status_code == 200:
```
michamos: you don't have to implement all of this error handling, you can use `raise_for_status` as explained here: http://docs.python-requests.org/en/master/user/quickstart/#response-status-codes.

inspirehep/modules/records/api.py
```diff
@@ -90,6 +91,50 @@ def _create_bucket(self, location=None, storage_class=None):
...
+                figures.
+        Example:
+            >>with open(my_file) as stream:
+                record.attach_file('NewFile', stream)
+        """
+
+        control_number = self.get('control_number')
+        normalized_file_name = u'{}_{}'.format(control_number, file_name)
+
+        for attachment in itertools.chain(
```
michamos: this is now quadratic in the number of files + figures. The number of figures can be quite large, so maybe you should have an `attach_files` function that takes no arguments and does this for all files?

zzacharo: this depends on how many times you call the `attach_file`. We want to handle in a single point the attachment of a file because we have the control_number logic and we use it in different places. So, what's the problem with this one, sorry I didn't understand it.

david-caro: How large is quite large?

inspirehep/modules/records/utils.py
zzacharo: so I can just first raise and then just do my logic after?

inspirehep/modules/records/api.py
david-caro: (in this case, it iterates the files first, and the images after, so the only resource that is being "wasted" is a bit of memory to hold the list of figures while waiting to iterate over them)

michamos: hundreds

zzacharo: @michamos https://github.com/ammirate/inspire-next/commit/26f9cfa67bdb8656a976efd8af5c66fe73125d50#diff-1bdc7b327ecbacc5d173e9ff21842a8dR91 here we iterate through different dataset and we are trying to be consistent with this too. For this PR is fine to include a resolver but for merger PR we cannot I think.

zzacharo:
I am closing this cause continues here #2899 .

PR #2624, opened 2017-08-15
add views to expose refextract for editor

harunurhan:


inspirehep/modules/editor/views.py
```diff
@@ -36,6 +43,56 @@
                       url_prefix='/editor',)
 
 
+@blueprint.route('/refextract/from_text', methods=['POST'])
```
jacquerie: Do you have a strong preference for `/editor/refextract/from_text` vs. `/editor/refextract/text` ?

PR #2669, opened 2017-08-23
Multieditor implementation

Glignos:
## Checklist:


inspirehep/modules/multieditor/views.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+@blueprint.route("/update", methods=['POST'])
+def index():
+    """Basic view."""
+    user_actions = request.json
+    return jsonify(actions.run_user_actions(user_actions))
+
+
+@blueprint.route("/search/<page_num>/<query_string>")
```
jmartinm: As mentioned IRL, the page and query string would be better query_params

inspirehep/modules/multieditor/views.py
```diff
@@ -0,0 +1,73 @@
...
+def search(page_num, query_string):
+    """Basic view."""
+    records = []
+    json_records = []
+    #  find record ids from elastic search
+    query_result = LiteratureSearch().query_from_iq(query_string).params(size=10, from_=int(page_num), _source=['control_number']).execute()
```
jmartinm: In the `from_` you want something like `(page_num - 1) * size`

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,184 @@
...
+
+def apply_to_array(record, key, regex, action, value, values_to_check):
+    """Applying action to array."""
+    for index, array_record in enumerate(record[key]):
+        if action == 'Update':
+            if regex and re.search(
+                    re.escape(values_to_check[0]),
+                    record[key][index]):
+                record[key][index] = value
+            elif array_record in values_to_check:
```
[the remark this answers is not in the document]
Glignos: Will we support to directly update an object?

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,209 @@
...
+            elif record[key][i] in values_to_check:
+                record[key][i] = value
+        elif action == 'Addition':
+            record[key].append(value)
+            return  # In that case we want to stop looping
+        elif action == 'Deletion':
+            if len(values_to_check) == 0 \
+                    or record[key][i] in values_to_check:
+                record[key].pop(i)
+                --i
```
jmartinm: This shows that there is no test covering this part, since otherwise you would catch that `--i` does not have any effect ;)

Glignos: Yup got it! I updated a test to get this case as well and I did some fixing on thsi code.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,222 @@
...
+            action_tuple = Action(new_keys, action.selected_action,
+                                  action.value, action.values_to_check, action.regex,
+                                  new_where_keys, action.where_value)
+            apply_action(new_schema, record[key], action_tuple)
+    if action.selected_action == 'Deletion':  # don't leave empty objects
+        clear_from_empty(record, key)
+
+
+def clear_from_empty(record, key):
+    if isinstance(record[key], list):
```
jmartinm: A better way of doing this is with a list comprehension:

```python
record = [item for item in record[key] if item]
```

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,281 @@
...
+            }},
+        "type": "object",
+    }
+    key = ['source']
+    value = 'success'
+    target_object = {'source': 'success'}
+    assert actions.create_schema_record(schema_2, key, value) == target_object
+
+
+def test_record_creation_3():
```
jmartinm: Name of the test should be more descriptive

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,281 @@
...
+    """should test sub_record handling for missing object"""
```
jmartinm: This is a leftover from previous test

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,281 @@
...
+        "abstracts": [
+            {
+                "value": "A dataset corresponding to $2.8~\\mathrm{fb}^{-1}$"
+            }
+        ],
+    }
+    target_object = {
+        "abstracts": [
+          {
+            "value": "A dataset corresponding to $2.8~\\mathrm{fb}^{-1}$"
```
jmartinm: Since you are showing that values are the same I would put a simpler example. Otherwise it takes a moment to make sure that formula is the same ;)

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,281 @@
...
+    }
+    curr_path = os.path.dirname(__file__)
+    with open(os.path.join(curr_path,
+                           'fixtures/schema.json'))\
+            as data_file:
+        schema = json.load(data_file)
+
+    actions_tuple = Action(values_to_check=['Rome'], keys=['authors', 'affiliations', 'value'],
+                           where_keys=['authors', 'signature_block'], selected_action="Update",
+                           where_value='BANARo', regex=True, value="Success")
```
jmartinm: Would be better if the `where_value` is an actual regex

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,308 @@
...
+                                   regex=self.where_regex, position=position):
+                    return
+                checked = True
+        if not record.get(key):
+            return
+        if position + 1 == len(self.keys):
+            if isinstance(record[key], list):
+                if self.update_regex:
+                    for value_to_check in self.values_to_check:
+                        record[key] = list(
```
jmartinm: I don't think you need explicit casting

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,308 @@
...
+                            filter(lambda x: (re.search(
+                                re.escape(value_to_check),
+                                x) or len(self.values_to_check) == 0),
+                                   record[key]))
+                else:
+                    record[key] = list(
+                        filter(lambda x: not (len(self.values_to_check) == 0
+                                              or x in self.values_to_check), record[key]))
+            else:
+                if len(self.values_to_check) == 0 \
```
jmartinm: I usually prefer:

```python
if not self.values_to_check
```

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,308 @@
...
+                        or record[key] in self.values_to_check:
+                    del record[key]  # Mark for deletion
```
jmartinm: You are deleting it, no need for the comment

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,308 @@
...
+                if self.update_regex:
```
jmartinm: Are `values_to_check` and `update_regex` incompatible? If you have both and you delete the key above then here you will get a key error in `record[key]`

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,308 @@
...
+                                del record[key]
+                return
+        else:
+            if isinstance(record[key], list):
+                for array_record in record[key]:
+                    self.apply_action(array_record, new_schema, position+1, checked)
+            else:
+                self.apply_action(record[key], new_schema, position+1, checked)
+        if isinstance(record[key], list):
+            record[key] = [item for item in record[key] if item]
```
jmartinm: The problem with doing `if item` is that if you get primitive values with value `False` or `0` you are going to remove them, but that might be valid values.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,308 @@
...
+        if not record[key]:
```
jmartinm: same as above, this might delete a valid value `False`

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,308 @@
...
+                                    re.escape(value_to_check),
+                                    array_record):
+                                record[key][index] = self.value
+                    else:
+                        if array_record in self.values_to_check:
+                            record[key][index] = self.value
+            else:
+                if len(self.values_to_check) == 0 \
+                        or record[key] in self.values_to_check:
+                    record[key] = self.value  # Mark for deletion
```
jmartinm: Outdated comment

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,308 @@
...
+            return
+        if position + 1 == len(self.keys):
+            if isinstance(record[key], list):
+                record[key].append(self.value)
+        else:
+            if isinstance(record[key], list):
+                for array_record in record[key]:
+                    self.apply_action(array_record, new_schema, position+1, checked)
+            else:
+                self.apply_action(record[key], new_schema, position+1, checked)
```
jacquerie: This is not a good class, because it's just a function call in disguise. Watch this talk to get what I mean: https://www.youtube.com/watch?v=o9pEzgHorH0.

jacquerie:
Before you continue writing code I'd recommend that you sit down and think carefully about the module that you're writing. I suggest in particular these questions:

1. What data do I need to represent the action that a user chose to do?
2. How should I represent a sequence of actions?
3. And how should I represent a set of records that the user selected?
4. What other data do I need to represent an execution of a sequence of actions on a set of records?
5. Which data needs to be persisted for later use?

After you're done with these questions, think about the API.

1. Which of the data that I chose to persist will the frontend need?
2. How do these translate to a set of REST endpoints?
3. What other things will the frontend need that look more like RPCs?
4. Roughy how many pages will I need to implement?

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,332 @@
...
+    assert actions.create_schema_record(schema_2, key, value) == target_object
+
+
+def test_record_creation_array():
+    """should test sub_record creation for missing object"""
+    key = ['authors']
+    value = {'full_name': 'success'}
+    target_object = {'authors': [{'full_name': 'success'}]}
+    curr_path = os.path.dirname(__file__)
+    with open(os.path.join(curr_path,
```
jmartinm: The schema snippet should be inline.

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,332 @@
...
+                           'fixtures/schema.json'))\
+            as data_file:
+        schema = json.load(data_file)
+    assert actions.create_schema_record(schema, key, value) == target_object
+
+
+def test_record_creation_3():
```
jmartinm: Name tests closer to what they do:

```
test_update_with_missing_keys
```

inspirehep/modules/multieditor/queries.py
```diff
@@ -0,0 +1,122 @@
...
+from __future__ import absolute_import, print_function, division
+from ..search import api
+from invenio_records.api import Record
+
+
+def get_records_from_query(query_string, size, page, index):
+    if index == 'hep':
+        query_result = api.LiteratureSearch().query_from_iq(query_string).params(
+            size=size,
+            from_=((page - 1) * 10),
```
jmartinm: At some point we will have to make the size of the page dynamic, so instead of hardcoding the `10` this value has to come from the request.

inspirehep/modules/multieditor/queries.py
```diff
@@ -0,0 +1,122 @@
...
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Module for quering in multi record editor used in http://inspirehep.net."""
+from __future__ import absolute_import, print_function, division
+from ..search import api
+from invenio_records.api import Record
+
+
+def get_records_from_query(query_string, size, page, index):
+    if index == 'hep':
```
jmartinm: A more compact way of doing all this:

```python
cls_map = {
    'hep': api.LiteratureSearch,
    'authors': api.AuthorsSearch,
    'data': api.DataSearch
}

query_result = cls_map[index]().query_from_iq(query_string).params(
            size=size,
            from_=((page - 1) * 10),
            _source=['control_number']
        ).execute()
```

inspirehep/modules/multieditor/queries.py
```diff
@@ -0,0 +1,122 @@
...
+    ids = [q.meta.id for q in query_result]
+    db_records = Record.get_records(ids)
+    return {
+        'json_records': db_records,
+        'total_records': query_result.hits.total
+    }
+
+
+def get_record_ids_from_query(query_string, index):
+    if index == 'hep':
```
jmartinm: Same as above to make it compact.

As we talked, here you are only going to get the default size of 10 ids in all queries.

inspirehep/modules/multieditor/views.py
```diff
@@ -0,0 +1,70 @@
...
+        index = searched_records['schema']
+    if all_selected:
+        ids = filter(lambda x: x not in checked_ids, ids)
+    else:
+        ids = checked_ids
+    actions.run_user_actions(user_actions, ids, index)
+    return 'success'  # fixme should return logged errors
+
+
+@blueprint.route("/search")
```
jmartinm: Add `methods`

inspirehep/modules/multieditor/views.py
```diff
@@ -0,0 +1,70 @@
...
+def search():
+    """Basic view."""
+    query_string = request.args.get('query_string', '')
+    page_num = int(request.args.get('page_num', 1))
+    index = request.args.get('index', '')
+    session['searched_records'] = {
```
jmartinm: Put a multieditor prefix: `multieditor_searched_records `

inspirehep/modules/multieditor/views.py
```diff
@@ -0,0 +1,70 @@
...
+    """Basic view."""
+    user_actions = request.json['user_actions']
+    checked_ids = request.json['ids']
+    all_selected = request.json['allSelected']
+    searched_records = session.get('searched_records', [])
+    if searched_records:
+        ids = searched_records['ids']
+        index = searched_records['schema']
+    if all_selected:
+        ids = filter(lambda x: x not in checked_ids, ids)
```
jmartinm: you have explained this to me a few times, but by reading the code is still not clear what `all_selected` means. I think you need a better naming.

tests/integration/multieditor/test_multiedit_api.py
```diff
@@ -0,0 +1,115 @@
...
+    with open(os.path.join(curr_path,
+                           'fixtures/test_records.json')) as data_file:
+        records = json.load(data_file)
+    for record in records:
+        db_records.append(InspireRecord.create(record))
+    db.session.commit()
+    es.indices.refresh('records-hep')
+    yield populate_db
+    for db_record in db_records:
+        db_record._delete(force=True)
```
jmartinm: did you make sure the database is in clean state after this fixture runs? I think you need `db.session.commit()`

tests/integration/multieditor/test_multiedit_api.py
```diff
@@ -0,0 +1,115 @@
...
+                        'f0746746-689f-4e64-b131-e23e37cc1ec7',
+                        '47e8b29b-48e0-4909-8063-97d6405c6e0b'],
+                'allSelected': False,
+                'index': 'hep'
+            }),
+        )
+    records = Record.get_records(['8569ad80-7996-43fb-a5b0-7b152a607668',
+                                 'f0746746-689f-4e64-b131-e23e37cc1ec7', '47e8b29b-48e0-4909-8063-97d6405c6e0b'])
+    for record in records:
+        assert 'success' in record.authors
```
jmartinm: I am not sure how this works, since 'success' is a value of a dictionary:

```python
In [1]: author = {
   ...: 'full_name': 'success',
   ...: 'foo': 'bar'
   ...: }

In [2]: 'success' in author
Out[2]: False
```

jmartinm:
Besides the review comments, some tests to add:

* Addition:
    * Add a top level key of a record (with and without where keys and where values)
    * Add to a primitive array (with and without where keys and values)
    * test addition to object and addition to array with where_keys and where_values
    * Test Additions with where_regex=True
* Similarly for updates and deletion some more combinations should be tested.
* `check_value` not tested
* For the test examples, it would be nice to have real snippets, like authors instead of made up keys.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,307 @@
...
+                if not check_value(record=record, keys=self.where_keys, values_to_check=self.where_values,
+                                   regex=self.where_regex, position=position):
+                    return
+                checked = True
+        if not record.get(key):  # adding to object as well
+            creation_keys = self.keys[position:]
+            record.update(create_schema_record(schema, creation_keys, self.value))
+            return
+        if position + 1 == len(self.keys):
+            if isinstance(record[key], list):
```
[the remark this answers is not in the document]
Glignos: this is working I made a test case for it

tests/integration/multieditor/test_multiedit_api.py
Glignos: Yes you were right it needs the commit for the db to change.

inspirehep/modules/multieditor/queries.py
```diff
@@ -0,0 +1,70 @@
...
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Module for quering in multi record editor used in http://inspirehep.net."""
+from __future__ import absolute_import, print_function, division
+from ..search import api
+from invenio_records.api import Record
+
+cls_map = {
```
jmartinm: just a naming convention, but since it is a global, name it in capital letters.

inspirehep/modules/multieditor/queries.py
```diff
@@ -0,0 +1,70 @@
...
+# along with INSPIRE; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Module for quering in multi record editor used in http://inspirehep.net."""
+from __future__ import absolute_import, print_function, division
```
jmartinm: Add one space before and after this line.

tests/integration/multieditor/test_multiedit_api.py
```diff
@@ -0,0 +1,116 @@
...
+from invenio_accounts.testutils import login_user_via_session
+from invenio_records.api import Record
+
+
+@pytest.fixture
+def populate_db(app):
+    db_records = []
+    curr_path = os.path.dirname(__file__)
+    with open(os.path.join(curr_path,
+                           'fixtures/test_records.json')) as data_file:
```
jmartinm: I would prefer if this was simplified to not use any fixtures and just use the demo records in the database.

Since the functionality will be unitested you just need to do some small tests here without modifying much the demo records. You can for example do an action and then the opposite action so that they stay the same.

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,527 @@
...
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, print_function, division
+import json
+import os
+
+from inspirehep.modules.multieditor import actions
```
jmartinm: No need for this import, just import all the needed functions in the next line.

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,527 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, print_function, division
```
jmartinm: Add space under

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,527 @@
...
+import json
+import os
+
+from inspirehep.modules.multieditor import actions
+from inspirehep.modules.multieditor.actions import Addition, Deletion, Update
+
+schema_1 = {
```
jmartinm: unused?

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,527 @@
...
+        "type": "array",
+        "uniqueItems": 'true'
+      }},
+  "type": "object",
+}
+
+curr_path = os.path.dirname(__file__)
+with open(os.path.join(curr_path, 'fixtures/schema.json')) \
+        as data_file:
+    schema = json.load(data_file)
```
jmartinm: If this is needed, it should be converted into a fixture that then is added as a parameter to the test that needs it.

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,527 @@
...
+        ],
+        "preprint_date": "2016"
+    }
+    add = Addition(keys=['preprint_date'], where_keys=['public_notes', 'value'], where_regex=False,
+                   where_values=['test'], value="2016")
+    add.apply_action(record, schema)
+    assert record == expected_map
+
+
+def test_addition_root_key_with_where_neg():
```
jmartinm: Don't know what `where_neg` means

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,527 @@
...
+      'key_b': {'key_c': {'key_d': 'val'}}
+    }
+    expected_map = {
+      'key_a': [{'key_c': ['val5', 'val4']},
+                {'key_c': ['val4']},
+                {'key_c': ['val3']}],
+      'key_b': {'key_c': {'key_d': 'val'}}
+    }
+
+    delete = Deletion(values_to_check=["val6", "val1"], keys=['key_a', 'key_c'], where_keys=[], where_regex=False,
```
jmartinm: There seems to be some funny indentation here.

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,527 @@
...
+    expected_map = {'abstracts': [{'not_source': 'success'}]}
+    update = Update(values_to_check=[], keys=['abstracts', 'source'], where_keys=[], where_regex=False,
+                    where_values=[], values_to_check_regex=False, value="success")
+    update.apply_action(record, {})
+    assert record == expected_map
+
+
+def test_record_creation_2():
+    """should test sub_record creation for missing object"""
+    schema_2 = {
```
jmartinm: you can just name it `schema`

jmartinm:
Tests are now covering most of the lines. Check https://coveralls.io/builds/13147059/source?filename=inspirehep%2Fmodules%2Fmultieditor%2Factions.py and you can write some more tests to cover the few missing lines.

jmartinm:
Also in all actions the `__init__` is mostly the same and the first ~15 lines of `apply_action` as well. This can be moved to a parent class `Action`

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,276 @@
...
+            if new_schema['items']['type'] == 'object':
+                new_schema = new_schema['items']['properties']
+            if len(path) == 1:
+                temp_record[key] = [value]
+                return record
+            else:
+                temp_record[key] = [{}]
+                temp_record = temp_record[key][0]
+    temp_record[path[-1]] = value
+    return record
```
jacquerie: I recommend reading http://okigiveup.net/arguments-against-json-driven-development/ to understand why this kind of code is problematic both in the short term and in the long run.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,299 @@
...
+    key = keys[position]
+    if not record.get(key):
+        if match_type == 'does not exist':
+            return True
+        else:
+            return False
+    temp_record = record[key]
+    if isinstance(temp_record, list):
+        for index, array_record in enumerate(temp_record):
+            if position + 1 == len(keys):
```
jmartinm: Like above, in Python to me it looks more natural: 

```python
if position == len(keys) - 1:
```

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,299 @@
...
+                if match_type == 'is equal to' and\
+                        array_record == value_to_check:
+                    return True
+                elif match_type == 'contains' and\
```
jmartinm: Even though it might go above the line limit, in this case it would be more readable to not break the line.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,299 @@
...
+                        value_to_check in array_record:
+                    return True
+                elif match_type == 'matches regular expression'and\
+                        re.search(
+                            re.escape(value_to_check),
+                            array_record):
+                            return True
+                elif match_type == 'does not exist' and\
+                        value_to_check == array_record:
```
jmartinm: Is this correct? if it does not exist I guess the check should be different?

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,299 @@
...
+                    temp_record == value_to_check:
+                return True
+            elif match_type == 'contains' and \
+                    value_to_check in temp_record:
+                return True
+            elif match_type == 'matches regular expression' and \
+                    re.search(
+                        re.escape(value_to_check),
+                        temp_record):
+                return True
```
jmartinm: Just checking, `does not exist` does not make sense in the context of checking an object?

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def get_actions(user_actions):
+    class_actions = []
+    where_actions = []
+    if not user_actions:
+        return
+
+    if user_actions.get('conditions'):
+        for action in user_actions.get('conditions'):
```
jmartinm: Since you do the `.get` above no need here.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,299 @@
...
+            if not action['key']:
+                continue
+            key = action['key'].split('/')
+            where_action = {'value': action['value'],
+                            'key': key,
+                            'match_type': action['matchType']}
+            where_actions.append(where_action)
```
jmartinm: The fact than on one side it is called `conditions` and then `where_actions` can be confusing, can the naming be unified?

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,299 @@
...
+
+    for user_action in user_actions['actions']:
```
jmartinm: So the actions key will always be there I assume, this is enforced on the frontend?

inspirehep/modules/multieditor/tasks.py
```diff
@@ -0,0 +1,45 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+from __future__ import absolute_import, print_function, division
```
jmartinm: Check spacing in other fields before/after `from __future__ import`

inspirehep/modules/multieditor/tasks.py
```diff
@@ -0,0 +1,45 @@
...
+@shared_task(ignore_result=True)
+def process_records(records_ids, user_actions, schema):
+    commit_record = False
+    records = Record.get_records(records_ids)
+    class_actions = get_actions(user_actions)
+    for record in records:
+        for class_action in class_actions:
+            class_action.apply_action(record, schema)
+        if class_action.changed:  # if the record has been touched commit it and reset value
+            commit_record = True
```
jmartinm: You don't need this variable, inside of this `if` you can add:

```python
record.commit()
```

inspirehep/modules/multieditor/tasks.py
```diff
@@ -0,0 +1,45 @@
...
+            class_action.changed = False
+        if commit_record:
```
jmartinm: And this if will not be needed.

inspirehep/modules/multieditor/views.py
```diff
@@ -0,0 +1,101 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2017 CERN.
```
jmartinm: Check license, they are all 2014-2017

inspirehep/modules/multieditor/views.py
```diff
@@ -0,0 +1,101 @@
...
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Module for backend of multi record editor used in http://inspirehep.net."""
+
+from __future__ import absolute_import, print_function, division
+
+from flask import Blueprint, request, jsonify, session
+from jsonschema import ValidationError
+from . import queries
```
jmartinm: Put all the local imports together. This one close to `from . import actions`

inspirehep/modules/multieditor/views.py
```diff
@@ -0,0 +1,101 @@
...
+blueprint = Blueprint(
+    'inspirehep_multieditor',
+    __name__,
+    url_prefix='/multieditor',
+)
+
+
+@blueprint.route("/update", methods=['POST'])
+def update():
+    """Basic view."""
```
jmartinm: Is it though? Change the docstring to something more meaninful

inspirehep/modules/multieditor/views.py
```diff
@@ -0,0 +1,101 @@
...
+    user_actions = request.json['userActions']
+    checked_ids = request.json['ids']
+    all_selected = request.json['allSelected']
+    searched_records = session.get('multieditor_searched_records', [])
+    if searched_records:
+        ids = searched_records['ids']
+        index = searched_records['schema']
+        schema = load_resolved_schema(index)
+    if all_selected:
```
jmartinm: Can this be `True` and `searched_records` empty? In which case `ids` will not be defined.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,299 @@
...
+                 where_regex=False, where_actions=None):
+        self.keys = keys
+        self.value = value
+        self.match_type = match_type
+        self.value_to_check = value_to_check
+        self.where_regex = where_regex
+        self.where_actions = where_actions
+        self.changed = False
+
+    def progress_keys(self, record, schema, position, checked):
```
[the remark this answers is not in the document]
Glignos: No, its progessing to the next position for the keys and schema.

inspirehep/modules/multieditor/actions.py
Glignos: Yes it will contain at least one action.

inspirehep/modules/multieditor/views.py
Glignos: Actually I removed this logic from the UI so now I will change this to remove the ids of the request from the total ones.

inspirehep/modules/multieditor/actions.py
Glignos: Actually the check is being made at the start of the code, the code for "does not exist" was wrong in the array.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,282 @@
...
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, print_function, division
+
+import re
+
+
+class Action(object):
+    def __init__(self, keys, value=None, match_type=None, value_to_check=None,
```
jmartinm: Since this is the base action, here we need a docstring that explains what each of the __init__ parameters is.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,282 @@
...
+        if schema:  # for testing purposes
+            if schema['type'] == 'object':
+                new_schema = schema['properties'][key]
+            elif schema['type'] == 'array':
+                new_schema = schema['items']['properties'][key]
+
+        if self.conditions:
+            for condition in self.conditions:
+                if position < len(condition['key']) and key != condition['key'][position]\
+                        and (position == 0 or self.keys[position - 1] == condition['key'][position - 1]):
```
jmartinm: In this `and`, the main key is equal to the condition key while on the comment you say 

```
# if the main key is different from the condition key for the first time
```

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,282 @@
...
+
+class Addition(Action):
+
+    def apply_action(self, record, schema, position=0, checked=0):
+        """Recursive function to add a record object."""
+        new_schema, checked, key, condition_failed = self.progress_keys(record, schema, position, checked)
+        if condition_failed:  # if the condition check was negative stop the action
+            return
+        if not record.get(key):
+            if self.conditions and checked < len(self.conditions):
```
jmartinm: instead of `checked`, that can be called `conditions_passed`

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,282 @@
...
+                return  # if the conditions that passed are less
+                #  than the total ones the subrecord of the condition is missing and the action wont be applied
```
jmartinm:
```
#  than the total ones, the subrecord of the condition is missing and the action won't be applied
```

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,282 @@
...
+        if position == len(self.keys) - 1:
+            if isinstance(record[key], list):
+                if self.match_type == 'regex':
+                        record[key] = filter(lambda x: (not re.search(
+                            re.escape(self.value_to_check),
+                            x)), record[key])
+
+                elif self.match_type == 'equal':
+                    record[key] = filter(lambda x: not x == self.value_to_check, record[key])
+
```
jmartinm: Remove extra space here and around all these if-elif

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,282 @@
...
+                temp_record[key] = [value]
+                return record
+            else:
+                temp_record[key] = [{}]
+                temp_record = temp_record[key][0]
+    temp_record[path[-1]] = value
+    return record
+
+
+def check_value(record, match_type, keys, value_to_check, position):
```
jmartinm: Are there tests for this function to make sure it is doing the correct thing?

inspirehep/modules/multieditor/tasks.py
```diff
@@ -0,0 +1,48 @@
...
+
+@shared_task(ignore_result=True)
+def process_records(records_ids, user_actions, schema):
+    commit_record = False
+    records = Record.get_records(records_ids)
+    class_actions = get_actions(user_actions)
+    for record in records:
+        for class_action in class_actions:
+            class_action.apply_action(record, schema)
+            if class_action.changed:  # if the record has been touched commit it and reset value
```
jmartinm: I think it is better to remove the comment to avoid confusion, since the code is quite self explanatory

inspirehep/modules/multieditor/views.py
```diff
@@ -0,0 +1,97 @@
...
+    searched_records = session.get('multieditor_searched_records', [])
+    if searched_records:
+        ids = searched_records['ids']
+        index = searched_records['schema']
+        schema = load_resolved_schema(index)
+        ids = filter(lambda x: x not in checked_ids, ids)
+    for i, chunk in enumerate(chunker(ids, 20)):
+        tasks.process_records.delay(records_ids=chunk, user_actions=user_actions, schema=schema)
+
+    return 'success'
```
jmartinm: it's better to return a JSON response for the moment. Then we will change this to use websockets.

inspirehep/modules/multieditor/actions.py
Glignos: Yes this function is being run for every condition, so all the conditions in the tests are triggering it.

inspirehep/modules/multieditor/__init__.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2017 CERN.
```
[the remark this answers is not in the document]
Glignos: A ok, misinterpreted the comment.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,348 @@
...
+       :type conditions: array(object)
+        """
+        self.keypath = keypath
+        self.value = value
+        self.match_type = match_type
+        self.value_to_check = value_to_check
+        self.conditions = conditions
+        self.changed = False
+
+    def process_keypath(self, record, position, conditions_passed):
```
harunurhan: what does this do actually, maybe more descriptive name would be better.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,348 @@
...
+        """Recursive function to add a record object."""
+        conditions_passed, key, condition_failed = self.process_keypath(record, position, conditions_passed)
+        if not schema:
+            return
+        if schema['type'] == 'object':
+            new_schema = schema['properties'][key]
+        elif schema['type'] == 'array':
+            new_schema = schema['items']['properties'][key]
+        if condition_failed:  # if the condition check was negative stop the action
+            return
```
harunurhan: this `return` should be at the beginning.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,348 @@
...
+        if not record.get(key):
+            if self.conditions and conditions_passed < len(self.conditions):
+                return  # if the conditions that passed are less
```
harunurhan: lose the comment it's kinda self explanatory.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,348 @@
...
+                                       value_to_check=condition.get('value', ''),
+                                       match_type=condition.get('match_type', ''), position=position):
+                        condition_failed = True
+                    conditions_passed = conditions_passed + 1  # number of conditions that passed successfully
+        return conditions_passed, key, condition_failed
+
+
+class Addition(Action):
+
+    def apply(self, record, schema, position=0, conditions_passed=0):
```
harunurhan: since we have class maybe it's ok store these states  (`position` and `condition_passed`) as an instance property instead of passing everytime.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,348 @@
...
+                for array_record in record[key]:
+                    self.apply(array_record, schema, position + 1, conditions_passed)
+            else:
+                self.apply(record[key], schema, position + 1, conditions_passed)
+
+
+def create_schema_record(schema, path, value):
+    """Object creation in par with the schema."""
+    record = {}
+    temp_record = record
```
harunurhan: I don't know python internals but this might end up `record` and `temp_record` having same references!
Maybe what you want actually `temp_record = {}`

harunurhan: now that checked again, you actually don't even use record at all, but modify `temp_record` which reflects on `record` then return `record`. So just remove `temp_record`

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,348 @@
...
+    new_schema = schema
```
harunurhan: just use `schema` instead of assigning it to `new_schema`.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,348 @@
...
+            elif match_type == 'regex' and re.search(
+                    value_to_check, temp_record):
+                return True
+        else:
+            return check_value(temp_record, match_type,
+                               keypath, value_to_check, position + 1)
+    return False
+
+
+# def get_actions(user_actions):
```
harunurhan: ?

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,348 @@
...
+#             class_action.apply(record, schema)
+#     return records
+#
+
+def compare_records(old_records, new_records, schema):
+    json_patches = []
+    errors = []
+    for index, new_record in enumerate(new_records):
+        json_patch = []
+        for dif in diff(old_records[index], new_record):
```
harunurhan: personally I don't like non english variable names such as `dif`

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,348 @@
...
+            dif = list(dif)
+            path = force_list(dif[1])  # in case the path is a string and not an array
+            if dif[0] == 'add':
+                if not path[0]:
+                    path = force_list(dif[2][0][0])
```
harunurhan: `dif[1][2][3][4]` looks terrible, I get that it's the library you're using but I suggest to wrap it and return a better format to use it here. 

so instead of `path = force_list(dif[2][0][0])` would be nice to see `path = diff_result.path`

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,348 @@
...
+def check_value(record, match_type, keypath, value_to_check, position):
+    """Function that checks the validity of the condition."""
+    key = keypath[position]
+    if not record.get(key):
+        return match_type == 'missing'
+    temp_record = record[key]
+    if isinstance(temp_record, list):
+        for index, array_record in enumerate(temp_record):
+            if position + 1 == len(keypath):
+                if match_type == 'exact' and array_record == value_to_check:
```
harunurhan: would be better to have a map where key is `match_type` and value is `match_function` so that we don't have to use `if` and more importantly duplicate them.

Example:
```python
{
  'exact': lambda a, b: a == b
...
}
```

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+
+@blueprint.route("/update", methods=['POST'])
+@multieditor_use_permission.require(http_exception=403)
+def update():
+    """Apply the user actions to the database records."""
+    user_actions = request.json['userActions']
+    all_selected = request.json['allSelected']
+    multieditor_session = session.get('multieditor_session', [])
+    if multieditor_session:
+        ids = multieditor_session['ids']
```
jmartinm: This will fail if above you assign to `multieditor_session` the default value of `[]`

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+            ids = set(ids) & set(request.json['ids'])
+    else:
+        return jsonify({'message': 'Please use the search before you apply actions'}), 400
+    actions = []
+    conditions = sanitize_user_conditions(user_actions.get('conditions', []))
+
+    for action in user_actions.get('actions', []):
+        actions.append(
+            ActionFactory.factory(action.get('actionName', ''),
+                                  **sanitize_user_action(action, conditions))
```
jmartinm: It could be better to be consistent and move all the sanitization above:

```python
conditions = sanitize_user_conditions(user_actions.get('conditions', []))
actions = sanitize_user_actions(user_actions.get('actions', []))
```

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+        else:
+            ids = set(ids) & set(request.json['ids'])
+    else:
+        return jsonify({'message': 'Please use the search before you apply actions'}), 400
+    actions = []
+    conditions = sanitize_user_conditions(user_actions.get('conditions', []))
+
+    for action in user_actions.get('actions', []):
+        actions.append(
+            ActionFactory.factory(action.get('actionName', ''),
```
jmartinm: I don't see the value of this being a class. A map `actionname: class_name` would be better.

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+    return jsonify({'message': 'Records are being updated'})
+
+
+@blueprint.route("/preview", methods=['POST'])
+@multieditor_use_permission.require(http_exception=403)
+def preview():
+    """Preview the user actions in the first (page size) records."""
+    user_actions = request.json['userActions']
+    query_string = request.json['queryString']
+    page_size = int(request.json['pageSize'])
```
jmartinm: Is `pageSize` always passed? If not `int(None)` will crash. You could also have a default value:

```python
page_size = request.form.get('pageSize', 10, type=int)
```
You can use the same method for the other arguments -  see http://werkzeug.pocoo.org/docs/0.12/datastructures/#werkzeug.datastructures.MultiDict.get

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+    page_num = request.json['pageNum']
+    multieditor_session = session.get('multieditor_session', [])
```
jmartinm: you expect `multieditor_session` to be a `dict` and not a `list`

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+
+    if not multieditor_session:
+        return jsonify({'message': 'Please use the search before you apply actions'}), 400
+
+    records = queries.get_records_from_query(query_string, page_size, page_num, multieditor_session['schema_name'])['json_records']
```
jmartinm: Above https://github.com/inspirehep/inspire-next/pull/2669/files#diff-6a582a185e685dc4bd77e3504aba3be7R79 you use `multieditor_session['schema']`, what is each used for?

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+
+    old_records = copy.deepcopy(records)
+    actions = []
+    conditions = sanitize_user_conditions(user_actions.get('conditions', []))
```
jmartinm: This line and the ones below are the same as in the function above, you can extract them into their own function.

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+@blueprint.route("/search", methods=['GET'])
+@multieditor_use_permission.require(http_exception=403)
+def search():
+    """Search for records using the query and store the result's ids"""
+    query_string = request.args.get('queryString', '')
+    page_num = int(request.args.get('pageNum', 1))
+    page_size = int(request.args.get('pageSize', 1))
+    schema_name = request.args.get('index', '')
```
jmartinm: `index` and `schema_name` are different concepts

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+    return jsonify({'json_records': records, 'errors': errors, 'json_patches': json_patches})
+
+
+@blueprint.route("/search", methods=['GET'])
+@multieditor_use_permission.require(http_exception=403)
+def search():
+    """Search for records using the query and store the result's ids"""
+    query_string = request.args.get('queryString', '')
+    page_num = int(request.args.get('pageNum', 1))
+    page_size = int(request.args.get('pageSize', 1))
```
jmartinm: 1 as default pageSize seems a bit small

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+    schema_name = request.args.get('index', '')
+    schema = load_schema(schema_name, resolved=True)
+    paginated_records = queries.get_records_from_query(query_string, page_size, page_num, schema_name)
+    if paginated_records['total_records'] > 10000:
```
jmartinm: I would make this number configurable.

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+        return jsonify({'message': 'Please narrow the results using a query'}), 400
```
jmartinm: And in this message if you show it in the UI, indicate that search results should be less than `10000`

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+def sanitize_user_action(action, conditions=[]):
+    action_kwargs = {}
+    keypath = action.get('mainKey', []).split('.')
+    if keypath:
+        action_kwargs['keypath'] = keypath
+        action_kwargs['value'] = action.get('value')
+        action_kwargs['match_type'] = match_type_map[action.get('matchType')]
+        action_kwargs['conditions'] = conditions
+        if action.get('updateValue'):
+            action_kwargs['value_to_check'] = action.get('updateValue')
```
jmartinm: why do you need a different name?

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+
+from inspire_schemas.api import load_schema
+from inspirehep.modules.multieditor import tasks
+from inspirehep.modules.migrator.tasks import chunker
+from .actions import ActionFactory, compare_records
+from . import queries
+from .permissions import multieditor_use_permission
+
+
+match_type_map = {
```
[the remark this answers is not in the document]
Glignos: Well if that would be the case then we need the mapping in the frontend since we want a different value in the UI.

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+@multieditor_use_permission.require(http_exception=403)
+def update():
+    """Apply the user actions to the database records."""
+    user_actions = request.json['userActions']
+    all_selected = request.json['allSelected']
+    multieditor_session = session.get('multieditor_session', [])
+    if multieditor_session:
+        ids = multieditor_session['ids']
+        if all_selected:
+            ids = set(ids) - set(request.json['ids'])
```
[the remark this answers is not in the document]
Glignos: In case that the all_selected is true then it takes a subset of all the records minus the ones that where desselected, I will make it clearer

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,160 @@
...
+        return jsonify({'message': 'Please narrow the results using a query'}), 400
+    session['multieditor_session'] = {
+        'ids': queries.get_record_ids_from_query(query_string, schema_name),
+        'schema': schema,
+        'schema_name': schema_name
+    }
+    return jsonify(paginated_records)
+
+
+def sanitize_user_action(action, conditions=[]):
```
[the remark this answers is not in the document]
Glignos: That's true, I understand that it's a better practice, but in my case that shouldn't have been a problem, right?(for reference reasons :) )

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,911 @@
...
+            }
+        ],
+        "document_type": ["book"]
+    }
+    add = Addition(keypath=['titles', 'subtitle'], value="success")
+    add.apply(record, get_schema)
+    assert record == expected_map
+
+
+def test_addition_array_with_condition_condition(get_schema):
```
david-caro: What is `condition_condition`? can you clarify in the name of the test? Don't be shy with test names, it's better to make them understandable than short named.

tests/unit/multieditor/test_actions.py
```diff
@@ -0,0 +1,911 @@
...
+from inspirehep.modules.multieditor.actions import Addition, Deletion, Update, create_schema_record
+
+
+@pytest.fixture
+def get_schema():
+    curr_path = os.path.dirname(__file__)
+    with open(os.path.join(curr_path, 'fixtures/schema.json')) \
+            as data_file:
+        schema = json.load(data_file)
+    return schema
```
david-caro: Instead of using a whole schema, it's better to use snippets of it for the example records you use, and add a validation step for them on the same test, like https://github.com/inspirehep/inspire-next/blob/master/tests/unit/utils/test_utils_record.py#L36

david-caro: That way the test will only be affected if that part of the schema changes, and when it does, we will know right away (reducing variability and ensuring early detection).

david-caro:
I forgot to mention, but it's also missing an alembic recipe.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,376 @@
...
+        :param conditions_passed: number of conditions that got checked and passed
+        :return: returns the new number of conditions passed,
+         the new key for this recursion and if a condition got checked and the failed the check
+        """
+        key = self.keypath[position]
+        condition_failed = False
+        if self.conditions:
+            for condition in self.conditions:
+                should_check = False
+                if position < len(condition.get('keypath')):
```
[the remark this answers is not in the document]
Glignos: This used the conditions as received from the frontend.
Now im sanitizing them as you can see in serializers.py.
So instead of .get im just going to access simply the key of the dict.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,376 @@
...
+class Addition(Action):
+    """Class that when applied adds a new primitive field or object to the selected path
+    if the conditions are satisfied"""
+    def apply(self, record, schema, position=0, conditions_passed=0):
+        """
+        :param record: record in which the field to be updated resides
+        :param schema: schema on which our record is based
+        :param position: level of recursion
+        :param conditions_passed: number of conditions that run and passed
+        :return:
```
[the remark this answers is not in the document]
Glignos: The action can still be used since im not creating a new instance of the class. Im just replacing the record to be processed in the existing one.Also in most cases I have no use of the old record and its more efficient for me when processing the records to have in the memory just the ones that are being processed.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,376 @@
...
+            if isinstance(record[key], list):
+                record[key].append(self.value)
+                self.changed = True
+        else:
+            new_schema = schema_progression(schema, key)
+            if isinstance(record[key], list):
+                for array_record in record[key]:
+                    self.apply(array_record, new_schema, position + 1, conditions_passed)
+            else:
+                self.apply(record[key], new_schema, position + 1, conditions_passed)
```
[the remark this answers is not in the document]
Glignos: There are no empty lists in our records, so if the case is that the list is not existing, it's key is going to be missing so im going to create it. :)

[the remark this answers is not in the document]
david-caro: That is something that might change in the future (not having empty lists), and it does not hurt now to make sure here to be able to handle those too, so better be on the safe side (it will be a mess to debug later in case we add empty lists).

inspirehep/modules/multieditor/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+                action.changed = False
+        if commit_record:
+            try:
+                record.commit()
+            except (ValidationError, Exception) as e:
+                errors.append(e.message)
+            else:
+                commit_session = True
+            commit_record = False
+    if commit_session:
```
david-caro: There should not be a need for this if, you should be able to commit the session even if empty afaik, make sure.

inspirehep/modules/multieditor/actions.py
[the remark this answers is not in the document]
david-caro: Exactly because you are not creating new instances, is why you can't reuse them, as once the `self.changed` is set, it will always be true. It's also only set in the `apply` method, so you don't really need a state to be passed around to other methods, or future calls, that's why it does not really fit as instance property, but more as return value of the `apply` method.
If what you want is to keep a flag about if any of the records processed so far has changed, I'd suggest changing the name of the class and the method, something like `ActionProcessor` for the basic one, `DeleteProcessor` for the delete, and then instead of `apply` should be `process`. That way it's clear that it's not applying an action to a record, but processing a set of records.

> Im just replacing the record to be processed in the existing one.Also in most cases I have no use of the old record and its more efficient for me when processing the records to have in the memory just the ones that are being processed.

I'm not sure what you mean with this. I don't see how passing the changed as return value modifies the way you process records. For example, in the `process_records` task, instead of:
```python
    for record in records:
        for action in actions:
            action.apply(record=record, schema=schema)
            if action.changed:
                commit_record = True
                action.changed = False
```
you'll have:
```python
    for record in records:
        for action in actions:
            if action.apply(record=record, schema=schema):
                commit_record = True
```

inspirehep/modules/multieditor/actions.py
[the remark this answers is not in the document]
david-caro: 👍

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,376 @@
...
+def serialize_value(value, schema):
+    if schema['type'] == 'array':
+        schema = schema['items']
+    if schema['type'] == 'integer':
+        serialized_value = int(value)
+    elif schema['type'] == 'boolean':
+        if value.lower() == 'true':
+            serialized_value = True
+        elif value.lower() == 'false':
+            serialized_value = False
```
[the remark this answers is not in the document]
Glignos: The frontend sends a stringified value, im not sure if I can use ```json.loads``` to get the right value, since for instance if a date contains only a year it's still not a number but a string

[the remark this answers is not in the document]
david-caro: So, we are sending stuff like:
```
{
    "date": 2017,
}
```
?

btw. If we are using some 'serializer' that's not actually json, we should at some point :/
If not, we are inventing our own...

inspirehep/modules/multieditor/bundles.py
```diff
@@ -0,0 +1,39 @@
...
+
+js = NpmBundle(
+    "node_modules/ng2-multi-record-editor/dist/inline.bundle.js",
+    "node_modules/ng2-multi-record-editor/dist/vendor.bundle.js",
+    "node_modules/ng2-multi-record-editor/dist/polyfills.bundle.js",
+    "node_modules/ng2-multi-record-editor/dist/main.bundle.js",
+    depends=("ng2-multi-record-editor/dist/*.js"),
+    output="gen/inspirehep-multi-record-editor.%(version)s.js",
+    npm={
+        "ng2-multi-record-editor": "dev",
```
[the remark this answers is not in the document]
Glignos: yes this is pointing to the dev releases of multieditor which we are releasing atm

inspirehep/modules/multieditor/ext.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self, app=None):
+        if app:
+            self.init_app(app)
+
+    def init_app(self, app):
+        self.init_config(app)
+        app.register_blueprint(api_blueprint)
+        app.extensions['inspire-multieditor'] = self
+
+    def init_config(self, app):
```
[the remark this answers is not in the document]
Glignos: Yes this configuration is only needed in the api side.Actually we dont need any app config because right now the only endpoint in app is the view template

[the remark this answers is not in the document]
david-caro: If it's not needed, then remove :), the best code is no code 👍

inspirehep/modules/multieditor/serializers.py
```diff
@@ -0,0 +1,122 @@
...
+    if not check_provided_key_path(keypath, schema):
+        return None
+    if keypath:
+        action_kwargs['keypath'] = keypath
+        action_kwargs['value'] = action.get('value')
+        action_kwargs['match_type'] = match_type_map[action.get('matchType')]
+        action_kwargs['conditions'] = conditions
+        if action.get('updateValue'):
+            action_kwargs['update_value'] = action.get('updateValue')
+    return action_kwargs
```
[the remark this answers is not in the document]
Glignos: I added a check for the matchtype so if there is a key and a matchtype I consider it a valid action, despite not affecting the records

inspirehep/modules/multieditor/serializers.py
```diff
@@ -0,0 +1,122 @@
...
+def sanitize_user_conditions(user_conditions, schema):
+    """
+    Sanitizes the user conditions and serializes them in the class format
+    :param user_conditions: user conditions as received from the frontend
+    :return: returns the serialized conditions
+    """
+    conditions = []
+    for condition in user_conditions:
+        if not condition.get('key'):
+            continue
```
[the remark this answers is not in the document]
Glignos: this is only in the case there is a condition with an empty key in which case it will just ignore it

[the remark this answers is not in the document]
david-caro: Are you sure? why would it be a condition with an empty key?

inspirehep/modules/multieditor/serializers.py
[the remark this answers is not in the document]
david-caro: What do you mean? If I use this I get:
```python
> sanitize_user_action(schema={}, action={'mainKey': {'keypath': ''}}) == {}
True
```
So if for whatever reason the frontend submits an actualy invalid action, the sanitize does not find it's bad no?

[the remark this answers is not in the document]
david-caro: hmmm... nop, `''.split('.')` gives a non-empty list `['']`...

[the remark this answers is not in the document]
david-caro: It will actually giving you a key error when trying to get the action from the`match_type_map` dict...  Not nice either :/

inspirehep/modules/multieditor/tasks.py
```diff
@@ -0,0 +1,61 @@
...
+def process_records(records_ids, user_actions, schema):
+    """
+    :param records_ids: ids of the records to be processed
+    :param user_actions: user actions as received from frontend
+    :param schema: corresponding schema for the records to be processed
+    """
+    commit_record = False
+    commit_session = False
+    errors = []
+    actions = get_actions(user_actions, schema)
```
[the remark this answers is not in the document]
Glignos: For the errors we will be using websockets but we havent emplemented that yet

inspirehep/modules/multieditor/serializers.py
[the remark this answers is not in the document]
david-caro: this will give you a strange action though:
```python
sanitize_user_action(
    schema={},
    action={'mainKey': '', 'matchType': 'contains'}},
)
```
That will give:
```
{
    'keypath': [''],
    'value', None,
    'match_type': 'contains',
    'conditions': None,
}
```

Is that a valid one?

inspirehep/modules/multieditor/actions.py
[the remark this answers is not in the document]
Glignos: Ah okay got it! Thanks :)

inspirehep/modules/multieditor/serializers.py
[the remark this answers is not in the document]
Glignos: This will pass throught the key validation of ```is_valid_key_path``` and get rejected as a non valid key thus rejecting the action as well.

inspirehep/modules/multieditor/serializers.py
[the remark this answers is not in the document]
Glignos: Well I talked about this with Stella and It's better for the users to have by a default a "sticky" condition even if they dont fill it when submitting.

tests/integration/multieditor/test_multiedit_api.py
```diff
@@ -0,0 +1,165 @@
...
+
+    expected_patch = [{u'path': [u'authors', 1], u'value': {u'full_name': u'success'}, u'op': u'add'}]
+    assert expected_patch == json.loads(response.data)['json_patches'][0]
+
+
+def test_multieditor_update_api(api_client):
+    login_user_via_session(api_client, email='cataloger@inspirehep.net')
+    response = api_client.get('/multieditor/search?pageNum=1&queryString=control_number:736770&index=hep')
+
+    api_client.post(
```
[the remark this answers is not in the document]
Glignos: Im actually reverting the actions in the next request and then asserting that the records are clean.

tests/integration/multieditor/test_multiedit_api.py
```diff
@@ -0,0 +1,165 @@
...
+                    }
+                ]
+            },
+            'ids': [],
+            'allSelected': True
+        }),
+    )
+    records = Record.get_records(json.loads(response.data)['uuids'])
+    if records[0].get('authors'):
+        assert 'success' not in records[0]['authors'][-1]['full_name']
```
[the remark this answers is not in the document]
Glignos: As I said in my previous comments I delete the changes that I made to the records, thus testing the deletion actions and reverting the records to theire initial state.

inspirehep/modules/multieditor/tasks.py
Glignos: Wouldnt that cost me in vain?If there is nothing to commit why should I commit the session. Well to be honest If its instantaneous it wont be a problem either.

inspirehep/modules/multieditor/actions.py
```diff
@@ -0,0 +1,376 @@
...
+    :param keypath: path in which the field should be checked
+    :param update_value: value to be matched with the conditions
+    :param position: current recursion level
+    :return:
+    """
+    key = keypath[position]
+    if not schema:
+        return False
+    if not record.get(key):
+        return match_type == 'missing'
```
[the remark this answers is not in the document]
Glignos: Well not really, if the conditions field is not there this just means that the condition failed.

Glignos:
ci test please

drjova:
For JSON Diff we can use <https://github.com/stefankoegl/python-json-patch>, it has otb support for diffs and it is used by invenio-records :)

``` python
a = {'a': 'hello', 'c': ['a', 'b', 'c']}
b = {'a': 'hello', 'c': ['b', 'c']}

patch = jsonpatch.make_patch(a, b)
print(json.dumps(patch.patch))
```

🎉 

```json
[
  {
    "op": "remove",
    "path": "/c/0"
  }
]

```

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,147 @@
...
+    session['multieditor_session'] = {
+        'uuids': uuids,
+        'schema': schema,
+    }
+    records_uuids, records = queries.get_paginated_records(page_number=page_number,
+                                                           page_size=page_size,
+                                                           uuids=uuids)
+    return jsonify({'total_records': total_records,
+                    'uuids': records_uuids,
+                    'json_records': records})
```
drjova: For the REST API we can use ``MethodView`` provided from Flask, we can separate the resources based on the logic, i.e.

```python
class MultiEditorRecordItemResource(MethodView):

  def post(self):
      """...."""

multieditor_record_item = MultiEditorRecordItemResource('multieditor_record_item')
blueprint.add_url_rule(
    '/multieditor/record',
    view_func= multieditor_record_item,
)
```

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,147 @@
...
+
+    return jsonify({'json_records': old_records, 'errors': errors, 'json_patches': json_patches, 'uuids': uuids})
+
+
+@blueprint.route("/paginate", methods=['GET'])
+@multieditor_use_permission.require(http_exception=403)
+def paginate():
+    """Get paginated records from the session"""
+    page_number = int(request.args.get('pageNum', 1))
+    page_size = int(request.args.get('pageSize', 10))
```
drjova: what about just ``page`` and ``size``? :)

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,157 @@
...
+                'schema': schema,
+            }
+            records_uuids, records = queries.get_paginated_records(number=number,
+                                                                   size=size,
+                                                                   uuids=uuids)
+        else:
+            multieditor_session = session.get('multieditor_session', {})
+            if not multieditor_session:
+                return jsonify({'message': 'Please refresh your page'}), 400
+            records_uuids, records = queries.get_paginated_records(number=number,
```
drjova: We could use ``queries.get_paginated_records`` for both cases

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,157 @@
...
+        json_patches, errors = compare_records(old_records, records, multieditor_session['schema'])
+
+        return jsonify({'json_records': old_records, 'errors': errors, 'json_patches': json_patches, 'uuids': uuids})
+
+
+preview_api = PreviewAPI.as_view('preview_api')
+blueprint.add_url_rule('/preview', view_func=preview_api, methods=['POST'])
+
+
+class SearchAPI(MethodView):
```
drjova: maybe a small description what it's ``MethodView`` does? :)

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,157 @@
...
+    decorators = [multieditor_use_permission.require(http_exception=403)]
+
+    def get(self):
+        """Search for records using the query and store the result's ids"""
+        query = request.args.get('q', None)
+        number = int(request.args.get('number', 1))
```
drjova:
```python
 request.args.get('number', 1, type=int)
``` 
:)

inspirehep/modules/multieditor/api.py
```diff
@@ -0,0 +1,157 @@
...
+from .actions import compare_records
+from . import queries
+from .permissions import multieditor_use_permission
+from .serializers import get_actions
+from .exceptions import InvalidValue, SchemaError, InvalidActions
+
+blueprint = Blueprint(
+    'inspirehep_multieditor',
+    __name__,
+    template_folder='templates',
```
drjova: probably we can remove it :)

inspirehep/modules/multieditor/config.py
```diff
@@ -0,0 +1,28 @@
...
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""ArXiv configuration."""
+
+from __future__ import absolute_import, division, print_function
+
+
+MULTI_MAX_RECORDS = 10000
```
drjova: Why don't we move this in the main ``config.py``? :)

inspirehep/modules/multieditor/config.py
```diff
@@ -0,0 +1,28 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""ArXiv configuration."""
```
drjova: Arxiv :P

inspirehep/modules/multieditor/exceptions.py
```diff
@@ -0,0 +1,38 @@
...
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+
+class InvalidActions(Exception):
```
drjova:
```python
class InvalidActions(Exception):
    """Invalid action."""

    def __init__(self):
        """Init with message."""
        super(InvalidActions, self).__init__("The actions that were provided are invalid")
```

inspirehep/modules/multieditor/queries.py
```diff
@@ -0,0 +1,87 @@
...
+}
+
+
+def get_total_records(query, index):
+    """
+    :param query: query string
+    :param index: index of the records to be searched
+    :return: returns the total records that match our query
+    """
+    query_result = INDEX_TO_QUERIES[index]()\
```
drjova: maybe a space before ``\`` :)

inspirehep/modules/multieditor/config.py
david-caro: This is the pattern that we are using for other modules too, but a broader discussion on how to architect it is welcome :)

inspirehep/modules/multieditor/exceptions.py
david-caro: Don't you think that this strips out usability from the exception? For example, you would not be able to do something like:
```python
if action not in available_actions:
    raise InvalidActions('Action "%s" is not a valid action, valid actions are %s' % (action, available_actions))
```

That would help a lot in debugging (as you can see the actual action value passed and the allowed ones).

david-caro: I agree though that if you directly use a superclass method, you must use super (it never hurts, but it solves possible issues).

Glignos:
Superseded by #3141

PR #2670, opened 2017-08-23
tests: add integration tests for the arxiv module

iulianav:
## Description

I have written a series of integration tests for all the possible xml response messages that the 'arxiv' module might have to deal with.

## Related Issue

Previously, this module was not tested at all. This is not good practice.

## Motivation and Context

Following TDD mantra prior to refactoring of the 'arxiv' module, i.e. reimplementing it with blueprints.

## Checklist:

### DIFF
--- a/tests/integration/arxiv/test_arxiv_views.py
+++ b/tests/integration/arxiv/test_arxiv_views.py
@@ -1,0 +1,151 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


tests/integration/arxiv/test_arxiv_views.py
```diff
@@ -0,0 +1,118 @@
...
+
+        requests_mocker.register_uri(
+            'GET',
+            'http://export.arxiv.org/oai2',
+            text=pkg_resources.resource_string(
+                __name__, os.path.join('fixtures', 'malformed_response_message.xml')),
+        )
+
+        assert app_client.get(
+            '/arxiv/search?arxiv=foo-bar-baz').status_code == 422
```
jacquerie: The same goes for the error messages (although I'm not sure if the JavaScript uses them to do stuff...)

PR #2673, opened 2017-08-24
config: use the right capitalization for LArSoft

jacquerie:
## Description:
Note that this commit requires to recreate the `collection` table in
production, otherwise the records in the LArSoft special collection
would be assigned to a no longer existing collection.

Because of this I'm not going to merge it now, as deployments are
still blocked, and it wouldn't make sense to pile up more manual ops.

CC: @david-caro @kaplun @jmartinm 

## Checklist:

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -258,11 +258,11 @@ INSPIRE_COLLECTIONS_DEFINITION = [
     },
     {
         'query': 'special_collections:LARSOFT-INTERNAL-NOTE',
-        'name': 'LARSOFT Internal Notes',
+        'name': 'LArSoft Internal Notes',
     },
     {
         'query': 'special_collections:LARSOFT-NOTE',
-        'name': 'LARSOFT Notes',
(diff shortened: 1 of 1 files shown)


michamos:
if I understood correctly what alembic was about, couldn't you write an alembic script for that?

jacquerie:
> if I understood correctly what alembic was about, couldn't you write an alembic script for that?

True, although (AFAIK) `alembic` is typically used for SQL schema migrations (e.g. adding or removing tables or columns) rather than data migrations.

This also requires a release of `invenio-collections`, otherwise this migration would have no base commit (since the alembic migrations of `invenio-collections` are still in an unreleased commit).

jacquerie:
Blocked by https://github.com/inveniosoftware/invenio-collections/pull/82.

jacquerie:
> Blocked by inveniosoftware/invenio-collections#82.

Not anymore, because commit 7a2e3a8 made searching on this field case-insensitive.

PR #2685, opened 2017-08-28
records: readd the affiliation suggester

harunurhan:


jacquerie:
@StellaCh this is the change needed from the backend. @harunurhan has to do some small changes to the frontend too, but his change should be quick.

PR #2688, opened 2017-08-29
workflows.matching: fix legacy existance check

david-caro:
* We were doing a query that never returned any results, so we thought
  the records were new when checking for their existance in the legacy
  system.

We were seeing duplicated records being submitted to legacy, this is
the first guess towards answering why.

### DIFF
--- a/inspirehep/modules/workflows/tasks/matching.py
+++ b/inspirehep/modules/workflows/tasks/matching.py
@@ -77,7 +77,7 @@ def match_by_arxiv_id(record):
     arxiv_id = get_arxiv_id(record)
 
     if arxiv_id:
-        query = '035:"{0}"'.format(arxiv_id)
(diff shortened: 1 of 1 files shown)


david-caro:
Fixes #2689

kaplun:
Any volunteer for adding a test where legacy is mocked with this interface?

david-caro:
How do you propose to write such a test? Hardcoding the query in 2 places (the code and the test) does not look like a good test to me. Legacy also allows sending any query, so a mock that accepts any query would not be useful either imo. I'm curious.

jacquerie:
Now it's no longer useful, but at the time in which this code was originally written writing this test would have prevented this error. In order to create the mock one would have sent a query to legacy, noticed that it didn't actually do its job, and corrected the code.

david-caro:
So the test itself would not have been useful, but the process of creating it?

PR #2691, opened 2017-08-29
crossref: refactor module

iulianav:
## Checklist:

### DIFF
--- a/inspirehep/modules/crossref/config.py
+++ b/inspirehep/modules/crossref/config.py
@@ -20,25 +20,16 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-"""Crossref extension."""
+"""Crossref configuration."""
 
 from __future__ import absolute_import, division, print_function
 
-from . import config
-from .views import blueprint
 
+CROSSREF_API_URL = 'http://api.crossref.org/works/'
 
-class InspireCrossref(object):
-    def __init__(self, app=None):
-        self.app = app
-        if app is not None:
-            self.init_app(app)
-
-    def init_app(self, app):
-        self.init_config(app)
-        app.register_blueprint(blueprint)
-
-    def init_config(self, app):
-        for k in dir(config):
-            if k.startswith('CROSSREF_'):
-                app.config.setdefault(k, getattr(config, k))
+CROSSREF_RESPONSE_CODES = {
+    'success': 200,
+    'notfound': 404,
+    'malformed': 422,
+    'multiplefound': 300,
+}
--- a/inspirehep/modules/crossref/core.py
+++ b/inspirehep/modules/crossref/core.py
@@ -20,25 +20,39 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
-"""Crossref extension."""
+"""Crossref core."""
 
 from __future__ import absolute_import, division, print_function
 
-from . import config
-from .views import blueprint
+import requests
+from flask import current_app, jsonify, request
+from six.moves.urllib.parse import urljoin
 
 
-class InspireCrossref(object):
-    def __init__(self, app=None):
-        self.app = app
-        if app is not None:
-            self.init_app(app)
+def get_response(crossref_doi):
+    response = requests.get(
+        urljoin(
+            current_app.config['CROSSREF_API_URL'],
+            '{term}'.format(term=crossref_doi.strip()),
+        ),
+    )
+    return response
 
-    def init_app(self, app):
-        self.init_config(app)
-        app.register_blueprint(blueprint)
 
-    def init_config(self, app):
-        for k in dir(config):
-            if k.startswith('CROSSREF_'):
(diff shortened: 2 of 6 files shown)


inspirehep/modules/crossref/views.py
```diff
@@ -0,0 +1,50 @@
...
+
+"""INSPIRE CrossRef blueprint."""
+
+from __future__ import absolute_import, division, print_function
+
+from flask import Blueprint, jsonify, request
+from flask_login import login_required
+
+from .core import get_json
+from .config  import response_code
```
jacquerie: In general you don't want to do this, as it prevents this configuration from being overridden at a later point. If instead you do it like this https://github.com/inspirehep/inspire-next/blob/6bc729117b8b5721736c150341c78870392c87a4/inspirehep/modules/hal/ext.py#L41-L44
then you'll find that you can modify things in the `current_app.config` any time you want.

inspirehep/modules/crossref/views.py
```diff
@@ -0,0 +1,50 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""INSPIRE CrossRef blueprint."""
```
jacquerie: Nit: apparently the correct capitalization is "Crossref". Their GitHub organization is called "CrossRef", though...

inspirehep/modules/crossref/views.py
```diff
@@ -0,0 +1,50 @@
...
+def search(doi=None):
+    """Search for given DOI."""
+    doi = doi or request.args.get("doi")
+
+    # query crossref
+    result = get_json(doi)
+
+    resp = jsonify(result)
+    resp.status_code = response_code.get(result['status'], 200)
+    return resp
```
jacquerie: You should configure your editor to leave newlines at the end of files automatically, otherwise someone else that touches this file will generate a spurious diff on this.

inspirehep/modules/crossref/ext.py
```diff
@@ -91,57 +82,4 @@ def init_app(self, app):
         """Initialize a Flask application."""
         app.config.setdefault("CROSSREF_API_URL",
                               "http://api.crossref.org/works/")
```
jacquerie: If you use the `init_config` pattern described below you can move this in `config.py`.

inspirehep/modules/crossref/core.py
```diff
@@ -0,0 +1,61 @@
...
+# or submit itself to any jurisdiction.
+
+"""CrossRef Core."""
+
+from __future__ import absolute_import, division, print_function
+
+from flask import current_app, request, jsonify
+
+import requests
+from urlparse import urljoin
```
jacquerie: Nit: it's not enforced by anything, but the imports are separated in sections according to some rules. In general:

1. Future imports
2. Standard library imports
3. Vendored from outside CERN
4. Vendored from inside CERN
5. INSPIRE packages (external)
6. INSPIRE packages (internal)

Moreover, imports are sorted alphabetically, but imports of whole modules come first. All in all, here it becomes:
```python

from urlparse import urljoin

import requests
from flask import current_app, jsonify, request
```

jacquerie: Although it's better for Python 3 compatibility to use `six` here, since `urlparse` was moved. So in fact:
```python

import requests
from flask import current_app, jsonify, request
from six.moves.urllib.parse import urljoin
```

inspirehep/modules/crossref/config.py
```diff
@@ -0,0 +1,31 @@
...
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""CrossRef Configuration."""
+
+from __future__ import absolute_import, division, print_function
+
+
+response_code = {'success': 200,
```
jacquerie: Constants should be in `ALL_CAPS_SEPARATED_BY_UNDERSCORES`, and if you use the `init_config` pattern described below this could become `CROSSREF_RESPONSE_CODES`.

On the other hand there's no need to obsess particularly about this, as we already know that this dictionary is going to  disappear soon with the better approach. Read this piece of documentation and try to figure out yourself how it should be: http://flask.pocoo.org/docs/0.12/api/#flask.Flask.make_response

jacquerie:
You should always check what Travis says about your PR! The tests were not passing: https://travis-ci.org/inspirehep/inspire-next/builds/270372252.

At the very least, you should run the unit tests locally before submitting, so that you're sure that your PR doesn't contain style issues. I know you were running the integration tests locally, but that was not enough: notice how the acceptance tests were failing.

Speaking of which: why? What did I change here in order to make them pass, and what will you need to change in order to make https://github.com/inspirehep/inspire-next/pull/2694 pass?

Hint: http://werkzeug.pocoo.org/docs/0.12/routing/

PR #2716, opened 2017-09-04
records: role based permissions

jmartinm:
* Permissions when accessing records will now take into account
  the view-restricted-collection action in roles.

* Adds back lowercase_analyzer for _collections field in ES mappings,
  which is needed for correct filtering of records.

* Adds extra permission tests.

## Checklist:

### DIFF
--- a/inspirehep/modules/records/mappings/records/authors.json
+++ b/inspirehep/modules/records/mappings/records/authors.json
@@ -11,6 +11,7 @@
                     "type": "string"
                 },
                 "_collections": {
+                    "analyzer": "lowercase_analyzer",
                     "type": "string"
                 },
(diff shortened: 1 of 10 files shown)


jmartinm:
The previous error was caused by the fact that integration tests have now a `SECRET_KEY` and things were being added to the session.

Since tests share the `SECRET_KEY` they can be affected by things that were added to the session by other tests (in this case causing the side effect that the anonymous user had some non-allowed collections added in the query filter).

Fix was to clear the cache in the fixture (see [here](https://github.com/inspirehep/inspire-next/pull/2716/files#diff-8eedb14f28e016fe0e021789e25878abR223))

PR #2721, opened 2017-09-06
workflows: use local scratch space

kaplun:
Reducing usage of EOS, when what is needed is only a temporary disk space.

Rather than using local .tar directory (currently on a networked
filesystem) as a scratch space for extracting plots and author.xml
file, uses a temporary local directory that is cleaned up at the end.

### DIFF
--- a/inspirehep/modules/authors/permissions.py
+++ b/inspirehep/modules/authors/permissions.py
@@ -23,7 +23,7 @@
 from __future__ import absolute_import, division, print_function
 
 from invenio_access.permissions import (
-    DynamicPermission,
+    Permission,
     ParameterizedActionNeed
 )
 
@@ -32,6 +32,6 @@ action_admin_holdingpen_authors = ParameterizedActionNeed(
     'admin-holdingpen-authors', argument=None
 )
 
-holdingpen_author_permission = DynamicPermission(
+holdingpen_author_permission = Permission(
     action_admin_holdingpen_authors
 )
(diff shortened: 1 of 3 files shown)


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -122,24 +124,28 @@ def arxiv_plot_extract(obj, eng):
     tarball = obj.files[filename]
 
     if tarball:
+        scratch_space = mkdtemp('plot_extract')
```
kaplun: I discovered we can globally set what root directory to use :+1: by simply setting: 
```python
import tempfile
tempfile.tempdir = '/bla'
```

in the config file.

kaplun:
cc: @jacquerie, @david-caro. Yay!

inspirehep/modules/workflows/tasks/arxiv.py
david-caro: It also respects the env vars 'TMPDIR' and similar.

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -122,24 +124,28 @@ def arxiv_plot_extract(obj, eng):
...
-            plots = process_tarball(tarball.file.uri)
-        except (InvalidTarball, NoTexFilesFound):
-            obj.log.error(
-                'Invalid tarball %s for arxiv_id %s', tarball.file.uri, arxiv_id)
-            return
-        except DelegateError as err:
-            obj.log.error('Error extracting plots for %s. Report and skip.', arxiv_id)
-            current_app.logger.exception(err)
-            return
+            try:
```
david-caro: This double try is ugly :/, maybe we can us ea context manager instead?

PR #2724, opened 2017-09-06
fixtures: extend user fixtures

jmartinm:
* Adds real INSPIRE use cases to user related fixtures.

* Amends permission tests to use the new fixtures.

## Checklist:


tests/integration/test_detailed_records.py
```diff
@@ -41,19 +41,3 @@ def test_all_records_are_valid(app):
...
-
-
-def test_all_records_are_there(app_client):
-    failed = []
-
-    for record in [record.json for record in RecordMetadata.query.all()]:
-        try:
-            absolute_url = record['self']['$ref']
-            relative_url = absolute_url.partition('api')[2]
-            response = app_client.get(relative_url)
```
jmartinm: I am removing this test since not all records found RecordMetadata.query.all() should be found without the appropriate permission (now that I added a restricted record to the demo records) and this functionality is already tested in `test_permissions.py`

jacquerie: But this is testing more than the permissions, this is testing that the detailed record page renders correctly for all records, that is, no exception is raised while rendering the templates. As we're going to extensively rewrite and refactor these templates soon, this test is still useful.

jmartinm:
Any more comments on this PR @jacquerie ? Otherwise I will fix the thing you mentioned and should be ready.

tests/integration/test_permissions.py
```diff
@@ -361,47 +162,64 @@ def test_inspire_search_filter(app, app_client, user_info, total_count, es_filte
...
+def test_inspire_search_filter_restricted_collection(app, app_client, user_info, total_count, es_filter):
+    """Test default inspire search filter."""
+
+    if user_info:
+        login_user_via_view(app_client, email=user_info['email'],
+                            password=user_info['password'],
+                            login_url='/login/?local=1')
+
+    # Doing a client request creates a request context that allows the
+    # assert to correctly use the logged in user.
```
jacquerie: Better: https://github.com/inspirehep/inspire-next/blob/fcb1263266d851ba4651a0123455ad2ff3974916/tests/unit/conftest.py#L75-L78

jacquerie: (You'll need to copy this from `tests/unit/conftest.py` to `tests/integration/conftest.py`.)

tests/integration/test_permissions.py
```diff
@@ -28,324 +28,125 @@
...
-    user_partially_allowed = User(
-        email='partially_allowed@inspirehep.net',
-        password=hashed_password,
-        active=True,
-    )
-    user_allowed = User(
-        email='allowed@inspirehep.net',
-        password=hashed_password,
-        active=True,
-    )
```
jacquerie: There's other places in which this refactoring has to happen. Grep for `users` in `tests` and you should find several other examples.

inspirehep/modules/fixtures/users.py
```diff
@@ -28,48 +28,96 @@
...
+        )
+    db.session.commit()
+
+
+def init_users():
+    """Sample users, not to be used in production."""
+    ds = current_app.extensions['invenio-accounts'].datastore
+    superuser = Role.query.filter_by(name='superuser').one()
+    cataloger = Role.query.filter_by(name='cataloger').one()
+    hermes_curator = Role.query.filter_by(name='hermescurator').one()
```
jacquerie: Nit: we should use consistently either `cataloger` or `curator`.

inspirehep/modules/fixtures/users.py
```diff
@@ -28,48 +28,96 @@
...
+    )
+    db.session.add(ActionRoles(
+        action='view-restricted-collection',
+        argument='HERMES Internal Notes',
+        role=hermes_collections)
+    )
     db.session.commit()
+
+
+def init_all_users():
```
jacquerie: Since this function is still initializing the permissions, the previous name was better. But what about adding an `init_all_fixtures` function that calls all the fixtures (collections, files, users, roles, permissions...) in the right order?

tests/integration/test_permissions.py
jmartinm: Tried this but does not work, at least just with `with app.test_request_context()`

PR #2729, opened 2017-09-08
editor: record revisions api

harunurhan:
* Adds views for getting revisions and reverting.

### DIFF
--- a/inspirehep/modules/editor/views.py
+++ b/inspirehep/modules/editor/views.py
@@ -26,8 +26,14 @@ from __future__ import absolute_import, division, print_function
 
 from flask import Blueprint, jsonify, request
 from flask_login import current_user
+from sqlalchemy_continuum import transaction_class, version_class
 
+from invenio_db import db
+from invenio_records.models import RecordMetadata
+
+from inspirehep.modules.pidstore.utils import get_pid_type_from_endpoint
 from inspirehep.modules.tools import authorlist
+from inspirehep.utils.record_getter import get_db_record
 from inspirehep.utils.references import (
     get_refextract_kbs_path,
     map_refextract_to_schema,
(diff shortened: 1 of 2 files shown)


inspirehep/modules/editor/views.py
```diff
@@ -42,6 +51,47 @@
...
 
 
+@blueprint.route('/revisions/<rec_uuid>/revert', methods=['PUT'])
+def revert_to_revision(rec_uuid):
+    """Revert given record to given revision"""
+    record = Record.get_record(rec_uuid)
+    revision_id = request.json['revision_id']
+    record.revert(revision_id)
+    db.session.commit()
+    return jsonify(succes=True)
```
jmartinm: Typo ;)

inspirehep/modules/editor/views.py
```diff
@@ -42,6 +51,56 @@
...
+
+    pid = PersistentIdentifier.query.filter_by(pid_value=recid).one()
+    record = get_db_record(pid.pid_type, recid)
+
+    revisions = []
+    for revision in record.revisions:
+        transaction_id = revision.model.transaction_id
+
+        user_email = db.session.query(Transaction).filter(
+            Transaction.id == transaction_id).one().user.email
```
harunurhan: can we have a transaction without a user? like first db insert! In that case that would crash.
@jmartinm

inspirehep/modules/editor/views.py
```diff
@@ -42,6 +51,56 @@
...
+
+    pid = PersistentIdentifier.query.filter_by(pid_value=recid).one()
+    record = get_db_record(pid.pid_type, recid)
+
+    revisions = []
+    for revision in record.revisions:
+        transaction_id = revision.model.transaction_id
+
+        user_email = Transaction.query.filter(
+            Transaction.id == transaction_id).one().user.email
```
harunurhan: can we have a transaction without a user? like first db insert! In that case that would crash.
@jmartinm

inspirehep/modules/editor/views.py
```diff
@@ -42,6 +51,59 @@
                       url_prefix='/editor',)
 
 
+@blueprint.route('/revisions/<rec_uuid>/revert', methods=['PUT'])
```
harunurhan: using `rec_uuid` here and `recid` in another route makes it a bit weird but avoids

```
pid = PersistentIdentifier.query.filter_by(pid_value=recid).one()
record = get_db_record(pid.pid_type, recid)
```

and uses more simple

```record = Record.get_record(rec_uuid)```

Comment (harunurhan) [reply inspirehep/modules/editor/views.py]: but also because of this, we have to return, `rec_uuid` from `get_revisions` method.

**Comment by harunurhan** [comment]
@jacquerie for the test I am actually not sure what to test, here except `'/revisions/<recid>'`'s logic with `for` and `if`s.

**Comment by jacquerie** [comment]
> @jacquerie for the test I am actually not sure what to test

What about the authentication, like @jmartinm suggested?

**Comment by jmartinm** [comment]
> What about the authentication, like @jmartinm suggested?

Once we merge https://github.com/inspirehep/inspire-next/pull/2724 I will create a PR with the permissions for the editor and then we can rebase this PR on top.

File: tests/integration/editor/test_editor_views.py [hunk 5]
```diff
@@ -287,7 +384,8 @@ def test_refextract_text(log_in_as_cataloger, api_client):
 
     assert response.status_code == 200
     assert validate(references, subschema) is None
-    assert get_value({'references': references}, 'references.reference.publication_info.journal_title')
+    assert get_value({'references': references},
+                     'references.reference.publication_info.journal_title')
```
jacquerie: Check your editor! Our `pytest.ini` ignores `E501`, so your style checks should be ignoring it as well.

harunurhan:
tests are passing when I rebased to #2830, so after it is merged this can be merged finally.

@jacquerie you can have a loot at the tests, if they are OK.

jacquerie:
@harunurhan @jmartinm in general, you should sign your commits.

If you don't want to remember adding `-s` you can add it to the configuration:
```shell
$ git config --local format.signoff true
```
(It's even part of the documentation! http://inspirehep.readthedocs.io/en/latest/dev_flow.html.)

jmartinm:
@harunurhan if you are amending the commits, I would also squash into one commit so that the functionality and the tests are in the same commit.

Issue #2687, opened 2017-08-29
541__e is populated differently between hepcrawl and submissions


kaplun:
## Expected Behavior
I'd expect that the field `541__e` sent to legacy would contain the ID of the holding pen entry that I can add to: `http://labs.inspirehep.net/holdingpen/XXX` in order to access the entry.

E.g.

```xml
<datafield tag="541" ind1=" " ind2=" ">
<subfield code="a">orcid:0000-0002-9075-015X</subfield>
<subfield code="b">fischler@zippy.ph.utexas.edu</subfield>
<subfield code="c">submission</subfield>
<subfield code="d">2017-08-28T23:43:50.572572</subfield>
<subfield code="e">708757</subfield>
</datafield>
```

->  https://labs.inspirehep.net/holdingpen/708757

## Current Behavior
In the case of `hepcrawl` here's what is stored instead:

```xml
<datafield tag="541" ind1=" " ind2=" ">
<subfield code="a">arXiv</subfield>
<subfield code="c">hepcrawl</subfield>
<subfield code="e">ac5c1012821e11e7887e02163e010a0e</subfield>
<subfield code="d">2017-08-16T03:03:41.606022</subfield>
</datafield>
```

jacquerie:
So `hepcrawl` is calling `add_acquisition_source` with some weird parameter (and that's a bug in `hepcrawl`), but it can't possibly know at that time the holdingpen entry id, so it must be added later by the worfklow (and that's the enhancement to do here).

PR #2753, opened 2017-09-14
workflows: fix submission number for harvests

iulianav:
## Description
The ``submission_number`` key in ``acquisition_source`` is currently
being populated differently between HEPCrawl and form submissions.
In fact, HEPCrawl puts there the Scrapy job id as it can't possibly
know the workflow object id since it will be generated later. This
workflow step makes sure that they will both contain the workflow id
as we'd might want to link from the record to the Holding Pen entry
from which it came from.

## Related Issue
Closes https://github.com/inspirehep/inspire-next/issues/2687.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -257,6 +257,29 @@ def populate_journal_coverage(obj, eng):
 
 
 @with_debug_logging
+def fix_submission_number(obj, eng):
+    """Ensure that the submission number contains the workflow object id.
+
+    Unlike form submissions, records coming from HEPCrawl can't know yet which
+    workflow object they will create, so they use the crawler job id as their
+    submission number. We would like to have there instead the id of the workflow
+    object from which they came from, so that, given a record, we can link to their
+    original Holding Pen entry.
+
+    Args:
+        obj: a workflow object.
+        eng: a workflow engine.
+
+    Returns:
+        None
+
+    """
+    method = get_value(obj.data, 'acquisition_source.method', default='')
+    if method == 'hepcrawl':
+        obj.data['acquisition_source']['submission_number'] = obj.id
+
+
+@with_debug_logging
 def submission_fulltext_download(obj, eng):
     submission_pdf = obj.extra_data.get('submission_pdf')
     if submission_pdf and is_pdf_link(submission_pdf):
--- a/inspirehep/modules/workflows/workflows/article.py
+++ b/inspirehep/modules/workflows/workflows/article.py
@@ -41,21 +41,22 @@ from inspirehep.modules.workflows.tasks.arxiv import (
 from inspirehep.modules.workflows.tasks.actions import (
     add_core,
     error_workflow,
+    fix_submission_number,
     halt_record,
-    is_record_relevant,
-    populate_journal_coverage,
-    is_record_accepted,
-    reject_record,
+    is_arxiv_paper,
     is_experimental_paper,
     is_marked,
+    is_record_accepted,
+    is_record_relevant,
     is_submission,
-    is_arxiv_paper,
     mark,
     normalize_journal_titles,
+    populate_journal_coverage,
     refextract,
+    reject_record,
+    save_workflow,
     set_refereed_and_fix_document_type,
     submission_fulltext_download,
-    save_workflow,
 )
 
 from inspirehep.modules.workflows.tasks.classifier import (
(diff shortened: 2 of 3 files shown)


iulianav:
@kaplun this is related to the issue you opened:  #2687 .

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -210,6 +210,17 @@ def is_submission(obj, eng):
...
 
 @with_debug_logging
+def add_holdingpen_entry_id(obj, eng):
+    """Add holdingpen entry id for this record in 'submission_number' """
+    source = obj.data.get('acquisition_source')
+    if source:
+        if source.get('method') == "hepcrawl":
+            source['submission_number'] = obj.id
+            return True
+    return False
```
jacquerie: Workflows steps don't need return values: they operate with side effects. You can remove this two lines.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -210,6 +210,17 @@ def is_submission(obj, eng):
 
 
 @with_debug_logging
+def add_holdingpen_entry_id(obj, eng):
+    """Add holdingpen entry id for this record in 'submission_number' """
+    source = obj.data.get('acquisition_source')
+    if source:
+        if source.get('method') == "hepcrawl":
+            source['submission_number'] = obj.id
```
jacquerie: More succinct:
```python
method = get_value(obj.data, 'acquisition_source.method', default='')
if method == 'hepcrawl':
    obj.data['acquisition_source']['submission_number'] = obj.id
```

iulianav:
https://github.com/inspirehep/inspire-next/issues/2687

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -210,6 +210,14 @@ def is_submission(obj, eng):
 
 
 @with_debug_logging
+def add_holdingpen_entry_id(obj, eng):
+    """Add holdingpen entry id for this record in 'submission_number' """
+    method = get_value(obj.data, 'acquisition_source.method', default='')
+    if method == 'hepcrawl':
+        obj.data['acquisition_source']['submission_number'] = obj.id
```
kaplun: I am not sure: don't we always want `obj.id` here?

iulianav: You are right, this method is being called only for "hepcrawl" records anyway, this is a bit of a redundant check. I will fix it!

kaplun: Mmh. So there are two paths that are setting `obj.id` independently?

ammirate:
This PR looks ok but we may want to apply first some refactor to the workflow definition. Moreover, the step  ``CHECK_IF_NOT_SUBMISSION_AND_ADD_HOLDINGPEN_ENTRY_ID`` could be moved inside an already existing step.

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -362,6 +367,7 @@ class Article(object):
         # use that instead of stopping
         CHECK_IF_MERGE_AND_STOP_IF_SO +
         CHECK_IF_SUBMISSION_AND_ASK_FOR_APPROVAL +
+        CHECK_IF_NOT_SUBMISSION_AND_ADD_HOLDINGPEN_ENTRY_ID +
```
ammirate: I would say, instead of creating a new step in the workflow, you could put this enhancement in POSTENHANCE_RECORD. In this way, we comply with the granularity of the current steps :)

ammirate: @iulianav what about this? I think now is the right moment to revive it :)

tests/unit/workflows/test_workflows_tasks_actions.py
```diff
@@ -398,6 +399,48 @@ def test_is_submission_returns_false_if_obj_has_falsy_acquisition_source():
...
+    }
+    extra_data = {}
+    assert validate(data['acquisition_source'], subschema) is None
+
+    obj = MockObj(data, extra_data)
+    eng = MockEng()
+
+    fix_submission_number(obj, eng)
+
+    assert obj.data['acquisition_source']['submission_number'] == 1
```
jacquerie: Mh, I missed the obvious bug here. Can you guess which one, @iulianav, and how we could have prevented it?

iulianav:
The bag consisted in the fact that the assertion does not respect the schema and tricks us into thinking we are populating the field with the right kind of value. The value in the field should be a string, thus when we populate it we should take care to cast the nr. from obj.id to an str. In this wrong case we are populating with a nr. and asserting against a nr., assertion which will pass even if it is wrong. The test fix is to first validate the returned result against the schema and then validate its actual value.

PR #2756, opened 2017-09-15
setup: fix superlance dependency

kaplun:

* Fixes superlance dependency in the worker-node.

* Cleans extra_require definitions.

Co-authored-by: David Caro <david@dcaro.es>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,6 @@ install_requires = [
     'refextract>=0.2.0',
     'orcid',
     'raven<=5.1.0',
-    'flower',
     'langdetect>=1.0.6',
     'librabbitmq>=1.6.1',
     'idutils>=0.2.1',
@@ -118,10 +117,17 @@ extras_require = {
     'postgresql': [
         'invenio-db[postgresql,versioning]>=1.0.0b2',
     ],
+    'build-node': [
+        'ipdb',
+    ],
     'web-node': [
         'gunicorn',
     ],
-    'crawler': [
+    'worker-node': [
+        'superlance',
+        'flower',
+    ],
+    'crawler-node': [
         'hepcrawl~=0.0,>=0.3.4',
     ],
     'tests': tests_require,


kaplun:
This is to clean up the situation where `memmon` is today not properly installed on worker nodes on production.

setup.py
```diff
@@ -118,10 +117,17 @@
     'postgresql': [
         'invenio-db[postgresql,versioning]>=1.0.0b2',
     ],
+    'build-node': [
+        'ipdb',
```
jacquerie: Have you tried `pdbpp` ?

setup.py
```diff
@@ -118,10 +117,17 @@
...
+    ]
```
jacquerie: Still missing a comma here.

setup.py
```diff
@@ -118,10 +117,17 @@
...
     'web-node': [
         'gunicorn',
     ],
-    'crawler': [
+    'worker-node': [
+        'superlance',
+        'flower',
+    ]
```
jacquerie: And here.

setup.py
kaplun: It's also very nice indeed. I've put `ipdb` because that's what was forced in via puppet before. Would you recommend it instead of ipdb?

setup.py
kaplun: D'oh!

setup.py
david-caro: the advantage of pdbpp was that has a 'published' pytest plugin (that you can actually just pip install) right?
Though I'm not used to it :/ (it's also a different shell than the one provided by `inspirehep` app).

jacquerie: I like it better because they have a more pragmatic approach than `ipdb`: they don't have to wait for IPython to implement a feature in order to have it, they just do it on their own. But it's not a blocker here!

kaplun: Actually we can simply have both :+1:

setup.py
```diff
@@ -118,10 +117,18 @@
     'postgresql': [
         'invenio-db[postgresql,versioning]>=1.0.0b2',
     ],
+    'build-node': [
+        'ipdb',
+        'pdbpp',
```
jacquerie: I'd rather have just `ipdb` than both!

setup.py
kaplun: And what about: https://pypi.python.org/pypi/pudb? Brings back so many memories.

PR #2759, opened 2017-09-15
workflows: handle client request to remove PDF

### LINKED ISSUE #2725: literaturesuggest: PDF always attached even if deselected in Holding Pen
@annetteholtkamp reports that when a user submission has a PDF, even if it the checkbox in the detailed Holding Pen page is deselected to indicate the PDF should not get attached to the record, this is ignored and the final record produced always has the PDF.

jmartinm:
* When handling the user action from the Holding Pen page, the
  checkbox to remove the user PDF was not being handled. (closes #2725)

## Checklist:

### DIFF
--- a/tests/integration/workflows/test_hep_approval.py
+++ b/tests/integration/workflows/test_hep_approval.py
@@ -1,0 +1,150 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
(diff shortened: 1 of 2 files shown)


tests/integration/workflows/test_hep_approval.py
```diff
@@ -0,0 +1,152 @@
...
+@pytest.fixture(scope='function')
+def workflow():
+    workflow_object = workflow_object_class.create(
+        data={},
+        id_user=1,
+        data_type="hep"
+    )
+    workflow_object.save()
+    db.session.commit()
+    workflow_object.continue_workflow = lambda **args: True
```
jmartinm: This is not super elegant, but not sure if there is a better way.

tests/integration/workflows/test_hep_approval.py
```diff
@@ -0,0 +1,148 @@
...
+    yield workflow_object
+    WorkflowObjectModel.query.filter_by(id=workflow_object.id).delete()
+    db.session.commit()
```
jacquerie: This is leaving some junk in the DB after the test have run, no?

tests/integration/workflows/test_hep_approval.py
```diff
@@ -0,0 +1,148 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2017 CERN.
```
jacquerie: Nit: wrong year. One day I will figure out where you get these wrong headers... : )

tests/integration/workflows/test_hep_approval.py
jmartinm: Well, this line is precisely removing the Workflow created above, so after each test it should be clean.

tests/integration/workflows/test_hep_approval.py
jmartinm: Oh, I copy paste them and then fix them :P Thought the idea was to have only the last year ;)

tests/integration/workflows/test_hep_approval.py
jacquerie: Whoops! I misread this as a save... In fact, what about separating with a space the `yield` line from the block that creates the `WorkflowObject` and the block that deletes it?

PR #2765, opened 2017-09-19
editor: add /authorlist/text view

michamos:
## Description

The authorlist tool is used by catalogers to parse author lists from PDFs (similar to refextract, but for authors instead of references). It used to produce MARC directly that could be pasted into bibedit.

Now it produces an inspire 3 record instead, that is seralized into MARC for legacy use but can also be used directly by the editor (@harunurhan, could you check that the editor API is OK for you?).

## Checklist:

### DIFF
--- a/inspirehep/modules/tools/templates/authorlist_form.html
+++ b/inspirehep/modules/tools/templates/authorlist_form.html
@@ -36,7 +36,16 @@
 {% if results %}
 <h4>{{ results_title }}</h4>
 <button class="btn btn-default" id="copy-button" data-clipboard-target="resultarea">Copy to clipboard</button>
-<pre><textarea class="form-control" form="authorlist" id="resultarea" rows="10">{{ results }}</textarea></pre>
+<pre><textarea class="form-control" form="authorlist" id="resultarea" rows="10">
+{% for author in results['authors'] %}
+{% set code = '100' if loop.index == 1 else '700' %}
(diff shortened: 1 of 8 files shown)


jacquerie:
Uhm, but we can't merge this before Labs becomes master, as people are using this tool on QA to do their job using the current editor... or am I missing something?

michamos:
yes, you are missing
https://github.com/michamos/inspire-next/blob/d79901dbd1006aff27059aeb3607c2dc8a3eee23/inspirehep/modules/tools/templates/authorlist_form.html#L40-L47

inspirehep/modules/tools/templates/authorlist_form.html
```diff
@@ -36,7 +36,16 @@
...
-<pre><textarea class="form-control" form="authorlist" id="resultarea" rows="10">{{ results }}</textarea></pre>
+<pre><textarea class="form-control" form="authorlist" id="resultarea" rows="10">
+{% for author in results['authors'] %}
+{% set code = '100' if loop.index == 1 else '700' %}
+{{ code }}__ $$a{{ author['full_name'] }}
+{%- for affiliation in author.get('raw_affiliations', []) %}
+$$v{{ affiliation.get('value') }}
+{%- endfor %}
+
+{% endfor %}
```
michamos: this is what you are missing.

PR #2778, opened 2017-09-22
records: add update_collections receiver

### LINKED ISSUE #2775: records: implement a lightweight update_collections receiver
The thing described in https://github.com/inspirehep/inspire-next/pull/2773#issuecomment-331075992.

jacquerie:
## Description:
Implements a lightweight receiver that copies and normalizes the
values of `special_collections` in `_collections`, replacing the
very expensive one removed in f460348 (closes #2775).

## Related Issue
Closes #2775 

## Checklist:

### DIFF
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -47,6 +47,27 @@ from inspirehep.modules.authors.utils import author_tokenize, phonetic_blocks
 from inspirehep.utils.date import create_earliest_date
 
 
+SPECIAL_COLLECTIONS_MAP = {
+    'BABAR-ANALYSIS-DOCUMENT': 'BABAR Analysis Documents',
+    'BABAR-INTERNAL-BAIS': 'BABAR Internal BAIS',
+    'BABAR-INTERNAL-NOTE': 'BABAR Internal Notes',
+    'CDF-INTERNAL-NOTE': 'CDF Internal Notes',
+    'CDF-NOTE': 'CDF Notes',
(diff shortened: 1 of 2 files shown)


inspirehep/modules/records/receivers.py
```diff
@@ -98,6 +119,20 @@ def assign_uuid(sender, *args, **kwargs):
             author['uuid'] = str(uuid.uuid4())
 
 
+@before_record_insert.connect
+@before_record_update.connect
+def update_collections(sender, *args, **kwargs):
```
michamos: Currently, you are only adding to `_collections`, never removing.

I think the behavior should be the following:
- if a record has `special_collections`, it should get the corresponding `_collections` without `Literature`.
- if a record has no `special_collections`, it should get `Literature` in `_special_collections`.

There is still the issue of how to assign stuff having both `Literature` and a special collection.

jacquerie: This is implementing only the behavior of [`copy_special_collections`](https://github.com/inspirehep/inspire-dojson/blob/8c02690ea5d997b46c8dc48e937f6017ddc05176/inspire_dojson/hep/model.py#L85-L90), because it assumes that whatever is in `_collections` is already correct... I think that this commit is correct, but maybe we need to discuss this in person.

jacquerie:
After IRL discussion with @michamos we decided to abandon this separation between `_collections` and `special_collections`: we can just write what we want to have in `_collections` without having to care about copying things around. So we can close this.

PR #2779, opened 2017-09-22
workflows: don't add curation status notes

michamos:
## Description

* After inspirehep/inspire-schemas#231 and inspirehep/inspire-dojson#94, the information about whether the record has been curated can be completely handled by means of the `curated` flag.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -107,19 +107,6 @@ def halt_record(action=None, message=None):
     return _halt_record
 
 
-@with_debug_logging
-def update_note(metadata):
-    """Check if the record was approved as CORE."""
-    new_notes = []
-    for note in metadata.get("public_notes", []):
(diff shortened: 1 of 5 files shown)


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -107,19 +107,6 @@ def _halt_record(obj, eng):
     return _halt_record
 
 
-@with_debug_logging
-def update_note(metadata):
```
michamos: interestingly, this was never called (outside of the tests).

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -279,7 +278,6 @@
 
 POSTENHANCE_RECORD = [
     add_core,
-    add_note_entry,
```
jacquerie: Uh, I didn't notice this before merging. Due to the known problem of migrating the workflow state, in order to deploy this we need to have empty-ish queues.

PR #2807, opened 2017-09-28
utils: use correct logger

chris-asl:
Following the recommendations at #1371, use per module loggers.

### DIFF
--- a/inspirehep/utils/normalizers.py
+++ b/inspirehep/utils/normalizers.py
@@ -22,9 +22,11 @@
 
 from __future__ import absolute_import, division, print_function
 
-from flask import current_app
+import logging
 
 from inspirehep.modules.search.api import JournalsSearch
+
+LOGGER = logging.getLogger(__name__)
 
 
 def normalize_journal_title(journal_title):
@@ -38,7 +40,5 @@ def normalize_journal_title(journal_title):
         try:
             normalized_journal_title = hits[0].short_title
         except (AttributeError, IndexError):
-            current_app.logger.debug(
-                "Failed to normalize journal title in: %s", repr(hits[0])
-            )
+            LOGGER.debug("Failed to normalize journal title in: %s", repr(hits[0]))
     return normalized_journal_title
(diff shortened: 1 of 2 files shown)


jmartinm:
As mentioned during standup, if I remember correctly, my observation last time I looked into loggers is that using the Flask `app` loggers is the correct thing to do (since they should be aware if the app is in debug mode, testing and so on) and after the last fixes I did on `invenio-logging`, that app logger should have the correct handlers (including Sentry).

So please have a look at each of the loggers (with and without Debug mode), for example:

```python
In [2]: app.logger
Out[2]: <flask.logging.DebugLogger at 0x1161e3890>

In [3]: app.logger.handlers
Out[3]:
[<flask.logging.DebugHandler at 0x116da8c10>,
 <flask.logging.ProductionHandler at 0x116ddac10>]

In [4]: app.logger.parent.handlers
Out[4]: []
```

To see what would be the difference between the 2 of them.

jmartinm:
Note that to see the Sentry handler, you need to add to your configuration:

```
SENTRY_DSN = "https://test@sentry.cern.ch/17?verify_ssl=0"
LOGGING_SENTRY_INCLUDE_WARNINGS = False
LOGGING_SENTRY_LEVEL = 'ERROR'
LOGGING_SENTRY_CELERY = True
```

And then you will see it:

```python
In [1]: app.logger
Out[1]: <flask.logging.DebugLogger at 0x10b4159d0>

In [2]: app.logger.handlers
Out[2]:
[<flask.logging.DebugHandler at 0x10d06b610>,
 <flask.logging.ProductionHandler at 0x10d06b6d0>,
 <raven.handlers.logging.SentryHandler at 0x10809d090>]
```

jacquerie:
This became stale, and we now have helpers for logging: https://github.com/inspirehep/inspire-utils/blob/f4f3a788a1c6c8300ecc3160ec2d09e5bbfc0910/inspire_utils/logging.py.

PR #2816, opened 2017-09-29
setup: add theme extension to the API application

szymonlopaciuk:
Inspire-theme added to API entry point

## Description
In setup.py inspire-theme added to api_apps.
Changes made to not render HTML error pages when
in the API application.

## Related Issue
N/A

## Motivation and Context
Enables to use jinja2 templates in the API.

## Checklist:

### DIFF
--- a/inspirehep/factory.py
+++ b/inspirehep/factory.py
@@ -29,14 +29,19 @@ import sys
 
 from invenio_base.app import create_app_factory
 from invenio_base.wsgi import create_wsgi_factory
-from invenio_config import create_conf_loader
+from invenio_config import create_config_loader
 
 from . import config
 
 
 env_prefix = 'APP'
 
-config_loader = create_conf_loader(config=config, env_prefix=env_prefix)
+config_loader = create_config_loader(config=config, env_prefix=env_prefix)
+
+
+def api_config_loader(app, **kwargs_config):
+    return config_loader(app, RESTFUL_API=True, **kwargs_config)
+
 
 instance_path = os.getenv(env_prefix + '_INSTANCE_PATH') or \
     os.path.join(sys.prefix, 'var', 'inspirehep-instance')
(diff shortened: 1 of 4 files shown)


inspirehep/factory.py
```diff
@@ -55,7 +55,7 @@
 """
 
 create_api = create_app_factory(
-    'inspirehep',
```
jmartinm: If I remember correctly, I once tried to change this name, but the consequence is that the configuration file will then not be found by the API application.

In `var/instance/inspirehep.cfg` (the name has to coincide with the app name) so this would not work for us.

ammirate:
I am not 100% sure this is the official way of distinguishing two flask apps, but I didn't find yet a better solution, if any. 
In the meantime, could you add some tests? e.g. calling a private api will return the expected json obj with 401.

ammirate:
As discussed with @david-caro, maybe it would be nice to have an attribute in the app's configuration, e.g. ``RESTFUL_API``: ``True`` in the api app, and then check it in ``error_render_response``. 
Wdyt?

szymonlopaciuk:
@ammirate I might be completely off here, but if I remember correctly part of the problem was that the apps share configuration, no?

david-caro:
I'm not 100% sure, but I think that they don't really share the config, they have a copy each, and we should be able to tweak that during the app creation (in the app factory somewhere).

jacquerie:
What about using the strategy described in https://github.com/pallets/flask/issues/1935#issuecomment-229308296 ? I like the current solution, but that one is even simpler...

david-caro:
> What about using the strategy described in pallets/flask#1935 (comment) ? I like the current solution, but that one is even simpler...

I don't really like relying on the url, but in this case it might be ok (I strongly recommend changing the strategy if we need to differenciate between apps anywhere else).

jacquerie:
> I strongly recommend changing the strategy if we need to differenciate between apps anywhere else

:+1: 

Modifying the configuration is definitely the cleaner approach if we need this information in multiple places, instead of littering the codebase of similar branching statements everywhere we need to know in which app we are.

jacquerie:
In the end I went for your solution, because `request.path` doesn't start with `/api/` if we're using the `api_client`, which would lead to some confusion. I rebased and squashed your commits, but I moved your tests to `tests/unit/theme/test_theme_views.py`, since this is where they belong according to the current naming/tests organization convention.

PR #2831, opened 2017-10-02
global: don't use refersto in queries

jacquerie:
## Description
inspirehep/inspire-query-parser#30 removed support for ``refersto:``,
so we will have to rewrite these queries using ``references__recid``.

## Checklist:

### DIFF
--- a/inspirehep/modules/records/serializers/impactgraph_serializer.py
+++ b/inspirehep/modules/records/serializers/impactgraph_serializer.py
@@ -54,8 +54,8 @@ class ImpactGraphSerializer(object):
         # Get citations
         citations = []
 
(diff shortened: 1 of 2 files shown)


kaplun:
OT: But are we sure users were not using such functionality? Also what is our plan to support n-order queries?

jacquerie:
This was just a workaround to fix the build, not something intended to be permanent.

> Also what is our plan to support n-order queries?

Nth-order queries require an additional visitor, which is currently not implemented. Currently `inspire-query-parser` is a map that transforms strings into ES queries. But an nth order query is not an ES query: it's something that wraps one. So `inspire-query-parser` has to become something that maps strings into things that wrap an ES query!

Issue #2712, opened 2017-09-04
literaturesuggest: be less strict in DOI input


michamos:
## Expected Behavior
the DOI input box should be less strict and attempt to normalize the DOI.
It should:

- strip whitespace
- strip leading `doi:`
- strip leading `(https?://)?(dx\.)?doi.org/`

## Current Behavior
A DOI needs to be entered without any prefix, that is starting with `10.`.

## Steps to Reproduce (for bugs)

1. Enter a DOI starting with `doi:`
2. Watch the exception on sentry: https://sentry.cern.ch/inspire-sentry/inspire-labs/group/865011/

jmartinm:
So there are a 2 different things being asked to do here:

1. Relax the validator on the doi input ([here](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/forms/validation_utils.py#L39)) in order to allow for `(https?://)?(dx\.)?doi.org/`.

2. Add a new normalizer ([here](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/literaturesuggest/views.py#L90)) that will clean up the DOI value to make it conform with the [schema](https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/hep.yml#L698)

iulianav:
@michamos I believe step 2 is wrong because the import happens before the submit and a get request is performed here https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/crossref/views.py#L41

PR #2835, opened 2017-10-04
forms: normalize DOIs at ingestion time

iulianav:
## Description

- I relaxed the form validator on DOI input in order to allow for ` (https?://)?(dx\.)?doi.org/ `.
- I added a new normalizer that cleans up the DOI value to make it conform with the schema.

## Related Issue

Closes https://github.com/inspirehep/inspire-next/issues/2712.

## Motivation and Context

The DOI input box should be less strict and attempt to normalize the DOI.
It should:
    - strip whitespace
    - strip leading doi:
    - strip leading ` (https?://)?(dx\.)?doi.org/ `

## Checklist:

### DIFF
--- a/inspirehep/modules/crossref/core.py
+++ b/inspirehep/modules/crossref/core.py
@@ -33,7 +33,7 @@ def get_response(crossref_doi):
     response = requests.get(
         urljoin(
             current_app.config['CROSSREF_API_URL'],
-            '{term}'.format(term=crossref_doi.strip()),
+            '{term}'.format(term=crossref_doi),
         ),
     )
     return response
--- a/inspirehep/modules/forms/validation_utils.py
+++ b/inspirehep/modules/forms/validation_utils.py
@@ -35,19 +35,18 @@ from requests import RequestException
 
 from wtforms.validators import StopValidation
 
+from idutils import doi_regexp
+
 
 class DOISyntaxValidator(object):
 
     """DOI syntax validator."""
-
-    pattern = "(^$|(doi:)?10\.\d+(.\d+)*/.*)"
 
     def __init__(self, message=None):
         """Constructor.
 
         :param message: message to override the default one.
         """
-        self.regexp = re.compile(self.pattern, re.I)
         self.message = message if message else (
             "The provided DOI is invalid - it should look similar to "
             "'10.1234/foo.bar'.")
@@ -59,7 +58,7 @@ class DOISyntaxValidator(object):
         :param form: validated form.
         """
         doi = field.data
-        if doi and not self.regexp.match(doi):
+        if doi and not doi_regexp.match(doi):
             # no point to further validate DOI which is invalid
             raise StopValidation(self.message)
 
--- a/inspirehep/modules/literaturesuggest/normalizers.py
+++ b/inspirehep/modules/literaturesuggest/normalizers.py
@@ -27,6 +27,7 @@ import json
 from sqlalchemy import text
 from sqlalchemy.orm.exc import NoResultFound
 
+from idutils import normalize_doi
 from invenio_accounts.models import User
 from invenio_db import db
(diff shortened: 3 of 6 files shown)


chris-asl:
Regarding the regex: the dot before `org` should be escaped, because currently it accepts `doi-org/`.

inspirehep/modules/crossref/core.py
```diff
@@ -56,3 +57,12 @@ def get_json(doi):
     data['query'] = query
 
     return data
+
+
+def normalize_doi(doi):
```
michamos: actually, I just found out about `idutils.normalize_doi` which does the right thing.

inspirehep/modules/forms/validation_utils.py
```diff
@@ -40,7 +40,7 @@ class DOISyntaxValidator(object):
 
     """DOI syntax validator."""
 
-    pattern = "(^$|(doi:)?10\.\d+(.\d+)*/.*)"
+    pattern = "(^$|((doi:)|((https?://)?(dx\.)?doi\.org/)?)?10\.\d+(.\d+)*/.*)"
```
michamos: here you can take `idutils.doi_regexp`

inspirehep/modules/literaturesuggest/normalizers.py
```diff
@@ -77,6 +80,16 @@ def normalize_formdata(obj, formdata):
     return formdata
 
 
+def normalize_doi(obj, formdata):
+    try:
+        doi = formdata.get('doi')
+        formdata['doi'] = re.search("(10\.\d+(.\d+)*/.*)", doi.strip()).group(0)
```
michamos: again, you can use the one from `idutils`.

tests/integration/crossref/test_crossref_views.py
```diff
@@ -85,7 +85,7 @@ def test_crossref_search_handles_the_response_when_the_request_is_valid(log_in_a
 def test_crossref_search_handles_the_response_when_the_request_asks_for_a_non_existing_doi(log_in_as_scientist, app_client):
     with requests_mock.Mocker() as requests_mocker:
         requests_mocker.register_uri(
-            'GET', 'http://api.crossref.org/works/does-not-exist',
+            'GET', 'http://api.crossref.org/works/None',
```
david-caro: This suspiciously looks like it's stringifying a `None` python object, are you sure that's correct? (maybe it should not be doing the request at all if the object is `None`).

iulianav: In that case I need to remove this test?

david-caro: I'd say no, better avoid doing that request at all, like just removing the `requests_mocker` part, and making sure it does not do the http request (as doing a request  like that will surely not work, we skip the whole external request at all).

iulianav:
@david-caro check my last commit: is that what you meant?

tests/integration/crossref/test_crossref_views.py
```diff
@@ -83,20 +83,12 @@ def test_crossref_search_handles_the_response_when_the_request_is_valid(log_in_a
...
 
 def test_crossref_search_handles_the_response_when_the_request_asks_for_a_non_existing_doi(log_in_as_scientist, app_client):
-    with requests_mock.Mocker() as requests_mocker:
-        requests_mocker.register_uri(
-            'GET', 'http://api.crossref.org/works/does-not-exist',
-            status_code=404,
-            text=pkg_resources.resource_string(
-                __name__, os.path.join('fixtures', 'does-not-exist')),
-        )
+    response = app_client.get('/doi/search?doi=does-not-exist')
```
david-caro: Here to make sure it does not do a request to crossreff, you can do the `with requests_mocker.Mocker()` thingie but not register any uri, so if it tries it will fail.

PR #2839, opened 2017-10-05
records: BibTeX citation serializer rewritten

szymonlopaciuk:
## Description

Rewrite of the BibTeX citation serializer.

Uses pybtex to generate BibTeX citations and inspire-utils get_value
for json field value extraction. New integration and unit tests created,
in the integration tests pybtex records are compared based on content, not
output produced (as ordering is irrelevant).

Made a change in utils, by adding an optional argument to get_page_artid
to support a different pages separator than just a hyphen: bibtex requires --,
an en dash.

## Related Issue

N/A

## Motivation and Context

Previous serializers went out-of-date in regards to the schema.

## Checklist:


inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -0,0 +1,337 @@
...
+        be represented in a format like: 'Korea, Republic of'. Thus we reverse the order of segments around
+        the comma to make sound it natural: 'Republic of Korea'.
+
+        In particular, we cannot pick the first segment only due to ambiguities like
+        South/North Korea or East/West Congo
+    """
+    country_name_for_code = {country.alpha_2: ' '.join(reversed(country.name.split(', '))) \
+                             # Order below matters, 'last wins' in dictionary comprehension
+                             # (https://docs.python.org/3/reference/expressions.html#dictionary-displays)
+                             for country in list(pycountry.historic_countries) + list(pycountry.countries)}
```
michamos: maybe use [`babel`](http://babel.pocoo.org/en/latest/api/core.html#babel.core.Locale.territories) instead?

inspirehep/modules/hal/utils.py
```diff
@@ -456,7 +457,7 @@ def get_page_artid(record):
     elif 'page_start' in publication_info and 'page_end' in publication_info:
         page_start = publication_info['page_start']
         page_end = publication_info['page_end']
-        return '{}-{}'.format(page_start, page_end)
+        return '{}{}{}'.format(page_start, separator, page_end)
```
michamos: this is unsafe when separator is unicode.

inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -0,0 +1,337 @@
...
+            return decorated_function
+        return decorator
+    extractor.store = store
+    return extractor
+
+
+def bibtex_document_type(doc_type, obj):
+    """
+    Args:
+        doc_type (string): INSPIRE document type.
```
michamos: the parentheses should contain actual types, so `str`.

inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -0,0 +1,337 @@
...
+        date_choice = publication_year or thesis_date or imprint_date
+
+    if date_choice:
+        return PartialDate.parse(unicode(date_choice))
+
+
+# Functions below describe where the non-obvious data is located:
+#
+# Args:
+#   data (dict): data partially serialized by the means of defined marshmallow schema only
```
michamos: outdated comment about marshmallow?

inspirehep/modules/records/serializers/pybtex_serializer_base.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        bib_dict = {}
+        for record in record_list:
+            texkey, entries = self.create_bibliography_entry(record)
+            bib_dict[texkey] = entries
+
+        bib_data = BibliographyData(bib_dict)
+        return self.get_writer().to_string(bib_data)
+
+    def get_writer(self):
```
michamos: why are these getters needed?

tests/integration/test_bibtex_exporting.py
```diff
@@ -22,124 +22,161 @@
...
-}'''
-    result = Bibtex(thesis).format()
-
-    assert expected == result
-
-
-@pytest.mark.xfail
+    proceedings = get_es_record('lit', 701585)
+    expected = ("Alekhin:2005dx", Entry('proceedings', [
+        ('key', '701585'),
```
michamos: this `key` should not be produced, it's used for ranking by bibtex.

tests/integration/test_bibtex_exporting.py
```diff
@@ -22,124 +22,161 @@
...
+        ('address', u"Geneva"),
+        ('pages', u"pp.1--326"),
```
michamos: `pp.` is not needed.

tests/integration/test_bibtex_exporting.py
```diff
@@ -22,124 +22,161 @@
...
+
 def test_format_book(app):
-    book = get_db_record('lit', 736770)
+    book = get_es_record('lit', 736770)
+    expected = ("Fecko:2006zy", Entry('book', [
+        ('key', '736770'),
+        ('publisher', u"Cambridge University Press"),
+        ('title', u"Differential geometry and Lie groups for physicists"),
+        ('year', u"2011"),
+        ('isbn', u"9780521187961, 9780521845076, 9780511242960"),
```
michamos: you should try to normalize the isbn in order to put back the dashes (it might fail as some are invalid).

inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -0,0 +1,337 @@
...
+    thesis_date = get_value(data, 'thesis_info.date')
+    imprint_date = get_value(data, 'imprints[0].date')
+
+    if doc_type.endswith('thesis'):
+        date_choice = thesis_date or publication_year or imprint_date
+    else:
+        date_choice = publication_year or thesis_date or imprint_date
+
+    if date_choice:
+        return PartialDate.parse(unicode(date_choice))
```
michamos: you could use `loads` instead of `parse` here as you are reading from a record.

inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -0,0 +1,337 @@
...
+    return get_value(data, 'publication_info[0].journal_issue')
+
+
+@extractor('pages')
+def get_pages(data, doc_type):
+    return get_page_artid(data, '--')
+
+
+@extractor('primaryClass')
+def get_primary_class(data, doc_type):
```
michamos: I would only do this for new-style archive identifiers, as the old-style ones have the category in their id already.

inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+@extractor('archivePrefix')
+def get_arxiv_prefix(data, doc_type):
+    if get_eprint(data, doc_type):
+        return "arXiv"
+
+
+@extractor('SLACcitation')
+def get_slac_citation(data, doc_type):
```
michamos: I am not sure we still need the `SLACcitation` bibtex field. @inspirehep/inspire-content, what do you think? is it just some left-over from SPIRES times?

inspirehep/modules/records/serializers/schemas/base.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        Args:
+            json: literature record from API
+
+        Returns:
+            Pybtex Entity representing the record
+        """
+        json = replace_refs(json, 'es')
+        doc_type, fields = bibtex_type_and_fields(json)
+        texkey = unicode(get_value(json, 'texkeys[0]'))
```
michamos: you should use `six.string_type` instead of `unicode` for Python 3 compatibility if this is really needed, but usually `str` -> `unicode` is transparent (if only ASCII is used like here).

inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -0,0 +1,337 @@
...
+@extractor('school')
+def get_school(data, doc_type):
+    schools = [school['name'] for school in get_value(data, 'thesis_info.institutions')]
+    if schools:
+        return ', '.join(schools)
+
+
+@extractor('address')
+def get_address(data, doc_type):
+    pubinfo_city = get_value(data, 'publication_info[0].conference_record.address[0].cities[0]')
```
michamos: this info is not present in a Literature record, so you need to resolve some references for this to work. Are you doing that?

szymonlopaciuk: Yes, in [inspirehep/modules/records/serializers/schemas/base.py#L42](https://github.com/szymonlopaciuk/inspire-next/blob/bibtex_rewrite/inspirehep/modules/records/serializers/schemas/base.py#L42) I replace_refs, I understand that's the correct thing to do?

tests/integration/test_bibtex_exporting.py
szymonlopaciuk: 'pp.1' is the value of page_start in the publication_info of that record. Should I be stripping it manually then?

michamos: no, that's something we were doing on legacy, that I forgot about, and that I should discuss with @jacquerie and @jmartinm.

inspirehep/modules/records/serializers/fields_export.py
michamos: :+1: I overlooked it.

inspirehep/modules/hal/utils.py
```diff
@@ -207,8 +209,11 @@ def get_conference_record(record):
         972464
 
     """
-    return replace_refs(get_value(
-        record, 'publication_info.conference_record[0]', default=None), 'db')
+    replaced = replace_refs(get_value(record, 'publication_info.conference_record[0]'), 'db')
+    if replaced:
+        return replaced
+    else:
+        return default
```
jacquerie: Commits should be *atomic*; while implementing a new feature one shouldn't modify other parts of the code, even if they realized that there's a bug or a missing feature.

inspirehep/modules/hal/utils.py
```diff
@@ -244,7 +249,7 @@ def get_conference_title(record):
         'Workshop on Neutrino Physics'
 
     """
-    return get_value(record, 'titles.title[0]', default='')
+    return get_value(record, 'titles.title[0]', default=default)
```
jacquerie: The same applies to this change.

inspirehep/modules/hal/utils.py
```diff
@@ -456,7 +462,7 @@ def get_page_artid(record):
     elif 'page_start' in publication_info and 'page_end' in publication_info:
         page_start = publication_info['page_start']
         page_end = publication_info['page_end']
-        return '{}-{}'.format(page_start, page_end)
+        return text_type('{}{}{}').format(page_start, text_type(separator), page_end)
```
jacquerie: And this change too.

inspirehep/modules/records/serializers/bibtex_serializer.py
```diff
@@ -24,30 +24,21 @@
...
-        :param links: Dictionary of links to add to response.
-        """
-        records = []
-        for hit in search_result['hits']['hits']:
-            records.append(Bibtex(record=hit['_source']).format())
-
-        return "\n".join(records)
+    def create_bibliography(self, record_list):
+        bibtex_string = super(BIBTEXSerializer, self).create_bibliography(record_list)
+        # Pybtex escapes '%' in Bibtex, however we don't want that to happen in SLACcitations.
```
jacquerie: Do we still need this now that SLACcitation is gone?

inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -0,0 +1,318 @@
...
+
+"""Instructions on where to find relevant bits of data in API."""
+
+from __future__ import absolute_import, division, print_function
+from idutils import is_arxiv_post_2007
+from inspire_utils.record import get_value
+from inspire_utils.date import PartialDate
+from inspirehep.modules.hal.utils import get_page_artid, get_conference_record, get_conference_title
+from babel import Locale
+from isbn_hyphenate import hyphenate, IsbnError
```
jacquerie: Wow, somebody created an entire package just for this function?

Well, we already have an implementation in IDUtils of this: https://github.com/inveniosoftware/idutils/blob/f159a4105e3b26558c2e8fbae17818e3a1e2e81d/idutils/__init__.py#L447-L450, and we should probably get it from there. Or is it not considering some edge case?

inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -0,0 +1,318 @@
...
+
+def bibtex_document_type(doc_type, obj):
+    """
+    Args:
+        doc_type (text_type): INSPIRE document type.
+        obj (dict): literature record.
+
+    Returns:
+        text_type: bibtex document type for the given INSPIRE entry.
+    """
```
jacquerie: PEP257 talks about the conventions of the Python community with regards to docstrings. Here's how to rewrite this one so that it conforms to them (https://www.python.org/dev/peps/pep-0257/):
```python
"""Return the BibTeX entry type.

Maps the INSPIRE ``document_type`` to a BibTeX entry type. Also checks
``thesis_info.degree_type`` in case it's a thesis, as it stores the information
on which kind of thesis we're dealing with.
"""
```

setup.py
```diff
@@ -89,11 +89,13 @@
     'invenio-workflows-files~=0.0,>=0.0.6',
     'invenio-workflows-ui~=1.0,>=1.1.1',
     'invenio-workflows~=6.0,>=6.1.0',
+    'isbn_hyphenate>=1.0.4',
     'langdetect~=1.0,>=1.0.7',
     'librabbitmq~=1.0,>=1.6.1',
     'nameparser~=0.0,>=0.5.3',
     'orcid~=0.0,>=0.7.0',
     'plotextractor~=0.0,>=0.1.6',
+    'pybtex>=0.21',
```
jacquerie: We are using a specific notation to denote dependencies. This should be `pybtext~=0.0,>=0.21`.

tests/unit/records/test_records_serializers_fields_export.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def test_bibtex_document_type(test_record):
+    expected = "mastersthesis"
+    result = bibtex_document_type("thesis", test_record)
+    assert expected == result
+
+
+def test_bibtex_document_type_prefers_article():
+    input = {"thesis_info": {"degree_type": "master"}, "document_type": ["thesis", "article"]}
```
jacquerie: Some unit tests use local data, some others use a fixture. Let's be consistent! Having self-contained tests has the advantage of giving very clear failure messages, while having common fixtures avoids repetition. In this case we shouldn't have repetition anyway (all tests are testing different fields), so I'd lean towards local data everywhere.

inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -0,0 +1,318 @@
...
+    'misc': ['howpublished', 'reportNumber', 'author'],
+    'mastersthesis': ['reportNumber', 'school', 'address', 'type', 'author'],
+    'proceedings': ['publisher', 'series', 'number', 'volume', 'reportNumber', 'editor', 'address', 'organization',
+                    'pages'],
+    'book': ['publisher', 'isbn', 'author', 'series', 'number', 'volume', 'edition', 'editor', 'reportNumber',
+             'address'],
+    'inbook': ['chapter', 'publisher', 'author', 'series', 'number', 'volume', 'edition', 'editor', 'reportNumber',
+               'address', 'type', 'pages'],
+    'article': ['author', 'journal', 'collaboration', 'number', 'volume', 'reportNumber', 'pages'],
+}
```
jacquerie: Mh, these global variables make me think that we should structure a little better the `records` module, so that this configuration lives in a single place. See: https://github.com/inspirehep/inspire-next/blob/67ff20df8c6ed9abb6a7d59a6a7c536d2610a925/inspirehep/modules/hal/config.py

jacquerie: This lack of structure also led to defining `MAX_AUTHORS_BEFORE_ET_AL` in two places!

inspirehep/modules/records/serializers/fields_export.py
szymonlopaciuk: It's doing the opposite, the function you pointed to strips the formatting. Micha asked me to format the ISBNs into the hyphenated format, and it turned out that it is actually a complicated set of rules to do that, hence the whole package.

inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -0,0 +1,318 @@
...
+        text_type: name of a country, or a code if no country found
+    """
+    try:
+        return Locale('en').territories[code]
+    except KeyError:
+        return default
+
+
+def get_date(data, doc_type):
+    publication_year = get_value(data, 'publication_info[0].year')
```
jacquerie: As we discussed on the board, in general `foo[0].bar` should be `foo.bar[0]`.

szymonlopaciuk: I was wary of a situation like this happening:

```python
publication_info = [
    {'journal_title': 'A journal'},
    {'journal_title': 'Another journal', 'journal_issue': '12'},
    {'journal_title': 'Yet another one', 'year': 2016},
]
```

In which case the final record will have mismatched data. Granted, what I have now is a bad solution, I could try to pick the "fullest" publication info entry?

jacquerie: > I could try to pick the "fullest" publication info entry?

This sounds much better :+1:

michamos: you should also exclude `material` that is not `publication` or empty, as you don't want to take errata here. By the way, are you handling errata like legacy https://inspirehep.net/record/374836/export/hx ?

szymonlopaciuk: Working on it. Only errata should be handled like this, or should addenda too?

michamos: :+1: addenda too

StellaCh:
What is pending to merge this?

jacquerie:
Superseded by https://github.com/inspirehep/inspire-next/pull/3046.

Issue #2259, opened 2017-04-21
converter: kill module


jacquerie:
Similar to https://github.com/inspirehep/inspire-next/issues/2258, this module should become a util.

PR #2846, opened 2017-10-09
converter: merge module inside of workflows/utils

iulianav:
## Description
As all the remaining uses of this module are in the ``workflows``
module, we might as well make it part of its utils.

## Related Issue
Closes https://github.com/inspirehep/inspire-next/issues/2259.

## Checklist:


inspirehep/modules/authors/tasks.py
```diff
@@ -27,16 +27,14 @@
...
 from invenio_accounts.models import User
 from invenio_oauthclient.models import UserIdentity
+from sqlalchemy.orm.exc import NoResultFound
 
-from inspire_dojson.utils import strip_empty_values
-from inspire_schemas.api import validate
 from inspirehep.modules.forms.utils import filter_empty_elements
-from inspirehep.modules.workflows.utils import with_debug_logging
-
+from inspirehep.modules.workflows.utils.utils import with_debug_logging
```
jacquerie: To avoid this ugly `utils.utils`, just put the stuff that was in `utils.py` before in `utils/__init__.py`.

inspirehep/modules/authors/tasks.py
```diff
@@ -27,16 +27,14 @@
 import os
 
 from flask import current_app, url_for
-from sqlalchemy.orm.exc import NoResultFound
-
+from inspire_dojson.utils import strip_empty_values
+from inspire_schemas.api import validate
 from invenio_accounts.models import User
 from invenio_oauthclient.models import UserIdentity
+from sqlalchemy.orm.exc import NoResultFound
```
jacquerie: I don't like the import changes you made in all this PR : /

Did you use an automatic tool for them?

inspirehep/modules/authors/tasks.py
iulianav: Thx, this confused me more than you think!

inspirehep/modules/authors/tasks.py
iulianav: I don't know to be honest why this happened...hmmm ok let me check

iulianav: sorry about that

jacquerie:
You also added a `dbs/hepcrawl.db` file to the PR, which probably didn't intend to do. Please add a commit to add it to `.gitignore`, in a new section called "Build artifacts".

inspirehep/modules/authors/tasks.py
jacquerie: No need to be sorry about that, because this gives me a chance to explain something!

The reason you should resist doing these large scale refactorings while you are modifying a tiny bit of code is that they might generate a conflict in a lot of other PRs, and conflict resolution always require a little bit of care. So in exchange for a little improvement in the code base you generate potential errors for everyone else, which is not a good trade off.

jacquerie:
> You also added a dbs/hepcrawl.db file to the PR, which probably didn't intend to do. Please add a commit to add it to .gitignore, in a new section called "Build artifacts".

I did that for you!

Issue #2587, opened 2017-08-03
submission: refextract doesn't run on references in textbox


annetteholtkamp:
## Expected Behavior
If a user copies/pastes a reference list in the text box refextract should run on these references, and the extracted references should be exported to legacy

## Current Behavior
Currently references in the text box are ignored.

## Context
Currently I have to copy/paste the references again in bibedit on legacy - a waste of time

jacquerie:
@jmartinm made me realize that in order to do this we might need at least part of the merger, because we need to merge together the references extracted from the PDF and the ones parsed from the comments.

Or would it be okay to just override the PDF-extracted ones?

annetteholtkamp:
I would use the references in the textbox only if no pdf is available. So nothing smart needed at the moment.

- Annette

On 10 Aug 2017, at 02:02, Jacopo Notarstefano <notifications@github.com<mailto:notifications@github.com>> wrote:


@jmartinm<https://github.com/jmartinm> made me realize that in order to do this we might need at least part of the merger, because we need to merge together the references extracted from the PDF and the ones parsed from the comments.

Or would it be okay to just override the PDF-extracted ones?

jacquerie:
> I would use the references in the textbox only if no pdf is available. So nothing smart needed at the moment.

We can quickly implement this :+1:

jacquerie:
Closed in https://github.com/inspirehep/inspire-next/pull/2854.

PR #2854, opened 2017-10-11
workflows: also run refextract on the textbox refs

iulianav:
## Description
Runs ``refextract`` on both the PDF in the workflow and the references
provided by the submitter, if any, then chooses the one that generated
the most and attaches them to the workflow object.

## Related Issue
https://github.com/inspirehep/inspire-next/issues/2587

## Motivation and Context
Previously references in the textbox were ignored and the curator had to copy/paste the references again in bibedit on legacy which is not efficient in terms of time.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/refextract.py
+++ b/inspirehep/modules/workflows/tasks/refextract.py
@@ -26,7 +26,11 @@ from __future__ import absolute_import, division, print_function
 
 from timeout_decorator import timeout
 
-from refextract import extract_journal_reference, extract_references_from_file
+from refextract import (
+    extract_journal_reference,
+    extract_references_from_file,
+    extract_references_from_string,
+)
 
 from inspire_schemas.utils import split_page_artid
 from inspire_utils.helpers import maybe_int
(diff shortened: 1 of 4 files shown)


inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -87,12 +91,24 @@ def extract_journal_info(obj, eng):
...
         override_kbs_files=get_refextract_kbs_path(),
-        reference_format=u'{title},{volume},{page}'
+        reference_format=u'{title},{volume},{page}',
+    )
+
+    return map_refextract_to_schema(extracted_references, source=source)
+
+
+@timeout(5 * 60)
+def extract_references_from_form_field(text, source=None, custom_kbs_file=None):
```
jacquerie: This function name is too specific: you only care about this being text, not that it comes from a form field.

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -87,12 +91,24 @@ def extract_journal_info(obj, eng):
...
+    """Extract references from PDF and return in INSPIRE format."""
```
jacquerie: ...and this docstring is lying : )

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -87,12 +91,24 @@ def extract_journal_info(obj, eng):
 
 
 @timeout(5 * 60)
-def extract_references(filepath, source=None, custom_kbs_file=None):
+def extract_references_from_PDF(filepath, source=None, custom_kbs_file=None):
```
jacquerie: > Function names should be lowercase

(https://www.python.org/dev/peps/pep-0008/#function-names)

jacquerie:
Apart from cosmetic/style issues, I believe that this is implementing what @annetteholtkamp requested.

jacquerie:
In the end I implemented a better idea, suggested by @michamos: run ``refextract`` on both the PDF and the references provided by the submitter, and choose the one that generated the most. In the future we might want to prefer the ones that generated the most *linked* references, as they are probably of higher quality.

PR #2856, opened 2017-10-11
mappings: exclude references from _all in hep

iulianav:
## Description

I set 'include_in_all' for the 'references' field of Literature collection records to false. I concluded that this was the right thing to do as follows:
1. I ran the search and noticed the high number of results
2. I copied the query which was available in the console in DEBUG mode
3. I opened Sentry and accessed the _explain endpoint in ES with the query
4. I investigated and noticed that most hits/matches occured in the '_all' which also contains the values from the  'references' field, specifically from 'full_name' in our case
5. I concluded that this is not what we need to happen as these matches are rather spurious, thus I configurated ES so that it shall not add the 'references' to '_all' by setting the 'include_in_all' field to false (by default it is true) int the Literature records mapping (hep.json).

## Related Issue

https://github.com/inspirehep/inspire-next/issues/2857

## Motivation and Context

When searching for an author's name in the Literature collection, only records whose authors match should be returned and the referenced authors matches should be ignored.

## Checklist:

### DIFF
--- a/inspirehep/modules/records/mappings/records/hep.json
+++ b/inspirehep/modules/records/mappings/records/hep.json
@@ -728,6 +728,7 @@
                     "type": "boolean"
                 },
                 "references": {
+                    "include_in_all": false,
                     "properties": {
                         "curated_relation": {
                             "type": "boolean"


jacquerie:
> I set 'include_in_all' for the 'references' field of Literature collection records to false.

While the code change is correct, the most important thing I want to see here is the "why": how did you determine that this is the right thing to exclude from `_all` ?

iulianav:
@jacquerie how about this description?

PR #2859, opened 2017-10-12
orcid: INSPIRE record to ORCID converter

### LINKED ISSUE #2798: Restore ORCID implementation
304cfcaead150ca5ab5e470dda5e51c3449550b8 removed the then current implementation of ORCID support. The first step to integrate ORCID into INSPIRE should be to restore that work and build/improve on it.

szymonlopaciuk:
## Description
Works on top of two classes. The converter takes out fields
from INSPIRE record and passes them to a builder which
is responsible for building a correct ORCID record based
on the information given.

## Related Issue
#2798 

## Motivation and Context
Previous implementation was deleted due to being outdated in relation to schemas.

## Checklist:

### DIFF
--- a/inspirehep/modules/orcid/builder.py
+++ b/inspirehep/modules/orcid/builder.py
@@ -1,0 +1,312 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Builds an ORCID work record."""
+
+from __future__ import absolute_import, division, print_function
+
+from six import text_type
+
+from lxml import etree
+from lxml.builder import ElementMaker
+
+_NAMESPACES = {
+    'work': 'http://www.orcid.org/ns/work',
+    'common': 'http://www.orcid.org/ns/common'
+}
+
+_WORK = ElementMaker(namespace=_NAMESPACES['work'], nsmap=_NAMESPACES)
+_COMMON = ElementMaker(namespace=_NAMESPACES['common'], nsmap=_NAMESPACES)
+
+_ELEMENT_MAKERS = {
+    'work': _WORK,
+    'common': _COMMON
+}
+
+
+class OrcidBuilder(object):
+    """Class used to build ORCID-compatible work records in JSON."""
+
+    def __init__(self):
+        self.record = _WORK.work()
+
+    def get_xml(self):
+        """Get an XML record.
+
+        Returns:
+            lxml.etree._Element: ORCID work record compatible with API v2.0
+        """
+        return self.record
+
+    def __str__(self):
+        """Get a string-serialized XML respresentation of a record.
+
+        Returns:
+            string: ORCID work record as an XML string
+        """
+        return etree.tostring(self.get_xml())
+
+    def add_title(self, title, subtitle=None, translated_title=None):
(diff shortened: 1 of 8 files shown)


inspirehep/modules/orcid/converter.py
```diff
@@ -0,0 +1,224 @@
...
+
+class OrcidConverter(object):
+    """Coverter for the Orcid format."""
+
+    # Maps INSPIRE author roles to ORCID contributor roles
+    # for full list see: https://git.io/vdKXv#L235-L245
+    INSPIRE_TO_ORCID_ROLES_MAP = {
+        'author': 'author',
+        'editor': 'editor',
+        'supervisor': 'support-staff'  # TODO: Is this correct?
```
michamos: no it's not. Since they don't have a role for supervisor, we should strip them (i.e. the authors who are supervisors) in the conversion.

inspirehep/modules/orcid/converter.py
```diff
@@ -0,0 +1,224 @@
...
+
+    # Maps INSPIRE document type to ORCID work types
+    # for full list see: https://git.io/vdKXv#L118-L155
+    INSPIRE_DOCTYPE_TO_ORCID_TYPE = {
+        'article': 'journal-article',
+        'book': 'book',
+        'book chapter': 'book-chapter',
+        'conference paper': 'conference-paper',
+        'note': 'other',
+        'proceedings': 'conference-abstract',  # TODO: Is this correct?
```
michamos: `edited-book`

inspirehep/modules/orcid/builder.py
```diff
@@ -0,0 +1,277 @@
...
+# or submit itself to any jurisdiction.
+
+"""
+Builds an ORCID work record.
+"""
+
+from __future__ import absolute_import, division, print_function
+
+
+class OrcidBuilder(object):
```
michamos: if what ORCID wants is XML, maybe the builder should build an XML document instead? you could use `lxml.etree` or maybe the [E-factory](http://lxml.de/tutorial.html#the-e-factory), whatever you think is nicer.

szymonlopaciuk: They do support both, it's just they only publish schemas for XML. I will look into these, it would be more straightforward to validate XML.

inspirehep/modules/orcid/converter.py
```diff
@@ -0,0 +1,245 @@
...
+        # Add a title
+        if self.title:
+            builder.add_title(self.title, self.subtitle, self.title_translation)
+
+        # Add a journal title
+        containing_publication_title = self.journal_title or self.conference_title or self.book_series_title
+        if containing_publication_title:
+            builder.add_journal_title(containing_publication_title)
+
+        # TODO: Add a citation
```
jacquerie: What's this? Is there somewhere where can I read what was supposed to be here?

szymonlopaciuk: I wrote this before we had the BibTex sorted, so now you actually just pass a citation (i.e. BibTex) string to `builder.add_citation`.

jacquerie: :+1:

tests/unit/orcid/test_orcid_builder.py
```diff
@@ -0,0 +1,307 @@
...
+
+def xml_parse(xml_string):
+    """Parse an ``xml_string`` into XML."""
+    parser = etree.XMLParser(remove_blank_text=True)
+    return etree.fromstring(xml_string, parser)
+
+
+def xml_compare(result, expected):
+    """Assert two XML nodes equal."""
+    assert etree.tostring(result, pretty_print=True) == etree.tostring(expected, pretty_print=True)
```
jacquerie: Mh, comparing XMLs by comparing their string representations is trickier than it looks because of things than attribute ordering, entity dereferencing and namespace representation (more painful details: https://www.w3.org/2001/tag/doc/xmlChunkEquality.html). I'm not sure if the `pretty_print=True` argument takes care of all these...

inspirehep/modules/orcid/builder.py
```diff
@@ -0,0 +1,313 @@
...
+        """Set a title of the work, and optionaly a subtitle.
+
+        Args:
+            title (string): title of the work
+            subtitle (string): subtitle of the work
+            translated_title (Tuple[string, string]): tuple consiting of the translated title and its language code
+        """
+        title_field = _WORK.title(_COMMON.title(title))
+
+        if subtitle:
```
puntonim: Do we support the case when we want subtitle to be the empty string ''?
If that's the case this clause should be: `if subtitle is not None`

michamos: no, as you can see in the [JSON schema](https://github.com/inspirehep/inspire-schemas/blob/master/inspire_schemas/records/elements/title.yml), it has to have length at least 1, that is, be non-empty. This is a general feature: all strings and arrays in the records have to be non-empty. Absence of a value should be represented as absence of the corresponding key.

PR #2870, opened 2017-10-17
figures: Append the figure extension in figure `key`

### LINKED ISSUE #2869: Correctly expose file extensions for figures
#2572 introduced exposing files (in particular to legacy). However for figures, URLs exposed are lacking the corresponding file extension. E.g.:
```json
{"figures": [
{
"caption": "{\\it Left panels:} SED decomposition on a selection of ZFOURGE sources using \\texttt{CIGALE} \\citep{Burgarella:2005bd, Noll:2009ed}. Yellow circles are the observed points, yellow arrows are the upper limits, and the black lines are the best-fit total models. We also show the AGN component in solid red lines, stellar-heated dust component in solid green lines, the unattenuated stellar emission in dashed blue lines, and the attenuated stellar emission in solid orange lines. The corresponding redshift and fraction of AGN emission to the $L_{\\mathrm{IR}}$ for each source are also provided. {\\it Right panels:} a comparison of \\texttt{CIGALE}-derived SFRs to those from a FIR-derived conversion of the bolometric 8-1000$\\mu$m IR luminosity calculated from a luminosity-independent conversion \\citep{Wuyts:2008hi, Wuyts:2011da} using PACS 160$\\mu$m fluxes. The solid black line is the 1:1 relation, while the dashed red line is the best fit. We also show the linear Pearson correlation coefficient, R.",
"key": "figure2",
"label": "fig:fig2",
"material": "preprint",
"source": "arxiv",
"url": "/api/files/390d5244-b4b9-4b63-8651-830f44499f98/figure2"
}]}
```

zzacharo:
(Closes #2869)

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/arxiv.py
+++ b/inspirehep/modules/workflows/tasks/arxiv.py
@@ -24,6 +24,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+import os
 import re
 from functools import wraps
 
@@ -147,7 +148,7 @@ def arxiv_plot_extract(obj, eng):
 
             lb = LiteratureBuilder(source='arxiv', record=obj.data)
             for index, plot in enumerate(plots):
-                plot_name = plot.get('name')
+                plot_name = os.path.basename(plot.get('url'))
                 files_keys = obj.files.keys
                 key = plot_name
                 if plot_name in files_keys:
--- a/tests/unit/workflows/test_workflows_tasks_arxiv.py
+++ b/tests/unit/workflows/test_workflows_tasks_arxiv.py
@@ -316,10 +316,10 @@ def test_arxiv_plot_extract_populates_files_with_plots(mock_os):
         assert arxiv_plot_extract(obj, eng) is None
 
         expected = [{
-            'url': '/api/files/0b9dd5d1-feae-4ba5-809d-3a029b0bc110/figure1',
+            'url': '/api/files/0b9dd5d1-feae-4ba5-809d-3a029b0bc110/figure1.png',
             'source': 'arxiv',
             'material': 'preprint',
-            'key': 'figure1',
+            'key': 'figure1.png',
             'caption': 'Difference (in MeV) between the theoretical and experimental masses for the 2027 selected nuclei as a function of the mass number.'
         }]
         result = obj.data['figures']


zzacharo:
@kaplun check this PR for the issue for the files. Also @michamos can have a look that everything is ok.

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -146,8 +146,13 @@ def arxiv_plot_extract(obj, eng):
                 return
 
             lb = LiteratureBuilder(source='arxiv', record=obj.data)
+            file_extension_pattern = re.compile('.+(\.[a-zA-Z]+)$')
             for index, plot in enumerate(plots):
-                plot_name = plot.get('name')
+                file_extension = ''
+                match = file_extension_pattern.match(plot.get('url'))
```
kaplun: I don't understand: what puts an extension in the URL in the first place?

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -146,8 +146,13 @@ def arxiv_plot_extract(obj, eng):
                 return
 
             lb = LiteratureBuilder(source='arxiv', record=obj.data)
+            file_extension_pattern = re.compile('.+(\.[a-zA-Z]+)$')
```
kaplun: Note that there is an official Python way for doing that: `os.path.splitext()`

michamos: you don't need to reinvent the wheel, you can use [`os.path.splitext`](https://docs.python.org/2/library/os.path.html#os.path.splitext). In general, Python comes "batteries included", so you can assume the basic stuff is already out there. What you probably want to use though is [`os.path.basename`](https://docs.python.org/2/library/os.path.html#os.path.basename) instead of the `name` for the key.

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -146,8 +146,13 @@ def arxiv_plot_extract(obj, eng):
...
             for index, plot in enumerate(plots):
```
michamos: IIRC, you don't need the `enumerate` as the `index` is not used.

inspirehep/modules/workflows/tasks/arxiv.py
zzacharo: I was not appending the extension in the popoluated url field. But in the plot there is the url field that has the extension of the file. So I just extract it and append it.

inspirehep/modules/workflows/tasks/arxiv.py
zzacharo: I updated!

inspirehep/modules/workflows/tasks/arxiv.py
zzacharo: it is used in the number ;) Here https://github.com/inspirehep/inspire-next/pull/2870/files/92b21ff84972aa6f31a116c91d13263c2805f475#diff-d5b37b2f12fac54ce9b26fffdb95b1ceL154

zzacharo:
@michamos @kaplun I updated

kaplun:
Mmh... Github doesn't show your update...

PR #2874, opened 2017-10-18
holdingpen: do not index record_matches

kaplun:
Fixes: https://sentry.cern.ch/inspire-sentry/inspire-labs/group/931989/

* Disables indexing of record_matches, which is not needed and has
  changed type in a 6e8432a thus making impossible to index
  both entries before the commit and after it.

### DIFF
--- a/inspirehep/modules/workflows/mappings/holdingpen/hep.json
+++ b/inspirehep/modules/workflows/mappings/holdingpen/hep.json
@@ -23,6 +23,11 @@
                                 }
                             },
                             "type": "object"
+                        },
+                        "record_matches": {
+                            "include_in_all": false,
+                            "enabled": false,
+                            "type": "object"
                         }
                     },
                     "type": "object"


kaplun:
@jacquerie @ammirate @david-caro I guess this is enough to move on and be able to index entries in the holding pen.

inspirehep/modules/workflows/mappings/holdingpen/hep.json
```diff
@@ -23,6 +23,11 @@
                                 }
                             },
                             "type": "object"
+                        },
+                        "record_matches": {
+                            "include_in_all": false,
+                            "enabled": false,
+                            "type": "object"
```
kaplun: I had to specify `object` even if the new type is `integer` otherwise I wasn't able to use `enabled` property. Note that if I was specifying the `type` `integer` then I would have had to use `"index": "no"`... but... it would have still refused to index `objects`. Instead right now seems that ES is truly ignoring the type I am passing in.

kaplun:
@jacquerie I'd like to merge this, so that ES will ignore the fields regardless of its structure. This should allow me to then reindex without touching holdingpen entries.

PR #2880, opened 2017-10-20
theme: change the style of modals

Dinika:
## Checklist:

### DIFF
--- a/inspirehep/modules/theme/static/scss/feedback/modal.scss
+++ b/inspirehep/modules/theme/static/scss/feedback/modal.scss
@@ -18,6 +18,54 @@
  * 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
  */
 
+.modal-header {
+    background-color: $dark-blue;
+    color: white;
+
+    .close {
+        color: white;
+        opacity: 1;
+    }
+}
+
+.modal-dialog {
+    width: 60%;
+    margin: 0px;
+    position: fixed;
+    top: 30%;
+    left: 50%;
+}
+
+.modal.fade .modal-dialog {
+    -webkit-transform: translate(-50%, -100%);
+    -ms-transform: translate(-50%, -100%);
+    -o-transform: translate(-50%, -100%);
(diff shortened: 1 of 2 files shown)


Dinika:
feedback
This is what it'll look like

jacquerie:
This reminds me of https://github.com/inspirehep/inspire-next/pull/888, where I tried to make all modals look the same, and of https://github.com/inspirehep/inspire-next/pull/599, where some screenshots show how the modal used to look when I originally created it : )

In fact, I believe we have other modals that could benefit from similar changes.

Dinika:
Aah! Okay, I'll update this PR to make these changes to the e-mail request modal and others as well.

Dinika:
@jacquerie  I noticed you added the send paper plane icon in the button. I removed that in this PR. Do you suggest I bring it back?

jacquerie:
> Do you suggest I bring it back?

I trust your design sense a lot more than mine!

jmartinm:
@Dinika indeed here best thing to do would be to modify the style of all modals to be the same. So instead of adding the CSS rule for `#feedbackModalHeader`, apply it to the Bootstrap classes.

Once you modify the style, check for example, the modals in `/literature/new` (after importing) and the modal that appears when exporting in https://labs.inspirehep.net/search to make sure that they are all consistent.

Dinika:
@jmartinm Updated this PR as per the suggestions.

inspirehep/modules/theme/static/scss/feedback/modal.scss
```diff
@@ -18,6 +18,53 @@
...
+      color: white; 
+      opacity: 1;
+  }
+}
+
+.modal-dialog {
+  width: 60%;
+  margin: 0px;
+  position: fixed;
+    top: 30%;
```
jacquerie: Indentation is off here, but that's ok: the real issue to fix here is adding some CSS linting step to our build. Perhaps we can postpone this to the general rearchitecture of the UI code.

PR #2881, opened 2017-10-20
WIP: adding references

jmartinm:
## Checklist:


jmartinm:
Early stage prototype for the handling of references when a record gets added/updated.

Some things missing:

* Add real inspire-matcher configuration

* Use the strategies mentioned in http://docs.celeryproject.org/en/latest/userguide/tasks.html#automatic-retry-for-known-exceptions in order to retry running the backwards reference matching only when a record is in the database.

* Add exhaustive tests

jacquerie:
This went stale. It will be resurrected when it will be needed (after we look at the data and see what needs to happen for the citation counts).

PR #2883, opened 2017-10-20
global: define JSONSCHEMAS_* config options

chris-asl:
https://github.com/inveniosoftware/invenio-jsonschemas/pull/64 adds two
configuration flags for registering or not the blueprints to the API
and UI apps. These are required before the defaults are read from the
config file of the dependency.
This code should be updated/removed when upstream dependency
integrates a fix.


david-caro:
@kaplun this needs to get in before the next deployment to prod

kaplun:
Do I need to do something on puppet side?

david-caro:
Nop, no need

Issue #2643, opened 2017-08-17
holdingpen: enrichment: fully taken journals


fschwenn:
When journal articles end up in the holdingpen for selection - the holdingpen should know whether that journal is taken completely for INSPIRE and display it accordingly.

### knowing what to take completely
The information which journal is taken completely is stored in the JOURNALS database (583__a at the moment). The holdingpen should access these informations to know which journals are completely in INSPIRE.
There are also cases where certain parts of a journal are taken completely which is not yet stored in the JOURNALS database but only hard coded at DESY, e.g. for Phys.Rev.Lett.:
  re.search('^..(23|25|27|48|16|18|20)',articlid) are taken completely
The ideal case would be to store such cases also in the JOURNALS database but it is not clear to me, which form would be the most efficient.
### acting on it
We see basically two ways to proceed
- Harvest directly to INSPIRE, similar to arXiv completely harvested categories, e.g. gr-qc.
Additional CORE-tagging, assignment of field-codes, etc. can be done later, see #2657.
In this case unwanted stuff like editorials has to be deleted later.
- Indicate if a journal article belongs to such a journal. In principle it should not be possible to reject such an article, but since sometimes editorials and other rubbish is disguised as regular articles the possibility to reject such an article has to remain (button smaller or are-you-sure?-pop-up or ...).

## Current Behavior
None in the actual holdingpen. In the DESY-workflow, the decision to take an article is made by typing a letter. For fully taken journal the form is pre-filled.

## Context
It is essential that our users can rely on INSPIRE covering certain journals completely. Therefore we need to make sure that none of the articles of the core journals falls through the cracks

ammirate:
Fixed by #2886.

PR #2886, opened 2017-10-23
DESY Workflow: add journal_coverage to extra_data

iulianav:
## Description

The workflow now adds in `extra_data` the`journal_coverage` field of the journal record that an
article belongs to. This way we can see whether the journal is 'full' or 'partial' covered, as these
are the possible values of the field. 

## Related Issue

https://github.com/inspirehep/inspire-next/issues/2643

## Motivation and Context

This information was requested in order to help with identifying core records and will impact the way the front-end of holdingpen displays the Reject button. One of the possibilities is that the Reject button will not be displayed for records that belong to fully covered journals. This will probably require modifications to inspire-search-js.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -249,7 +249,11 @@ def get_journal_coverage(obj, eng):
         cast(RecordMetadata.json['short_title'], String) == type_coerce(a_journal_title, JSON))
 
     result = db.session.execute(query).fetchone()
-    journal_coverage = result.records_metadata_json['_harvesting_info']['coverage']
+
+    if result:
+        journal_coverage = result.records_metadata_json['_harvesting_info']['coverage']
+    else:
+        journal_coverage = None
 
     if journal_coverage:
         obj.extra_data['journal_coverage'] = journal_coverage
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -26,12 +26,11 @@ from __future__ import absolute_import, division, print_function
 
 import os
 from sqlalchemy import (
-     JSON,
-     String,
-     cast,
-     type_coerce,
- )
-from sqlalchemy import text
+    JSON,
+    String,
+    cast,
+    type_coerce,
+)
 
 from functools import wraps
 import re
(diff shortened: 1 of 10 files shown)


iulianav:
@StellaCh what should happen on the front-end side? The issue is not explicit enough.

StellaCh:
Based on the requirements, the Reject button should stay:
>In principle it should not be possible to reject such an article, but since sometimes editorials and other rubbish is disguised as regular articles the possibility to reject such an article has to remain

But you can display this information as a tag in the button area. Let's discuss this in person!

StellaCh:
Following a discussion with @ksachs : 
For fully taken journals:
reject

- Reject button: make opacity: 0.3 with no border. The button should still be clickable
- When you hover over the button, add the tooltip "The article belongs to a fully taken journal"

For partially taken journals:
- No UI change

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -200,6 +210,33 @@ def is_submission(obj, eng):
     return False
 
 
+@with_debug_logging
+def _get_journal_title(obj, eng):
```
ammirate: I would move this function in `get_journal_coverage`, or remove it completely, since it's not used anywhere else ;)

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -200,6 +210,33 @@ def is_submission(obj, eng):
...
+    if publication:
+        return publication[0].get('journal_title')
+
+
+@with_debug_logging
+def get_journal_coverage(obj, eng):
+    """Return the journal coverage that this article belongs to."""
+    a_journal_title = _get_journal_title(obj, eng)
+
+    query = RecordMetadata.query.filter(
```
ammirate: if there's no journal title I'd avoid query the DB :P

inspirehep/modules/workflows/tasks/actions.py
michamos: I wouldn't. These nested functions are not really Pythonic (`Flat is better than nested.`) and are wasteful, as they are redefined on every function invocation.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -200,6 +210,33 @@ def is_submission(obj, eng):
...
+        cast(RecordMetadata.json['short_title'], String) == type_coerce(a_journal_title, JSON))
+
+    result = db.session.execute(query).fetchone()
+
+    if result:
+        journal_coverage = result.records_metadata_json['_harvesting_info']['coverage']
+    else:
+        journal_coverage = None
+
+    obj.extra_data['journal_coverage'] = journal_coverage
```
ammirate: A small refactor:
```python
if result:
    obj.extra_data['journal_coverage'] = get_value(result.records_metadata_json, "_harvesting_info.coverage")
```

inspirehep/modules/workflows/tasks/actions.py
iulianav: So which approach should I choose in the end?

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -200,6 +210,33 @@ def is_submission(obj, eng):
     return False
 
 
+@with_debug_logging
+def _get_journal_title(obj, eng):
+    """Return the journal title of the publication."""
+    publication = obj.data.get('publication_info')
+    if publication:
+        return publication[0].get('journal_title')
```
iulianav: @michamos this is wrong, right? As there can be more journals? I think I made this mistake in the past too. Also, if there are more journals in `publication_info`, then which one do I consider for the coverage?

michamos: I don't think you should do a query in `get_journal_coverage`. If this runs after `normalize_journal_title` from #2938, then there is already the `$ref` to the journal and you can use a strategy similar to #2951.
If any journal has full coverage, then the `reject` button should be greyed out.

ammirate:
Closes #2643.

PR #2887, opened 2017-10-23
mappings: rename facet_inspire_subjects

chris-asl:
Rename facet_inspire_subjects so it follows the field it takes its
value from, i.e. facet_inspire_categories.

## Motivation and Context
The goal is to ensure uniformity in naming several mappings fields.
As we discussed with @jacquerie IRL, mappings naming can be improved (e.g. has inconsistencies). 
The goal is to gather (in time) some naming schemes/guidelines for anyone editing the mappings to follow.

For example, this PR is fixing the inconsistency of `facet_inspire_subjects` which got its values from `inspire_categories`.  

## Checklist:

### DIFF
--- a/inspirehep/modules/authors/rest/stats.py
+++ b/inspirehep/modules/authors/rest/stats.py
@@ -68,7 +68,7 @@ class AuthorAPIStats(object):
                 "citation_count",
                 "control_number",
                 "facet_inspire_doc_type",
-                "facet_inspire_subjects",
+                "facet_inspire_categories",
                 "keywords",
             ]
         )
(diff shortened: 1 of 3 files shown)


inspirehep/modules/authors/rest/stats.py
```diff
@@ -68,7 +68,7 @@ def serialize(self, pid, record, links_factory=None):
                 "citation_count",
                 "control_number",
                 "facet_inspire_doc_type",
-                "facet_inspire_subjects",
+                "facet_inspire_categories",
```
jacquerie: Note how this code didn't work at all, as `facet_inspire_categories` is not part of the `_source` we are originally sending to ES, but is the result of a `copy_to`. But let's keep it, since the Author API will see some love in the future.

chris-asl: Will create an issue so that we have it somewhere concretely. 

Also, if this change is ok, then we can merge it?

PR #2898, opened 2017-10-24
alembic: Create inspire branch for alembic

zzacharo:
## Checklist:


setup.py
```diff
@@ -225,6 +225,9 @@
         'invenio_db.models': [
             'inspire_workflows_audit = inspirehep.modules.workflows.models',
         ],
+        'invenio_db.alembic': [
+            'inspirehep = inspirehep:alembic',
+        ],
```
jacquerie: Let's keep our entry points alphabetically sorted! This should go before `invenio_db.models`.

inspirehep/alembic/0da6eba8fbee_create_inspirehep_branch.py
```diff
@@ -0,0 +1,56 @@
...
+depends_on = 'dbdbc1b19cf2'
+
+
+def upgrade():
+    """Upgrade database."""
+    op.create_table(
+        'workflows_record_sources',
+        sa.PrimaryKeyConstraint('record_id', 'source'),
+        sa.Column('source', sa.Text, default="", nullable=False),
+        sa.Column('record_id', UUIDType, sa.ForeignKey("records_metadata.id", ondelete='CASCADE'), nullable=False),
```
jacquerie: Nit: let's use either single quotes or double quotes consistently.

jmartinm:
We could do `Invenio` style where there is a first alembic recipe creating the branch (https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/alembic/1095cdf9f350_create_records_branch.py), then a recipe creating the initial tables

jmartinm:
I think it's better if we create a first PR with the items mentioned above and merge it.

Then the merger PR if it has a new table it can include the recipe to create it.

david-caro:
@jmartinm yep, we were actually kind of lost here without you :)

ammirate:
@zzacharo so, can we push this on top of #2904 ?

zzacharo:
@ammirate the previous recipe that I created for the merger should be moved to the merger branch or manual merger based on which branch needs the table it was creating. So, the concept is every modification will be a new revision under this one.

zzacharo:
@jmartinm have you checked that in order to merge it?

zzacharo:
depends on https://github.com/inveniosoftware-contrib/invenio-workflows-ui/pull/30 and https://github.com/inveniosoftware-contrib/invenio-workflows-files/pull/8 and https://github.com/inspirehep/inspire-crawler/pull/23 . For the last one should first integrate alembic.

tests/alembic/conftest.py
```diff
@@ -0,0 +1,62 @@
...
+from sqlalchemy_utils.functions import create_database, database_exists
+
+
+@pytest.fixture(scope='module')
+def app():
+    """Flask application.
+
+    Creates a Flask application with a simple testing configuration,
+    then creates an application context and inside of it recreates
+    all databases and indices from the fixtures. Finally it yields,
```
jmartinm: This `app` does not seem to recreate databases and indices?

zzacharo: yep leftover. I am changing it.

zzacharo:
@david-caro you can check it too and merge it if is ok.

jacquerie:
As I wrote on Gitter:
> creating a new test suite for alembic is not a good idea
> because Travis limits us to 5 concurrent jobs
> so if you add another test suite for it you add >5 minutes to the build time
> besides, what's the need? it probably fits in integration

zzacharo:
@jacquerie in my tests for alembic I destroy the database to check if migrations are applied as expected. Inside integration suite they all share the same database and things are getting messy because the app fixture is persistent across the test session. Any thoughts?

jmartinm:
👍 I would approve it again. Nice that you could make it work inside the integration tests.

tests/integration/conftest.py
```diff
@@ -121,6 +121,22 @@ def small_app():
         yield app
 
 
+@pytest.fixture(scope='module')
+def alembic_app():
+    """Flask application with no records and module scope."""
+    app = create_app(
+        SQLALCHEMY_DATABASE_URI='postgresql+psycopg2://inspirehep:dbpass123@localhost:5432/inspirehep_alembic'
```
david-caro: 👍 awesome xd

PR #2899, opened 2017-10-24
record: add documents/figures handling methods

david-caro:
~This PR is just a POC about something we discussed in #2623, it
started as quick patch but looks complete :/, will push to that one
if we agree to use this in it's tead.~

~Opened a PR just to see if it passes the travis tests.~

This has become the main PR now.

### DIFF
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -55,9 +55,16 @@ class InspireRecord(Record):
 
     @classmethod
     def create(cls, *args, **kwargs):
-        """
+        """Override the default ``create``.
 
-        Args:
+        To handle also the docmuments and figures retrieval.
+
+        Note:
+
+            Might create an extra revision in the record if it had to download
+            any documents or figures.
+
+        Keyword Args:
 
             files_src_record(InspireRecord): if passed, it will try to get the
                 files for the documents and figures from this record's files
@@ -71,8 +78,11 @@ class InspireRecord(Record):
         return new_record
 
     def update(self, *args, **kwargs):
-        """
-        Args:
+        """Override the default ``update``.
+
+        To handle also the docmuments and figures retrieval.
+
+        Keyword Args:
 
             files_src_record(InspireRecord): if passed, it will try to get the
                 files for the documents and figures from this record's files
@@ -150,6 +160,32 @@ class InspireRecord(Record):
         file_name=None,
         key=None,
     ):
+        """Add a document or figure to the record.
+
+        Args:
+
+            metadata(dict): metadata of the document or figure, see the schemas
+                for more details, will be validated.
+            stream(file like object): if passed, will extract the file contents
+                from it.
+            is_document(bool): if the given information is for a document,
+                set to ```False``` for a figure.
+            file_name(str): Name of the file, used as a basis of the key for
+                the files store.
+            key(str): if passed, will use this as the key for the files store
+                and ignore ``file_name``, use it to overwrite existing keys.
+
+
+        Returns:
+
+            dict: metadata of the added document or figure.
+
+
+        Raises:
+
+            TypeError: if not ``file_name`` nor ``key`` are passed (one of
+                them is required).
+        """
         if not key and not file_name:
             raise TypeError(
                 'No file_name and no key passed, at least one of them is '
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -24,6 +24,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+import copy
 from datetime import datetime
 
 import arrow
@@ -31,30 +32,74 @@ from elasticsearch.exceptions import NotFoundError
 from flask import current_app
 
 from inspire_schemas.api import validate
+from inspire_schemas.builders import LiteratureBuilder
 from invenio_files_rest.models import Bucket
 from invenio_pidstore.errors import PIDDoesNotExistError
 from invenio_pidstore.models import PersistentIdentifier
 from invenio_records_files.api import Record
 from invenio_db import db
 
+from inspirehep.modules.records.utils import open_url_or_path
 from inspirehep.utils.record_getter import (
     RecordGetterError,
     get_es_record_by_uuid
 )
 
 
+MAX_UNIQUE_KEY_COUNT = 50000
+
+
 class InspireRecord(Record):
 
     """Record class that fetches records from DataBase."""
 
+    @classmethod
+    def create(cls, *args, **kwargs):
+        """
+
+        Args:
+
+            files_src_record(InspireRecord): if passed, it will try to get the
+                files for the documents and figures from this record's files
+                iterator before downloading them, for example to merge existing
+                records.
+        """
+        files_src_record = kwargs.pop('files_src_record', None)
+        new_record = super(InspireRecord, cls).create(*args, **kwargs)
+        new_record.download_documents_and_figures(src_record=files_src_record)
+        new_record.commit()
+        return new_record
+
+    def update(self, *args, **kwargs):
+        """
+        Args:
+
+            files_src_record(InspireRecord): if passed, it will try to get the
+                files for the documents and figures from this record's files
+                iterator before downloading them, for example to merge existing
+                records.
+        """
+        files_src_record = kwargs.pop('files_src_record', None)
+        super(InspireRecord, self).update(*args, **kwargs)
+        self.download_documents_and_figures(
+            only_new=True,
+            src_record=files_src_record,
+        )
+
     def merge(self, other):
         """Redirect pidstore of current record to the other InspireRecord.
 
-        :param other: The record that self(record) is going to be redirected.
-        :type other: InspireRecord
+        Args:
+
+            other(InspireRecord): The record that self(record) is going to be
+                redirected.
         """
-        pids_deleted = PersistentIdentifier.query.filter(PersistentIdentifier.object_uuid == self.id).all()
-        pid_merged = PersistentIdentifier.query.filter(PersistentIdentifier.object_uuid == other.id).one()
+        pids_deleted = PersistentIdentifier.query.filter(
+            PersistentIdentifier.object_uuid == self.id
+        ).all()
+        pid_merged = PersistentIdentifier.query.filter(
+            PersistentIdentifier.object_uuid == other.id
+        ).one()
         with db.session.begin_nested():
             for pid in pids_deleted:
                 pid.redirect(pid_merged)
@@ -62,7 +107,9 @@ class InspireRecord(Record):
 
     def delete(self):
         """Mark as deleted all pidstores for a specific record."""
-        pids = PersistentIdentifier.query.filter(PersistentIdentifier.object_uuid == self.id).all()
+        pids = PersistentIdentifier.query.filter(
+            PersistentIdentifier.object_uuid == self.id
+        ).all()
         with db.session.begin_nested():
             for pid in pids:
                 pid.delete()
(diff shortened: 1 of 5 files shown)


david-caro:
@jacquerie @ammirate this one once passing the tests (locally it does) is ready to be merged, chips out a couple of commits from the merger branch.

inspirehep/modules/records/api.py
```diff
@@ -24,45 +24,92 @@
...
+MAX_UNIQUE_KEY_COUNT = 50000
+
+
 class InspireRecord(Record):
 
     """Record class that fetches records from DataBase."""
 
+    @classmethod
+    def create(cls, *args, **kwargs):
+        """
```
jacquerie: Something's missing here!

inspirehep/modules/records/api.py
```diff
@@ -24,45 +24,92 @@
...
+                records.
+        """
+        files_src_record = kwargs.pop('files_src_record', None)
+        new_record = super(InspireRecord, cls).create(*args, **kwargs)
+        new_record.download_documents_and_figures(src_record=files_src_record)
+        new_record.commit()
+        return new_record
+
+    def update(self, *args, **kwargs):
+        """
```
jacquerie: Here too!

inspirehep/modules/records/api.py
david-caro: what do you mean?

tests/integration/editor/test_editor_views.py
```diff
@@ -125,8 +125,8 @@ def test_get_revisions(log_in_as_cataloger, record_with_two_revisions, api_clien
 
     result = json.loads(response.data)
 
-    assert result[0]['revision_id'] == 1
-    assert result[1]['revision_id'] == 0
+    assert result[0]['revision_id'] == 2
+    assert result[1]['revision_id'] == 1
```
jacquerie: Who's creating the extra revision?

inspirehep/modules/records/api.py
```diff
@@ -24,45 +24,92 @@
...
+                redirected.
         """
-        pids_deleted = PersistentIdentifier.query.filter(PersistentIdentifier.object_uuid == self.id).all()
-        pid_merged = PersistentIdentifier.query.filter(PersistentIdentifier.object_uuid == other.id).one()
+        pids_deleted = PersistentIdentifier.query.filter(
+            PersistentIdentifier.object_uuid == self.id
+        ).all()
+        pid_merged = PersistentIdentifier.query.filter(
+            PersistentIdentifier.object_uuid == other.id
+        ).one()
```
jacquerie: Nit: why these changes?

We don't enforce line length limits in INSPIRE (`E501` is ignored in https://github.com/inspirehep/inspire-next/blob/5f1a16289fde4753099ebd2febbddac5e9038c4a/pytest.ini#L25).

david-caro: But my editor does, and complains, and pollutes the errors list. This way it's cleaner and I don't mistakenly ignore any errors at editing time.

tests/integration/editor/test_editor_views.py
david-caro: The download_documents, when there are documents, as I don't want to reimplement the invenio record `create` but just use it, if I do any changes after the `super` runs, I have to create a new revision, any ideas there are welcome.

inspirehep/modules/records/api.py
jacquerie: Then your editor is misconfigured...

inspirehep/modules/records/api.py
jacquerie: From https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings:
> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description.

The summary line is missing.

inspirehep/modules/records/api.py
david-caro: to some extent yes, but it seems to be more work to properly make it fetch the config from the current project than doing the small innocuous changes you see above.

inspirehep/modules/records/api.py
david-caro: I will not add a spurious one line comment there because some 'directive' says it should. Imo we are better without comments than with "Creates a record." comment. That's why we are not using the docstring checker invenio uses.

david-caro: From that same page:
```
The aim of this PEP is to standardize the high-level structure of docstrings: what they should contain, and how to say it (without touching on any markup syntax within docstrings). The PEP contains conventions, not laws or syntax.
```

jacquerie: > Imo we are better without comments than with "Creates a record." comment.

And I agree with that, because this is not what it should say! Wouldn't it be better to have
```
"""Override the ``create`` method.

[explanation of what is wrong with the overridden method]

Args:
    [...]
"""
```
?

inspirehep/modules/records/api.py
```diff
@@ -24,45 +24,92 @@
...
+
 class InspireRecord(Record):
 
     """Record class that fetches records from DataBase."""
 
+    @classmethod
+    def create(cls, *args, **kwargs):
+        """
+
+        Args:
```
jacquerie: Napoleon includes a specific section for kwargs, named not surprisingly `Keyword Args`: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/index.html#sections.

inspirehep/modules/records/api.py
david-caro: The first phrase there still does not say anything...

inspirehep/modules/records/api.py
david-caro: Good to know 👍

tests/integration/editor/test_editor_views.py
david-caro: Added a note in the create record method.

inspirehep/modules/records/api.py
```diff
@@ -24,45 +24,100 @@
...
+
+
 class InspireRecord(Record):
 
     """Record class that fetches records from DataBase."""
 
+    @classmethod
+    def create(cls, *args, **kwargs):
+        """Override the default ``update`` to handle also the docmument and
+        figures retrieval.
```
david-caro: This is way more useful, but it takes more than one line, is that ok?

inspirehep/modules/records/api.py
```diff
@@ -24,45 +24,102 @@
...
+MAX_UNIQUE_KEY_COUNT = 50000
+
+
 class InspireRecord(Record):
 
     """Record class that fetches records from DataBase."""
 
+    @classmethod
+    def create(cls, *args, **kwargs):
+        """Override the default ``update``.
```
david-caro: this should be 'create'

inspirehep/modules/records/api.py
```diff
@@ -24,45 +24,102 @@
...
+
 class InspireRecord(Record):
 
     """Record class that fetches records from DataBase."""
 
+    @classmethod
+    def create(cls, *args, **kwargs):
+        """Override the default ``create``.
+
+        To handle also the docmument and figures retrieval.
```
jmartinm: typo ;)

inspirehep/modules/records/api.py
```diff
@@ -95,6 +152,243 @@ def validate(self):
...
         validate(self)
 
+    def add_document_or_figure(
+        self,
+        metadata,
+        stream=None,
+        is_document=True,
+        file_name=None,
+        key=None,
+    ):
```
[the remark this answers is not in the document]
jmartinm: And is this a public method? Seems to only be called from `_download_to_docs_or_figs`

inspirehep/modules/records/api.py
```diff
@@ -95,6 +152,243 @@ def validate(self):
...
+
+        It will fill it up with the local url in case it's found in the given
+        src_record.
+        It will leave any document that does not need to be downloaded with the
+        url intact, pointing to the local files api endpoint.
+        """
+        doc_or_fig_obj = copy.deepcopy(doc_or_fig_obj)
+        key = doc_or_fig_obj['key']
+
+        if not doc_or_fig_obj['url'].startswith('/api/files/'):
```
jmartinm: So this method only does something for `invenio-files` URLs. What is the return value and why is so much code needed when the only line that seems to 'do something' is [this one](https://github.com/inspirehep/inspire-next/pull/2899/files#diff-3cf20a9f11cd40bad5e31856ef439299R239)?

inspirehep/modules/records/api.py
```diff
@@ -95,6 +152,243 @@ def validate(self):
...
+        figure=None,
+        src_record=None,
+        only_new=False,
+    ):
+        if not document and not figure:
+            raise TypeError(
+                'No document nor figure passed, at least one is needed.'
+            )
+
+        is_document = bool(document)
```
jmartinm: Just wondering if this is enough? What does a document look like?

inspirehep/modules/records/api.py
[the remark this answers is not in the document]
david-caro: yep, though it's the entry point if anybody wants to individually add a document or figure, adding docs.

inspirehep/modules/records/api.py
david-caro: It tries to resolve the `/api/files` url's, and raises the correct exceptions if unable, properly hinting on what is wrong with the data passed, that's why there are so many ifs, otherwise it will raise only one generic exception with no hint.
Taking into account too if this is just an update or not (`update_only` param).

Then, in the case that you passed a `src_record`, and the `key` is not in `self.files` but it's on `src_record`, it uses the local path to that file in the returned document metadata, so later, we can download it safely.

The return object is a dict with:
* In the case it has to be downloaded: the metadata of the document that we can actually download (the `url` is set to something that you can use `open_url_or_path` on). 
* In the case it does not have to be downloaded: the metadat of the document with a local api `url` (/api/files/...).

I'll clarify the docs.

The alternative code with no hints would be:
```python
if only_new and src_record and key in src_record and key not in self.files:
    doc_or_fig_obj['url'] = src_record.files[key].file.uri
    return doc_or_fig_obj

if not src_record and only_new and key in self.files:
    return doc_or_fig_obj

raise Exception("There's something wrong with the files.")
```
but the conditions being that 'complex' I preferred raising a sensible exception.

Hmm... I can split the two things, something like:
```python
if only_new and src_record and key in src_record and key not in self.files:
    doc_or_fig_obj['url'] = src_record.files[key].file.uri
    return doc_or_fig_obj

if not src_record and only_new and key in self.files:
    return doc_or_fig_obj

### then after this, try to give a nice exception
if not src_record and not only_new:
    raise Exception('you did not pass a record to get files from...')

if not src_record  and only_update and key not in self.files:
    raise Exception('Bad document ...')    
```

Hmm... refactoring!

inspirehep/modules/records/api.py
david-caro: It's a dictionary with some keys, it's defined in the schemas, I just rely in the fact that there's at least on key required. Should be enough.

inspirehep/modules/records/api.py
```diff
@@ -95,6 +152,310 @@ def validate(self):
...
+        if _already_there(self, src_record, only_new):
+            return doc_or_fig_obj
+
+        raise _get_meaningful_exception(
+            self,
+            key,
+            src_record,
+            only_new,
+            doc_or_fig_obj,
+        )
```
david-caro: @jmartinm looks a lot more readable :)

inspirehep/modules/records/api.py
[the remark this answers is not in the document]
david-caro: done

inspirehep/modules/records/api.py
jmartinm: 👍

inspirehep/modules/records/api.py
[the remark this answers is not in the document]
jmartinm: 👍

PR #2906, opened 2017-10-26
workflows: decrease log level for invalid tarball

jacquerie:
## Description:
Some arXiv records just don't have a tarball, as they are allowed to
have anything there, so we shouldn't log an error whenever we don't
see one.

Note also that logging an error eats rapidly into our ``sentry.io``
quota, as we see roughly a thousand such events per day per instance.

## Related Issues:
https://sentry.io/inspirehep/inspire-prod/issues/388912388/
https://sentry.io/inspirehep/inspire-qa/issues/388908905/

## Checklist:


kaplun:
https://sentry.io/inspirehep/inspire-prod/issues/388912388/

Issue #2785, opened 2017-09-24
global: rename all CFG configuration variables


jacquerie:
Configuration variables starting with `CFG` are a relic from the past, and don't satisfy the Flask convention for configuration variable names, leading to potential clashes and obscure names (next year who is going to know what `CFG_BIBCATALOG` is?)

In fact, configuration variables relevant only inside module `foo` should start with `FOO_` ([example](https://github.com/inspirehep/inspire-next/blob/eb08c999b2190a2c64f392623df63deaee97068c/inspirehep/modules/hal/config.py)), while configuration variables used all across INSPIRE should start with `INSPIRE_`.

PR #2917, opened 2017-10-28
utils.InspireRT: allow set the ssl verification in config

david-caro:
Added new CFG_BIBCATALOG_SYSTEM_RT_VERIFY_SSL config option to allow
disabling ssl verification for the rt server (useful for tests and
qa).


jacquerie:
One day (soon) we should improve the names of these configuration variable. See: https://github.com/inspirehep/inspire-next/issues/2785.

david-caro:
This is another extra commit from the json-merger branch that can be merged separately.

PR #2920, opened 2017-10-30
records: use validate from inspire_schemas

michamos:
## Description:
This ensures that we have a single implementation of schema
validation (for constraints that do not require a DB) and
that we validate ``format``s.

## Checklist:

### DIFF
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -30,6 +30,7 @@ import arrow
 from elasticsearch.exceptions import NotFoundError
 from flask import current_app
 
+from inspire_schemas.api import validate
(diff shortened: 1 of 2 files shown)


inspirehep/modules/records/api.py
```diff
@@ -30,6 +30,7 @@
 from elasticsearch.exceptions import NotFoundError
 from flask import current_app
 
+from inspire_schemas.api import validate as inspire_schemas_validate
```
jacquerie: I don't think we need the `[...] as inspire_schemas_validate`, as the `validate` method inside `InspireRecord` does not shadow the outer one.

Issue #2894, opened 2017-10-24
Synchronizing ORCID Tokens from legacy


kaplun:
Currently users are authenticating and authorizing INSPIRE to perform ORCID operation on Legacy. Soon though we will need to stop pushing records from legacy because ORCID will switch off the 1.2 API used by legacy. Therefore there will be an interim period where user are still granting rights in legacy but publications will be pushed from Labs. For this we need to sync between legacy and labs the OAUTH access tokens.

These have just been set up on legacy side. Labs can access ORCID tokens by reading the `legacy_orcid_tokens` key from Redis. This key is structured as a list of tuples of the form:
`(orcid, token, email, name)` all objects being unicode objects.

A celery task should be created that:
* pops all rows from this key
* for each row: 
  * check if the corresponding user does not yet exist (using the `orcid` as identifier)
  * populates corresponding tables in order to simulate as if the user actually authenticated and granted Labs to push things to ORCID.

PR #2922, opened 2017-10-30
orcid: sync OAUTH tokens from legacy

szymonlopaciuk:
Celery task pulls ORCID token data from legacy through redis, as described in issue #2894, and adds the data to inspire-next.

## Description

Will try to match an existing user to add the token to, upon failing will silently create a new user for that token. If the user already has a token assigned, this will leave it untouched.

## Related Issue

implements #2894

## Motivation and Context

ORCID API ver. 1.2 used in legacy will be switched off. In the meantime, until inspire-next is fully functional we could use inspire-next to push records from legacy using API ver. 2.0. To do this we need to copy over users' OAUTH tokens. See: #2894.

## Checklist:

### DIFF
--- a/inspirehep/modules/orcid/__init__.py
+++ b/inspirehep/modules/orcid/__init__.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""ORCID integration module."""
+
+from __future__ import absolute_import, division, print_function
+
+from .ext import InspireOrcid  # noqa: F401
--- a/inspirehep/modules/orcid/ext.py
+++ b/inspirehep/modules/orcid/ext.py
@@ -1,0 +1,34 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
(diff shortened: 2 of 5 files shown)


tests/integration/test_orcid_import_legacy_tokens.py
```diff
@@ -0,0 +1,203 @@
...
+
+def test_linked_user_with_token_exists(app):
+    """Ignore token, if already has one."""
+    assert_db_has_n_legacy_records(0, SAMPLE_USER)
+
+    # Register sample user
+    _register_user(orcid=SAMPLE_USER[0],
+                   token=SAMPLE_USER[1],
+                   email=SAMPLE_USER[2],
+                   name=SAMPLE_USER[3])
```
jacquerie: You probably want to use a `dict` to store the user, and then use the double splat operator (`**`):
```python
SAMPLE_USER = {
    'orcid': ORCID,
    'token': TOKEN,
    [...]
}

[...]

_register_user(**SAMPLE_USER)
```
as this will actually call `_register_user` with the right keyword arguments.

jacquerie: (The same applies for all the uses that follow!)

setup.py
```diff
@@ -221,6 +221,7 @@
             'inspire_migrator = inspirehep.modules.migrator.tasks',
             'inspire_records = inspirehep.modules.records.tasks',
             'inspire_refextract = inspirehep.modules.refextract.tasks',
+            'inspire_orcid = inspirehep.modules.orcid.tasks',
```
jacquerie: Let's keep these in alphabetical order! This should go just after `inspire_migrator`.

inspirehep/modules/orcid/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""INSPIRE migrator module."""
```
jacquerie: Well, not really : )

inspirehep/modules/orcid/ext.py
```diff
@@ -0,0 +1,34 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Migrator extension."""
```
jacquerie: Also somewhat false : )

inspirehep/modules/orcid/tasks.py
```diff
@@ -0,0 +1,144 @@
...
+    Yields:
+        tuple: user data in the form of (orcid, token, email, name)
+    """
+    redis_url = current_app.config.get('CACHE_REDIS_URL')
+    r = StrictRedis.from_url(redis_url)
+    lock = Lock(r, 'import_legacy_orcid_tokens', expire=120, auto_renewal=True)
+    if lock.acquire(blocking=False):
+        try:
+            while r.llen('legacy_orcid_tokens'):
+                yield literal_eval(r.lrange('legacy_orcid_tokens', 0, 1)[0])
```
kaplun: As discussed IRL, that's unfortunate due to my wrong assumption on Redis serialization. Let's stick to exchange data using JSON serialized string (in this case it will be a simple JSON array with positional values).

PR #2928, opened 2017-11-01
Add new arxiv/doi validators to the literature form.

david-caro:
Adds some extra checks to the form when passing an arxiv id or doi:
    * Check if is already ni the holdingpen with the same source.
    * Check locally (in the DB) if there's a record that matches.

In those cases, it will not allow to pass it.

### DIFF
--- a/inspirehep/modules/forms/validators/simple_fields.py
+++ b/inspirehep/modules/forms/validators/simple_fields.py
@@ -29,12 +29,16 @@ from urllib import urlencode
 import requests
 from flask import current_app
 from flask_login import current_user
+from idutils import is_arxiv
+from invenio_search import current_search_client as es
 from wtforms.validators import ValidationError, StopValidation
 
-from idutils import is_arxiv
 
 from inspire_schemas.utils import load_schema
+from inspire_matcher.api import match
+from inspire_utils.dedupers import dedupe_list
 from inspire_utils.record import get_value
+
 from inspirehep.utils.url import is_pdf_link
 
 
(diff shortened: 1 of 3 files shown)


inspirehep/modules/forms/validators/simple_fields.py
```diff
@@ -111,53 +160,162 @@ def duplicated_orcid_validator(form, field):
...
+        property_name='DOI',
+        value=doi,
+    )
+
+
+def arxiv_id_already_pending_in_holdingpen_validator(form, field):
+    """Check if there's a submission in the holdingpen with the same arXiv ID.
+    """
+    arxiv_id = field.data
+    # TODO: local check for duplicates
```
david-caro: remove the todo

michamos:
maybe you could include also #2835?

david-caro:
They don't really collide or anything, I think it's better to keep them separated.

david-caro:
(no strong opinion though)

tests/unit/forms/test_forms_validators_simple_fields.py
```diff
@@ -40,6 +45,11 @@ def __init__(self, data):
         self.data = data
 
 
+class MockForm(object):
+    def __init__(self, roles=None):
+        self.roles = roles or []
```
jacquerie: Unused?

david-caro: yep, not sure where it comes from :/

PR #2932, opened 2017-11-02
workflow: change logic of matches in holdingpen

### LINKED ISSUE #1724: Rejected records
When a literature record is rejected we need to store the minimal amount of metadata that allows it to be matched in the future (see other issues related to record matching) so that should an update come again the workflow can immediately decide whether to reject it.
* [ ] implement metadata stripper to just keep matching related information (i.e. publication information and title and say first author)
* [ ] implement collection for rejected record
* [ ] implement storing of rejected records
* [ ] implement matching step against rejected records and automatic rejection.
* [ ] in case of safe match, the new record would simply replace the matched one.

### LINKED ISSUE #1799: workflows: When an article is marked as 'already-ingested' there's nothing shown on the holdingpen ui
When an article submission is tagged as 'already-ingested' the workflow is created, and it shows on the holding pen, but it has no indication towards why it's not actually going on, it just shows as 'workflow completed' with no other info or actions available.

### LINKED ISSUE #1800: workflow: the logic to check if a record was previously rejected is not ok
Right now we just check if it's 5 days old, nothing else:

`inspirehep/modules/workflows/tasks/matching.py`
```python
 def previously_rejected(days_ago=None):
    """Check if record exist on INSPIRE or already rejected."""
    @wraps(previously_rejected)
    def _previously_rejected(obj, eng):
        if current_app.config.get('PRODUCTION_MODE'):
            if days_ago is None:
                _days_ago = current_app.config.get('INSPIRE_ACCEPTANCE_TIMEOUT', 5)
            else:
                _days_ago = days_ago

            if is_too_old(obj.data, days_ago=_days_ago):
                obj.log.info("Record is likely rejected previously.")
                return True
        return False

    return _previously_rejected
```

ammirate:
Change the workflow's logic for handling matches in the holdingpen

## Description
This pull request changes the workflow's logic for rejecting, stopping or continuing workflows when the payload matches an existing workflow in the holdingpen.
The new business logic is:
* if matches in holdingpen but have different sources then continue.
* if matches in holdingpen but different sources, reject if the matched was previously rejected.
* if matches in holdingpen but different sources, stop the matched workflow if it wasn't rejected and continue. 

This PR also refactors some workflow's parts, adding marks in the main decisional points. Such marks are used in integration tests to check that the right control flow has been executed. 

## Checklist:

Closes #1800, closes #1799, closes #1724.

Co-Authored-By: David Caro <david@dcaro.es>

### DIFF
--- a/inspirehep/modules/workflows/tasks/matching.py
+++ b/inspirehep/modules/workflows/tasks/matching.py
@@ -30,12 +30,13 @@ from functools import wraps
 from flask import current_app
 
 from invenio_db import db
-from invenio_workflows import workflow_object_class
+from invenio_workflows import workflow_object_class, WorkflowEngine
 
 from inspire_matcher.api import match
 from inspire_utils.dedupers import dedupe_list
 from inspirehep.utils.datefilter import date_older_than
-from inspirehep.utils.record import get_arxiv_categories, get_arxiv_id
+from inspirehep.utils.record import get_arxiv_categories, get_arxiv_id, get_value
+from inspirehep.modules.workflows.tasks.actions import mark
 
 from ..utils import with_debug_logging
 
@@ -151,12 +152,12 @@ def previously_rejected(days_ago=None):
     return _previously_rejected
 
 
-@with_debug_logging
-def pending_in_holding_pen(obj, eng):
-    """Return ``True`` if the record is already present in the Holding Pen.
+def match_non_completed_wf_in_holdingpen(obj, eng):
+    """Return ``True`` if a matching wf is processing in the HoldingPen.
 
     Uses a custom configuration of the ``inspire-matcher`` to find duplicates
-    of the current workflow object in the Holding Pen.
+    of the current workflow object in the Holding Pen not in the
+    COMPLETED state.
 
     Also sets ``holdingpen_matches`` in ``extra_data`` to the list of ids that
     matched.
@@ -167,7 +168,59 @@ def pending_in_holding_pen(obj, eng):
 
     Returns:
         bool: ``True`` if the workflow object has a duplicate in the Holding
-        Pen, ``False`` otherwise.
+        Pen that is not COMPLETED, ``False`` otherwise.
+
+    """
+    def _non_completed(base_record, match_result):
+        return not get_value(match_result, '_source._workflow.status') == 'COMPLETED'
+
+    matched_ids = _pending_in_holding_pen(obj, _non_completed)
+    obj.extra_data['holdingpen_matches'] = matched_ids
+    return bool(matched_ids)
+
+
+def match_previously_rejected_wf_in_holdingpen(obj, eng):
+    """Return ``True`` if matches a COMPLETED and rejected wf in the HoldingPen.
+
+    Uses a custom configuration of the ``inspire-matcher`` to find duplicates
+    of the current workflow object in the Holding Pen in the
+    COMPLETED state, marked as ``approved = False``.
+
+    Also sets ``holdingpen_matches`` in ``extra_data`` to the list of ids that
+    matched.
+
+    Arguments:
+        obj: a workflow object.
+        eng: a workflow engine.
+
+    Returns:
+        bool: ``True`` if the workflow object has a duplicate in the Holding
+        Pen that is not COMPLETED, ``False`` otherwise.
+
+    """
+    def _rejected_and_completed(base_record, match_result):
+        return get_value(match_result, '_source._workflow.status') == 'COMPLETED' and \
+            get_value(match_result, '_source._extra_data.approved') is False
+
+    matched_ids = _pending_in_holding_pen(obj, _rejected_and_completed)
+    obj.extra_data['previously_rejected_matches'] = matched_ids
+    return bool(matched_ids)
+
+
+@with_debug_logging
+def _pending_in_holding_pen(obj, validation_func):
+    """Return the list of matching workflows in the holdingpen.
+
+    Matches the holdingpen records by their ``arxiv_eprint``, their ``doi``,
+    and by a custom validator function.
+
+    Args:
+        obj: a workflow object.
+        validation_func: a function used to filter the matched records.
+
+    Returns:
+        (list): the ids matching the current ``obj`` that satisfy
+        ``validation_func``.
 
     """
     config = {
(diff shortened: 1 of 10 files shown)


david-caro:
> One strange thing is the need to call save_workflow. By construction invenio-workflows was saving upon every halting. So it's suspicious to have the need to explicitly save as a step of the workflow.

The reason is that at some steps you are doing lookups in the db and elasticsearch, and you need the actual latest data to have been persisted, so you have to save before that. Also adding the save after adding marks, makes sure that the current state of the workflow is persisted too, to do searches for broken/failed/suspicious workflows.
Making it explicit helps understanding that the saving is not done in the background, and then being able to understand what would be the status of a workflow and debug it.

ammirate:
> ...adding the save after adding marks, makes sure that the current state of the workflow is persisted too...

Actually, I am removing the explicit ``save`` from the workflow definition, since it is saved everytime we call ``mark``.

david-caro:
> Actually, I am removing the explicit save from the workflow definition, since it is saved everytime we call mark.

Then I don't like it, as it will become 'hidden' and hard to know when it's being saved or not, that will make debugging harder, and easy to forget to save when modifying the workflows when it's needed (as you'll swap around tasks without having a clear idea of which one is saving).

ammirate:
The fact is that the workflow is saved twice in a row, and we can be aware of the ``save`` operation with good documentation. 
In fact, my proposal is: if we agree on this style of implementing the workflow using explicit marks, testing by marked branch and other conventions like these, we should write them down to make any future change compliant with such spec. 

Step by step I believe we can improve both readability and testability of the workflow.

Moreover, I agree in making things explicit, but if we explicitly document that every time you ``mark`` the workflow it is saved, then after marking there is no point in saving again.

I am open to discussion and proposal to improve the workflow, but I'd like to remember that this is out of the scope of this PR, which does some refactoring but wants mainly change some business logic that right now is missing/wrong in production.

david-caro:
* Check the `source` and not the `method` when updating the existing workflow.
* Use the query validator as https://github.com/inspirehep/inspire-next/pull/2950

you can play with:
https://qa.inspirehep.net/holdingpen/809287

ammirate:
> Check the source and not the method when updating the existing workflow.
Ops, the code on QA is old, here this is done in ``holdingpen_match_with_same_source``.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -132,31 +132,42 @@ def already_harvested(obj, eng):
...
+    """
+    def _non_completed(base_record, match_result):
+        return not get_value(match_result, '_source._workflow.status') == 'COMPLETED'
+
+    matched_ids = _pending_in_holding_pen(obj, _non_completed)
+    if matched_ids:
+        obj.extra_data['holdingpen_matches'] = matched_ids
+        return True
+    obj.extra_data['holdingpen_matches'] = []
+    return False
```
david-caro:
```python
obj.extra_data['holdingpen_matches'] = matched_ids
return bool(matched_ids)
```

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -210,8 +243,19 @@ def delete_self_and_stop_processing(obj, eng):
...
+        obj: a workflow object.
+        eng: a workflow engine.
+
+    Returns:
+        None
+    """
+    eng.stop()
 
 
 @with_debug_logging
```
david-caro: This can be deleted

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -240,3 +284,65 @@ def update_existing_workflow_object(obj, eng):
...
+
+    Returns:
+        bool: True if the matched workflow object has the same source,
+        otherwise False.
+    """
+    try:
+        pending_wf_id = obj.extra_data['holdingpen_matches'][0]
+    except IndexError:
+        # no non-COMPLETED workflows matched
+        pending_wf_id = obj.extra_data['previously_rejected_matches'][0]
```
david-caro:
```python
def _get_first_same_source(list_recs):
    for rec in recs:
        if get_value... == get_value...:
            return rec
    return None

matched_rec =  _get_first_same_source(obj.extra_data['holdingpen_matches'])
if not matched_rec:
     matched_rec =  _get_first_same_source(obj.extra_data['previously_rejected_matches'])

return bool(matched_rec)
```

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -317,26 +262,180 @@
...
-    ),
+    )
+]
+
+
+INIT_MARKS = [
+    mark('already-ingested', False),
+    mark('already-in-holding-pen', False),
+    mark('previously_rejected', False),
+    mark('is-update', False),
```
david-caro: s/False/None and set to False in the check too.

david-caro: and put here all the marks initialized to None

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -317,26 +262,180 @@
...
+            mark('is-update', True),
+            save_workflow,
+        ]
+    )
+]
+
+
+STOP_IF_ALREADY_HARVESTED_ON_LEGACY = [
+    IF(
+        already_harvested,
```
david-caro: already_harvested should check too the day limit until merger is in (or in another better named step here too).

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -317,26 +262,180 @@
...
+STOP_IF_EXISTING_SUBMISSION = [
     IF(
-        article_exists,
+        is_submission,
+        IF(
+            is_marked('is-update'),
+            NOTIFY_ALREADY_EXISTING
+        )
+    )
+]
```
david-caro: This can be removed.

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -317,26 +262,180 @@
...
+
+
+HALT_FOR_APPROVAL = [
```
david-caro: Remove this (or CHECK_IF_SUBMISSION_AND_ASK_FOR_APPROVAL), and maybe rename.

ammirate: removed CHECK_IF_SUBMISSION_AND_ASK_FOR_APPROVAL

inspirehep/modules/workflows/workflows/article.py
ammirate: I removed `CHECK_IF_MERGE_AND_STOP_IF_SO` that is doing something different compared to what it says

PR #2937, opened 2017-11-07
workflows: fix the extract_journal_info task

### LINKED ISSUE #2936: handle the journal split in extract_journal_info
## Expected Behavior

The `extract_journal_info` workflow step https://github.com/inspirehep/inspire-next/blob/9d9ecfe0dee3ddbacb2e478b7ae197cb010e9dc1/inspirehep/modules/workflows/tasks/refextract.py#L46-L90 correctly puts the volume letter in the `journal_title`.
## Current Behavior

It still uses the old scheme where the journal letter is put in the `volume`.

jacquerie:
## Description:
Ensures that ``extract_journal_info`` is correctly populating the
``journal_title`` and ``journal_volume`` keys after the journal
split.

## Related Issue:
Closes #2936.

## Checklist:

### DIFF
--- a/tests/unit/workflows/test_workflows_tasks_refextract.py
+++ b/tests/unit/workflows/test_workflows_tasks_refextract.py
@@ -51,8 +51,8 @@ def test_extract_journal_info():
     eng = MockEng()
 
     assert extract_journal_info(obj, eng) is None
-    assert validate(obj.data['publication_info'], subschema) is None
-    assert obj.data['publication_info'] == [
+
+    expected = [
         {
             'artid': '082102',
             'journal_title': 'J. Math. Phys.',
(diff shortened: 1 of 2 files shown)


tests/unit/workflows/test_workflows_tasks_refextract.py
```diff
@@ -80,17 +83,51 @@ def test_extract_journal_info_handles_year_an_empty_string():
     eng = MockEng()
 
     assert extract_journal_info(obj, eng) is None
-    assert validate(obj.data['publication_info'], subschema) is None
-    assert obj.data['publication_info'] == [
+
+    expected = [
         {
             'artid': '134',
             'journal_title': 'Astrophys. J.',
```
michamos: So here it looks like it's normalizing the journal name, but it puts spaces after the `.`s nonetheless.

jacquerie: I believe that this is the reason: https://github.com/inspirehep/refextract/blob/19dd41af327512c98323634e0aed78fe3cffc1ec/refextract/references/kbs/journal-titles.kb#L1306

jacquerie: (This is not a problem on Labs because we regenerate `journal-titles.kb` from the data.)

Issue #2862, opened 2017-10-13
workflows: normalize journal title


michamos:
## Expected Behavior

The `publication_info.journal_title` should be normalized by the workflow to match the `short_title` of the journal. The fields that should be considered for this are the same as for `refextract`: https://github.com/inspirehep/inspire-next/blob/67ff20df8c6ed9abb6a7d59a6a7c536d2610a925/inspirehep/modules/refextract/tasks.py#L56-L74. Also the `journal_record` field could then be populated based on the record ref of the matched journal record.

## Current Behavior

A partial implementation exists in https://github.com/inspirehep/inspire-next/blob/67ff20df8c6ed9abb6a7d59a6a7c536d2610a925/inspirehep/utils/normalizers.py#L30-L44, but uses ES whereas refextract uses the DB directly. @jacquerie wanted to use the latter, he can comment on the reasons. However it is used in `literaturesuggest` only: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/literaturesuggest/normalizers.py#L117. It should be moved to the workfows so that it runs also for hepcrawl harvests.

## Context

This is needed if we want to obtain the correct journal titles for the hepcrawl harvests.

iulianav:
@michamos does the `journal_record` field exist at the moment?

ammirate:
Closed by #2938.

PR #2938, opened 2017-11-07
Workflows: normalize journal title

iulianav:
## Description

The publication_info.journal_title should be normalized by the workflow to match the short_title of the journal. The fields that should be considered for this are the same as for refextract. . Also the journal_record field could then be populated based on the record ref of the matched journal record.

## Related Issue

https://github.com/inspirehep/inspire-next/issues/2862

## Motivation and Context

A partial implementation exists in `inspire-next/inspirehep/utils/normalizers.py`, but uses ES whereas refextract uses the DB directly. @jacquerie wanted to use the latter, he can comment on the reasons. However it is used in literaturesuggest only: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/literaturesuggest/normalizers.py#L117. It should be moved to the workfows so that it runs also for hepcrawl harvests. This is needed if we want to obtain the correct journal titles for the hepcrawl harvests.

## Checklist:


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -285,3 +301,47 @@ def refextract(obj, eng):
...
+
+            for line in lines:
+                normalized_title_mapping = line.split('---')
+                variant_title = normalized_title_mapping[0]
+                normalized_title = normalized_title_mapping[1]
+
+                if journal_title == variant_title:
+                    journal_title = normalized_title.strip('\n')
+
+                    obj.data['publication_info'][index]['journal_title'] = journal_title
```
michamos: This index is wrong if there is a `publication_info` with no `journal_title`.

michamos: This is still wrong, the problem is the following. If the record contains something like
```
{
    'publication_info': [
        {
	    'cnum': 'C01-01-01',
	},
	{
	    'journal_title': 'Journal of Foo',
	    'journal_volume': '42',
	    'artid': 12345,
	}
    ]
}
```
then `enumerate(get_value(obj.data, 'publication_info.journal_title'))` is `[(0, 'Journal of Foo')]`, but `obj.data['publication_info'][0]['journal_title']` does not exist.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -285,3 +300,51 @@ def refextract(obj, eng):
...
+    if not s:
+        return
+
+    result = RE_ALPHANUMERIC.sub(' ', s)
+    result = ' '.join(result.split())
+    result = result.upper()
+
+    if not result:
+        return
+    return result
```
ammirate: `return` will return `None`, so you don't need to differentiate between the two cases :)

ammirate:
For me this PR looks kind of ok, I would just suggest to move the tests somewhere else, since you are not running a complete workflow but just asserting that calling a function on a workflow object does what you expect. Thus, you can also avoid to have the fixture that creates and run the workflow. Using `workflow_object_class.create()` should be enough :)
WDYT?

iulianav:
@ammirate I will try to do what you suggested with the `workflow_object_class.create()`

inspirehep/modules/workflows/tasks/actions.py
jacquerie: But `''` (empty string) is falsy without being `None`, so this change would change behavior.

PR #2948, opened 2017-11-13
InspireRecord: allow skipping the files downloading

david-caro:
We need a way to skip downloading the records files to speed up the
migrations and not use any space, as those migrations will be used
only for it's metadata until labs becomes master.

Added two methods to skip the files downloading:
* Using a global setting RECORDS_SKIP_FILES
* Using a parameter to `InspireRecord.create` and
  `InspireRecord.update` called `skip_files`.

The latter has precedence. By default the previous behavior
(downloading the files) is preserved.

### DIFF
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -72,9 +72,19 @@ class InspireRecord(Record):
                 records.
         """
         files_src_record = kwargs.pop('files_src_record', None)
+        skip_files = kwargs.pop(
+            'skip_files',
+            current_app.config.get('RECORDS_SKIP_FILES', False),
+        )
+
         new_record = super(InspireRecord, cls).create(*args, **kwargs)
-        new_record.download_documents_and_figures(src_record=files_src_record)
-        new_record.commit()
+
+        if not skip_files:
+            new_record.download_documents_and_figures(
+                src_record=files_src_record,
+            )
+            new_record.commit()
+
         return new_record
 
     def update(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


david-caro:
This requires a change in the puppet generated configuration and a change in the ingestion workflows, or a change in the migration task (probably the migration task is more suitable).
As I guess that we still want to keep getting files from the workflows (maybe arxiv can be skipped, but all the others probably not, specially for the future but near publisher ones).

inspirehep/modules/records/api.py
```diff
@@ -72,9 +72,19 @@ def create(cls, *args, **kwargs):
                 records.
         """
         files_src_record = kwargs.pop('files_src_record', None)
+        skip_files = kwargs.pop(
+            'skip_files',
+            current_app.config.get('RECORDS_SKIP_FILES', False),
+        )
```
jacquerie: Uhm, the fact that the value of the parameter takes precedence means that the configuration variable is ignored unless the method was explicitly using the parameter, which doesn't really work as a workaround... probably it should be the opposite.

david-caro: The global config is ignored only when the parameter is passed, not the other way around (that allows locally oberriding the config value, if need be).

jacquerie: Whoops, right, I was wrong. My problem was that I put the configuration file in the wrong place, in fact...

jacquerie:
Superseded by https://github.com/inspirehep/inspire-next/pull/2986.

PR #2950, opened 2017-11-13
forms: filter out deleted records when checking dupes

david-caro:
We are currently blocking submissions of records that were deleted,
we should let them go to the holding pen.

This depends on https://github.com/inspirehep/inspire-matcher/pull/31

### DIFF
--- a/inspirehep/modules/forms/validators/simple_fields.py
+++ b/inspirehep/modules/forms/validators/simple_fields.py
@@ -88,6 +88,9 @@ def does_exist_in_inspirehep(query, collections=None):
 
 
 def duplicated_validator(property_name, property_value):
+    def _is_not_deleted(base_record, match_result):
+        return not get_value(match_result, '_source.deleted', default=False)
+
     config = {
         'algorithm': [
             {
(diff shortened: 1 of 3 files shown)


inspirehep/modules/forms/validators/simple_fields.py
```diff
@@ -88,6 +88,9 @@ def does_exist_in_inspirehep(query, collections=None):
 
 
 def duplicated_validator(property_name, property_value):
+    def _is_not_deleted(base_record, match_result):
+        return not match_result.get('_source', {}).get('deleted', False)
```
jacquerie: Chained `get`s can always be replaced by `get_value`. For example, here it would be:
```python
return not get_value(match_result, '_source.deleted', default=False)
```

jacquerie:
> This depends on inspirehep/inspire-matcher#31

Wouldn't it be cleaner to bump the minimal requirement of `inspire-matcher` in `setup.py` to reflect this?

Also, something is wrong with the existing tests: we're using an unreleased feature of `inspire-matcher`, and yet the tests are not failing. This usually is a symptom of too aggressive mocking.

david-caro:
I couldn't know the version until it's merged.

About the tests, inspire-matcher will silently fallback to the default validator if what you pass is not what it expects, so it's not unusual that they don't fail and that it behaves as before. I think though that I missed commiting the new tests for this feature :/

david-caro:
@ammirate all yours

ammirate:
lgtm

Issue #2863, opened 2017-10-13
workflows: set refereed/conference paper based on journal information


michamos:
## Expected Behavior

after the normalization in #2862 has run, the following has to happen based on the contents of the journal records (plural as in general there can be more than one publication_info) it is published in:

* if all journals have `proceedings = True` but none has `refereed = True`, if `article` is present in `document_type` remove it and add `conference paper`;
* if any journal has both `proceedings = True` and `refereed = True`, and `conference paper` is not present in `document type`, set `refereed = True`;
* if any journal has `refereed = True` but no `proceedings = True`, set `refereed = True`;
* if all journals have `refereed = False`, set `refereed = False`.

@inspirehep/inspire-content do you agree about these rules? In MARC terms, they are are taking the contents of `690` to automatically tag the corresponding journal articles and refereed for a paper means `980:PUBLISHED`.

## Current Behavior

this information is not populated based on the journal information.

## Context

This is needed to properly handle hepcrawl harvests.

michamos:
@iulianav this is what I was referring to this morning

PR #2951, opened 2017-11-14
Workflows: set `refereed` and conference paper in `document_type`

iulianav:
## Description

After the workflow [step](https://github.com/inspirehep/inspire-next/issues/2862) that normalizes the journal title, set refereed/document_type fields according to the journals information accessed in `publication_info` and a set of rules.

## Related Issue

https://github.com/inspirehep/inspire-next/issues/2863

## Motivation and Context

 This is needed to properly handle hepcrawl harvests.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -233,20 +233,6 @@ def is_submission(obj, eng):
 
 
 @with_debug_logging
-def get_journal_coverage(obj, eng):
-    """Return the journal coverage that this article belongs to."""
-    journals = replace_refs(get_value(obj.data, 'publication_info.journal_record'), 'db')
-
-    if not journals:
-        return
-
-    if any(journal['_harvesting_info'].get('coverage') == 'full' for journal in journals):
-        obj.extra_data['journal_coverage'] = 'full'
-    else:
-        obj.extra_data['journal_coverage'] = 'partial'
-
-
-@with_debug_logging
 def submission_fulltext_download(obj, eng):
     submission_pdf = obj.extra_data.get('submission_pdf')
     if submission_pdf and is_pdf_link(submission_pdf):
@@ -308,6 +294,7 @@ def refextract(obj, eng):
 
     Returns:
         None
+
     """
     pdf_references, text_references = [], []
     source = get_value(obj.data, 'acquisition_source.source')
@@ -416,3 +403,52 @@ def normalize_journal_titles(obj, eng):
 
             if result:
                 obj.data['publication_info'][index]['journal_record'] = result.records_metadata_json['self']
+
+
+@with_debug_logging
+def set_document_type_and_refereed(obj, eng):
+    """Set the `refereed` field to the appropriate value based on the journals' `refereed`
+    and `proceedings` fields values. Also, modify the `document_type` field accordingly.
+
+    The journal information is retrieved using the `journal_record` field value which is a
+    json reference to the journal records.
+
+    Note:
+        If some of the journals present in the `publication_info` have no `journal_record`
+        json reference associated, only the ones that do have the associated reference
+        are taken into account in making the decisions. If there are no journals or none
+        have an associated json reference to the corresponding journal record, the function
+        return  None and no further changes are being made.
+
+    Args:
+        obj: a workflow object.
+        eng: a workflow engine.
+
+    Returns:
+        None
+
+    """
+    journals = replace_refs(get_value(obj.data, 'publication_info.journal_record'), 'db')
+
+    if not journals:
+        return
+
+    if any(journal.get('refereed') and journal.get('proceedings') for journal in journals) \
+            and 'conference paper' not in obj.data.get('document_type'):
+                    obj.data['refereed'] = True
+
+    if any(journal.get('refereed') and not journal.get('proceedings') for journal in journals):
+        obj.data['refereed'] = True
+
+    if all(journal.get('proceedings') for journal in journals) \
+            and all(not journal.get('refereed') for journal in journals):
+        doc_types = obj.data.get('document_type')
+        for index, doc_type in enumerate(doc_types):
+            if doc_type == 'article':
+                del obj.data['document_type'][index]
+                if 'conference paper' not in doc_types:
+                    obj.data['document_type'].append('conference paper')
+                break
+
+    if all(not journal.get('refereed', True) for journal in journals):
+                obj.data['refereed'] = False
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -233,6 +233,20 @@ def is_submission(obj, eng):
 
 
 @with_debug_logging
+def get_journal_coverage(obj, eng):
+    """Return the journal coverage that this article belongs to."""
+    journals = replace_refs(get_value(obj.data, 'publication_info.journal_record'), 'db')
+
+    if not journals:
+        return
+
+    if any(journal['_harvesting_info'].get('coverage') == 'full' for journal in journals):
+        obj.extra_data['journal_coverage'] = 'full'
+    else:
+        obj.extra_data['journal_coverage'] = 'partial'
+
+
+@with_debug_logging
 def submission_fulltext_download(obj, eng):
     submission_pdf = obj.extra_data.get('submission_pdf')
     if submission_pdf and is_pdf_link(submission_pdf):
--- a/inspirehep/modules/workflows/workflows/article.py
+++ b/inspirehep/modules/workflows/workflows/article.py
@@ -54,6 +54,7 @@ from inspirehep.modules.workflows.tasks.actions import (
     normalize_journal_titles,
     prepare_update_payload,
     refextract,
+    set_document_type_and_refereed,
     submission_fulltext_download,
     save_workflow,
 )
@@ -205,6 +206,7 @@ POSTENHANCE_RECORD = [
     filter_keywords,
     prepare_keywords,
     remove_references,
+    set_document_type_and_refereed,
 ]
 
 
--- a/tests/integration/workflows/fixtures/jou_record_not_refereed_and_no_proceedings.json
+++ b/tests/integration/workflows/fixtures/jou_record_not_refereed_and_no_proceedings.json
@@ -1,0 +1,16 @@
+{
+    "$schema":"http://localhost:5000/schemas/records/journals.json",
+    "_collections":[
+        "Journals"
+    ],
+    "control_number":1936477,
+    "journal_title":{
+        "title":"No Proceedings and Not Refereed J."
+    },
+    "self":{
+        "$ref":"http://localhost:5000/api/journals/1936477"
+    },
+    "proceedings":false,
+    "refereed":false,
+    "short_title":"No.Pro.No.Ref.J."
+}
--- a/tests/integration/workflows/fixtures/jou_record_not_refereed_and_proceedings.json
+++ b/tests/integration/workflows/fixtures/jou_record_not_refereed_and_proceedings.json
@@ -1,0 +1,16 @@
+{
+    "$schema":"http://localhost:5000/schemas/records/journals.json",
+    "_collections":[
+        "Journals"
+    ],
+    "control_number":1936478,
+    "journal_title":{
+        "title":"Proceedings and Not Refereed J. A"
+    },
+    "self":{
+        "$ref":"http://localhost:5000/api/journals/1936478"
+    },
+    "proceedings":true,
+    "refereed":false,
+    "short_title":"Pro.No.Ref.J.A."
+}
--- a/tests/integration/workflows/fixtures/jou_record_not_refereed_and_proceedings_2.json
+++ b/tests/integration/workflows/fixtures/jou_record_not_refereed_and_proceedings_2.json
@@ -1,0 +1,16 @@
+{
+    "$schema":"http://localhost:5000/schemas/records/journals.json",
+    "_collections":[
+        "Journals"
+    ],
+    "control_number":1936479,
+    "journal_title":{
+        "title":"Proceedings and Not Refereed J. B"
+    },
+    "self":{
+        "$ref":"http://localhost:5000/api/journals/1936479"
+    },
+    "proceedings":true,
+    "refereed":false,
+    "short_title":"Pro.No.Ref.J.B."
+}
--- a/tests/integration/workflows/fixtures/jou_record_refereed.json
+++ b/tests/integration/workflows/fixtures/jou_record_refereed.json
@@ -5,17 +5,17 @@
     ],
     "control_number":1936475,
     "journal_title":{
-        "title":"A Test Journal1"
+        "title":"No Proceedings and Refereed J."
     },
     "title_variants":[
-        "A Test Journal1 Variant 1",
-        "A Test Journal1 Variant 2",
-        "A Test Journal1 Variant 3"
+        "No Proceedings and Refereed J. Variant 1",
+        "No Proceedings and Refereed J. Variant 2",
+        "No Proceedings and Refereed J. Variant 3"
     ],
     "self":{
         "$ref":"http://localhost:5000/api/journals/1936475"
     },
     "proceedings":false,
     "refereed":true,
-    "short_title":"Test.Jou.1"
+    "short_title":"No.Pro.Ref.J."
 }
--- a/tests/integration/workflows/fixtures/jou_record_refereed_and_proceedings.json
+++ b/tests/integration/workflows/fixtures/jou_record_refereed_and_proceedings.json
@@ -5,17 +5,17 @@
     ],
     "control_number":1936476,
     "journal_title":{
-        "title":"A Test Journal2"
+        "title":"Proceedings and Refereed J."
     },
     "title_variants":[
-        "A Test Journal2 Variant 1",
-        "A Test Journal2 Variant 2",
-        "A Test Journal2 Variant 3"
+        "Proceedings and Refereed J. Variant 1",
+        "Proceedings and Refereed J. Variant 2",
+        "Proceedings and Refereed J. Variant 3"
     ],
     "self":{
         "$ref":"http://localhost:5000/api/journals/1936476"
     },
     "proceedings":true,
     "refereed":true,
-    "short_title":"Test.Jou.2"
+    "short_title":"Pro.Ref.J."
 }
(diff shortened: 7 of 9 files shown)


iulianav:
I know I branched off the journal title normalization branch here and then just deleted every change introduced by that branch because I realized it was not a good idea! I will rebase the last commit at some point if this is a problem!

michamos:
what you did seems like a bad idea :) better to `rebase -i master` and remove the commits you don't want to have on your current branch instead of undoing the changes by hand. You could `git rebase -i master` and then squash all your changes now.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -285,3 +295,47 @@ def refextract(obj, eng):
...
+    journals = filter(lambda x: 'journal_title' in x, publications)
+
+    modify_doc_type = False
+    modify_ref_f = False
+
+    if not journals:
+        return None
+
+    for index, journal in enumerate(journals):
+        query = RecordMetadata.query.filter(
```
michamos: you don't need to do all this. In principle, after #2938, the `journal_record` will have been written in the `publication_info`, so you can just get the record from the DB instead of doing this complicated query.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -51,6 +58,105 @@
...
+
+    yield workflow_object
+
+    WorkflowObjectModel.query.filter_by(id=workflow_object.id).delete()
+    db.session.commit()
+
+
+@pytest.fixture(scope='function')
+def insert_journals_in_db(workflow_app):
+    """Temporarily add the HAL id to the CERN record."""
```
michamos: wrong docstring

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -285,3 +295,47 @@ def refextract(obj, eng):
...
         obj.log.info('Extracted %d references from text.', len(text_references))
         obj.data['references'] = text_references
+
+
+@with_debug_logging
+def set_document_type_and_refereed(obj, eng):
+    publications = obj.data.get('publication_info')
+    journals = filter(lambda x: 'journal_title' in x, publications)
+
+    modify_doc_type = False
```
michamos: `is_conference_paper`

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -285,3 +295,47 @@ def refextract(obj, eng):
...
+    modify_ref_f = False
```
michamos: `is_refereed`

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -285,3 +295,47 @@ def refextract(obj, eng):
...
+    if not journals:
+        return None
+
+    for index, journal in enumerate(journals):
+        query = RecordMetadata.query.filter(
+            RecordMetadata.json['_collections'].op('?')('Journals')).filter(
+            cast(RecordMetadata.json['short_title'], String) == type_coerce(journal['journal_title'], JSON))
+        result = db.session.execute(query).fetchone()
+
+        if result:
```
michamos: please read the spec in the issue: it talks about **all**/**any** journal. Here you are overwriting your decisions on every journal you encounter. There's also a difference between a flag not being present and being false. You can write this in a such a way that it reads exactly like the spec. E.g. for the first point
```
if any(journal.get('refereed') for journal in journals) and all(not journal.get('proceedings') for journal in journals):
    # add conference paper to document_type
```

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -285,3 +295,47 @@ def refextract(obj, eng):
...
+                obj.data['refereed'] = True
+
+            if result.records_metadata_json['refereed'] and not result.records_metadata_json['proceedings']:
+                obj.data['refereed'] = True
+
+    if modify_doc_type:
+        doc_types = obj.data.get('document_type')
+
+        for index_d, type in enumerate(doc_types):
+            if type == 'article':
```
michamos: `type` is a builtin you are shadowing

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -285,3 +295,47 @@ def refextract(obj, eng):
...
+                del obj.data['document_type'][index_d]
+                obj.data['document_type'].append('conference paper')
```
michamos: you can combine these two lines by overwriting instead of deleting and appending: `obj.data[index_d] = 'conference paper'`

iulianav: @michamos i was doing that at first, but then i thought what if there are both 'article' and 'conference paper' in there? I am not sure whether that ever happens tho

michamos: good point :+1: the first rule of inspire metadata is "anything that can happen will hapen".

inspirehep/modules/workflows/tasks/actions.py
iulianav: I am not sure how i can get it with just the reference?

inspirehep/modules/workflows/tasks/actions.py
michamos: However, you are not checking if `conference paper` is already present.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -233,3 +379,256 @@ def test_harvesting_arxiv_workflow_manual_accepted(
...
+            {
+                'journal_title': 'Test.Jou.4',
+                'journal_record':
+                    {
+                        '$ref': 'http://localhost:5000/api/journals/1936478'
+                    }
+            }
+        ],
+    }
+
```
ammirate: I would put this record, as well as the others in the tests above, in a fixture to make it more readable

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -285,3 +286,52 @@ def refextract(obj, eng):
...
+            and all(not journal.get('refereed') for journal in journals):
+        doc_types = obj.data.get('document_type')
+        for index, doc_type in enumerate(doc_types):
+            if doc_type == 'article':
+                del obj.data['document_type'][index]
+                if 'conference paper' not in doc_types:
+                    obj.data['document_type'].append('conference paper')
+                break
+
+    if all(not journal.get('refereed') for journal in journals):
```
michamos: `journal.get('refereed', True)` to be sure that if any journal has no `refereed`, this flag doesn't get set.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -233,3 +313,66 @@ def test_harvesting_arxiv_workflow_manual_accepted(
         obj = eng.processed_objects[0]
         # It was accepted
         assert obj.status == ObjectStatus.COMPLETED
+
+
+def test_workflows_change_document_type_article(workflow_app, insert_journals_in_db, workflow):
```
michamos: the convention is to have `test_{function_name}_{extra_details}`, so something like `test_set_document_type_and_refereed_all_proceedings_none_refereed`. This applies to all your tests.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -233,3 +313,66 @@ def test_harvesting_arxiv_workflow_manual_accepted(
...
+    workflow.data = json.loads(pkg_resources.resource_string(
+                __name__, os.path.join('fixtures', 'lit_record_article_all_journals_proceedings_and_none_refereed.json')))
+
+    set_document_type_and_refereed(workflow, None)
+
+    assert 'conference paper' in workflow.data['document_type']
```
michamos: as this function is setting both `refereed` and `conference paper`, you should assert something about both in all tests.

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -192,6 +193,7 @@
             refextract,
         ]
     ),
+    set_document_type_and_refereed,
```
michamos: @ammirate not sure if this is the right place for it, ideally it would run after a merge and halt for curation, so it can take into account all info in the record.

ammirate: I think @iulianav put it here to be coherent with the enhancing and, in particular, for `extract_journal_info` have some benefit from it. But yes, have it when the record is completed makes sense.
At this point, I would put it in `POSTENHANCE_RECORD`.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -285,3 +286,52 @@ def refextract(obj, eng):
...
+
+@with_debug_logging
+def set_document_type_and_refereed(obj, eng):
+    """Set the `refereed` field to the appropriate value based on the journals' `refereed`
+    and `proceedings` fields values. Also, modify the `document_type` field accordingly.
+
+    The journal information is retrieved using the `journal_record` field value which is a
+    json reference to the journal records.
+
+    Note:
```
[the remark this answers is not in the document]
jacquerie: `Note:` is correct! It's part of the sections recognized by Napoleon: https://pypi.python.org/pypi/sphinxcontrib-napoleon#docstring-sections

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -285,3 +286,52 @@ def refextract(obj, eng):
...
+        eng: a workflow engine.
+
+    Returns:
+        None
+
+    """
+    journals = replace_refs(get_value(obj.data, 'publication_info.journal_record'), 'db')
+
+    if not journals:
+        return None
```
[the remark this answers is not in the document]
iulianav: But `Pythin in a nutshell` says it's a good practice to return `None`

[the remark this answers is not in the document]
michamos: [PEP8](https://www.python.org/dev/peps/pep-0008/) disagrees:
>Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as return None, and an explicit return statement should be present at the end of the function (if reachable).

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -233,3 +313,73 @@ def test_harvesting_arxiv_workflow_manual_accepted(
...
+    assert 'refereed' not in workflow.data
+
+
+def test_set_document_type_and_refereed_all_proceedings_none_refereed_no_effect(workflow_app, insert_journals_in_db, workflow):
+    workflow.data = json.loads(pkg_resources.resource_string(
+                __name__, os.path.join('fixtures', 'lit_record_all_journals_proceedings_and_none_refereed.json')))
+
+    set_document_type_and_refereed(workflow, None)
+
+    assert 'conference paper' not in workflow.data['document_type']
```
michamos: these tests are really hard to check: one needs to look at this file for assertions, then at the `lit` fixture for the content of the record, then at 2 journal fixtures to check what's in them, so 4 files to understand exactly what is being asserted. May I suggest:
1. Inline the `lit` fixtures: they are not being reused, so it's fine to have the content of each record inside the test.
2. Name the journals in a meaningful way: `Refereed J.`,`Refereed and Proceedings J.`, etc., with some letters (`A`, `B`, etc.) at the end if you need to have more than one of the same type.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -233,3 +313,73 @@ def test_harvesting_arxiv_workflow_manual_accepted(
...
+    workflow.data = json.loads(pkg_resources.resource_string(
+                __name__, os.path.join('fixtures', 'lit_record_article_all_journals_proceedings_and_none_refereed.json')))
+
+    set_document_type_and_refereed(workflow, None)
+
+    assert 'conference paper' in workflow.data['document_type']
+    assert 'refereed' not in workflow.data
+
+
+def test_set_document_type_and_refereed_all_proceedings_none_refereed_no_effect(workflow_app, insert_journals_in_db, workflow):
```
michamos: `no_effect` -> `not_article`, as that's the reason there is no effect.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -48,7 +57,66 @@
...
+    yield workflow_object
+
+    WorkflowObjectModel.query.filter_by(id=workflow_object.id).delete()
+    db.session.commit()
+
+
+@pytest.fixture(scope='function')
+def insert_journals_in_db(workflow_app):
+    """Temporarily add few journals in the DB"""
+    from inspirehep.modules.migrator.tasks import record_insert_or_replace  # imported here because it is a Celery task
```
michamos: you don't import this here. It's the other way around: if you need to import stuff inside a celery task, the import should be done inside the task, as it will be run somewhere else.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -48,7 +57,66 @@
...
+        record_insert_or_replace(journal_pro_and_ref)
+        record_insert_or_replace(journal_no_pro_and_not_ref)
+        record_insert_or_replace(journal_pro_and_no_ref_a)
+        record_insert_or_replace(journal_pro_and_no_ref_b)
+    db.session.commit()
+    es.indices.refresh('records-journals')
+
+    yield
+
+    _delete_record('jou', 1936475)
```
michamos: do you have to harcdode this? would be better to get this from the return value of `record_insert_or_replace`. Alternatively, @jacquerie suggests using real journals to be added to the demo records.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -233,3 +302,253 @@ def test_harvesting_arxiv_workflow_manual_accepted(
...
+
+def test_set_document_type_and_refereed_any_refereed_and_proceedings_but_no_conference_paper(workflow_app, insert_journals_in_db, workflow):
+    workflow.data = {
+        "_collections": [
+            "Literature"
+        ],
+        "titles": [
+            "A title"
+        ],
+        "document_type": [
```
michamos: can you add `article`? results should be the same, but you are testing more.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -233,3 +302,253 @@ def test_harvesting_arxiv_workflow_manual_accepted(
...
+            },
+            {
+                "journal_title": "Pro.No.Ref.J.B."
+            }
+        ]
+    }
+
+    set_document_type_and_refereed(workflow, None)
+
+    assert 'conference paper' not in workflow.data['document_type']
```
michamos: this should come after.

tests/integration/workflows/test_arxiv_workflow.py
iulianav: But the problem is I do not know how to look for the journals that have the exact field that i need. It seems a bit time consuming? That is why I chose to mock them like this.

tests/integration/workflows/test_arxiv_workflow.py
iulianav: Moving the import outside of the test makes it stuck. @ammirate has the same problem.  @jacquerie @michamos are there some updates regarding this mistery?

tests/integration/workflows/test_arxiv_workflow.py
ammirate: IIRC, you found the journals you needed in inspirehep.net using the notation `CODE_FIELD:**`, right? If yes, we can proceed as the above comments were suggesting ;)

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -233,3 +302,253 @@ def test_harvesting_arxiv_workflow_manual_accepted(
         obj = eng.processed_objects[0]
         # It was accepted
         assert obj.status == ObjectStatus.COMPLETED
+
+
+def test_set_document_type_and_refereed_all_proceedings_none_refereed(workflow_app, insert_journals_in_db, workflow):
```
ammirate: Here is happening the same as commented in #2938. Maybe we can avoid running workflows, just instantiating a workflow object, since you are asserting that calling `set_document_type_and_refereed` on a workflow object does the right thing.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -416,3 +417,52 @@ def normalize_journal_titles(obj, eng):
...
+    journals = replace_refs(get_value(obj.data, 'publication_info.journal_record'), 'db')
+
+    if not journals:
+        return
+
+    if any(journal.get('refereed') and journal.get('proceedings') for journal in journals) \
+            and 'conference paper' not in obj.data.get('document_type'):
+                    obj.data['refereed'] = True
+
+    if any(journal.get('refereed') and not journal.get('proceedings') for journal in journals):
```
ammirate: Since this leads to the same action if the previous `if`, I would put it in `or` with the other condition statement.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -416,3 +417,52 @@ def normalize_journal_titles(obj, eng):
...
+        doc_types = obj.data.get('document_type')
+        for index, doc_type in enumerate(doc_types):
+            if doc_type == 'article':
+                del obj.data['document_type'][index]
+                if 'conference paper' not in doc_types:
+                    obj.data['document_type'].append('conference paper')
+                break
+
+    if all(not journal.get('refereed', True) for journal in journals):
+                obj.data['refereed'] = False
```
jacquerie: The logic of this method is quite complicated, and could benefit from extracting some variables and giving them meaningful names: https://sourcemaking.com/refactoring/extract-variable.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -416,3 +417,52 @@ def normalize_journal_titles(obj, eng):
...
+
+    if all(journal.get('proceedings') for journal in journals) \
+            and all(not journal.get('refereed') for journal in journals):
+        doc_types = obj.data.get('document_type')
+        for index, doc_type in enumerate(doc_types):
+            if doc_type == 'article':
+                del obj.data['document_type'][index]
+                if 'conference paper' not in doc_types:
+                    obj.data['document_type'].append('conference paper')
+                break
```
jacquerie: That's quite the convoluted way for replacing `article` with `conference paper`!
```python
>>> arr = ['foo', 'bar', 'baz']
>>> arr.remove('foo')
>>> arr
['bar', 'baz']
```
but you also need to handle the case in which `article` is not in `document_type`.

PR #2953, opened 2017-11-14
tasks: merge records based on `deleted_records`

### LINKED ISSUE #2912: invert merge_merged_records
## Expected Behavior

`deleted_records` is looked at to determine which records have been merged into the current one and mark them as merged: 

## Current Behavior

The opposite thing happens, the lookup is based on the deleted records. As we are not going to migrate all deleted records (because some of them lack enough information to be valid), this will miss some deleted records.
https://github.com/inspirehep/inspire-next/blob/9b6b0e13897930314aaa76243fc689ee3708c7ac/inspirehep/modules/records/tasks.py#L133-L150

ammirate:
Closes #2912.

## Description
Merging records now works in the opposite direction. All the records having ``deleted_records`` are retrieved from ES. Then, all the records referred in that field are loaded and merged into the enriched record.

## Checklist:

### DIFF
--- a/inspirehep/modules/records/tasks.py
+++ b/inspirehep/modules/records/tasks.py
@@ -37,7 +37,7 @@ from invenio_search import current_search_client as es
 from inspire_dojson.utils import get_recid_from_ref
 from inspirehep.modules.records.api import InspireRecord
 from inspirehep.modules.records.utils import get_endpoint_from_record
-from inspirehep.utils.record_getter import get_db_record
+from inspirehep.modules.pidstore.utils import get_pid_type_from_schema
 
 
 logger = get_task_logger(__name__)
@@ -132,30 +132,37 @@ def get_records_to_update(old_ref):
 @shared_task
 def merge_merged_records():
     """Merge all records that were marked as merged."""
-    def _get_record(recid):
-        pid = PersistentIdentifier.query.filter_by(pid_value=str(recid)).one()
-        return get_db_record(pid.pid_type, recid)
+    records = get_merged_records()
 
-    records = get_records_to_merge()
+    for record in records:
 
-    with db.session.begin_nested():
-        for record in records:
-            recid = record['control_number']
-            other_recid = get_recid_from_ref(record['new_record'])
-            other_record = _get_record(other_recid)
+        record_pid = PersistentIdentifier.query.filter_by(object_uuid=record.id).one()
+        deleted_ids = [get_recid_from_ref(ref) for ref in record['deleted_records']]
 
-            logger.info('Merged records: %d and %d', recid, other_recid)
-            record.merge(other_record)
-            record.commit()
+        for deleted_id in deleted_ids:
+                deleted_pid = PersistentIdentifier.query.filter_by(
+                    pid_value=str(deleted_id)
+                ).one_or_none()
+
+                if not deleted_pid:
+                    deleted_pid = PersistentIdentifier.create(
+                        pid_type=get_pid_type_from_schema(record['$schema']),
+                        pid_value=deleted_id,
+                        object_type='rec'
+                    )
+                    deleted_pid.register()
+                    db.session.add(deleted_pid)
+
+                deleted_pid.redirect(record_pid)
     db.session.commit()
 
 
-def get_records_to_merge():
+def get_merged_records():
     def _get_uuids_to_merge():
         body = {
             'query': {
                 'exists': {
-                    'field': 'new_record',
+                    'field': 'deleted_records',
                 },
             },
         }
@@ -166,10 +173,7 @@ def get_records_to_merge():
         for result in query:
             yield result['_id']
 
-    def _get_records_to_merge(uuids):
-        return InspireRecord.get_records(uuids)
-
     uuids = _get_uuids_to_merge()
-    records = _get_records_to_merge(uuids)
+    records = InspireRecord.get_records(uuids)
 
     return records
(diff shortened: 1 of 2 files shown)


inspirehep/modules/records/tasks.py
```diff
@@ -136,26 +136,28 @@ def _get_record(recid):
...
+            for deleted_id in deleted_ids:
+                rec_to_merge = _get_record(deleted_id)
+                logger.info('Merged records %d in %d', deleted_id, recid)
+                rec_to_merge.merge(rec_to_merge)
+                rec_to_merge.commit()
     db.session.commit()
 
 
-def get_records_to_merge():
+def get_enriched_records():
```
michamos: `get_merged_records`?

inspirehep/modules/records/tasks.py
```diff
@@ -166,10 +168,10 @@ def _get_uuids_to_merge():
         for result in query:
             yield result['_id']
 
-    def _get_records_to_merge(uuids):
+    def _get_enriched_records(uuids):
```
michamos: `_get_merged_records`?

inspirehep/modules/records/tasks.py
```diff
@@ -136,26 +136,28 @@ def _get_record(recid):
...
-            other_recid = get_recid_from_ref(record['new_record'])
-            other_record = _get_record(other_recid)
 
-            logger.info('Merged records: %d and %d', recid, other_recid)
-            record.merge(other_record)
-            record.commit()
+            deleted_ids = [get_recid_from_ref(ref) for ref in record['deleted_records']]
+
+            for deleted_id in deleted_ids:
+                rec_to_merge = _get_record(deleted_id)
```
michamos: here the problem is that there isn't necessarily a record corresponding to this `recid`, because some deleted records don't migrate properly. If there is, this should be fine, but otherwise, you should register the PID for redirection without there being an underlying record.

ammirate:
This PR is still in WIP, that's why is lacking tests

david-caro:
I think we should avoid getting the point where we have to make "consistent" the database by running a periodical task like this one (not actually being called anywhere).
I think it would be better to do this linking at migration time, when the record comes in, if it was deleted by some other record, then try to link it, if that record does not yet exist in the db, just skip it, then when the record finally gets migrated, it will update that first deleted record to link it to itself.

Also, might be a good time to change the `new_record` and `deleted_records` keys, as they are a bit misleading, imo they should be something like `replaced_by` and `replaced_records`.

Wdyt?

tests/integration/test_records.py
```diff
@@ -306,6 +307,26 @@ def test_merged_records_stay_merged(api_client, merged_records):
...
 
 
+def test_merge_record_with_non_existing_pid(api_client, merged_records):
+    def get_pid_entry(recid):
+        return PersistentIdentifier.query.filter_by(pid_value=str(recid)).one_or_none()
+
+    merged_record = get_db_record('lit', 111)
+
+    assert get_recid_from_ref(merged_record['deleted_records'][0]) == 222
+    # remove it so it doesn't exists anymore
```
david-caro: s/exists/exist/

tests/integration/test_records.py
```diff
@@ -306,6 +307,26 @@ def test_merged_records_stay_merged(api_client, merged_records):
...
+    db.engine.execute("delete from pidstore_pid where pid_value='222'")
```
david-caro: Maybe you should use here the 'usual' way to delete a record, as it might do something more than just delete a row in a table.

tests/integration/test_records.py
```diff
@@ -306,6 +307,26 @@ def test_merged_records_stay_merged(api_client, merged_records):
...
+
+    pid_for_222 = get_pid_entry(222)
+    assert pid_for_222 is None
+
+    merge_merged_records.delay()
```
david-caro: delay will try to run the task in the background, better just call it without delay no?

tests/integration/test_records.py
ammirate: doh!

david-caro: I usually do the typos by swapping letters, like 'eixsts' or 'rwapper' xd

tests/integration/test_records.py
ammirate: The problem is that ``delete`` and ``_delete`` functions delete the record but not the PID, which is exactly what I need.

david-caro: Then you might want to use the pid model then? (to delete the pid)

ammirate: Same, ``delete`` marks the PID as deleted (if it's already registered as in this case).
Any suggestion?

https://invenio-pidstore.readthedocs.io/en/latest/api.html#module-invenio_pidstore.models

ammirate:
Ready!

inspirehep/modules/records/tasks.py
```diff
@@ -132,30 +132,38 @@ def _get_records_to_update(uuids):
...
-            logger.info('Merged records: %d and %d', recid, other_recid)
-            record.merge(other_record)
-            record.commit()
+        for deleted_id in deleted_ids:
+                # the record has not been migrated. Redirect only its pid
+                deleted_pid = PersistentIdentifier.query.filter_by(
+                    pid_value=str(deleted_id)
+                ).one_or_none()
+
+                if not deleted_pid:
```
david-caro:
```python
ids_to_redirect = (deleted_id in deleted_ids if has_been_migrated(deleted_id))
for deleted_id in ids_to_redirect:
    deleted_pid = PersistentIdentifier.create(
        pid_type=get_pid_type_from_schema(record['$schema']),
        pid_value=deleted_id,
        object_type='rec'
      )
      deleted_pid.register()
      db.session.add(deleted_pid)
```

PR #2955, opened 2017-11-15
mget: handle empty results

### LINKED ISSUE #2954: search/api.py:mget does not handle properly empty search results
https://sentry.inspirehep.net/inspire-sentry/prod/issues/179/

```
KeyError: '_source'
(12 additional frame(s) were not displayed)
...
  File "inspirehep/modules/records/serializers/response.py", line 44, in view
    serializer.serialize(pid, record, links_factory=links_factory),
  File "inspirehep/modules/records/serializers/impactgraph_serializer.py", line 99, in serialize
    'earliest_date'
  File "inspirehep/utils/record_getter.py", line 89, in get_es_records
    return search_class.mget(uuids, **kwargs)
  File "inspirehep/modules/search/api.py", line 93, in mget
    results = [document['_source'] for document in documents['docs']]

KeyError: '_source'
```

david-caro:

Fixes #2954

### DIFF
--- a/inspirehep/modules/search/api.py
+++ b/inspirehep/modules/search/api.py
@@ -90,7 +90,10 @@ class SearchMixin(object):
                 body={'ids': uuids},
                 **kwargs
             )
-            results = [document['_source'] for document in documents['docs']]
+            results = [
+                document['_source'] for document in documents['docs']
+                if document.get('found')
+            ]
         except RequestError as e:
             logger.exception(e)
 


inspirehep/modules/search/api.py
```diff
@@ -90,7 +90,10 @@ def mget(self, uuids, **kwargs):
                 body={'ids': uuids},
                 **kwargs
             )
-            results = [document['_source'] for document in documents['docs']]
+            results = [
+                document['_source'] for document in documents['docs']
+                if document.get('found')
```
jmartinm: So when there are no real matches, there is still one document but with no `found` key?

david-caro: kind of:
```
{
    u'_id': u'7f6bd778-c2f8-4ddf-a98d-34db4024544c',
    u'_index': u'records-hep',
    u'_type': u'hep',
    u'found': False
}
```

jmartinm: Ah I see 👍

david-caro:
You might be right that that error is happening, but it any case I don't think that this function is the one that should be responsible of knowing that, should be the function calling this the one that should know if the uuids it's looking for should or should not exist and throw an error (or handle however needed).

jacquerie:
> You might be right that that error is happening, but it any case I don't think that this function is the one that should be responsible of knowing that, should be the function calling this the one that should know if the uuids it's looking for should or should not exist and throw an error (or handle however needed).

Thing is, we shouldn't need this function at all, because this could just be a `get_records`. I need a diagram to explain the design error that led to the creation of this function... will you be back on Monday?

david-caro:
yep

jacquerie:
This can be closed for the reasons contained in https://github.com/inspirehep/inspire-next/pull/2955#pullrequestreview-78792449.

PR #2958, opened 2017-11-15
mappings-hep: populate two diff. name variations

chris-asl:
Populate two distinct name variations field, one containing all
variations and one ignoring single name variations (i.e. lastnames).
That way depending on the user's query we can select which field to
filter with. E.g. `a Ellis` will filter using the "all variations"
field, while `a Ellis, John` will query the more specific one.

## Motivation and Context
Attempts to imitate better the behaviour of legacy to get more exact results. 

### Note 
The way the no lastnames list (_i.e. more specific one_) is generated (ignoring variations which after being `.split()` have length more than one) will result in adding also name variations for authors with many lastnames.  

For example, `Caro-Estevez` lastname, becomes  a lastname variation: `['Caro', 'Caro Estevez']`.
The way it's implemented now, it will add `'Caro Estevez'` in the no lastnames list, but will not add `Caro`.  

## Checklist:

### DIFF
--- a/inspirehep/modules/records/mappings/records/hep.json
+++ b/inspirehep/modules/records/mappings/records/hep.json
@@ -230,8 +230,17 @@
                             "type": "completion"
                         },
                         "name_variations": {
-                            "index": "not_analyzed",
-                            "type": "string"
+                            "properties": {
+                                "all": {
+                                    "index": "not_analyzed",
+                                    "type": "string"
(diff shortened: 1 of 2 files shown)


chris-asl:
Superseded by https://github.com/inspirehep/inspire-next/pull/3009 and probably the two name variations will not be needed after all.

PR #2961, opened 2017-11-16
templates: fix the latex in literature title search results

### LINKED ISSUE #2855: Fix LaTeX in titles
In the search results, the LaTeX is not shown properly in the title: latex

whereas it appears correctly in the detailed page:
latex2

eg. https://labs.inspirehep.net/search?page=1&size=25&q=refersto:recid:1372575&sort=-mostrecent

Dinika:
Signed-off-by: Dinika <dinika.saxena@cern.ch>

## Description
I tried adding a setTimeout without specifying the time delay of 500ms but it wouldn't work. 

## Related Issue
fixes #2855 

## Checklist:

### DIFF
--- a/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_general_macros.tpl
+++ b/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_general_macros.tpl
@@ -187,7 +187,7 @@
   require([
     "jquery",
     ], function ($) {
-      $(document).ready(function () {
+       setTimeout(function () {
         var mathJaxURL = "//cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js?config=TeX-AMS-MML_HTMLorMML";
         $.getScript(mathJaxURL, function() {
           MathJax.Hub.Config({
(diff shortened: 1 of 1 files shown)


inspirehep/modules/theme/templates/inspirehep_theme/format/record/Inspire_Default_HTML_general_macros.tpl
```diff
@@ -198,7 +198,7 @@
           });
           MathJax.Hub.Queue(["Typeset", MathJax.Hub]);
         });
-      })
+      }, 500)
```
jacquerie: Uhm, this doesn't seem very robust...

Dinika: I see your point. I tested this code on a slower 3G network and of course, it fails. I'm looking into another solution.

jacquerie: Thanks for checking! Yes, slower network is exactly the scenario I had in mind.

Perhaps there's a better hook for MathJax than DOMContentLoaded? Something in Angular's lifecycle that fires after Angular has finished its first render, or something like that...

PR #2967, opened 2017-11-17
Setting 'marionette' capability in selenium to False

bavaria95:
Disabling 'marionette' capability in selenium

## Motivation and Context
After update of the `pytest-selenium` library from `1.11.1` to `1.11.2` acceptance tests started to fail due to absence of `geckodriver` (`WebDriverException: Message: 'geckodriver' executable needs to be in PATH.`).
New version of this library now uses defined `DesiredCapabilities` from the Selenium, among which there is `"marionette": True` flag (which was not set up in the previous version). Thus, it needs to be set up as False to make it work again.

## Description
It's possible to set capability via pytest arguments, but it accepts everything as a string, thus by providing `False` it will assume only that it's `'False'`. The workaround is to put there an empty string, so python considers it as a False value.

## Checklist:

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -82,7 +82,7 @@ services:
   acceptance:
     extends:
       service: test-service_base
-    command: py.test --driver Remote --host selenium --port 4444 --capability browserName firefox --html=selenium-report.html tests/acceptance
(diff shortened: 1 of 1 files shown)


david-caro:
Related to pytest-dev/pytest-selenium/issues/139#issuecomment-345254286 let's see if there's a better way

david-caro:
This was not really needed in the end

PR #2981, opened 2017-11-21
receivers: populate bookautocomplete in Python

iulianav:
## Description
 
When populating the "bookautocomplete" field of Literature records, retrieving the values supposedly stored in "bookautocomplete" was done as if ES stores JSON objecs, i.e. using `json.get("bookautocomplete", [])`. This always returns `None`, therefore I had to remove `"copy-to":
"bookautocomplete"` from the hep.json mapping. Instead, I retrieve the required values that are assumed to be stored there manually and then appropriately update "bookautocomplete".

## Checklist:

### DIFF
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -131,13 +131,13 @@ def enhance_after_index(sender, json, *args, **kwargs):
 
     .. note::
 
-       ``populate_recid_from_ref`` **MUST** come before ``add_book_autocomplete``
+       ``populate_recid_from_ref`` **MUST** come before ``populate_bookautocomplete``
        because the latter puts a JSON reference in a completion payload, which
        would be expanded to an incorrect ``payload_recid`` by the former.
 
     """
     populate_recid_from_ref(sender, json, *args, **kwargs)
-    add_book_autocomplete(sender, json, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


inspirehep/modules/records/receivers.py
```diff
@@ -155,18 +155,26 @@ def add_book_autocomplete(sender, json, *args, **kwargs):
...
+        'imprints.date',
+        'imprints.publisher',
+        'isbns.value',
+    }
 
-    result = json.get('bookautocomplete', [])
-    result.extend(authors)
-    result.extend(titles)
+    authors = force_list(get_value(json, 'authors.full_name', []))
+    titles = force_list(get_value(json, 'titles.title', []))
```
jacquerie: I liked the proposal you had in chat that put everything in `paths`... what was the reason to have some there and some outside of it?

inspirehep/modules/records/receivers.py
```diff
@@ -155,18 +155,26 @@ def add_book_autocomplete(sender, json, *args, **kwargs):
...
+    input_values = filter(input_values)
 
     ref = get_value(json, 'self.$ref')
 
     json.update({
         'bookautocomplete': {
-            'input': result,
+            'input': input_values,
             'payload': {
                 'authors': authors,
```
iulianav: @jacquerie this is why

PR #2986, opened 2017-11-22
records: allow to skip downloading the files

jacquerie:
## Description
Adds two ways to skip downloading the files:

* Setting the configuration variable ``RECORDS_SKIP_FILES``.
* Passing the ``skip_files`` parameter to the ``create`` and
  ``update`` methods in ``InspireRecord``.

Note that the latter has precedence, and that by default the
previous behavior (downloading the files) is preserved.

## Related Issue
Supersedes https://github.com/inspirehep/inspire-next/pull/2948.

## Checklist:

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -1136,6 +1136,16 @@ RECORDS_VALIDATION_TYPES = {
     'array': (list, tuple),
 }
 
+RECORDS_SKIP_FILES = False
+"""Disable the downloading of files at record creation and update times.
+
+Note:
+
+  The ``skip_files`` parameter passed to ``InspireRecord.create`` or
+  ``InspireRecord.update`` takes precedence on this config variable.
+
+"""
+
 JSONSCHEMAS_HOST = "localhost:5000"
 JSONSCHEMAS_REPLACE_REFS = True
 JSONSCHEMAS_LOADER_CLS = 'inspirehep.modules.records.json_ref_loader.SCHEMA_LOADER_CLS'
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -71,13 +71,23 @@ class InspireRecord(Record):
                 record in the list that has it in it's files iterator before
                 downloading them, for example to merge existing
                 records.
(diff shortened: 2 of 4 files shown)


inspirehep/modules/records/api.py
```diff
@@ -91,13 +101,23 @@ def update(self, *args, **kwargs):
...
-            src_records=files_src_records,
-        )
+        skip_files = kwargs.pop('skip_files', config_skip_files)
+
+        record = super(InspireRecord, self).update(*args, **kwargs)
+
+        if not skip_files:
+            record.download_documents_and_figures(src_records=files_src_records, only_new=True)
+
+        return record
```
jacquerie: Well, here I made a silly mistake. I'm going to fix it tomorrow, as I only needed the `create` method to work for the HAL push.

inspirehep/config.py
```diff
@@ -1136,6 +1136,11 @@
     'array': (list, tuple),
 }
 
+# If set, will disable downloading records files by default (can be overriden
+# by the `skip_files` parameter of `InspireRecord.create` and
+# `InspireRecord.update`)
```
jacquerie: A docstring after the configuration variable will show up when building the docs (e.g. http://inspirehep.readthedocs.io/en/latest/_api/inspirehep.modules.hal.html#module-inspirehep.modules.hal.config), while a comment will be ignored. So I'd say it's better to document variables with docstrings.

tests/integration/helpers/utils.py
```diff
@@ -43,3 +48,19 @@ def _delete_record(pid_type, pid_value):
...
+            requests_mocker.register_uri(
+                requests_mock.ANY,
+                re.compile('.*(indexer|localhost).*'),
+                real_http=True,
+            )
+
+        for address in addresses:
+            requests_mocker.register_uri(**address)
+
+        yield
```
jacquerie: I will also try to see if this (unrelated) refactor can be spun out in its own commit.

jacquerie: Practicality beat purity and I left it in.

Issue #2273, opened 2017-04-22
global: uniform the handling of names


jacquerie:
According to the schema names:
> Should be of the form 'family names, first names'
(https://github.com/inspirehep/inspire-schemas/blob/1db9777634844724c522728521ab1f876b25ca8e/inspire_schemas/records/hep.yml#L328-L339)

Logic to deal with this, either by normalizing some input we get (legacy, forms, harvesting) or producing the output other services expect (ORCID, HAL) is scattered all over the codebase. Often it is wrong. For example: https://github.com/inspirehep/inspire-next/blob/c14f5327de85f56da9a6565997a10fd6b3c71c43/tests/unit/dojson/test_dojson_hep_bd9xx.py#L924

It should be centralized, and we should probably something semi-smart like https://pypi.python.org/pypi/nameparser to handle the parsing for us, instead of directly monkeying with strings.

PR #2904, opened 2017-10-25
workflows: add manual merger

ammirate:
## Description
Add a new workflow to perform a manual merge between two records.
The workflow behaves in the following way:
* read the records from the DB, 
* check their sources
* merge them and halt, always
* when resumed, saves the roots, add some metadata, and save the merged record.

## Checklist:

- [x] I have all the information that I need (if not, move to `RFC` and look for it).
- [x] I linked the related issue(s) in the corresponding commit logs.
- [x] I wrote [good commit log messages](https://github.com/torvalds/subsurface-for-dirk/blob/5f15ad5a86ada3c5e574041a5f9d85235322dabb/README#L92-L119).
- [x] My code follows the code style of this project.
- [x] I've added any new docs if API/utils methods were added.
- [x] I have updated the existing documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

Signed-off-by: Antonio Cesarano cesarano2607@gmail.com

PR #2992, opened 2017-11-25
hal: use inspire-utils to parse names

jacquerie:
## Related issues:
Addresses https://github.com/inspirehep/inspire-next/issues/2273, supersedes https://github.com/inspirehep/inspire-next/pull/2990.

## Checklist:


jacquerie:
The record that doesn't pass validation in https://travis-ci.org/inspirehep/inspire-next/jobs/307298427 is because of https://github.com/inspirehep/inspire-schemas/issues/280.

jacquerie:
And all the failing tests in https://travis-ci.org/inspirehep/inspire-next/jobs/307298424 might be because of the same reason, as `pdf_url_correct` gets put in `urls.value`, but it isn't a valid URI.

tests/unit/workflows/test_workflows_tasks_upload.py
```diff
@@ -72,7 +75,8 @@ def test_set_schema_builds_a_full_url_for_the_schema():
 
     data = {'$schema': 'hep.json'}
     extra_data = {}
-    assert validate(data['$schema'], subschema) is None
+    with pytest.raises(ValidationError):
+        validate(data['$schema'], subschema) is None
```
jacquerie: This is notable: `inspire-dojson` puts something invalid there, which `set_schema` makes valid. This is the only exception to the usual task invariant (valid in, valid out).

jacquerie:
This requires the general dependencies bumping done in https://github.com/inspirehep/inspire-next/pull/2904.

PR #2995, opened 2017-11-27
wip editor files upload view

zzacharo:

## Checklist:

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -164,6 +164,7 @@ call `register_signals()` on our own.
 
 RECORD_EDITOR_INDEX_TEMPLATE = 'inspirehep_theme/invenio_record_editor/index.html'
 RECORD_EDITOR_PREVIEW_TEMPLATE_FUNCTION = get_detailed_template_from_record
+RECORD_EDITOR_FILE_UPLOAD_FOLDER = 'inspirehep/modules/editor/temp'
 
 # Path to where journal kb file is stored from `inspirehep.modules.refextract.tasks.create_journal_kb_file`
 # On production, if you enable celery beat change this path to point to a shared space.
--- a/inspirehep/modules/editor/views.py
+++ b/inspirehep/modules/editor/views.py
@@ -24,9 +24,11 @@
 
 from __future__ import absolute_import, division, print_function
 
-from flask import Blueprint, jsonify, request
+import os
+from flask import Blueprint, jsonify, request, current_app
 from flask_login import current_user
 from sqlalchemy_continuum import transaction_class, version_class
+from werkzeug.utils import secure_filename
 
 from invenio_db import db
 from invenio_records.models import RecordMetadata
@@ -47,6 +49,7 @@ from refextract import (
 from .permissions import editor_permission, editor_use_api_permission
 from ...utils import tickets
 
+MAX_UNIQUE_KEY_COUNT = 100
 
 blueprint = Blueprint(
     'inspirehep_editor',
@@ -250,3 +253,24 @@ def _simplify_ticket_response(ticket):
         owner=ticket['Owner'],
         date=ticket['Created'],
         link=ticket['Link'])
+
+
+@blueprint.route('/upload', methods=['POST'])
+@editor_use_api_permission.require(http_exception=403)
+def upload_files():
+    if 'file' not in request.files:
+        return jsonify(success=False, message='No file part'), 400
+    attachment = request.files['file']
+    if attachment:
+        filedir = current_app.config['RECORD_EDITOR_FILE_UPLOAD_FOLDER']
+        if not os.path.isdir(filedir):
+            os.makedirs(filedir)
+        filename = secure_filename(attachment.filename)
+        filepath = os.path.join(filedir, filename)
+        if os.path.isfile(filepath):
+            count = 1
+            while os.path.isfile(filepath) and count < MAX_UNIQUE_KEY_COUNT:
+                filepath = '%s_%s' % (filepath, count)
+                count += 1
+        attachment.save(filepath)
+        return jsonify(dict(path=filepath))
--- a/.gitignore
+++ b/.gitignore
@@ -134,3 +134,6 @@ twistd.pid
 
 # Build artifacts
 dbs/
+
+# editor upload folder
+inspirehep/modules/editor/temp/
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -61,6 +61,7 @@ def app():
         CELERY_CACHE_BACKEND='memory',
         CELERY_EAGER_PROPAGATES_EXCEPTIONS=True,
         SECRET_KEY='secret!',
+        RECORD_EDITOR_FILE_UPLOAD_FOLDER='tests/integration/editor/temp',
         TESTING=True,
     )
 
--- a/tests/integration/editor/test_editor_views.py
+++ b/tests/integration/editor/test_editor_views.py
@@ -24,12 +24,14 @@ from __future__ import absolute_import, division, print_function
 
 import os
 import json
+import shutil
 
 import pkg_resources
 import pytest
 import requests_mock
 
 from mock import patch
+from StringIO import StringIO
 
 from invenio_accounts.models import SessionActivity
 from invenio_accounts.testutils import login_user_via_session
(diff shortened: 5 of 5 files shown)


inspirehep/modules/editor/views.py
```diff
@@ -225,3 +228,38 @@ def _simplify_ticket_response(ticket):
...
+            filedir,
+            filename
+        )
+        file.save(filepath)
+        return jsonify(filepath)
+
+
+@blueprint.route('/<recid>/upload/clean', methods=['DELETE'])
+@editor_permission
+def clean_files(recid):
```
harunurhan: better to do cron job, otherwise another weird logic on the frontend.
@jacquerie @david-caro @michamos

inspirehep/modules/editor/views.py
```diff
@@ -225,3 +228,38 @@ def _simplify_ticket_response(ticket):
...
+@blueprint.route('/<recid>/upload', methods=['POST'])
+@editor_permission
+def upload_files(recid):
+    # check if the post request has the file part
+    if 'file' not in request.files:
+        return jsonify(success=False, message='No file part')
+    file = request.files['file']
+    # if user does not select file, browser also
+    # submit a empty part without filename
+    if file.filename == '':
```
harunurhan: ditto

inspirehep/modules/editor/views.py
```diff
@@ -225,3 +228,38 @@ def _simplify_ticket_response(ticket):
         owner=ticket['Owner'],
         date=ticket['Created'],
         link=ticket['Link'])
+
+
+@blueprint.route('/<recid>/upload', methods=['POST'])
+@editor_permission
+def upload_files(recid):
+    # check if the post request has the file part
+    if 'file' not in request.files:
```
harunurhan: handled by frontend but can stay. maybe someone else should have look.

inspirehep/modules/editor/views.py
```diff
@@ -225,3 +228,38 @@ def _simplify_ticket_response(ticket):
...
+        filedir = os.path.join(current_app.config['RECORD_EDITOR_FILE_UPLOAD_FOLDER'], recid)
+        if not os.path.isdir(filedir):
+            os.makedirs(filedir)
+        filename = secure_filename(file.filename)
+        filepath = os.path.join(
+            filedir,
+            filename
+        )
+        file.save(filepath)
+        return jsonify(filepath)
```
harunurhan: should be wrapped with a dict, `dict(path=filepath)`

inspirehep/modules/editor/views.py
```diff
@@ -225,3 +228,38 @@ def _simplify_ticket_response(ticket):
...
         link=ticket['Link'])
+
+
+@blueprint.route('/<recid>/upload', methods=['POST'])
+@editor_permission
+def upload_files(recid):
+    # check if the post request has the file part
+    if 'file' not in request.files:
+        return jsonify(success=False, message='No file part')
+    file = request.files['file']
```
michamos: `file` is a Python builtin, you should use another name.

inspirehep/modules/editor/views.py
```diff
@@ -225,3 +228,38 @@ def _simplify_ticket_response(ticket):
...
+    if file.filename == '':
+        return jsonify(success=False, message='No selected file')
+    if file:
+        filedir = os.path.join(current_app.config['RECORD_EDITOR_FILE_UPLOAD_FOLDER'], recid)
+        if not os.path.isdir(filedir):
+            os.makedirs(filedir)
+        filename = secure_filename(file.filename)
+        filepath = os.path.join(
+            filedir,
+            filename
```
michamos: this fits on one line.

zzacharo:
@harunurhan @michamos I updated with your comments.

inspirehep/modules/editor/views.py
```diff
@@ -225,3 +228,22 @@ def _simplify_ticket_response(ticket):
         owner=ticket['Owner'],
         date=ticket['Created'],
         link=ticket['Link'])
+
+
+@blueprint.route('/upload', methods=['POST'])
+@editor_use_api_permission.require(http_exception=403)
+def upload_files(recid):
```
harunurhan: leftover, and causes `TypeError`

inspirehep/modules/editor/views.py
```diff
@@ -225,3 +228,22 @@ def _simplify_ticket_response(ticket):
...
+    # check if the post request has the file part
+    if 'file' not in request.files:
+        return jsonify(success=False, message='No file part')
+    attachment = request.files['file']
+    # if user does not select file, browser also
+    # submit a empty part without filename
+    if attachment:
+        filedir = os.path.join(current_app.config['RECORD_EDITOR_FILE_UPLOAD_FOLDER'], recid)
```
harunurhan: also now no need for a folder for each record since cleanup will be done by a cron job.

inspirehep/modules/editor/views.py
```diff
@@ -225,3 +228,22 @@ def _simplify_ticket_response(ticket):
...
+    if 'file' not in request.files:
+        return jsonify(success=False, message='No file part')
+    attachment = request.files['file']
+    # if user does not select file, browser also
+    # submit a empty part without filename
+    if attachment:
+        filedir = current_app.config['RECORD_EDITOR_FILE_UPLOAD_FOLDER']
+        if not os.path.isdir(filedir):
+            os.makedirs(filedir)
+        filename = secure_filename(attachment.filename)
```
harunurhan: seems like if you upload a file with same name, it'll override. I think file name should be generated or sth.

zzacharo: @harunurhan check it again. I updated with check for file existence.

inspirehep/modules/editor/views.py
```diff
@@ -225,3 +228,27 @@ def _simplify_ticket_response(ticket):
         owner=ticket['Owner'],
         date=ticket['Created'],
         link=ticket['Link'])
+
+
+@blueprint.route('/upload', methods=['POST'])
+@editor_use_api_permission.require(http_exception=403)
+def upload_files():
+    # check if the post request has the file part
```
jmartinm: This comment is not needed

inspirehep/config.py
```diff
@@ -164,6 +164,7 @@
 
 RECORD_EDITOR_INDEX_TEMPLATE = 'inspirehep_theme/invenio_record_editor/index.html'
 RECORD_EDITOR_PREVIEW_TEMPLATE_FUNCTION = get_detailed_template_from_record
+RECORD_EDITOR_FILE_UPLOAD_FOLDER = 'inspirehep/modules/editor/temp'
```
jmartinm: What was the idea behind making this configurable?  Not enough with using https://docs.python.org/2/library/tempfile.html#tempfile.mkstemp ?

inspirehep/modules/editor/views.py
```diff
@@ -225,3 +228,27 @@ def _simplify_ticket_response(ticket):
...
+    # if user does not select file, browser also
+    # submit a empty part without filename
+    if attachment:
+        filedir = current_app.config['RECORD_EDITOR_FILE_UPLOAD_FOLDER']
+        if not os.path.isdir(filedir):
+            os.makedirs(filedir)
+        filename = secure_filename(attachment.filename)
+        filepath = os.path.join(filedir, filename)
+        if os.path.isfile(filepath):
+            count = 1
```
jmartinm: Since this is the second time this pattern is going to be used (first one https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/api.py#L488-L495) it would be good to add a utility function for it and that would make it also easier to test.

inspirehep/modules/editor/views.py
```diff
@@ -225,3 +228,27 @@ def _simplify_ticket_response(ticket):
...
         date=ticket['Created'],
         link=ticket['Link'])
+
+
+@blueprint.route('/upload', methods=['POST'])
+@editor_use_api_permission.require(http_exception=403)
+def upload_files():
+    # check if the post request has the file part
+    if 'file' not in request.files:
+        return jsonify(success=False, message='No file part')
```
jmartinm: Add http error return code.

inspirehep/modules/editor/views.py
jmartinm: Is it important to conserve the name passed from the UI or append an index to the name? Otherwise by using the filename as sufix in https://docs.python.org/2/library/tempfile.html#tempfile.mkstemp you could avoid checking for duplicate files.

inspirehep/config.py
david-caro: We were not sure if we wanted a local temporary file (mkstemp) or one on eos, as if not using eos it requires the sessions to be sticky in the load balancer.

inspirehep/modules/editor/views.py
zzacharo: This name I think would be in the final url for the file and we needed it to be user readable. Where should I put the utility function you think?

inspirehep/modules/editor/views.py
zzacharo: 400 ?

inspirehep/modules/editor/views.py
jmartinm: we could create `utils/files.py` for filename/directory related utils

inspirehep/modules/editor/views.py
jmartinm: Yes, that sounds good

inspirehep/modules/editor/views.py
zzacharo: @jmartinm I used this pattern in a bit different way and in my opinion the pattern inside the while loop is not worth moving to a util function. I tried to integrate the whole loop into a function but the conditions in the 2 cases are different.

jmartinm:
Continued in https://github.com/inspirehep/inspire-next/pull/3014

Issue #2642, opened 2017-08-16
forms: module should contain all forms


jacquerie:
Currently code related to forms is spread in at least 5 modules:

- `forms`, which actually contains a mini framework to build forms (see https://github.com/inspirehep/inspire-next/issues/2641)
- `authors`, which contains the author forms
- `literaturesuggest`, which contains the literature form
- `arxiv`, which contains support code for `literaturesuggest`
- `crossref`, which also contains support code for `literaturesuggest`

This is not ideal. If we assume that https://github.com/inspirehep/inspire-next/issues/2641 has happened then we can put all this code in a `forms` module, which contains further subfolders for the various forms we need (and the future forms that we will create).

Issue #2641, opened 2017-08-16
forms: spin out module as inspire-forms


jacquerie:
The `forms` module is basically a mini framework for forms that adapts `FlaskWTF` to INSPIRE.

This is not ideal, because it means that we're mixing in `inspire-next` library code and application code, which leads to people not knowing where to put code: close to the application, or close to the library? (Not knowing where to put code leads to poor code organization, from which INSPIRE currently suffers).

So we could spin out `forms` as `inspire-forms` and use the classes it provides to build the forms we need in the application.

Issue #2260, opened 2017-04-21
demosite: merge with fixtures


jacquerie:
Related to https://github.com/inspirehep/inspire-next/issues/2253: we no longer have a "demosite", what we have are actually fixtures that we use during development/testing. So the demosite should become part of the fixtures.

PR #2997, opened 2017-11-27
Added the needed changes to support EOS

david-caro:
There were a few places where we were expecting to be able to open
invenio files uris as local files, just needed to add the proper
abstraction by using PyFilesystem.

### DIFF
--- a/inspirehep/modules/editor/views.py
+++ b/inspirehep/modules/editor/views.py
@@ -24,30 +24,33 @@
 
 from __future__ import absolute_import, division, print_function
 
-import os
 from flask import Blueprint, jsonify, request, current_app
 from flask_login import current_user
+from fs.opener import fsopendir
 from sqlalchemy_continuum import transaction_class, version_class
 from werkzeug.utils import secure_filename
 
 from invenio_db import db
 from invenio_records.models import RecordMetadata
-
-from inspirehep.modules.pidstore.utils import get_pid_type_from_endpoint
-from inspirehep.modules.tools import authorlist
-from inspirehep.modules.workflows.workflows.manual_merge import start_merger
-from inspirehep.utils.record_getter import get_db_record
-from inspirehep.utils.references import (
-    local_refextract_kbs_path,
-    map_refextract_to_schema,
-)
 from refextract import (
     extract_references_from_string,
     extract_references_from_url,
 )
 
-from .permissions import editor_permission, editor_use_api_permission
-from ...utils import tickets
+from inspirehep.modules.editor.permissions import (
+    editor_permission,
+    editor_use_api_permission,
+)
+from inspirehep.modules.pidstore.utils import get_pid_type_from_endpoint
+from inspirehep.modules.tools import authorlist
+from inspirehep.utils import tickets
+from inspirehep.utils.record_getter import get_db_record
+from inspirehep.utils.references import (
+    local_refextract_kbs_path,
+    map_refextract_to_schema,
+)
+from inspirehep.utils.url import copy_file
+from inspirehep.modules.workflows.workflows.manual_merge import start_merger
 
 MAX_UNIQUE_KEY_COUNT = 100
 
@@ -262,17 +265,25 @@ def _simplify_ticket_response(ticket):
 def upload_files():
     if 'file' not in request.files:
         return jsonify(success=False, message='No file part'), 400
+
     attachment = request.files['file']
-    if attachment:
-        filedir = current_app.config['RECORD_EDITOR_FILE_UPLOAD_FOLDER']
-        if not os.path.isdir(filedir):
-            os.makedirs(filedir)
-        filename = secure_filename(attachment.filename)
-        filepath = os.path.join(filedir, filename)
-        if os.path.isfile(filepath):
-            count = 1
-            while os.path.isfile(filepath) and count < MAX_UNIQUE_KEY_COUNT:
-                filepath = '%s_%s' % (filepath, count)
-                count += 1
-        attachment.save(filepath)
-        return jsonify(dict(path=filepath))
+    if not attachment:
+        return
+
+    filedir = current_app.config['RECORD_EDITOR_FILE_UPLOAD_FOLDER']
+    fs = fsopendir(filedir, create_dir=True)
+    filename = secure_filename(attachment.filename)
+    base_filename = filename
+    count = 1
+    while fs.isfile(filename) and count < MAX_UNIQUE_KEY_COUNT:
+        filename = '%s_%s' % (base_filename, count)
+        count += 1
+
+    with fs.open(filename, mode='wb') as remote_file:
+        copy_file(attachment.stream, remote_file)
+
+    full_url = fs.getpathurl(filename, allow_none=True)
+    if not full_url:
+        full_url = fs.getsyspath(filename)
+
+    return jsonify({'path': full_url})
--- a/tests/integration/editor/test_editor_views.py
+++ b/tests/integration/editor/test_editor_views.py
@@ -107,6 +107,13 @@ def record_with_two_revisions(app):
     _delete_record('lit', 111)
 
 
+@pytest.fixture(scope='function')
+def clean_uploads_folder(app):
+    yield
+    uploads = app.config['RECORD_EDITOR_FILE_UPLOAD_FOLDER']
+    shutil.rmtree(uploads)
+
+
 def test_get_revisions(log_in_as_cataloger, record_with_two_revisions, api_client):
     response = api_client.get(
         '/editor/literature/111/revisions',
@@ -439,7 +446,7 @@ def test_manual_merge(mock_start_merger, log_in_as_cataloger, api_client):
     assert expected == result
 
 
-def test_upload(app, log_in_as_cataloger, api_client):
+def test_upload(app, log_in_as_cataloger, api_client, clean_uploads_folder):
 
     response = api_client.post(
         '/editor/upload',
@@ -448,8 +455,10 @@ def test_upload(app, log_in_as_cataloger, api_client):
             'file': (StringIO('my file contents'), 'attachment.pdf'),
         },
     )
-    tempfolder = app.config['RECORD_EDITOR_FILE_UPLOAD_FOLDER']
-    assert json.loads(response.data)['path'] == os.path.join(tempfolder, 'attachment.pdf')
+    uploads_folder = app.config['RECORD_EDITOR_FILE_UPLOAD_FOLDER']
+    result = json.loads(response.data)['path']
+    expected = os.path.realpath(os.path.join(uploads_folder, 'attachment.pdf'))
+    assert result == expected
     assert response.status_code == 200
 
     response = api_client.post(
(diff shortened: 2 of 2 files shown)


inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -147,12 +149,26 @@ def _try_to_log(logfn, *args, **kwargs):
     return _decorator
 
 
+def get_uri(uri, outdir=None):
```
jacquerie: Mh, from the name of the function I would expect that this function returns an URI, while instead it returns a file from a URI... what about `get_file_from_uri` ?

tests/unit/workflows/test_workflows_tasks_actions.py
```diff
@@ -426,8 +427,12 @@ def test_prepare_update_payload_overwrites():
...
 @patch('inspirehep.modules.workflows.tasks.actions.get_pdf_in_workflow')
 def test_refextract_from_pdf(mock_get_pdf_in_workflow):
-    mock_get_pdf_in_workflow.return_value = pkg_resources.resource_filename(
-        __name__, os.path.join('fixtures', '1704.00452.pdf'))
+    mock_get_pdf_in_workflow.return_value = get_uri(
+        pkg_resources.resource_filename(
+            __name__,
+            os.path.join('fixtures', '1704.00452.pdf'),
+        )
+    )
```
jacquerie: It sounds like this wasn't correct before... why would `get_pdf_in_workflow` return a filename? Perhaps it should have been `pkg_resources.resource_file` here, and then we no longer need the `get_uri` ?

inspirehep/modules/workflows/utils/__init__.py
david-caro: It's not returning a file, but just a path. Better `retrieve_uri`

tests/unit/workflows/test_workflows_tasks_actions.py
david-caro: `get_pdf_in_workflow` retrieves the filename of the pdf that is in the workflow. `get_uri` is required there because the `refextract` task expects that file to be a temporary one, and removes it when done.

inspirehep/config.py
```diff
@@ -91,7 +91,7 @@
 BASE_FILES_LOCATION = os.path.join(sys.prefix, 'var/data')
 
 # This is needed in order to be able to use EOS files locations
-MAX_CONTENT_LENGTH = 100 * 1024 * 1024 # 100 MiB
+MAX_CONTENT_LENGTH = 100 * 1024 * 1024  # 100 MiB
```
kaplun: This can be squashed with first commit

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -263,12 +264,15 @@ def refextract(obj, eng):
     pdf_references, text_references = [], []
     source = get_value(obj.data, 'acquisition_source.source')
 
-    pdf = get_pdf_in_workflow(obj)
-    if pdf:
+    tmp_pdf = get_pdf_in_workflow(obj)
```
kaplun: Why this?

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -263,12 +264,15 @@ def refextract(obj, eng):
...
+    if tmp_pdf:
         try:
-            pdf_references = extract_references_from_pdf(pdf, source)
+            pdf_references = extract_references_from_pdf(tmp_pdf, source)
         except TimeoutError:
             obj.log.error('Timeout when extracting references from PDF.')
+        finally:
+            if os.path.exists(tmp_pdf):
+                os.unlink(tmp_pdf)
```
kaplun: Is this safe? Aren't we deleting files from under invenio-files?

inspirehep/utils/url.py
```diff
@@ -65,3 +69,17 @@ def is_pdf_link(url):
     correct_magic_number = magic_number.startswith('%PDF')
 
     return correct_magic_number
+
+
+def retrieve_uri(uri, outdir=None):
+    """Retrieves the given uri and stores it in a temporary file."""
+    fd, local_file = tempfile.mkstemp(prefix='inspire', dir=outdir)
```
kaplun: You can use https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile with `deleted=False` which is slightly easier.

inspirehep/utils/url.py
```diff
@@ -65,3 +69,17 @@ def is_pdf_link(url):
...
+    os.close(fd)
+
+    with fsopen(uri, 'rb') as remote_fd:
+        data = remote_fd.read()
```
kaplun: This is not safe. It must be bufferized.

inspirehep/modules/workflows/tasks/actions.py
david-caro: to make explicit that the pdf that you expect to get is a temporary one (and you'll have to clean it up).

inspirehep/modules/workflows/tasks/actions.py
david-caro: the`tmp_pdf` is the locally downloaded copy of the pdf, not really the original one.

inspirehep/utils/url.py
david-caro: I was not expecting to have big files (GB order), but sure, better safe than sorry.

inspirehep/utils/url.py
david-caro: This and the next will be solved with:

```
  7 def retrieve_uri(uri, outdir=None):                                              
  6     """Retrieves the given uri and stores it in a temporary file."""             
  5     fd, local_file = tempfile.mkstemp(prefix='inspire', dir=outdir)              
  4     os.close(fd)                                                                 
  3                                                                                  
  2     fs.copy.copy_file(uri, local_file)                                                   
  1                                                                                  
81      return local_file 
```

as the copy_file requires a string or fs.FS instance, not an open file (will open it with whatever opener is needed for it).
is that ok? I've checked the code underneath, and it uses chunks of:
```
In [1]: import io

In [2]: io.DEFAULT_BUFFER_SIZE
Out[2]: 8192
```

inspirehep/utils/url.py
kaplun: There might be hundredths of MB indeed. And also we might have some malicious user pointing to something huge...

OK, very improbable, but given the small size of our VMs... :-)

inspirehep/utils/url.py
david-caro: dammit... this requires fs>=2, but it's capped by invenio-files-rest, see https://github.com/inveniosoftware/invenio-files-rest/issues/130

lnielsen: You can do this with PyFSFileStorage which has all the support for chunking,  size limit checking etc:

```python
>>> import os
>>> import tempfile
>>> import requests
>>> from invenio_files_rest.storage import PyFSFileStorage
>>> fd, local_file = tempfile.mkstemp(prefix='inspire')
>>> os.close(fd)
>>> dst = PyFSFileStorage(local_file)
>>> dst.save(requests.get(uri, stream=True).raw, size_limit=1*1024*1024, chunk_size=2*1024*1024)
```

lnielsen: For local disk 8kb chunk size is ok....for EOS you definitely want to go to MB size for best performance.

david-caro: Thanks for the tip, though I think it's a bit premature to strive for performance (and probably it's at the pyfs level where we want to handle each different fs type tweaking, here we know one of the files is local, but the other we are not sure).

inspirehep/modules/refextract/tasks.py
```diff
@@ -73,19 +74,55 @@ def create_journal_kb_file():
...
+                        normalized_journal_title,
+                        row['short_title'],
+                    )
+            )
+            kb_data.append(encoded_journal_title)
+
+        if len(kb_data) == batch_size:
+            with fsopen(refextract_journal_kb_path, mode='wb') as fd:
+                fd.write(''.join(kb_data))
+            kb_data = []
```
ammirate: I would suggest putting this code in a function since the pattern used is the same in three points.
Something like
```
. . . 
    for row in titles_query:
        _add_normalized_title(kb_data, row['short_title'])
        _add_normalized_title(kb_data, row['journal_title'])
        _flush_kb_data(batch_size, kb_data, refextract_journal_kb_path)
. . .

File: inspirehep/modules/workflows/tasks/arxiv.py [hunk 9]
```diff
@@ -41,16 +41,14 @@
...
-
 from plotextractor.api import process_tarball
 from plotextractor.converter import untar
 from plotextractor.errors import InvalidTarball, NoTexFilesFound
 
 from ..errors import DownloadError
-from ..utils import download_file_to_workflow, with_debug_logging
+from ..utils import download_file_to_workflow, with_debug_logging, convert
+from ....utils.record import get_arxiv_categories, get_arxiv_id
+from ....utils.url import is_pdf_link, retrieve_uri
```
ammirate: nit but...it's super hard to me reading relative imports :P

jacquerie: ([Google's Python style guide](https://google.github.io/styleguide/pyguide.html#Imports) actually disallows relative imports)

ammirate:
Apart from the above comments, I don't notice any other major change. If somebody has some, please provide them, since this is solving an high priority issue.

inspirehep/modules/workflows/tasks/arxiv.py
jacquerie: Also here this is escaping the current module and reaching the `utils` folder, which is just confusing...

david-caro: Changed.

The fact that we have those "modules" (that actually are python subpackages, don't confuse with python modules) is by itself confusing. And imo a symptom of a wider structural issue in the general organization of the code.

david-caro:
I've added the adaptation of the latest editor file upload, and refactored according to the comments. Ready again for review.

inspirehep/modules/workflows/tasks/arxiv.py
jacquerie: What I would like to have here is a fairly common architecture of large Flask applications based on blueprints, each implementing a distinct feature of INSPIRE: `editor`, `holdingpen`, `search`, `theme`...

This is what Explore Flask calls a "divisional structure": https://exploreflask.com/en/latest/blueprints.html#divisional. I detailed in several issues some of the steps needed to get there: https://github.com/inspirehep/inspire-next/issues/2251, https://github.com/inspirehep/inspire-next/issues/2260, https://github.com/inspirehep/inspire-next/issues/2641, https://github.com/inspirehep/inspire-next/issues/2642

david-caro: One thing that I see there is that it has, tops, a main package (top level) and supackages (one depth level) of python modules. That's something that's completely missed on all those issues.

jacquerie: If what you want is to remove the `modules` folder I agree with that, but that's a much simpler (although quite wide) change than splitting code in their own blueprint-driven modules.

Mind you, we didn't even start with blueprints: in the past some modules were actually single-file Flask applications that were injecting themselves in the main application through internal Flask APIs...

david-caro:
Once this passes the tests, it's ready for review/merge.

david-caro:
Passed the tests, ready :)

Issue #2866, opened 2017-10-16
submission form: journal autocomplete - show more than 5


annetteholtkamp:
## Expected Behavior
The journal autocomplete should show more than 5 candidates.

## Current Behavior
Currently only 5 suggestions are shown. E.g. trying Philosophical Magazine shows 5 entries of the form Phil.Mag.Ser.x but Phil.Mag. is not shown - and cannot be made to appear by typing a lengthier string.

## Context
One has to consult the journal db to find the correct form.

jacquerie:
I tested manually on a local dump of the Journals database that @iulianav's PR fixes this issue, in particular that typing "Philosophical Magazine" will show "Phil.Mag.". It will be deployed soon!

PR #3002, opened 2017-11-28
 config: make journal suggester return 10 hits

iulianav:
## Description

The journal title suggester only displayed 5 results, and not
necessary the most relevant ones. This is due to the fact that
ES completion suggesters are implemented in such way that
it does not support relevance ordering for the suggesters.
Therefore, the current solution is to simply display more
results, i.e. 10 instead of 5.

## Related Issue

Closes https://github.com/inspirehep/inspire-next/issues/2866.

## Checklist:

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -788,7 +788,8 @@ RECORDS_REST_ENDPOINTS = dict(
         },
         suggesters=dict(
             journal_title=dict(completion=dict(
-                field='title_suggest'
+                field='title_suggest',
+                size=10,
             ))
         ),
         list_route='/journals/',


inspirehep/config.py
```diff
@@ -788,7 +788,8 @@
         },
         suggesters=dict(
             journal_title=dict(completion=dict(
-                field='title_suggest'
+                field='title_suggest',
+                size=10
```
jacquerie: As it can be seen from this diff, not adding trailing commas means that, whenever we add a new field, we get "dirty" diffs. Adding a trailing comma here will prevent this from happening in a future commit that will modify this configuration.

jacquerie:
Build https://travis-ci.org/inspirehep/inspire-next/jobs/308471146 was unlucky: https://travis-ci.org/inspirehep/inspire-next/jobs/308471146#L645 was raised because `setuptools` wasn't installed due to a transient PyPI failure (https://travis-ci.org/inspirehep/inspire-next/jobs/308471146#L637), which meant that an old version was present, which doesn't support some more modern features we use in our packaging script.

PR #3003, opened 2017-11-28
migrator: synchronize claims from legacy

### LINKED ISSUE #2971: Claim sync
For the intermediate period where Labs will be master of Literature, but legacy will be master of claims (i.e. confirmed associations between signatures on papers and authors), we need a task on Labs that will receive claims from legacy and apply them back on labs records.

Legacy is currently using the REDIS shared key technique by pushing onto the `legacy_claims` redis key the following JSON Array:
```
[<INSPIRE BAI>, <author recid>, <literature recid>, <signature>, <flag>]
```

Where:
* `INSPIRE BAI` is the the author profile identifier as e.g. `B.Storaci.1`. See: http://inspire-schemas.readthedocs.io/en/latest/schemas/records/elements/id.html#inspire-bai, 
* `author recid` is the integer representing the record ID of the corresponding author, 
* `literature recid` is the integer representing the record ID of the literature record from which the claim originates,
* `signature` is actually the `full_name` as in: http://inspire-schemas.readthedocs.io/en/latest/schemas/records/hep.html#hep-json-properties-authors-items-properties-full-name. On legacy there is no way to identify a particular signature within a paper. (On Labs this is possible thanks to: http://inspire-schemas.readthedocs.io/en/latest/schemas/records/hep.html#uuid) so the task being created will have to pick the first signature that matches the `full_name`. This is sad, but it's the best thing we can do.
* `flag`: this will be either 2 or -2. If it's -2 it can be ignored (it means this is anti-claim, i.e. a it's stating that signature MUST NOT be connected with this author) See also: #1071

## Expected Behavior
The task should poll the above `legacy_claims` entry, and pop one claim at a time and applying it to the corresponding record.
Applying it means: 
* getting the corresponding literature record, 
* identifing the right signature (if possible), by matching with `full_name`s, 
* enriching the signature with the INSPIRE BAI information (by overwriting an already existing INSPIRE BAI one or append it to the list of existing IDs),
* enriching the `$ref` of the corresponding author record (by overwriting it if already existing),
* setting the `curated_flag` to 1

Note: if after applying a claim the record is in fact not modified (because all the information was already there), we shall not save the record, so that we avoid creating unnecessary noise in the history.

szymonlopaciuk:
## Description
This synchronizes claims from legacy, coming through redis, as specified in #2971.
If a record is unchanged as a result of applying a claim, no changes are committed.

## Related Issue
Closes #2971 

## Checklist:

### DIFF
--- a/inspirehep/modules/migrator/tasks.py
+++ b/inspirehep/modules/migrator/tasks.py
@@ -393,15 +393,11 @@ def apply_claim(inspire_bai, author_recid, lit_recid, signature, flag):
     if flag == -2:  # FIXME: Handle the flag
         return
 
-    try:
-        lit_record = get_db_record('lit', lit_recid)
-        for author in get_value(lit_record, 'authors', []):
-            if author.get('full_name') == signature:
-                if claim_author_field(author, inspire_bai, author_recid):
-                    lit_record.commit()
-    except PIDDoesNotExistError:
-        # Either author or lit doesn't exist
-        pass
+    lit_record = get_db_record('lit', lit_recid)
+    for author in get_value(lit_record, 'authors', []):
+        if author.get('full_name') == signature:
+            if claim_author_field(author, inspire_bai, author_recid):
+                lit_record.commit()
 
 
 def claim_author_field(author, inspire_bai, author_recid):
--- a/inspirehep/modules/migrator/tasks.py
+++ b/inspirehep/modules/migrator/tasks.py
@@ -29,7 +29,7 @@ import re
 import zlib
 from collections import Counter
 from itertools import chain
-from simplejson import loads
+from json import loads
 
 import click
 from celery import group, shared_task
@@ -53,7 +53,7 @@ from invenio_search import current_search_client as es
 from invenio_search.utils import schema_to_index
 
 from inspire_dojson.processors import overdo_marc_dict
-from inspire_dojson.utils import get_recid_from_ref
+from inspire_dojson.utils import get_recid_from_ref, get_record_ref
 from inspire_schemas.builders import SignatureBuilder
 from inspire_utils.dedupers import dedupe_list
 from inspire_utils.helpers import force_list
@@ -62,6 +62,7 @@ from inspirehep.modules.pidstore.minters import inspire_recid_minter
 from inspirehep.modules.pidstore.utils import get_pid_type_from_schema
 from inspirehep.modules.records.api import InspireRecord
 from inspirehep.modules.records.receivers import index_after_commit
+from inspirehep.utils.record_getter import get_db_record
 
 from .models import InspireProdRecords
 
@@ -389,12 +390,11 @@ def apply_claim(inspire_bai, author_recid, lit_recid, signature, flag):
         signature (string): name as written on the paper
         flag (string): record can be ignored if flag == -2 (claim reject)
     """
-    if flag == -2:
+    if flag == -2:  # FIXME: Handle the flag
         return
 
     try:
-        lit_pid = PersistentIdentifier.get('lit', lit_recid)
-        lit_record = InspireRecord.get_record(lit_pid.object_uuid)
+        lit_record = get_db_record('lit', lit_recid)
         for author in get_value(lit_record, 'authors', []):
             if author.get('full_name') == signature:
                 if claim_author_field(author, inspire_bai, author_recid):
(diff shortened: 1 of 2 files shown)


inspirehep/modules/migrator/tasks.py
```diff
@@ -29,10 +29,12 @@
 import zlib
 from collections import Counter
 from itertools import chain
+from simplejson import loads
```
david-caro: why simplejson?

szymonlopaciuk: I vaguely remember it being used somewhere else, no reason beyond that. Should I change it to `json`?

inspirehep/modules/migrator/tasks.py
```diff
@@ -356,3 +359,72 @@ def migrate_and_insert_record(raw_record):
...
+    if lock.acquire(blocking=False):
+        try:
+            while r.llen('legacy_claims'):
+                claim = loads(r.lrange('legacy_claims', 0, 1)[0])
+                apply_claim(*claim)
+                r.lpop('legacy_claims')
+        finally:
+            lock.release()
+    else:
+        logger.info('Import_legacy_claims already executed. Skipping.')
```
david-caro: I think this should be at least an error, so it shows up on sentry, @kaplun it should not happen right? and if it does, we should actually take a look.

I think that's better than ignoring it completely, as we would not notice that it got stuck.

inspirehep/modules/migrator/tasks.py
```diff
@@ -356,3 +359,72 @@ def migrate_and_insert_record(raw_record):
...
+    try:
+        lit_pid = PersistentIdentifier.get('lit', lit_recid)
+        lit_record = InspireRecord.get_record(lit_pid.object_uuid)
+        for author in get_value(lit_record, 'authors', []):
+            if author.get('full_name') == signature:
+                if claim_author_field(author, inspire_bai, author_recid):
+                    lit_record.commit()
+    except PIDDoesNotExistError:
+        # Either author or lit doesn't exist
+        pass
```
david-caro: Same here, I'm not sure if we should ignore these.

inspirehep/modules/migrator/tasks.py
```diff
@@ -356,3 +359,72 @@ def migrate_and_insert_record(raw_record):
...
+        lit_recid (string): Inspire literature record identifier
+        signature (string): name as written on the paper
+        flag (string): record can be ignored if flag == -2 (claim reject)
+    """
+    if flag == -2:
+        return
+
+    try:
+        lit_pid = PersistentIdentifier.get('lit', lit_recid)
+        lit_record = InspireRecord.get_record(lit_pid.object_uuid)
```
jacquerie: `get_db_record` already does this job for you:
```python
record = get_db_record(control_number, 'lit')
```

inspirehep/modules/migrator/tasks.py
```diff
@@ -356,3 +359,72 @@ def migrate_and_insert_record(raw_record):
...
+        for author in get_value(lit_record, 'authors', []):
```
jacquerie: `get_value` is useful only when you do chained `get`s. In this case it's faster to do:
```python
for author in record.get('authors', []):
```

inspirehep/modules/migrator/tasks.py
kaplun: Nope it can happily happen. When the currently executing one is too slow to operate. E.g. say that it take 10 minutes to ingest all the latest claims (very improbable), and that the task is scheduled to run every minute in celery beat: then after 1 minute a second identical task will be executed and we don't want that. But it's not an error. It's just the system being unnecessarily slow...

inspirehep/modules/migrator/tasks.py
```diff
@@ -356,3 +359,72 @@ def migrate_and_insert_record(raw_record):
...
+    author_copy = deepcopy(author)
+
+    builder = SignatureBuilder(author)
+    builder.set_uid(inspire_bai, schema='INSPIRE BAI')
+
+    server = current_app.config['SERVER_NAME']
+    if not server.startswith('http://'):
+        server = 'http://{}'.format(server)
+    ref_url = '{}/api/authors/{}'.format(server, author_recid)
+    builder.set_record({'$ref': ref_url})
```
jacquerie: This is reimplementing `get_record_ref`: https://github.com/inspirehep/inspire-dojson/blob/dd6fb27d14f5b024d53d713a1340bf815a5ab09a/inspire_dojson/utils/__init__.py#L120-L128

inspirehep/modules/migrator/tasks.py
kaplun: Yep. Also the comment is wrong. There won't be a PIDDoesNotExistError error in case the author signature can't be found.

inspirehep/modules/migrator/tasks.py
```diff
@@ -356,3 +359,72 @@ def migrate_and_insert_record(raw_record):
...
+
+    Args:
+        inspire_bai (string): author profile identifier as BAI
+        author_recid (int): Inspire author record identifier
+        lit_recid (string): Inspire literature record identifier
+        signature (string): name as written on the paper
+        flag (string): record can be ignored if flag == -2 (claim reject)
+    """
+    if flag == -2:
+        return
```
kaplun: Can you add here a FIXME to say that at some point we might be able to store this info?

michamos:
As we are not going to sync from labs to legacy, we don't need to synchronize claims in the other direction.

PR #3005, opened 2017-11-28
workflows: add manual_merge workflow

jacquerie:
## Related Issue
Supersedes #2904.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/manual_merging.py
+++ b/inspirehep/modules/workflows/tasks/manual_merging.py
@@ -1,0 +1,171 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Tasks related to manual merging."""
+
+from __future__ import absolute_import, division, print_function
+
+import json
+
+from invenio_db import db
+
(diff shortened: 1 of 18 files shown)


inspirehep/modules/workflows/tasks/manual_merging.py
```diff
@@ -0,0 +1,157 @@
...
+    head.clear()
+    head.update(obj.data)    # head's content will be replaced by merged
+    update.merge(head)       # update's uuid will point to head's uuid
+    update.delete()          # mark update record as deleted
+
+    # add schema contents to refer deleted record to the merged one
+    update['new_record'] = get_record_ref(
+        head['control_number'],
+        endpoint='record'
+    )
```
jacquerie: This is not right, as it would generate a JSON reference like
```JSON
{"$ref": "http://localhost:5000/api/record/4328"}
```
which isn't correct, because we don't redirect routes like `/api/record/<int:control_number>`, just `/record/<int:control_number>`.

jacquerie: As other parts of the code assume that we're merging Literature records (most notably, we don't have configuration otherwise in `inspire-json-merger`) it's ok to hardcode it here too.

inspirehep/modules/workflows/tasks/merging.py
```diff
@@ -0,0 +1,50 @@
...
+# or submit itself to any jurisdiction.
+
+"""Tasks related to record merging."""
+
+from __future__ import absolute_import, division, print_function
+
+from inspirehep.modules.workflows.models import WorkflowsRecordSources
+
+
+def get_head_source(head_uuid):
```
jacquerie: This function is in a dedicated `tasks` module, but isn't a task... It should probably be moved to `utils`.

tests/integration/tasks/test_merging.py
```diff
@@ -0,0 +1,57 @@
...
+        "control_number": 1
+    }
+
+
+def test_get_head_source_no_rec_on_the_db_gives_none(app):
+    assert get_head_source('7753a30b-1111-2222-3333-d5020069b3ab') is None
+
+
+def test_get_head_source_return_arxiv_when_one_arxive_source_present(app, simple_record):
+    rec = record_insert_or_replace(simple_record)
```
jacquerie: This test doesn't clean after itself: it creates this record, but then it doesn't delete it.

tests/integration/workflows/test_manual_merge_workflow.py
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+
+def test_manual_merge_existing_records(workflow_app):
+
+    json_head = fake_record('This is the HEAD', 1)
+    json_update = fake_record('While this is the update', 2)
+
+    head = record_insert_or_replace(json_head)
+    update = record_insert_or_replace(json_update)
```
jacquerie: Same observation as https://github.com/inspirehep/inspire-next/pull/3005/files#r153618433.

PR #3007, opened 2017-11-29
global: bump inspire-schemas to version ~56.0

michamos:
## Checklist:


inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -153,7 +153,11 @@ def _is_arxiv_url(url):
         else form_fields.get('language')
     builder.add_language(language=language)
 
-    builder.add_title_translation(title=form_fields.get('title_translation'))
+    if form_fields.get('title_translation'):
+        builder.add_title_translation(
+            title=form_fields['title_translation'],
+            language='en',
+        )
```
jacquerie: `formdata_to_model` here was calling `add_title_translation` without `language`, which is required by the schema, so it should be a required parameter of the builder method: https://github.com/inspirehep/inspire-schemas/pull/288

inspirehep/wsgi_with_coverage.py
```diff
@@ -89,7 +89,7 @@ def mock_literature_validate():
 
 @app.route('/submit/author/validate', endpoint='inspirehep_authors.validate', methods=['POST'])
 def mock_author_validate():
-    if request.json.get('orcid') == '1111-1111-1111-1111':
+    if request.json.get('orcid') == '0000-0002-1825-0097':
```
jacquerie: ORCiD has a specific profile for tests: https://orcid.org/0000-0002-1825-0097. We might as well use it, instead of inventing our own data.

tests/unit/records/test_records_receivers.py
```diff
@@ -769,108 +769,6 @@ def test_populate_title_suggest_with_all_inputs():
...
-        'payload': {
-            'full_title': ''
-        }
-    }
-
-    result = record['title_suggest']
-
-    assert expected == result
-
-
```
jacquerie: We can safely remove these tests, as they protect against something that this schema version disallows.

PR #3012, opened 2017-11-30
workflows: handle duplicate plot names in arXiv tarball

jmartinm:
* In some cases, the same plot name can be found in different
  directories of an arXiv tarball. This change ensures that all
  plots get attached with unique names.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/arxiv.py
+++ b/inspirehep/modules/workflows/tasks/arxiv.py
@@ -147,12 +147,16 @@ def arxiv_plot_extract(obj, eng):
                 return
 
             if 'figures' in obj.data:
+                for figure in obj.data['figures']:
+                    del obj.files[figure['key']]
                 del obj.data['figures']
 
             lb = LiteratureBuilder(source='arxiv', record=obj.data)
-            for plot in plots:
+            for index, plot in enumerate(plots):
                 plot_name = os.path.basename(plot.get('url'))
                 key = plot_name
+                if plot_name in obj.files.keys:
+                    key = '{number}_{name}'.format(number=index, name=plot_name)
                 with open(plot.get('url')) as plot_file:
                     obj.files[key] = plot_file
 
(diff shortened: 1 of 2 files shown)


jmartinm:
@david-caro now this should handle both the case of rerunning the task and having duplicate names in the tarball. Is deleting files like [this](https://github.com/inspirehep/inspire-next/pull/3012/files#diff-d5b37b2f12fac54ce9b26fffdb95b1ceR151) ok?

david-caro:
@jmartinm that should be ok yep:
    https://github.com/inveniosoftware/invenio-records-files/blob/master/invenio_records_files/api.py#L185

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -147,12 +147,16 @@ def arxiv_plot_extract(obj, eng):
                 return
 
             if 'figures' in obj.data:
+                for figure in obj.data['figures']:
+                    del obj.files[figure['key']]
                 del obj.data['figures']
 
             lb = LiteratureBuilder(source='arxiv', record=obj.data)
-            for plot in plots:
+            for index, plot in enumerate(plots):
```
david-caro: This is still a bit confusing, but it works now that all the figures get deleted beforehand.

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -391,6 +391,50 @@ def test_arxiv_plot_extract_is_safe_to_rerun(mock_os):
...
+            }),
+        }),
+    })
+    assert validate(data['arxiv_eprints'], subschema) is None
+
+    obj = MockObj(data, extra_data, files=files)
+    eng = MockEng()
+
+    try:
+        temporary_dir = mkdtemp()
```
david-caro: You can use pytests tmpdir: https://docs.pytest.org/en/latest/tmpdir.html

jmartinm: That fixture is a bit weird. It creates in the temporary directory a folder structure like `pytest-of-jmartinm/pytest-1/test_arxiv_plot_extract_handle0` and then it fails with  `EEXIST: [File exists]:` when creating a file...

PR #3008, opened 2017-11-29
records: add MARCXML serializer

kaplun:
## Checklist:
- [x] I have all the information that I need (if not, move to `RFC` and look for it).
- [ ] I linked the related issue(s) in the corresponding commit logs.
- [x] I wrote [good commit log messages](https://github.com/torvalds/subsurface-for-dirk/blob/5f15ad5a86ada3c5e574041a5f9d85235322dabb/README#L92-L119).
- [x] My code follows the code style of this project.
- [x] I've added any new docs if API/utils methods were added.
- [ ] I have updated the existing documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #3013, opened 2017-11-30
records: push to legacy

kaplun:

* Registers a receiver to monitor records being updated in order
  to ping legacy (through Redis) to come and fetch updated MARCXML
  information. (After #3008 is integrated)

### DIFF
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -46,6 +46,7 @@ from inspire_utils.helpers import force_list
 from inspire_utils.name import generate_name_variations
 from inspire_utils.record import get_value
 from inspirehep.modules.authors.utils import phonetic_blocks
+from .utils import get_endpoint_from_record
 
 
 #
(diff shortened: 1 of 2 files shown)


inspirehep/modules/records/receivers.py
```diff
@@ -103,6 +103,30 @@ def assign_uuid(sender, record, *args, **kwargs):
...
+    def _get_redis_connection():
+        if redis_connection:
+            return redis_connection
+        redis_url = current_app.config.get('CACHE_REDIS_URL')
+        return StrictRedis.from_url(redis_url)
+
+    for model_instance, dummy in changes:
+        if isinstance(model_instance, RecordMetadata):
+            recid = model_instance.json.get('control_number')
+            if recid:
```
kaplun: Here we might also want to check if the schema corresponds to HEP or whatever configurable list of records we want to push back to legacy.

inspirehep/modules/records/receivers.py
```diff
@@ -103,6 +103,30 @@ def assign_uuid(sender, record, *args, **kwargs):
...
+
+    def _get_redis_connection():
+        if redis_connection:
+            return redis_connection
+        redis_url = current_app.config.get('CACHE_REDIS_URL')
+        return StrictRedis.from_url(redis_url)
+
+    for model_instance, dummy in changes:
+        if isinstance(model_instance, RecordMetadata):
+            recid = model_instance.json.get('control_number')
```
kaplun: Note: this assumes that the recid is already there. Such recid needs to be registered on legacy first though. It is the responsibility of the workflow to correctly register it first.

inspirehep/modules/records/receivers.py
```diff
@@ -103,6 +103,30 @@ def assign_uuid(sender, record, *args, **kwargs):
...
+    sharing this information via Redis. Then Legacy will fetch the actual
+    record through subsequent HTTP requests.
+    """
+    redis_connection = None
+
+    def _get_redis_connection():
+        if redis_connection:
+            return redis_connection
+        redis_url = current_app.config.get('CACHE_REDIS_URL')
+        return StrictRedis.from_url(redis_url)
```
kaplun: This is a cheap way to cache the redis connection. If you have a better one, please do!

inspirehep/modules/records/receivers.py
```diff
@@ -103,6 +103,30 @@ def assign_uuid(sender, record, *args, **kwargs):
...
+
+    for model_instance, dummy in changes:
+        if isinstance(model_instance, RecordMetadata):
+            recid = model_instance.json.get('control_number')
+            if recid:
+                redis_connection = _get_redis_connection()
+                redis_connection.sadd('records_to_sync_into_legacy', recid)
```
kaplun: Actually: there is currently no API that given just the recid, it will give back a record. One also needs to know the document type.

E.g.: https://qa.inspirehep.net/api/record/1405870 currently returns `404` while https://qa.inspirehep.net/api/iterature/1405870 correctly works.

So either we make the 404 to actually be a redirect, or we send to legacy not only the recid but also literature.

For the time being legacy will assume that the the set element in redis will contain `1234,literature` (i.e. comma-separated integer and document type).

michamos:
we are not going to push labs records to legacy (only workflows), so closing this.

PR #3023, opened 2017-12-04
global: bump inspire-dojson to version ~57.0

jacquerie:
## Checklist:

### DIFF
--- a/inspirehep/modules/migrator/cli.py
+++ b/inspirehep/modules/migrator/cli.py
@@ -24,31 +24,31 @@
 
 from __future__ import absolute_import, division, print_function
 
-import os
 import csv
-import traceback
+import os
 import sys
-
+import traceback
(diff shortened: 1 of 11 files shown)


inspirehep/modules/migrator/cli.py
```diff
@@ -140,13 +140,13 @@ def get_collection(marc_record):
     results_length = results.count()
     with click.progressbar(results.yield_per(100), length=results_length) as bar:
         for obj in bar:
-            marc_record = marc_create_record(obj.marcxml, keep_singletons=False)
+            marc_record = create_record(obj.marcxml, keep_singletons=False)
```
jacquerie: This is the only remaining use of upstream's `create_record`. We could avoid it, but since this function is going to die there's no point in making it nice.

PR #3034, opened 2017-12-06
forms: collapse links when importing from arXiv

### LINKED ISSUE #1997: Collapsing links when importing from arXiv
When a user suggest some content from arXiv with the literature suggestion form, we really do not need the "Links" tab to be displayed because we already have a URL for the PDF (the arXiv one) and a URL for the abstract (the arXiv one).

So we could de-clutter the screen by collapsing that panel by default in case of import from arXiv.

Dinika:
## Related Issue:
Closes #1997 

## Notes:
There appears to have been an error while publishing the image for
rabbitmq 3.7.0, so I added a commit to pin the previous version until
a fix is released.

## Checklist:

### DIFF
--- a/inspirehep/modules/literaturesuggest/static/js/literaturesuggest/literature_submission_form.js
+++ b/inspirehep/modules/literaturesuggest/static/js/literaturesuggest/literature_submission_form.js
@@ -528,8 +528,10 @@ define(function(require, exports, module) {
       this.$formWrapper.show('blind', 1000);
       $actionBar.show('blind', 1000);
 
-      // run checkDepositionType() only when showing the rest of the form
+      // run checkDepositionType() and collapseLinksPanel()
+      // only when showing the rest of the form
       this.checkDepositionType();
+      this.collapseLinksPanel();
     },
 
     /**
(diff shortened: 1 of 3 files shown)


jmartinm:
Very small thing, but when collapsing in that way, the arrow does not show properly. It shows pointing down instead of:

screen shot 2017-12-07 at 4 37 50 pm

Dinika:
@jmartinm I have updated this PR. I deal with the issue that you mentioned by adding the `collapsed` class to the link the header of the Links panel. If there's a better approach to deal with this problem, please let me know.

inspirehep/modules/literaturesuggest/static/js/literaturesuggest/literature_submission_form.js
```diff
@@ -612,6 +614,13 @@ define(function(require, exports, module) {
       });
     },
 
+    // Collapse the links panel by default if the record is imported from arXiv
+    collapseLinksPanel: function collapseLinksPanel() {
+      if($.trim(this.$arxiv_id_field.val().length) > 0) {
+        $('#collapse-3').prev().find('a.panel-toggle').click();
```
jacquerie: I checked manually that this works, but code like this is **really** brittle... Unfortunately, we will have to live with this until we will have time to refactor the forms.

PR #3046, opened 2017-12-11
 records: rewrite BibTeX serializer using pybtex

jacquerie:
## Checklist:


inspirehep/modules/hal/utils.py
```diff
@@ -458,16 +491,7 @@ def get_page_artid(record):
...
-    if 'artid' in publication_info:
-        artid = publication_info['artid']
-        return artid
-    elif 'page_start' in publication_info and 'page_end' in publication_info:
-        page_start = publication_info['page_start']
-        page_end = publication_info['page_end']
-        return '{}-{}'.format(page_start, page_end)
-
-    return ''
+    return get_page_artid_for_publication_info(publication_info, separator)
```
jacquerie: I dislike all the changes in this file, as they:
1. Break the consistency of the previous API, which previously had only functions of type `record -> list(str)` or `record -> str`.
1. Modify code that was tested on 4000+ records during the HAL exports.

So please don't depend anywhere else on these utils, as they are going to be refactored ASAP!

inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -0,0 +1,376 @@
...
+
+from inspire_utils.date import PartialDate
+from inspire_utils.record import get_value
+from inspirehep.modules.hal.utils import (
+    get_page_artid_for_publication_info,
+    get_conference_record,
+    get_conference_title,
+)
+
+from .config import COMMON_FIELDS_FOR_ENTRIES, FIELDS_FOR_ENTRY_TYPE
```
jacquerie: Importing from the configuration means that the only way of overriding is modifying the code, which is highly undesiderable. A better way is to make this configuration part of the global configuration, and then request che configuration from the `current_app`.

PR #3054, opened 2017-12-13
WIP: tools: new authortool

ksachs:
* allows digits, letters and symbols as affiliation IDs

* authortool.py separated off - is not using invenio

## Description
This tool works with digits, letters or symbols (e.g. dagger) too.
Some errors replaced by warnings, so cataloger has output to fix.
Requires empty line between authors and affiliations.

Almost all tests have to be re-written, some have to be added.

## Questions

- Separate tests for 
  create_authors (testing conversion and logic)
  authorlist (conversion to inspire record)
  ?
  
- How can I test on the warning messages?
  
- Can I run the tests without the full machinery? I don't have inspire-next installed. 

## Motivation and Context

https://labs.inspirehep.net/tools/authorlist
works only with digits as affiliation-IDs. 

Letters are common too, symbols are rare but nasty to work with.

## Checklist:


michamos:
@ksachs To run the tests, you just need the dependencies of the code you are testing plus the tests (so inspire-schemas and refextract), and you need `pytest` which actually runs the test.
Then you should be able to do
```
$ py.test tests/unit/tools/test_tools_utils.py
```

Alternatively you can just let Travis run the tests for you. This happens automatically for every new commit added to the pull request. You can see the results of the build by clicking on "Details" on the bottom of the page:

screenshot-2017-12-13 inspirehep inspire-next

Except for style errors, the tests fail because you are doing something wrong with unicode. What are you trying to achieve?

inspirehep/modules/tools/utils.py
```diff
@@ -22,134 +22,30 @@
...
+    text = text.replace('\r', '')  # Input from the form contains unwanted \r's
 
-def authorlist_without_affiliations(text):
-    """Return a record containing the authors."""
-    builder = LiteratureBuilder()
-    authors = text.replace(' and ', ', ').split(', ')
-    for author in authors:
-        builder.add_author(builder.make_author(author))
+    for fullname, author_affs in create_authors(text):
+        if author_affs:
```
michamos: The `if` is not needed, the builder can deal with empty arguments.

inspirehep/modules/tools/authortool.py
```diff
@@ -0,0 +1,311 @@
...
+    """
+    Split text in (useful) blocks, sepatated by empty lines.
+    1 block: no affiliations
+    2 blocks: authors and affiliations
+    more blocks: authors grouped by affiliation (not implemented yet)
+
+    Return list of (author_fullname, [author_affiliations])
+    """
+
+    text = unicode(text, 'utf-8')
```
michamos: This is making the tests fail. What are you trying to do?

ksachs: Test-case:
† 
CERN
*
DESY

I read the lines and get (print len(line), line)
3 †
4 CERN
1 *
4 DESY


after calling unicode I get
1 †
4 CERN
1 *
4 DESY

and I need  † to match regex '.'

inspirehep/modules/tools/utils.py
ksachs: great - thanks

inspirehep/modules/tools/authortool.py
ksachs: I can move unicode(text, 'utf-8') to the main code where I read the file, hoping it works with input from the web-form.

michamos: This is happening because at the beginning you are dealing with a series of bytes containing UTF-8 encoded unicode. `len` is counting the number of bytes, but UTF-8 is a variable-length encoding, meaning that one character may be encoded either in one byte (ASCII) or several (the `†`). On the other hand calling `len` on unicode strings counts the number of characters.

What you need to do is call `decode` at the beginning, if `text` contains bytes and not unicode strings. So you should add `import six` (for python 3 compatibility) at the top, and add at the beginning of `parse_authors`:
```
if isinstance(text, six.binary_type):
    text = text.decode('utf-8')
```

michamos:
A significant part of your code seems to be handling proper formatting of author names. You don't need to do this, as the builder is already taking care of it:
```
In [1]: from inspire_schemas.api import LiteratureBuilder

In [2]: l = LiteratureBuilder()

In [3]: l.make_author('J Smith')
Out[3]: {'full_name': u'Smith, J.'}
```

inspirehep/modules/tools/authortool.py
ksachs: OK. I'll ask Florian to install six.

ksachs:
The significant part is guessing if there are left-over IDs and raise warnings.
But I can get rid of the 5 lines of name formatting.

inspirehep/modules/tools/authortool.py
```diff
@@ -0,0 +1,311 @@
...
+                    fullname = '%s, %s' % \
+                        (author_names[-1], ' '.join(author_names[:-1]))
+                else:
+                    fullname = author_names[0]
+                fullname = fullname.replace('. ', '.')
+                author_names = []
+            if word in aff_keys:
+                author_affs.append(affiliations[word])
+            elif not word == ',':
+                warnings.warn(
```
michamos: you should probably be raising exceptions instead of warnings. The former will be caught and shown to the user, the latter might end up in some webserver logs at best.

ksachs: Then the user has nothing.
It's better to have something and fix the mistakes manually.

I was hoping it's possible to display the warnings the same way errors are now returned + the result.

ksachs: Otherwise it might be better to ignore the warnings.

michamos: I asked @jacquerie and there doesn't seem to be a way to display warnings. Exceptions are shown to the user, who can correct them and try again. I would say that's better than silently going on and maybe lose some authors.

ksachs: too bad ...
If I parse something like

    P.Carniti a, b L. di Caprio b a  M. Faverzani b
    a  A.Giachero bb, a 
    
    a
    INFN
    b
    Universita di Roma
    bb
    Universita di Milano-Bicocca
    
It works OK, but I would like to alert the cataloger to have a closer look whether 'di' is part of a name or not. Throwing an error without the result is not an option. What if one author really has no affiliation?

You know more  about the integration.
What if I write a new function that returns both the result-record and some text?
The record can be used to pre-fill the editor. And in the current web-interface you would first print the text followed by the record. Is that feasible?

jacquerie: > What if I write a new function that returns both the result-record and some text?

Yes, you have to return something like that. For example you can return an object like
```JSON
{
    "authors": [
        {
            "affiliations": [
                "INFN",
                "Universita di Roma"
            ],
            "full_name": "Carniti, P."
        },
        ...
    ],
    "warnings": [
        {"message": "Author number 2 contains critical word 'di'. Please check if it's actually part of a name."},
        ...
    ]
}
```

inspirehep/modules/tools/authortool.py
```diff
@@ -0,0 +1,297 @@
...
+        warning += text
+    elif len(text_blocks) == 2:
+        affiliations = parse_affiliations(text_blocks[1])
+        authors, text = parse_authors(text_blocks[0], affiliations)
+        warning += text
+    else:
+#        authors = parse_blocks(text_blocks)
+        authors = []
+
+    return  authors, warning
```
jacquerie: Minor thing: dict-like returns like the one I proposed in https://github.com/inspirehep/inspire-next/pull/3054#discussion_r156929280 are more extensible than returning tuples, because you can always add keys to a dictionary without breaking a client, and that's not the case with sequence types.

ksachs:
Do I have to add docstrings and newlines and fix the invalid function names in the test functions to make Travis happy?

Do we want to keep the old functions? Otherwise I could just delete those tests.

michamos:
* It's not complaining about docstrings, tests should not have docstrings as the test name can be made explicit enough. They are some small style things about consistent use of whitespace mostly.

* You can replace the old `authorlist` by your new function (now called `authortool`), like you were doing at first.
The old unit tests will fail, but you should check the failures and modify the tests/the implementation so that they pass with the new implementation. In this way, we are certain that all cases that were handled correctly in the past are still handled correctly by your new implementation. For the integration tests, you can let them fail as they test that the author tool is correctly integrated into the rest of inspire, and that will need to change anyway because of the warning handling. I will take care of that part.

ksachs:
I will rename utils.authortool to authorlist as you suggested and delete the old code. There will be only the interface left over that does the integration between authortool.py and inspire.
I would like to keep the inspire-independent authortool.py and its tests which I can run at DESY.
Shall I rename it to authorlist.py?

I will delete almost all tests in test_tools_utils.py, keeping just 2 or 3 to make sure integration with the builder works and some UTF-8 problems that are solved by refextract.documents.pdf.replace_undesirable_characters

All test cases are copied to test_tools_authortool.py, modified where necessary and they pass.
I'll add some more tests for the new features.

ksachs:
in case of no warning do you prefer 
<code>'warning':''</code> or no warning key as return from utils.authorlist?

michamos:
I would say calling everything authorlist is more consistent, and no `warning` key when there is no warning too.

ksachs:
the only thing missing might be to add the warnings to the output of the web-form. 
@michamos  can you do that?

Issue #3048, opened 2017-12-11
Parser: requirements for title keyword


StellaCh:
Searching by title in INSPIRE is about discovering literature. We anticipate that Labs will return more results than legacy and that's ok. However, these results need to be prioritized. For the first version, we will implement a simple solution similar to legacy and we'll reiterate based on user feedback.

Example of queries (from user logs) that should work in the first version:

- [ ] ***find t Search for standard model Higgs boson production in association with a W boson using a matrix element technique*** 
It works!
- [ ] ***find t A First Course in String Theory***
Legacy returns only 1 result, whereas Labs doesn't return the right paper on the 1st page. The paper should appear first in Labs too.
- [ ] ***f t N-body and t separable***
In this case, the results should return any title with "N-body" AND "separable" in the first page. Legacy returns only 3 results, whereas in Labs, the first results are either with N-body or with separable.
- [ ] ***f t qq production LHC***
As a user, I expect results that have qq **and** production **and** LHC in the title, so this: http://pcrcsis09.cern.ch:5000/search?page=1&size=25&cc=literature&q=titles.title:qq%20AND%20titles.title:production%20AND%20titles.title:LHC 
- [ ] ***Find T g-2***
As a user, I expect to find "g-2" in the title, as in legacy
- [ ] ***find t uv/ir mixing***
Same as above, the search should return results with "uv/ir mixing" in the title
- [ ] ***find t /wigner function/***
The parser should support / /, as in legacy: http://inspirehep.net/info/hep/search-tips#exactphrase 
- [ ] ***t SU(2)***
SU(2) should be in the title

## Future improvements
- [ ] stemming
- [ ] ***t feynrules 2*** (the right result is in the first page but is 7th - should be 1st)

StellaCh:
@jacquerie @michamos feel free to add any comments/extra requirements

chris-asl:
@StellaCh Regarding the 2nd one, I can see the [correct paper as the first result](https://labs.inspirehep.net/search?page=1&size=25&q=%20find%20t%20A%20First%20Course%20in%20String%20Theory).

StellaCh:
[...]

PR #3056, opened 2017-12-13
mappings: hep-titles improvements

chris-asl:
* Add ``titles.full_title.search`` which uses 2-shingles so that we
  have improved relevance. e.g. a search for 'n-body' should not match
  'Exact n-d scattering...three-body...'.
* In ``title`` analyzer, enable stemming (add ``porter_stem`` in token
  filters).

## Related Issue
https://github.com/inspirehep/inspire-next/issues/3048

## Motivation and Context
The second change is for being able to get results for ``separation`` when searching for ``separable``, as in the related issue.

Specifically, both of the chages are directed towards this query `f t N-body and t separable`, but in general improve our searching, as at least the 2-shingles (i.e. bigrams) are suggested by [ElasticSearch for finding associated words](https://www.elastic.co/guide/en/elasticsearch/guide/current/shingles.html).

## Checklist:

### DIFF
--- a/inspirehep/modules/records/mappings/records/hep.json
+++ b/inspirehep/modules/records/mappings/records/hep.json
@@ -1036,6 +1036,16 @@
                     "properties": {
                         "full_title": {
                             "analyzer": "title",
+                            "fields": {
+                                "search": {
+                                    "analyzer": "title_search_analyzer",
+                                    "type": "string"
+                                },
+                                "raw": {
+                                    "index": "not_analyzed",
+                                    "type": "string"
+                                }
+                            },
                             "type": "string"
                         },
                         "source": {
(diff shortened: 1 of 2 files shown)


jacquerie:
> The second change is for being able to get results for separation when searching for separable, as in the related issue.

This is the problem: stemming is nice for dealing with normal, everyday, text, but is prone to generating false positives when dealing with specialized, technical, text.

For example: a [separated space](https://en.wikipedia.org/wiki/Hausdorff_space) and a [separable space](https://en.wikipedia.org/wiki/Separable_space) are very different concepts in topology, but they would both stem to "separ space" using the Porter stemmer. But if I am searching for theorems that apply to separable spaces, I wouldn't want to see results that apply to separated spaces!

Sadly the only real solution I know about is to build HUGE lists of synonyms (in the above example, "separated" and "Hausdorff"). That said, we can show this solution to users, but we should be aware of this pitfall, and be prepared to abandon stemming as soon as they report similar, irrelevant results.

michamos:
a different case where stemming is problematic: "electron detector" vs "electronic detector".

StellaCh:
Stemming requires a wider discussion, thus it won't be supported in the first version that we'll show to the users, but it's listed in the future improvements: https://github.com/inspirehep/inspire-next/issues/3048 
So, no need to have stemming in the first version.

chris-asl:
As stemming isn't good for our use-case and synonyms won't be included in the first version, this isn't needed anymore.

PR #3063, opened 2017-12-15
search: show the selected sort option in the select box

### LINKED ISSUE #3047: sort by display
## Current Behavior
The `Sort  by` results dropdown is broken on Labs. Although the results are updated, 'sort by' stays empty when you select an option from the dropdown

## Screenshots (if appropriate):

sort

Dinika:
Signed-off-by: Dinika <dinika.saxena@cern.ch>

## Description
Along with [#56](https://github.com/inspirehep/inspirehep-search-js/pull/56/files) in inspirehep-search-js this pr fixes #3047

## Related Issue
closes #3047 

## Motivation and Context
[Invenio-search](https://github.com/inveniosoftware/invenio-search-js/blob/master/src/invenio-search-js/directives/invenioSearchSelectBox.js#L96) sets the value of the selected option as the current value without the dash prepending it. This results in modified option.value. However, since Angular relies on the original value (with the dash), the title of the selected option is not rendered in the select box.
To evade the mis-match between the previous and the new option.value, I decided to not add the dash to the sort options in the first place. 

## Checklist:

### DIFF
--- a/inspirehep/modules/search/views.py
+++ b/inspirehep/modules/search/views.py
@@ -123,7 +123,7 @@ def sorted_options(sort_options):
     return [
         dict(
             title=v['title'],
-            value=('-{0}'.format(k)
+            value=('{0}'.format(k)
                    if v.get('default_order', 'asc') == 'desc' else k),
         )
         for k, v in


Dinika:
@jacquerie do you think removing the dash might have some unwanted side effects?

chris-asl:
So, incidentally I stumbled upon this issue.
@Dinika @jacquerie 
The dash is used for doing a `desc` sort on the fields. 
Which doesn't make much sense since we have the `"default_order": 'desc'` option set [here](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/config.py#L1075).

Actually the `-` is used in the [URL](https://labs.inspirehep.net/search?page=1&size=25&sort=-mostcited) and it's working fine. If I remove the dash I am getting the ascending sorting.
It seems as if we're duplicating the logic of sorting with both the `-` sign and the `default_order` setting.

Related to that:
Another thing that I couldn't find, is where `invenio-records-rest` is getting/using the options for [`fields` parameter](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/config.py#L1074). 
The only place in `invenio-records-rest` that's using this config option is [here](https://github.com/inveniosoftware/invenio-records-rest/blob/f223d86ff285bb4edcc50da01e8557ce68a4a8c5/invenio_records_rest/views.py#L285-L286) - which uses only `title` and `default_order`.

chris-asl:
[Here's](https://github.com/inveniosoftware/invenio-records-rest/blob/master/invenio_records_rest/sorter.py#L149) where there `fields` parameter is actually used!

PR #3064, opened 2017-12-18
records: populate full_name_unicode_normalized

iulianav:
## Description

Add `author.full_name_normalized` not analysed field to hep.json. This field is populated by the receivers with the `NFKC` unicode normalized and lowercased `full_name` value. This value is already normalized by `inspire_utils.normalize_name`.

## Related Issue

https://github.com/inspirehep/inspire-query-parser/issues/63

## Related PR
https://github.com/inspirehep/inspire-query-parser/pull/64

## Motivation and Context

This field is needed in order to  support `exact-author` queries that perform
diacritic expansion and lowercase normalization.

## Checklist:

### DIFF
--- a/inspirehep/modules/records/mappings/records/hep.json
+++ b/inspirehep/modules/records/mappings/records/hep.json
@@ -211,6 +211,10 @@
                             },
                             "type": "string"
                         },
+                        "full_name_unicode_normalized": {
+                            "index": "not_analyzed",
+                            "type": "string"
+                        },
                         "ids": {
                             "properties": {
                                 "schema": {
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -25,6 +25,7 @@
 from __future__ import absolute_import, division, print_function
 
 import uuid
+from unicodedata import normalize
 from itertools import chain
 
 import six
@@ -141,6 +142,7 @@ def enhance_after_index(sender, json, *args, **kwargs):
     populate_abstract_source_suggest(sender, json, *args, **kwargs)
     populate_affiliation_suggest(sender, json, *args, **kwargs)
     populate_author_count(sender, json, *args, **kwargs)
+    populate_authors_full_name_unicode_normalized(sender, json, *args, **kwargs)
     populate_earliest_date(sender, json, *args, **kwargs)
     populate_inspire_document_type(sender, json, *args, **kwargs)
     populate_name_variations(sender, json, *args, **kwargs)
@@ -412,3 +414,17 @@ def populate_author_count(sender, json, *args, **kwargs):
         if 'supervisor' not in author.get('inspire_roles', [])
     ]
     json['author_count'] = len(authors_excluding_supervisors)
+
+
+def populate_authors_full_name_unicode_normalized(sender, json, *args, **kwargs):
+    """Populate the ``authors.full_name_normalized`` field of Literature records."""
+    if 'hep.json' not in json.get('$schema'):
+        return
+
+    authors = json.get('authors', [])
+
+    for index, author in enumerate(authors):
+        full_name = six.text_type(author['full_name'])
+        json['authors'][index].update({
+            'full_name_unicode_normalized': normalize('NFKC', full_name).lower()
+        })
--- a/tests/unit/records/test_records_receivers.py
+++ b/tests/unit/records/test_records_receivers.py
@@ -38,6 +38,7 @@ from inspirehep.modules.records.receivers import (
     populate_recid_from_ref,
     populate_title_suggest,
     populate_author_count,
+    populate_authors_full_name_unicode_normalized,
 )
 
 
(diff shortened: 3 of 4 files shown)


inspirehep/modules/records/receivers.py
```diff
@@ -412,3 +414,21 @@ def populate_author_count(sender, json, *args, **kwargs):
         if 'supervisor' not in author.get('inspire_roles', [])
     ]
     json['author_count'] = len(authors_excluding_supervisors)
+
+
+def populate_authors_full_name_normalized(sender, json, *args, **kwargs):
+    """Populate the ``authors.full_name.raw`` field of Literature records."""
```
chris-asl: Should be `full_name_normalized`

inspirehep/modules/records/receivers.py
```diff
@@ -412,3 +414,21 @@ def populate_author_count(sender, json, *args, **kwargs):
...
+    if 'hep.json' not in json.get('$schema'):
+        return
+
+    authors = json.get('authors', [])
+
+    for index, author in enumerate(authors):
+        full_name = author.get('full_name')
+        if full_name:
+            if not isinstance(full_name, unicode):
```
chris-asl: This is not Python 3 compliant, it's better to cover these cases for when we make the change.  
Look into `six` library and especially their `text_type`.

inspirehep/modules/records/receivers.py
```diff
@@ -412,3 +414,21 @@ def populate_author_count(sender, json, *args, **kwargs):
...
+                unicode_full_name = unicode(full_name, 'utf-8')
```
chris-asl: Same goes for this, Python3 doesn't know the name `unicode`. 
Every `str`ing in Python 3 is converted into unicode code points.

tests/unit/records/test_records_receivers.py
```diff
@@ -1066,3 +1067,44 @@ def test_populate_author_count_does_nothing_if_record_is_not_literature():
...
+        'authors': [
+            {
+                'full_name': 'Vurës, John',
+            },
+            {
+                'full_name': 'Vures, John',
+            },
+        ],
+    }
+    assert validate(record['authors'], subschema) is None
```
chris-asl: IIRC @jacquerie told me that these assertions are not needed anymore in receiver tests.

jacquerie: No, those are good assertions to have, as they protect the tested code against changes in the schema. What were bad were the assertions on the `$schema` property, which I removed in https://github.com/inspirehep/inspire-next/commit/08231012d665677c84c7e4ef4a9e34dc55feb15d#diff-d73f938cbaa6c2c71bc00274ac594f14.

inspirehep/modules/records/mappings/records/hep.json
```diff
@@ -211,6 +211,10 @@
                             },
                             "type": "string"
                         },
+                        "full_name_normalized": {
+                            "index": "not_analyzed",
+                            "type": "string"
+                        },
```
jacquerie: Why can't this be a field of `full_name`, instead of another key?

inspirehep/modules/records/receivers.py
jacquerie: At this point in the code we shouldn't have to deal with bytes; everything should already be decoded (Unicode sandwich!). You had to add these two lines because your tests are using bytes, but this doesn't happen in real data.

inspirehep/modules/records/mappings/records/hep.json
jacquerie: Ah, because this is not a variant of full_name, is a completely new thing that you are creating in Python...

tests/unit/records/test_records_receivers.py
```diff
@@ -1066,3 +1067,44 @@ def test_populate_author_count_does_nothing_if_record_is_not_literature():
...
+
+    populate_authors_full_name_normalized(None, record)
+
+    assert record['authors'][0]['full_name_normalized'] == u'vur\xebs, john'
+    assert record['authors'][1]['full_name_normalized'] == u'vures, john'
```
jacquerie: Because we are testing receivers in isolation from each other, you can write more in your assertions:
```python
expected = [
    {
        'full_name': 'foo',
        'full_name_normalized': 'foo',
    },
]
result = record['authors']

assert expected == result
```

bavaria95:
ci test please

bavaria95:
ci test please

iulianav:
Related to https://github.com/inspirehep/inspire-query-parser/pull/64

iulianav:
@jacquerie this needs merging for the exact-author query. I hope it is good enough now and ready to go.

inspirehep/modules/records/receivers.py
chris-asl: @jacquerie From the [build log](https://travis-ci.org/inspirehep/inspire-next/jobs/329388387#L1571) I guess when migrating the `full_name` is not in unicode. 
So, we're violating the "unicode sandwich" rule at our boundaries, which should be either in literature builder or dojson.  

I would say @iulianav can add the "check-unicode-barrier" there and at a later point fix it at our system boundaries.

tests/unit/records/test_records_receivers.py
```diff
@@ -1066,3 +1067,44 @@ def test_populate_author_count_does_nothing_if_record_is_not_literature():
     populate_author_count(None, record)
 
     assert 'author_count' not in record
+
+
+def test_populate_populate_authors_full_name_normalized():
```
chris-asl: `populate_populate` :P

chris-asl:
Integrated the change as discussed IRL with @michamos and @jacquerie and rebased with master.

PR #3066, opened 2017-12-19
workflows: store the record only when we have recid

david-caro:

Until we use labs as master, we need the recid to be minted on legacy
first, and then store the record locally.


david-caro:
Tested on QA 👍

david-caro:
This is currently blocking the deployment, merging (for speed and stability, we might want in the future to just mint the recid on legacy instead of robotuploading the whole record).

PR #3067, opened 2017-12-20
workflow: auto accept core and non core articles

ammirate:

Articles belonging to core and non core categories should be auto accepted by the workflow, and not stopped as it happens now.

## Checklist:


tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -148,9 +148,9 @@ def test_harvesting_arxiv_workflow_already_on_legacy(
         eng = WorkflowEngine.from_uuid(workflow_uuid)
         obj = eng.processed_objects[0]
 
-        assert obj.status == ObjectStatus.COMPLETED
-        assert 'already-ingested' in obj.extra_data
-        assert obj.extra_data['already-ingested']
+        assert obj.extra_data['approved'] is True
+        assert obj.extra_data['auto-approved'] is True
+        assert obj.data['core'] is True  # the record is core
```
david-caro: The comment here is also redundant.

tests/unit/workflows/test_workflows_tasks_matching.py
```diff
@@ -89,115 +91,40 @@ def test_is_being_harvested_on_legacy_returns_true_when_there_is_one_core_catego
             },
         ],
     }
-    assert validate(record['arxiv_eprints'], subschema) is None
+    assert validate(obj.data['arxiv_eprints'], subschema) is None
+    assert belongs_to_relevant_category(obj, eng) is True
+    assert get_coreness(obj.data) is True
```
david-caro: You shoud be patching the ARXIV_CATEGORIES here, as you are relying on the default value.

tests/unit/workflows/test_workflows_tasks_matching.py
```diff
@@ -89,115 +91,40 @@ def test_is_being_harvested_on_legacy_returns_true_when_there_is_one_core_catego
...
-    eng = MockEng()
-
-    extra_config = {
-        'ARXIV_CATEGORIES_ALREADY_HARVESTED_ON_LEGACY': [
-            'astro-ph.CO'
-        ]
-    }
+    assert validate(obj.data['arxiv_eprints'], subschema) is None
+    assert belongs_to_relevant_category(obj, eng) is True
+    assert get_coreness(obj.data) is False
```
david-caro: same here.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -107,30 +107,46 @@ def article_exists(obj, eng):
...
+    """
+    arxiv_categories = set(get_arxiv_categories(record))
+    relevant_categories = current_app.config.get('ARXIV_CATEGORIES', [])
+
+    if arxiv_categories & set(relevant_categories['core']):
         return True
+    elif arxiv_categories & set(relevant_categories['non-core']):
+        return False
+    else:
+        return None
```
david-caro: I don't like this `True`-`False`-`None` triplet :/, maybe we can use an enum instead, something like:
```python
from enum import Enum
class Coreness(Enum):
    core = True
    non_core = False
    non_relevant = None

# then return one of the properties in get_coreness
def get_coreness(...):
    ...
    if ...:
        coreness = Coreness.core
    elif ...:
        coreness = Coreness.non_core
    else:
        coreness = Coreness.non_relevant
    return coreness
```

It requires `enum34` on python 2 though, that we indirectly require, but I find it way more readable.

Then when setting it you just need:
```python
coreness = get_coreness(obj.data)
obj.extra_data['core'] = coreness.value
```

ammirate: Great, that was the feedback I wanted! I don't like it too.

david-caro:
Looks ok, I'll test + rebase and merge

Issue #3069, opened 2017-12-20
forms: store the formdata under `source_data` instead when creating the workflow.


david-caro:
In order to be able to restart a wokrflow reusing the same data that was used to start it the first time, we have to unify between submissions and harvests where to put that data, and that key will be from now on `sorce_data` under the workflow extra_data.

david-caro:
This is a bit more complex than I though, the `source_data` must have information about both, the extra_data that it would set in the workflow + the data, so it must have the structure:
```
"source_data": {
    "extra_data": {
        ...
    },
    "data": {
        ...
    }
}
```

david-caro:
It also requires adding a step/call in the restart to restore it.

PR #3081, opened 2017-12-21
submission: pass source_data to the initial workflow

david-caro:
This partially addresses #3069

### DIFF
--- a/inspirehep/modules/authors/views.py
+++ b/inspirehep/modules/authors/views.py
@@ -24,6 +24,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+import copy
 import os
 import re
 
@@ -279,6 +280,10 @@ def submitupdate():
     )
     workflow_object.data = data
     workflow_object.extra_data = extra_data
+    workflow_object.extra_data['source_data'] = {
+        'data': copy.deepcopy(data),
+        'extra_data': copy.deepcopy(extra_data),
+    }
     workflow_object.save()
     db.session.commit()
 
@@ -314,6 +319,10 @@ def submitnew():
     )
     workflow_object.data = data
     workflow_object.extra_data = extra_data
+    workflow_object.extra_data['source_data'] = {
+        'data': copy.deepcopy(data),
+        'extra_data': copy.deepcopy(extra_data),
+    }
     workflow_object.save()
     db.session.commit()
 
@@ -372,7 +381,6 @@ def reviewhandler():
     visitor.visit(form)
 
     workflow_object = workflow_object_class.get(objectid)
-    workflow_object.data = formdata_to_model(workflow_object, visitor.data)
     data, _ = formdata_to_model(
         formdata=visitor.data,
         id_workflow=workflow_object.id,
@@ -381,7 +389,6 @@ def reviewhandler():
     workflow_object.data = data
     workflow_object.extra_data["approved"] = True
     workflow_object.extra_data["ticket"] = request.form.get('ticket') == "True"
-    workflow_object.extra_data['formdata'] = visitor.data
     workflow_object.save()
     db.session.commit()
 
--- a/inspirehep/modules/authors/views.py
+++ b/inspirehep/modules/authors/views.py
@@ -24,7 +24,6 @@
 
 from __future__ import absolute_import, division, print_function
 
-import copy
 import os
 import re
 
@@ -53,7 +52,7 @@ from inspirehep.modules.forms.form import DataExporter
 
 from .forms import AuthorUpdateForm
 from .permissions import holdingpen_author_permission
-from .tasks import formdata_to_model
+from .utils import formdata_to_model
 
 
 blueprint = Blueprint(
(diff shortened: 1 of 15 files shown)


david-caro:
ci test please

david-caro:
ci test please

inspirehep/modules/authors/views.py
```diff
@@ -362,10 +381,15 @@ def reviewhandler():
...
     workflow_object = workflow_object_class.get(objectid)
+    data, _ = formdata_to_model(
+        formdata=visitor.data,
+        id_workflow=workflow_object.id,
+        id_user=current_user.get_id(),
+    )
+    workflow_object.data = data
     workflow_object.extra_data["approved"] = True
     workflow_object.extra_data["ticket"] = request.form.get('ticket') == "True"
     workflow_object.extra_data['formdata'] = visitor.data
```
david-caro: I have to remove this

david-caro:
ci test please

david-caro:
@bavaria95 this one gave a false negative, can you check and fix the jobs if need be? Thanks!

david-caro:
ci test please

bavaria95:
ci test please

bavaria95:
ci test please

bavaria95:
ci test please

david-caro:
Tested locally, @michamos I decided to remove the form provided references from the tickets created for the curators, as it does not seem to be used much (clicked around between tickets, did not see any with them) and the references are better displayed in the workflow or the record itself.

michamos:
not sure I understand correctly. You removed the references from the RT ticket, but didn't change anything about references in the rest of the workflow?

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -22,203 +22,37 @@
...
-        orcid=form_fields.get('orcid'),
-        method='submitter'
+def _get_user_comment(workflow_object):
+    user_comment = next(
+        (
+            note.get('value')
+            for note in workflow_object.data.get('_private_notes', [])
+            if note.get('source') == 'submitter'
+        ),
+        '',
```
ammirate: You already defined this in inspirehep/modules/authors/tasks.py

inspirehep/modules/literaturesuggest/tasks.py
```diff
@@ -249,22 +83,20 @@ def curation_ticket_context(user, obj):
...
-    references = obj.extra_data.get('formdata').get('references')
-    user_comment = obj.extra_data.get('formdata', {}).get('extra_comments', '')
+    user_comment = _get_user_comment(obj)
 
     return dict(
         recid=recid,
         record_url=record_url,
         link_to_pdf=link_to_pdf,
         email=user.email,
-        references=references,
```
ammirate: Why remove this field?

david-caro: See the old comment https://github.com/inspirehep/inspire-next/pull/3081#issuecomment-353637476

The reason behind it is to be able to move `formdata_to_model` to a non-task function, and thus, simplify it's usage so we don't need any dummy mocks in the tests for it anymore.

If we can avoid having it as a task, we should :), even if we need it as a task, I'd leave the task as a very simple wrapper on top of the actual function that does the stuff.

david-caro: At some point we might want to create a 'builder' class, or just a class for the formdata, same as we do for the record, that would help using the formdata, and testing it too. So you can do stuff like:
```python
my_formdata = Formdata(field1='bla', field2='bla')
function_that_uses_formdata(Formdata)

# or
my_formdata = Formdata(invalid_field='dummy value')
# raises TypeError exception with 'got unexpected keyword argument invalid_field'
```

That way, every time we modify it we will notice if anyone that uses it is expecting it to have certain fields.

inspirehep/modules/literaturesuggest/tasks.py
david-caro: I think that when I started the PR they were different... looking to see if I messed up a copy-paste or something.

david-caro: yes, this looks like a messed up copy-paste.

And the goal here is to do the `formdata_to_model` only once at the start of the workflow, then save the results in the `source_data` so when we restart the workflow, there will be no need to rerun the `fromdata_to_model` again, decoupling the workflow from the form, and allowing a generic handling of them.
Otherwise the source truth for the workflow will not always be the `data` and `extra_data` under `source_data` but sometimes they would be generated by the other `formdata`.

david-caro: @ammirate So if I fix it, I end up with the same method in both places... but I'm not sure if it's worth putting it anywhere else.

It's very specific, so I'm reluctant on putting it in inspire-utils, adding it to inspirehep/modules/tools does not look right either :/, I would not put it in inspirehep/modules/__init__.py either.

Maybe under inspirehep/modules/forms/utils.py? wdyt?

david-caro: hmpf... it looks to me also out of context there :/

david-caro:
@michamos I finally ended up adding to the curator ticket all the 'submitter' private notes (that might be `'extra_comments', 'nonpublic_note', 'hidden_notes', 'conf_name', 'references'` given from the form).

david-caro:
This will retaken after refinement if needed, currently already has conflicts.

PR #3082, opened 2017-12-21
Cleanup the workflow object at the start

david-caro:
If there's the data with what the workflow was started on it, use it
to clean up any changes that were made on the previous run.

Closes 3071
Depends on #3081 (not really, but without it it's useless).

### DIFF
--- a/inspirehep/modules/authors/utils.py
+++ b/inspirehep/modules/authors/utils.py
@@ -25,11 +25,22 @@
 from __future__ import absolute_import, division, print_function
 
 import re
-
+import copy
+import datetime
 
 import numpy as np
+from flask import url_for
 from beard.utils.strings import asciify
 from beard.clustering import block_phonetic
+from sqlalchemy.orm.exc import NoResultFound
+
+from invenio_accounts.models import User
+from invenio_oauthclient.models import UserIdentity
+
+from inspire_dojson.utils import strip_empty_values
+from inspire_schemas.api import validate
+from inspirehep.modules.forms.utils import filter_empty_elements
+from .dojson.model import updateform
 
 
 _bai_parentheses_cleaner = \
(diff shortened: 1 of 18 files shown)


david-caro:
damn!!!!

david-caro:
depends on #3081

inspirehep/modules/authors/utils.py
```diff
@@ -126,3 +137,79 @@ def phonetic_blocks(full_names, phonetic_algorithm='nysiis'):
...
+        datetime=datetime.datetime.utcnow().isoformat(),
+        method="submitter",
+        orcid=orcid,
+        submission_number=str(id_workflow),
+        internal_uid=int(id_user),
+    )
+
+    data = strip_empty_values(data)
+
+    validate(data, 'authors')
```
iulianav: Maybe validation won't be needed here in the future if it will happen at the beginning of a workflow the same it is now happening for hep records. This way, even if the data is not valid, it will not be lost, but saved in an ERRORed workflow so as to make the problem visible and, why not, easy to fix.

inspirehep/modules/authors/utils.py
```diff
@@ -126,3 +137,79 @@ def phonetic_blocks(full_names, phonetic_algorithm='nysiis'):
...
+
+    if is_update is not None:
```
iulianav: Maybe this can also be done later in the workflow for consistency reasons using [this](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/tasks/actions.py#L66). Maybe doing something like [this](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/workflows/article.py#L436) and then [this](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/workflows/article.py#L160) .

inspirehep/modules/authors/utils.py
david-caro: Yep, though we don't have validation on the authors workflow yet, so it should probably still be there until we tackle those (in any case I think it should be low prio, as the data that the forms send should always be valid, as it's not something that is populated by external sources like the hepcrawl does).

inspirehep/modules/authors/utils.py
david-caro: You are 100% right, that should be a mark in the workflow. though I think it's out of the scope of this PR to refactor it.

We should tackle this when doing the refactor of the author workflow.

(btw. fyi this PR is rebased on top of #3081, this one only adds the cleanup at the end).

inspirehep/modules/authors/utils.py
jacquerie: > in any case I think it should be low prio, as the data that the forms send should always be valid

50 users in the past three months have been victim of this assumption: https://sentry.inspirehep.net/inspire-sentry/prod/issues/48/

david-caro:
Then the problem is way bigger than that, is that we don't do proper validation in the form :(

--

CERN - RCS-SIS
http://inspirehep.net

On Feb 9, 2018 7:38 AM, Jacopo Notarstefano <notifications@github.com> wrote:

@jacquerie commented on this pull request.

________________________________

In inspirehep/modules/authors/utils.py<https://github.com/inspirehep/inspire-next/pull/3082#discussion_r167150035>:

> +            method='orcid'
+        ).one().id
+    except NoResultFound:
+        orcid = ''
+    data['acquisition_source'] = dict(
+        email=user_email,
+        datetime=datetime.datetime.utcnow().isoformat(),
+        method="submitter",
+        orcid=orcid,
+        submission_number=str(id_workflow),
+        internal_uid=int(id_user),
+    )
+
+    data = strip_empty_values(data)
+
+    validate(data, 'authors')


in any case I think it should be low prio, as the data that the forms send should always be valid

50 users in the past three months have been victim of this assumption: https://sentry.inspirehep.net/inspire-sentry/prod/issues/48/

david-caro:
This should be retaken if needed after new refinement, the pr got stalled.

PR #3077, opened 2017-12-20
migrator: use a special queue for the one-off migrations

david-caro:
This will help in order to avoid starving the regular workflows.
Requires changes in the celery config in puppet.

- [x] I have all the information that I need (if not, move to `RFC` and look for it).
- [x] I linked the related issue(s) in the corresponding commit logs.
- [x] I wrote [good commit log messages](https://github.com/torvalds/subsurface-for-dirk/blob/5f15ad5a86ada3c5e574041a5f9d85235322dabb/README#L92-L119).
- [x] My code follows the code style of this project.
- [x] I've added any new docs if API/utils methods were added.
- [x] I have updated the existing documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Signed-off-by: David Caro <david@dcaro.es>

PR #3093, opened 2018-01-09
procfile: add queue celery,migrator option

chris-asl:
Update Procfile so that it matches our Docker configuration as per
inspirehep/inspire-next#3077.

## Description
This was the issue that caused the migration of records hanging in macOS. Records were being inserted into the default `celery` named queue, but in #3077 they were inserted into a `migrator` named queue.

### Suggestion
@david-caro @jacquerie
Since the commands that are executed in the Procfile are the same (or _should be_ the same) in each Docker Compose service, we could have them centralized in a configuration file, from which our `docker-compose.yml` and `Procfile` are either _reading_ or _being generated from_.

## Checklist:

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -22,7 +22,7 @@
 
 web: gunicorn inspirehep.wsgi -c gunicorn.cfg
 cache: redis-server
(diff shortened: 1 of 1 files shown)


michamos:
maybe fix also https://github.com/inspirehep/inspire-next/pull/3077#discussion_r160404827 ?

PR #3095, opened 2018-01-09
migrator: ensure $schema is uri in reporterrors

michamos:
## Description

As the `$schema` key of records is now required by the schemas to be a
valid `uri` but inspire-dojson only puts the schema name there, the
`reporterrors` CLI command has to fix this. Otherwise, the invalid
`$schema` format masks the real validation error.

## Checklist:

### DIFF
--- a/inspirehep/modules/migrator/cli.py
+++ b/inspirehep/modules/migrator/cli.py
@@ -33,7 +33,6 @@ from itertools import dropwhile
 import click
 import jsonschema
 import requests
-from flask import url_for
 from flask_cli import with_appcontext
 
 from dojson.contrib.marc21.utils import create_record
@@ -42,6 +41,7 @@ from inspire_dojson import marcxml2record
 from inspire_schemas.api import validate
 from inspire_utils.helpers import force_list
 
+from inspirehep.utils.schema import ensure_valid_schema
 from .models import InspireProdRecords
 from .tasks import (
     add_citation_counts,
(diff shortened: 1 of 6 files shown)


inspirehep/modules/migrator/cli.py
```diff
@@ -134,6 +135,13 @@ def get_collection(marc_record):
                 return collection
         return 'HEP'
 
+    def fix_schema(json_record):
```
ammirate: Apparently, this code already exists at least in another place, like https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/authors/tasks.py#L65. Shall we move it somewhere else to make it reusable?

michamos: There's also https://github.com/inspirehep/inspire-next/blob/a731e4f43bbded5cf5663e814f00c05e687c45aa/inspirehep/modules/workflows/tasks/upload.py#L90-L95. So it makes sense to pull this out into a separate function. What's a good place?

michamos:
@ammirate did the refactoring

michamos:
ci test please

michamos:
@bavaria95 acceptance-authors is failing (https://jenkins-inspire.web.cern.ch/blue/organizations/jenkins/inspire-tests/detail/inspire-tests/895/pipeline/37#step-42-log-845 ) but Jenkins thinks it went fine.

bavaria95:
@michamos 
Are you talking about this graph?
image

Here (for some reasons and I don't think it's possible to change it here) it shows only whether step is finished or not. It can be either pending or done. Result doesn't matter in this case.

Where it matters, it gives feedback to GH that it failed:
image
And you can check selenium report as build artifact (also shows screenshot in case it is available) :
[https://jenkins-inspire.web.cern.ch/job/inspire-tests/895/artifact/selenium-report-authors.html](selenium-report-authors)

Also it should be possible to check report from pytest, but right now playing with this feature and broke it. Should fix it soon, but it should give an output like here (another build, but something like this):
[https://jenkins-inspire.web.cern.ch/job/inspire-tests/820/testReport/junit/tests.acceptance.literature/test_literature_new_form/test_literature_create_article_journal_manually](url)

michamos:
yes, I was thinking of the green checks. That's confusing, thanks for explaining!

PR #3096, opened 2018-01-10
serializers: handle theses without thesis_info

szymonlopaciuk:
## Description
This fixes an issue where BibTeX generation fails if a thesis record contains no `thesis_info` field.

Example records displaying this issue:

- https://labs.inspirehep.net/api/literature/1646337
- https://labs.inspirehep.net/api/literature/754840

## Related Issue
https://sentry.inspirehep.net/inspire-sentry/prod/issues/1773/

## Checklist:

### DIFF
--- a/inspirehep/modules/records/serializers/fields_export.py
+++ b/inspirehep/modules/records/serializers/fields_export.py
@@ -90,7 +90,7 @@ def bibtex_document_type(doc_type, obj):
     if doc_type in DOCUMENT_TYPE_MAP:
         return DOCUMENT_TYPE_MAP[doc_type]
     # Theses need special treatment, because bibtex differentiates between their types:
-    elif doc_type == 'thesis' and obj['thesis_info']['degree_type'] in ('phd', 'habilitation'):
+    elif doc_type == 'thesis' and get_value(obj, 'thesis_info.degree_type') in ('phd', 'habilitation'):
         return 'phdthesis'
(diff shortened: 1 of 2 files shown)


inspirehep/modules/records/serializers/fields_export.py
```diff
@@ -63,6 +63,23 @@ def decorator(decorated_function):
...
+    First checks thesis_info then public_notes to determine if a record
+    is a PhD thesis.
+    """
+    if get_value(obj, 'thesis_info.degree_type') in ('phd', 'habilitation'):
+        return True
+
+    public_notes = get_value(obj, 'public_notes.value', [])
+    for note in public_notes:
+        if "phd thesis" in note.lower():
+            return True
```
jacquerie: @michamos do we want this?

michamos: no, reading `thesis_info.degree_type` is sufficient. If it so happens that many theses don't have that field but contain `phd thesis` in `public_notes`, we need to clean up the metadata.

szymonlopaciuk: Seems I was unlucky: first record from arXiv I checked had this issue, but looking at more now, it doesn't seem all that common. I'll remove this then.

michamos: anyway, when metadata is not where you expect it, better to ask :)

bavaria95:
ci test please

PR #3104, opened 2018-01-11
WIP: refactor of the acceptance tests

david-caro:
This is in order to allow easily to create a basic set of end to end
tests to run in CI/QA as deploy gatekeepers.


inspirehep/bat/pages/literature_submission_form.py
```diff
@@ -0,0 +1,708 @@
...
+
+    if 'abstract' in input_data:
+        Arsenic().find_element_by_id('abstract').send_keys(
+            input_data['abstract']
+        )
+
+    _populate_report_numbers(input_data.get('report_numbers', []))
+
+
+def _deprecated_populate_basic_info(input_data):
```
david-caro: remove this, is not being used anymore :)

jacquerie:
My opinion of the first commit hasn't changed from what I said a few months ago: https://github.com/inspirehep/inspire-next/pull/2517#issuecomment-335137553
> [...] this PR goes deeper in the bad direction: instead of simply writing assertions like `assert 'this text' in this_field` these tests had the complicated mechanism of `has_error` and closures and what not, and this PR makes that even more complicated by adding another similar mechanism.

david-caro:
I see a some big disadvantages of that approach:
* It requires a substantial amount of extra development.
* It spreads the knowledge of the internal workings of the pages through the tests, for example, each of the tests that has to submit a record, has to check if it was successful, with your approach, each of them should have hard-written the expected string, and know on how to extract the expected 'error message' entity from the page. That imo is a big push back.
* It does not improve the readability, actually, it will affect negatively, as it will replace something like `submit_record.assert_has_no_errors()` with a bunch of `assert 'random flaky string' in fetch_field('random flaky xpath')`. The next step then would be to put all those in a function, but then again, we get to this PR's approach.

PR #3105, opened 2018-01-12
workflows: set refereed and fix the document_type

### LINKED ISSUE #2863: workflows: set refereed/conference paper based on journal information

jacquerie:
## Description:
After the workflow step that normalizes journal titles has run, set
the ``refereed`` and fix the ``document_type`` keys if needed. This
is required to properly handle HEPCrawl harvests.

## Related Issue
Closes #2863 and closes #2951 

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -25,19 +25,18 @@
 from __future__ import absolute_import, division, print_function
 
 import os
+import re
+from functools import wraps
+
+from flask import current_app
 from sqlalchemy import (
     JSON,
     String,
     cast,
     type_coerce,
 )
-
-from functools import wraps
-import re
-
-from flask import current_app
-from werkzeug import secure_filename
 from timeout_decorator import TimeoutError
+from werkzeug import secure_filename
 
 from invenio_db import db
 from invenio_workflows import ObjectStatus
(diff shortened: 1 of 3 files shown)


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -427,3 +426,56 @@ def normalize_journal_titles(obj, eng):
...
+
+    Returns:
+        None
+
+    """
+    journals = replace_refs(get_value(obj.data, 'publication_info.journal_record'), 'db')
+    if not journals:
+        return
+
+    is_published_in_a_refereed_journal = any(
```
iulianav: @jacquerie `is_published_in_a_refereed_journal` should be called `is_published_in_a_refereed_journal_that_does_not_publish_proceedings` in order to be accurate.

PR #3110, opened 2018-01-14
mappings: make publication_info.year a date

chris-asl:
For being able to support range queries with ES date rounding, while
also having all date fields, with date type (see
inspirehep/inspire-query-parser#75).

## Related Issue
https://github.com/inspirehep/inspire-query-parser/pull/75

## Checklist:

### DIFF
--- a/inspirehep/modules/records/mappings/records/hep.json
+++ b/inspirehep/modules/records/mappings/records/hep.json
@@ -702,7 +702,8 @@
                             "type": "string"
                         },
                         "year": {
-                            "type": "integer"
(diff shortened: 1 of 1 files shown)


jacquerie:
We don't need if for queries, but for matching I think we also need to change this other guy: https://github.com/inspirehep/inspire-next/blob/d1338939caf3c52f150c4545cb65318561258b3c/inspirehep/modules/records/mappings/records/hep.json#L883-L885

chris-asl:
Ok then, I will change also this and will push again!

PR #3112, opened 2018-01-15
theme: remove auto-complete by default from the search bar

### LINKED ISSUE #2979: Search box text is overwritten when search contains `'` or `"`
Search queries are being overwritten when they contain `'` or `"` while focus out and then back focus in to the `inspire-next` browser tab.

## Expected Behavior
The text should stay the same at all times.

## Current Behavior
Searching for `a "S.Mele.1"` with the cursor left after the `S`, focusing out and then back in, the search box text has been transformed into `a "Sa "S.Mele.1"Mele.1"`.

Additionally, if you search for `a: 's.mEle'` and you leave the cursor after `E` then it gets transformed into `a: 's.me`. 

Another example is when searching for `a: 's.mele' and TITLE INSPIRE` with the cursor left before the `TITLE`, then the text is transformed into `a: 's.mele' and a: 's.mele' and TITLE INSPIRETITLE INSPIRE`.

Dinika:
Signed-off-by: Dinika Saxena <dinika.saxena@cern.ch>

## Description
Allows typeahead.js to autocomplete only when the `allowAutocomplete` attribute is set to true.

## Related Issue
fixes #2979

## Checklist:

### DIFF
--- a/inspirehep/modules/theme/templates/inspirehep_theme/typeahead.html
+++ b/inspirehep/modules/theme/templates/inspirehep_theme/typeahead.html
@@ -80,15 +80,6 @@ require(
               scope.placeholder = attrs.placeholder;
               scope.updateQuery = updateQuery;
               scope.collection = attrs.collection;
-
-              $("#search").searchTypeahead({
-                // A server returning a list of values is needed
-                // Also CORS needs to be enabled, e.g with a browser extension
-                value_hints_url: "{{config.get('SEARCH_TYPEAHEAD_HINT_URL')|safe}}",
-                options_sets: parser_config,
-                default_set: "{{config.get('SEARCH_TYPEAHEAD_DEFAULT_SET')}}",
-                suggestion_fn: suggestion_fn
-              });
             }
             function templateUrl(element, attrs) {
               return attrs.template;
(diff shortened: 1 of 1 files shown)


Dinika:
I'm not quite sure when these lines are executed. I thought they were required for the forms but that's not the case.

inspirehep/modules/theme/templates/inspirehep_theme/header.html
```diff
@@ -194,7 +194,8 @@
           <invenio-search-autocomplete-bar
             template="{{ url_for('static', filename='node_modules/inspirehep-search-js/dist/templates/searchbar.html') }}"
             placeholder="Search {{collection_name}}"
-            collection={{collection_name}}>
+            collection={{collection_name}}
+            allowAutocomplete="false">
```
Dinika: I've named the atrribute `allowAutocomplete` because just ` autocomplete` was generating errors.

drjova: 👍 we can name it ``allow-autocomplete``, angular automatically transform it to camel case ``allowAutocomplete`` inside the directive :)

PR #3122, opened 2018-01-17
migrator: only set prod record as failed when it's there

### LINKED ISSUE #3121: UnboundLocalError: local variable 'prod_record' referenced before assignment
https://sentry.inspirehep.net/inspire-sentry/prod/issues/2095/

```
UnboundLocalError: local variable 'prod_record' referenced before assignment
  File "inspirehep/modules/migrator/tasks.py", line 184, in continuous_migration
    skip_files=skip_files,
  File "inspirehep/modules/migrator/tasks.py", line 371, in migrate_and_insert_record
    prod_record.valid = False

UnboundLocalError: local variable 'prod_record' referenced before assignment
```

david-caro:

Fixes #3121

### DIFF
--- a/inspirehep/modules/migrator/tasks.py
+++ b/inspirehep/modules/migrator/tasks.py
@@ -340,13 +340,15 @@ def migrate_and_insert_record(raw_record, skip_files=False):
 
     try:
         json_record = marcxml2record(raw_record)
-        if '$schema' in json_record:
-            ensure_valid_schema(json_record)
     except Exception as e:
         LOGGER.exception('Migrator DoJSON Error')
         error = e
         recid = 'No recid extracted'
+        prod_record = None
     else:
+        if '$schema' in json_record:
+            ensure_valid_schema(json_record)
+
         recid = json_record['control_number']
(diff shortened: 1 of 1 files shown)


ammirate:
Perhaps adding some tests would make this function more stable, we already did a quick fix for this... wdyt? :/

david-caro:
That it needs refactoring, I think it tries to do too many things, and it strange ways (keeping the error var around, and such). I'd create another PR for it.

ammirate:
So it's not worth neither adding the record that triggered this unknown behavior as a test? That would be at least something.

david-caro:
That in specific will not help, as it will stop triggering the error as soon as dojson is fixed. To test this, I think that the strategy should be mocking the marcxml2record and forcing the throw of an exception there. Also for the record_insert_or_replace, and then make sure that it handles it gracefully.

david-caro:
btw. if you agree, I'll merge, this is hiding some errors (as it shadows the original error).

ammirate:
"Move fast, break things"

Issue #3124, opened 2018-01-18
workflows: deleted inspire categories shouldn't reappear in updates


michamos:
## Expected Behavior

Inspire categories should not be changed by automatic updates when harvesting the arXiv update.
A category derived automatically from an arXiv category and manually deleted shouldn't reappear .
A category manually added shouldn't be deleted. 

## Current Behavior

All arXiv categories in the update are converted to inspire categories, so previously deleted categories reappear. There is no way to prevent this.

michamos:
@ksachs I hope this is a faithful translation of the issue you raised by email. If not please edit it or reply with a correction.

The problem is actually worse than you mention: currently, updates to arXiv preprints that fall in the 30 day window after the first version was created are sent to legacy as `correct`, so potentially overwriting curator work. This is a terrible bug that was caused by increasing the window for accepting arXiv harvests from 5 days to 30 days (see https://github.com/inspirehep/inspire-next/issues/2524#issuecomment-339677349 ). @ammirate interrupted his work to fix this before tomorrow's harvest, and we will try to find all affected records and revert them if curator work has been overwritten.

Once the merger is integrated (#2760), we will be able to handle updates correctly.

ksachs:
what about stopping updates from labs completely until the merger is in place?

I assume legacy can handle the basic updates.
The rest was ignored the last years, so it doesn't matter if we ignore it a few weeks/month longer.

ksachs:
I guess those updates could also wipe out information from journal harvest.

michamos:
> what about stopping updates from labs completely until the merger is in place?

that's exactly what @ammirate is implementing.

> I guess those updates could also wipe out information from journal harvest.

wipe out no, the updates were done in `correct` mode, so they would overwrite MARC fields, but not erase existing MARC fields that were not in the arXiv metadata. This could happen on top of a journal harvest only if

1.the journal harvest occurred within the 30 day window after initial arXiv submission, and
2. there was an arXiv update after the journal harvest, still in the 30 day window.

Authors usually submit to the journals at the same time as to arXiv ar after and it takes more than a month for the article to go through the whole editorial process, so while definitely possible, it has probably not happened very often.

michamos:
This should be fixed by #3126.

PR #3126, opened 2018-01-18
workflow: don't send updates to legacy and approve them on labs

ammirate:

This PR removes the window time in which articles are automatically rejected since we can recognize them now as updates, using the matcher.
Such updates are now being sent back to legacy, potentially overwriting human work.
Thus, updates in this PR are not sent to legacy at all.

## Related Issue
https://github.com/inspirehep/inspire-next/issues/3124

## Checklist:

### DIFF
--- a/tests/integration/workflows/test_arxiv_workflow.py
+++ b/tests/integration/workflows/test_arxiv_workflow.py
@@ -373,3 +373,42 @@ def test_match_in_holdingpen_different_sources_continues(
     assert obj.extra_data['holdingpen_matches'] == [wf_to_match]
     assert obj.extra_data['previously_rejected'] is False
     assert not obj.extra_data.get('stopped-matched-holdingpen-wf')
+
+
+@mock.patch(
+    'inspirehep.modules.workflows.tasks.arxiv.is_pdf_link',
+    return_value=True
+)
+@mock.patch(
+    'inspirehep.modules.workflows.tasks.arxiv.download_file_to_workflow',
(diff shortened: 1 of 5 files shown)


tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -373,3 +375,46 @@ def test_match_in_holdingpen_different_sources_continues(
...
+)
+@mock.patch(
+    'inspirehep.modules.workflows.tasks.beard.json_api_request',
+    side_effect=fake_beard_api_request,
+)
+@mock.patch(
+    'inspirehep.modules.workflows.tasks.magpie.json_api_request',
+    side_effect=fake_magpie_api_request,
+)
+def test_arxiv_update_overwrites_record_on_labs_but_not_in_legacy(
```
michamos: why should it overwrite the record on labs? don't we want to maintain it in sync with legacy?

ammirate: For now yes, and in fact, after the robot upload the record comes back from legacy and overwrites whatever there is.

michamos: but if it's an update, there is no robotupload, so the record might never come back.

ammirate: That's true, you are right. So, let's throw them away.

ammirate:
ci test please

PR #3127, opened 2018-01-18
migrator: fix add_citation_counts for PID with non numeric values

puntonim:
Fix add_citation_counts task for PID with non numeric values.

## Description

We are about to introduce PID with string values (like ISBN, URN, ArXiv ID, ...) and the add_citation_counts task assumes that the PID value is always numeric.

## Checklist:


ammirate:
@puntonim What is missing here? Seems that we have to mock Elastic Search for the unit tests, is it right?

puntonim:
@ammirate it's all set now.
It required me to write integration tests instead of unit tests (I didn't get that at first).
Then It was blocked by test db isolation.

Issue #2527, opened 2017-07-12
HoldingPen: arXiv reappears


ksachs:
## Expected Behavior
harvests of an article we harvested previously should be filtered

I'm not sure whether this is already_harvested or previously_rejected.
In any case: @fschwenn and I don't want to see an article more than once in the holdingpen.

## Current Behavior / Example
An arxiv article (1707.02681) was harvested and rejected yesterday (https://labs.inspirehep.net/holdingpen/675354). Today another harvest of the same article is again waiting in the HP (https://labs.inspirehep.net/holdingpen/676378). 

https://labs.inspirehep.net/holdingpen/list/?page=1&size=10&q=1707.02681

## Screenshots (if appropriate):
![hp_reharvest](https://user-images.githubusercontent.com/7591594/28106412-094a74a4-66e4-11e7-8ab4-fe2fea5e9af6.png)

fschwenn:
This seems to happen even for pending articles.
https://labs.inspirehep.net/holdingpen/677472 was harvested while
https://labs.inspirehep.net/holdingpen/676633 was already waiting for a decision. It has not been updated on arXiv (https://arxiv.org/abs/1707.02998) meanwhile.

kaplun:
This is normal, until we merge: #2423. The current situation is that we don't know if something is an update or not. So there is a stupid check: if something has a creation date older than 2 weeks it's rejected assuming it's duplicate. 
Once #2423 is merged (which will happen as soon as we implement the filtering of updated Vs. new records I mentioned on standup) then you will see updates treated as updates.

kaplun:
Re: what @fschwenn  says... mmh...

kaplun:
The record you mention has been modified last time the 12th according to arXiv, so it make sense that it has been harvested on the 13th... (assuming my explanation that we are not able yet to distinguish updates).

ksachs:
we don't need the merger. We need to know whether we saw it before. I.e. it should be enough to search for the arXiv-id in INSPIRE - if it's there it's an update - do whatever was done on legacy; if not in INSPIRE query holdingpen - if it's there reject automatically.
At least reduce the time-window to max. 4 days. Even on holidays and weekends I guess we get the first harvest within 4 days.

ksachs:
there is no difference in the metadata for 676633 and 677472, maybe a status-change triggered the update.
[...]

PR #3129, opened 2018-01-19
workflow: integrate fuzzy matcher in the `article` workflow

ammirate:
## Description

Changed the step in the workflow that checks if the ingested article is an update, by
performing a fuzzy match. If there is at least a match, the workflow halts, waiting
for human matching confirmation.

All the matches are now stored in `obj.extra_data.matches`, which is a dictionary containing
`exact` matches, `fuzzy` matches, and eventually the `approved` fuzzy match.

A new action called `MatchApproval` has been added to resume the workflow when is halted for
matching approval.

Solves #2527.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/mappings/holdingpen/hep.json
+++ b/inspirehep/modules/workflows/mappings/holdingpen/hep.json
@@ -12,7 +12,7 @@
                         "is-update": {
                             "type": "boolean"
                         },
-                        "record_matches": {
+                        "matches": {
                             "enabled": false,
                             "include_in_all": false,
                             "type": "object"
--- a/inspirehep/modules/workflows/tasks/matching.py
+++ b/inspirehep/modules/workflows/tasks/matching.py
@@ -78,13 +78,13 @@ def is_too_old(record, days_ago=5):
 
 
 @with_debug_logging
-def article_exists(obj, eng):
+def exact_match(obj, eng):
     """Return ``True`` if the record is already present in the system.
 
     Uses the default configuration of the ``inspire-matcher`` to find
     duplicates of the current workflow object in the system.
 
-    Also sets the ``record_matches`` property in ``extra_data`` to the list of
+    Also sets the ``matches.exact`` property in ``extra_data`` to the list of
     control numbers that matched.
 
     Arguments:
@@ -96,14 +96,57 @@ def article_exists(obj, eng):
         ``False`` otherwise.
 
     """
-    matches = dedupe_list(match(obj.data))
+    exact_match_config = current_app.config['EXACT_MATCH']
+    matches = dedupe_list(match(obj.data, exact_match_config))
     record_ids = [el['_source']['control_number'] for el in matches]
-    if record_ids:
-        obj.extra_data['record_matches'] = record_ids
-        return True
+    obj.extra_data.setdefault('matches', {})['exact'] = record_ids
+    return bool(record_ids)
 
-    obj.extra_data['record_matches'] = []
-    return False
+
+@with_debug_logging
+def fuzzy_match(obj, eng):
+    """Return ``True`` if a similar record is found in the system.
+
+    Uses a custom configuration for ``inspire-matcher`` to find records
+    similar to the current workflow object's payload in the system.
+
+    Also sets the ``matches.fuzzy`` property in ``extra_data`` to the list of
+    control numbers that matched.
+
+    Arguments:
+        obj: a workflow object.
+        eng: a workflow engine.
+
+    Returns:
+        bool: ``True`` if the workflow object has a duplicate in the system
+        ``False`` otherwise.
+
+    """
+    fuzzy_match_config = current_app.config['FUZZY_MATCH']
+    matches = dedupe_list(match(obj.data, fuzzy_match_config))
+    record_ids = [el['_source']['control_number'] for el in matches]
+    obj.extra_data.setdefault('matches', {})['fuzzy'] = record_ids
+    return bool(record_ids)
+
+
+@with_debug_logging
+def is_fuzzy_match_approved(obj, eng):
+    """Check if a fuzzy match has been approved by a human."""
+    return obj.extra_data.get('fuzzy_match_approved_id')
+
+
+@with_debug_logging
+def set_fuzzy_match_approved_in_extradata(obj, eng):
+    """Set the human approved match in `matches.approved` in extra_data."""
+    approved_match = obj.extra_data.get('fuzzy_match_approved_id')
+    obj.extra_data.setdefault('matches', {})['approved'] = approved_match
+
+
+@with_debug_logging
+def set_exact_match_as_approved_in_extradata(obj, eng):
+    """Set the best match in `matches.approved` in extra_data."""
+    best_match = obj.extra_data['matches']['exact'][0]
+    obj.extra_data.setdefault('matches', {})['approved'] = best_match
 
 
 def auto_approve(obj, eng):
@@ -201,7 +244,7 @@ def match_non_completed_wf_in_holdingpen(obj, eng):
     def _non_completed(base_record, match_result):
         return not get_value(match_result, '_source._workflow.status') == 'COMPLETED'
 
-    matched_ids = _pending_in_holding_pen(obj, _non_completed)
+    matched_ids = pending_in_holding_pen(obj, _non_completed)
     obj.extra_data['holdingpen_matches'] = matched_ids
     return bool(matched_ids)
 
@@ -229,13 +272,13 @@ def match_previously_rejected_wf_in_holdingpen(obj, eng):
         return get_value(match_result, '_source._workflow.status') == 'COMPLETED' and \
             get_value(match_result, '_source._extra_data.approved') is False
 
-    matched_ids = _pending_in_holding_pen(obj, _rejected_and_completed)
+    matched_ids = pending_in_holding_pen(obj, _rejected_and_completed)
     obj.extra_data['previously_rejected_matches'] = matched_ids
     return bool(matched_ids)
 
 
 @with_debug_logging
-def _pending_in_holding_pen(obj, validation_func):
+def pending_in_holding_pen(obj, validation_func):
     """Return the list of matching workflows in the holdingpen.
 
     Matches the holdingpen records by their ``arxiv_eprint``, their ``doi``,
--- a/inspirehep/modules/workflows/tasks/upload.py
+++ b/inspirehep/modules/workflows/tasks/upload.py
@@ -38,14 +38,14 @@ from inspirehep.utils.schema import ensure_valid_schema
 def store_record(obj, eng):
     """Insert or replace a record."""
     def _get_updated_record(obj):
-        """TODO: use only head_uuid once we have them merger."""
+        """TODO: use only head_uuid once we have the merger."""
         if 'head_uuid' in obj.extra_data:
             updated_record = InspireRecord.get_record(
                 obj.extra_data['head_uuid'],
             )
         else:
             pid_type = get_pid_type_from_schema(obj.data['$schema'])
-            updated_record_id = obj.extra_data['record_matches'][0]
+            updated_record_id = obj.extra_data['matches']['approved']
             updated_record = get_db_record(pid_type, updated_record_id)
 
         return updated_record
--- a/inspirehep/modules/workflows/workflows/article.py
+++ b/inspirehep/modules/workflows/workflows/article.py
@@ -75,7 +75,11 @@ from inspirehep.modules.workflows.tasks.matching import (
     stop_processing,
     match_non_completed_wf_in_holdingpen,
     match_previously_rejected_wf_in_holdingpen,
-    article_exists,
+    exact_match,
+    fuzzy_match,
+    is_fuzzy_match_approved,
+    set_exact_match_as_approved_in_extradata,
+    set_fuzzy_match_approved_in_extradata,
     previously_rejected,
     has_same_source,
     stop_matched_holdingpen_wfs,
@@ -262,16 +266,6 @@ STOP_IF_EXISTING_SUBMISSION = [
 
 
 HALT_FOR_APPROVAL_IF_NEW_OR_STOP_IF_NOT_RELEVANT = [
-    IF_NOT(
-        is_record_relevant,
-        [
-            reject_record("Article automatically rejected"),
-            mark('approved', False),  # auto reject
-            save_workflow,
-            stop_processing,
-        ]
-    ),
-
     IF_ELSE(
         is_marked('is-update'),
         [
@@ -285,7 +279,16 @@ HALT_FOR_APPROVAL_IF_NEW_OR_STOP_IF_NOT_RELEVANT = [
                 message="Submission halted for curator approval.",
             )
         ),
-    )
+    ),
+    IF_NOT(
+        is_record_relevant,
+        [
+            reject_record("Article automatically rejected"),
+            mark('approved', False),  # auto reject
+            save_workflow,
+            stop_processing,
+        ]
+    ),
 ]
 
 
@@ -395,11 +398,33 @@ PROCESS_HOLDINGPEN_MATCHES = [
 ]
 
 
-MARK_IF_UPDATE = [
+CHECK_IS_UPDATE = [
     IF_ELSE(
-        article_exists,
-        mark('is-update', True),
-        mark('is-update', False),
+        exact_match,
+        [
+            set_exact_match_as_approved_in_extradata,
+            mark('is-update', True),
+            mark('exact-matched', True),
+        ],
+        IF_ELSE(
+            fuzzy_match,
+            [
+                halt_record(
+                    action="match_approval",
+                    message="Halted for matching approval.",
+                ),
+                IF_ELSE(
+                    is_fuzzy_match_approved,
+                    [
+                        set_fuzzy_match_approved_in_extradata,
+                        mark('fuzzy-matched', True),
+                        mark('is-update', True),
+                    ],
+                    mark('is-update', False),
+                )
+            ],
+            mark('is-update', False),
+        )
     ),
     save_workflow,
 ]
@@ -466,7 +491,7 @@ class Article(object):
         PRE_PROCESSING +
         NOTIFY_IF_SUBMISSION +
         MARK_IF_MATCH_IN_HOLDINGPEN +
-        MARK_IF_UPDATE +
+        CHECK_IS_UPDATE +
         PROCESS_HOLDINGPEN_MATCHES +
         ENHANCE_RECORD +
         STOP_IF_EXISTING_SUBMISSION +
--- a/tests/acceptance/conftest.py
+++ b/tests/acceptance/conftest.py
@@ -25,6 +25,7 @@ from __future__ import absolute_import, division, print_function
 import pytest
 
 from invenio_db import db
+from invenio_records.models import RecordMetadata
 from invenio_search import current_search_client as es
 from invenio_workflows import workflow_object_class
 
@@ -116,3 +117,13 @@ def cleanup_workflows_tables(app):
                 obj.delete()
 
         db.session.commit()
+
+
+@pytest.fixture(autouse=True, scope='function')
+def cleanup_literatures(app):
+    recs = RecordMetadata.query.filter(RecordMetadata.json['_collections'].op('?')('Literature')).all()
+    for lit_record in recs:
+        db.session.delete(lit_record)
+
+    db.session.commit()
+    es.indices.refresh('records-hep')
--- a/tests/integration/workflows/test_arxiv_workflow.py
+++ b/tests/integration/workflows/test_arxiv_workflow.py
@@ -32,6 +32,7 @@ from invenio_db import db
 from invenio_workflows import (
     ObjectStatus,
     WorkflowEngine,
+    WorkflowObject,
     start,
     workflow_object_class,
 )
@@ -42,6 +43,7 @@ from calls import (
     do_accept_core,
     do_webcoll_callback,
     do_robotupload_callback,
+    do_resolve_matching,
     generate_record,
 )
 from mocks import (
@@ -76,11 +78,12 @@ from utils import get_halted_workflow
     return_value=[],
 )
 def test_harvesting_arxiv_workflow_manual_rejected(
-    mocked_arxiv_download,
     mocked_refextract_extract_refs,
     mocked_api_request_magpie,
-    mocked_api_request_beard,
-    mocked_package_download,
+    mocked_beard_api,
+    mocked_actions_download,
+    mocked_is_pdf_link,
+    mocked_arxiv_download,
     workflow_app,
     mocked_external_services,
 ):
@@ -109,6 +112,7 @@ def test_harvesting_arxiv_workflow_manual_rejected(
     obj = workflow_object_class.get(obj_id)
     # It was rejected
     assert obj.status == ObjectStatus.COMPLETED
+    assert obj.extra_data["approved"] is False
 
 
 @mock.patch(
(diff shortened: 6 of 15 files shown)


inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -105,12 +105,83 @@ def article_exists(obj, eng):
...
+                                'path':  'abstracts',
+                            },
+                            {
+                                'boost': 10,
+                                'path':  'authors[:3]',
+                            },
+                            {
+                                'boost': 10,
+                                'path':  'titles',
+                            },
```
jacquerie: @jmartinm derived a different configuration: https://github.com/inspirehep/invenio-matcher-benchmark/blob/83caef0aaddb59269cb4769fd1fc9d72fb8234f2/config.py

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -98,12 +98,87 @@ def article_exists(obj, eng):
...
+            {
+                'queries': [
+                    {
+                        'clauses': [
+                            {
+                                'boost': 20,
+                                'path':  'abstracts',
+                            },
+                            {
+                                'boost': 10,
```
ammirate: @jacquerie I have updated the configuration tin according to https://github.com/inspirehep/invenio-matcher-benchmark/blob/83caef0aaddb59269cb4769fd1fc9d72fb8234f2/config.py, but removing the `boost` from `authors` I couldn't match the basic record I am using. Shall we go for some benchmarks? If yes, I think they are out of the scope of this PR.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -98,12 +98,87 @@ def article_exists(obj, eng):
...
+    return bool(record_ids)
+
+
+@with_debug_logging
+def is_fuzzy_match_approved(obj, eng):
+    """Check if a fuzzy match has been approved.
+
+    Also, if the property `fuzzy_match_approved_id` in obj.extra_data
+    contains an id, it is added to the property `matches.approved` in
+    obj.extra_data.
```
david-caro: I think that the name should be changed, according to this. I'd expect a step called 'is_...' not to change anything.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -98,12 +98,87 @@ def article_exists(obj, eng):
...
+        obj: a workflow object.
+        eng: a workflow engine.
+
+    Returns:
+        bool: ``True`` if the property fuzzy_match_approved_id
+        in obj.extra_data contains an integer id, ``False`` otherwise.
+    """
+    approved_match_id = obj.extra_data.get('fuzzy_match_approved_id')
+    obj.extra_data.setdefault('matches', {})['approved'] = approved_match_id
+    return bool(isinstance(approved_match_id, (int, long)))
```
david-caro: the bool is not needed here, isinstance always returns bool.

david-caro: The fact that you are checking the type of the value looks strange. Should we care about that? If it's a string or anything? I think we should only care if there's any, not what it is.

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -392,11 +394,32 @@
...
-        mark('is-update', False),
+        exact_match,
+        [
+            mark('is-update', True),
+            mark('exact-matched', True),
+        ],
+        IF_ELSE(
+            fuzzy_match,
+            [
+
```
david-caro: extra newline

david-caro: If you added it to make it more readable, I suggest using a new array var instead (the equivalent of moving to a new function).

inspirehep/modules/workflows/tasks/matching.py
david-caro: Maybe we can split it in `handle_fuzzy_match_decision` that does the changes in the workflow, and `is_there_an_approval`.

The second one might not care about if the approval is fuzzy match or not, but just that there's a result in the `matches.approved`.

ammirate: Do you mean in the future, we may want to halt even for exact matches and reuse this part? Sounds reasonable

david-caro: Yep, but I was focusing more on those two actions being different things.

inspirehep/modules/workflows/workflows/article.py
ammirate: just extra

inspirehep/modules/workflows/tasks/matching.py
ammirate: Depends on how we define the API. If we specify the front-end has to pass an integer, then we can check if it is compliant to the spec.

tests/unit/workflows/test_workflows_tasks_matching.py
```diff
@@ -23,21 +23,25 @@
 from __future__ import absolute_import, division, print_function
 
 from mock import patch
+from inspire_utils.record import get_value
 
 from inspire_schemas.api import load_schema, validate
 from inspirehep.modules.workflows.tasks.matching import (
-    article_exists,
     has_fully_harvested_category,
     physics_data_an_is_primary_category,
```
iulianav: These are not in order.

tests/unit/workflows/test_workflows_tasks_matching.py
```diff
@@ -279,3 +283,79 @@ def test_pending_in_holding_pen_returns_false_if_nothing_matched(mock_match):
...
+    assert 'matches' in obj.extra_data
+
+    expected = []
+    result = get_value(obj.extra_data, 'matches.fuzzy')
+
+    assert expected == result
+
+
+@patch('inspirehep.modules.workflows.tasks.matching.match')
+def test_is_fuzzy_match_approved_returns_returns_true_if_there_is_a_match_approved(mock_match):
```
iulianav: 2 'returns' in the test name.

tests/unit/workflows/test_workflows_tasks_matching.py
```diff
@@ -279,3 +283,79 @@ def test_pending_in_holding_pen_returns_false_if_nothing_matched(mock_match):
...
+    assert 'matches' in obj.extra_data
+
+    expected = 4328
+    result = get_value(obj.extra_data, 'matches.approved')
+
+    assert expected == result
+
+
+@patch('inspirehep.modules.workflows.tasks.matching.match')
+def test_is_fuzzy_match_approved_returns_returns_False_if_there_is_not_a_match_approved(mock_match):
```
iulianav: Again 2 'returns'.

tests/unit/workflows/test_workflows_actions.py
```diff
@@ -0,0 +1,80 @@
...
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+from invenio_workflows import ObjectStatus
+
+from inspirehep.modules.workflows.actions import MatchApproval, MergeApproval
+from mocks import MockObj
```
iulianav: I think this should be placed right after the '__future__' import.

tests/integration/workflows/test_workflows_manual_merge.py
```diff
@@ -25,7 +25,7 @@
 import pytest
 
 from invenio_workflows import ObjectStatus, workflow_object_class
-from invenio_records.api import RecordMetadata
+from invenio_records.models import RecordMetadata
```
iulianav: These are not in alphabetical order.

tests/acceptance/conftest.py
```diff
@@ -116,3 +117,11 @@ def cleanup_workflows_tables(app):
                 obj.delete()
 
         db.session.commit()
+
+
+@pytest.fixture(autouse=True, scope='function')
+def cleanup_literatures(app):
+    recs = RecordMetadata.query.filter(RecordMetadata.json['_collections'].op('?')('Literature')).all()
```
iulianav: Can't you maybe do a straight 'delete' here with a filter or smth? I am not sure.

inspirehep/modules/workflows/actions/__init__.py
```diff
@@ -27,3 +27,4 @@
 from .hep_approval import HEPApproval        # noqa: F401
 from .merge_approval import MergeApproval    # noqa: F401
 from .author_approval import AuthorApproval  # noqa: F401
+from .match_approval import MatchApproval  # noqa: F401
```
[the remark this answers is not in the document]
ammirate: :P

inspirehep/modules/workflows/tasks/upload.py
```diff
@@ -45,7 +46,7 @@ def _get_updated_record(obj):
             )
         else:
             pid_type = get_pid_type_from_schema(obj.data['$schema'])
-            updated_record_id = obj.extra_data['record_matches'][0]
+            updated_record_id = get_value(obj.extra_data, 'matches.approved')
```
drjova: we've to double check if the same behaviour for ``exact_match`` and ``fuzzy_match``

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -98,12 +98,77 @@ def article_exists(obj, eng):
     """
     matches = dedupe_list(match(obj.data))
     record_ids = [el['_source']['control_number'] for el in matches]
-    if record_ids:
-        obj.extra_data['record_matches'] = record_ids
-        return True
+    obj.extra_data.setdefault('matches', {})['exact'] = record_ids
```
drjova: has to be the first record of the list in ``matched.approved``

ammirate: Great!

inspirehep/modules/workflows/mappings/holdingpen/hep.json
```diff
@@ -12,7 +12,7 @@
                         "is-update": {
                             "type": "boolean"
                         },
-                        "record_matches": {
+                        "matches": {
```
jacquerie: OPS: the content of this key should be migrated as well.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -98,12 +98,83 @@ def article_exists(obj, eng):
...
+                            },
+                        ],
+                        'type': 'fuzzy',
+                    }
+                ]
+            }
+        ],
+        'doc_type': 'hep',
+        'index': 'records-hep',
+    }
```
jacquerie: This should be part of the configuration, so that it can be overridden in the OPS repo (if need arises).

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -98,12 +98,83 @@ def article_exists(obj, eng):
     """
     matches = dedupe_list(match(obj.data))
     record_ids = [el['_source']['control_number'] for el in matches]
```
jacquerie: Since https://github.com/inspirehep/inspire-matcher/commit/3d9cec0df0944acd7733949e776f1e95248a13be one can avoid this and just pass the `source` parameter to the `match` call to do filtering already on ES (which is much faster, as it reduces network communication).

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -98,12 +98,83 @@ def article_exists(obj, eng):
...
+                    }
+                ]
+            }
+        ],
+        'doc_type': 'hep',
+        'index': 'records-hep',
+    }
+
+    matches = dedupe_list(match(obj.data, fuzzy_match_configuration))
+    record_ids = [el['_source']['control_number'] for el in matches]
```
jacquerie: Same as https://github.com/inspirehep/inspire-next/pull/3129/files#r171280155.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -98,12 +98,83 @@ def article_exists(obj, eng):
     """
     matches = dedupe_list(match(obj.data))
```
jacquerie: I was sloppy here, and relied on the default configuration of the matcher, but it should be passed as a parameter (fetched from the configuration, as observed in https://github.com/inspirehep/inspire-next/pull/3129/files#r171278799).

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -98,12 +98,83 @@ def article_exists(obj, eng):
...
+
+
+def set_fuzzy_match_approved_in_extradata(obj, eng):
+    """Set the human approved match in `matches.approved` in extra_data."""
+    approved_match = obj.extra_data.get('fuzzy_match_approved_id')
+    obj.extra_data.setdefault('matches', {})['approved'] = approved_match
+
+
+def set_exact_match_as_approved_in_extradata(obj, eng):
+    """Set the human approved match in `matches.approved` in extra_data."""
```
jacquerie: Rogue Ctrl-C Ctr-V ?

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -98,12 +98,83 @@ def article_exists(obj, eng):
...
+    best_match = get_value(obj.extra_data, 'matches.exact')[0]
```
jacquerie: You can just `get_value(obj.extra_data, 'matches.exact[0]')`.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -98,12 +98,83 @@ def article_exists(obj, eng):
...
+    return bool(record_ids)
+
+
+@with_debug_logging
+def is_fuzzy_match_approved(obj, eng):
+    """Check if a fuzzy match has been approved by a human."""
+    return obj.extra_data.get('fuzzy_match_approved_id')
+
+
+def set_fuzzy_match_approved_in_extradata(obj, eng):
```
jacquerie: Missing `with_debug_logging` decorator.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -98,12 +98,83 @@ def article_exists(obj, eng):
...
+    """Set the human approved match in `matches.approved` in extra_data."""
+    approved_match = obj.extra_data.get('fuzzy_match_approved_id')
+    obj.extra_data.setdefault('matches', {})['approved'] = approved_match
+
+
+def set_exact_match_as_approved_in_extradata(obj, eng):
```
jacquerie: Missing `with_debug_logging` decorator.

tests/acceptance/conftest.py
ammirate: mmh nope, the result of the query relies on the _collection field, which is in the json column of the table

inspirehep/modules/workflows/tasks/matching.py
ammirate: Moving it in `inspirehep/config.py`.

inspirehep/modules/workflows/tasks/matching.py
ammirate: The filtering works as expected, but it still returns a minimal object, from which we have to extract only the `control_number`, so this statement is still required.

This is the object returned by `match`:
```python
{
  u'_id': u'c672d49e-f633-45dd-89d3-78d653098350',
  u'_index': u'records-hep',
  u'_score': 1.292094,
  u'_source': {u'control_number': 1234},
  u'_type': u'hep'
}
```

inspirehep/modules/workflows/tasks/upload.py
david-caro: Break if the approved is None or not there (key not there).

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -395,11 +399,33 @@
 ]
 
 
-MARK_IF_UPDATE = [
+HALT_FOR_MATCH_APPROVAL_OR_MARK_IF_UPDATE = [
```
david-caro: CHECK_IS_UPDATE

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -76,11 +78,12 @@
     return_value=[],
 )
 def test_harvesting_arxiv_workflow_manual_rejected(
-    mocked_arxiv_download,
     mocked_refextract_extract_refs,
     mocked_api_request_magpie,
-    mocked_api_request_beard,
     mocked_package_download,
```
david-caro: s/mocked_package_download/mocked_beard_api or whatever

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -136,11 +140,10 @@ def test_harvesting_arxiv_workflow_manual_rejected(
     return_value=[],
 )
 def test_harvesting_arxiv_workflow_core_record_auto_accepted(
-    mocked_download,
-    mocked_is_pdf,
```
david-caro: Looks like you removed one too many, strange it did not fail though (checking locally).

david-caro: Tested locally, yes, it's overriding the workflow_app. My guess is that one of the other fixtures (probably mocked_external_services) is already requiring the workflow app, and it's doing so inside an app context, so the magick mock `workflow_app.app_context()` is not really doing anything, and it does not fail just because it's lucky.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -261,10 +263,11 @@ def test_harvesting_arxiv_workflow_manual_accepted(
     side_effect=fake_magpie_api_request,
 )
 def test_match_in_holdingpen_stops_pending_wf(
-    mocked_download_arxiv,
-    mocked_api_request_beard,
     mocked_api_request_magpie,
+    mocked_api_request_beard,
     mocked_package_download,
+    mocked_is_pdf_link,
+    mocked_download_arxiv,
```
david-caro: This one was using one less param as needed... it's the reverse case than the other one :), now it's ok.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -498,3 +503,109 @@ def test_article_workflow_continues_when_record_is_valid(workflow_app):
...
+    mocked_api_request_beard,
+    mocked_api_request_magpie,
+    workflow_app,
+    mocked_external_services,
+    record_from_db,
+):
+    """Test update article fuzzy matched.
+
+    In this test the `WorkflowObject.continue_workflow` is mocked because in
+    the test suite celery is run in eager mode. This prevent celery to switch
```
david-caro: s/prevent/prevents/

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -498,3 +503,109 @@ def test_article_workflow_continues_when_record_is_valid(workflow_app):
...
+    assert obj.extra_data['already-in-holding-pen'] is False
+    assert obj.extra_data['holdingpen_matches'] == []
+    assert obj.extra_data['matches']['fuzzy'] == [rec_id]
+
+    WorkflowObject.continue_workflow = continue_wf_patched_context(workflow_app)
+    do_resolve_matching(workflow_app, obj.id, rec_id)
+
+    obj = workflow_object_class.get(obj_id)
+    assert obj.extra_data['matches']['approved'] == rec_id
+    assert obj.extra_data['is-update']
```
david-caro: Should this be approved/completed? It's not really clear to me what we are testing here though :/, we tend to mix integration with end-to-end with acceptance with ui...

tests/integration/workflows/test_workflows_manual_merge.py
david-caro: why this change?

tests/integration/workflows/test_workflows_manual_merge.py
```diff
@@ -73,7 +71,12 @@ def test_manual_merge_existing_records(workflow_app):
         current_user_id=1,
     )
 
-    do_resolve_manual_merge_wf(workflow_app, obj_id)
+    # assuming MergeApproval.resolve() is called by the api
```
david-caro: and this change? you adapted the function in https://github.com/inspirehep/inspire-next/pull/3129/files#diff-45d44c66955f78a020f91bbab856b800R65

tests/acceptance/conftest.py
david-caro:
```python
RecordMetadata.query.filter(RecordMetadata.json['_collections'].op('?'), 'Literature')).delete(synchronize_session=False)
```

check http://docs.sqlalchemy.org/en/latest/orm/query.html?highlight=delete#sqlalchemy.orm.query.Query.delete for details.

tests/unit/workflows/test_workflows_tasks_matching.py
```diff
@@ -23,40 +23,45 @@
 from __future__ import absolute_import, division, print_function
 
 from mock import patch
+from inspire_utils.record import get_value
 
 from inspire_schemas.api import load_schema, validate
 from inspirehep.modules.workflows.tasks.matching import (
-    article_exists,
+    _pending_in_holding_pen,
```
david-caro: This should be renamed to not `_`, or not taken into account in the tests imo.

david-caro: I was there already though :/

tests/unit/workflows/test_workflows_tasks_matching.py
```diff
@@ -23,40 +23,45 @@
...
 )
 
 from mocks import MockEng, MockObj
 
 
 @patch('inspirehep.modules.workflows.tasks.matching.match')
-def test_article_exists_returns_true_if_something_matched(mock_match):
-    mock_match.return_value = iter([{'_source': {'control_number': 4328}}])
+def test_exact_match_returns_true_if_something_matched(mock_match):
+    mock_match.return_value = iter([{'_source': {'control_number': 4238}}])
```
david-caro: Is this control number relevant?

tests/unit/workflows/test_workflows_tasks_matching.py
```diff
@@ -65,11 +70,11 @@ def test_article_exists_returns_false_if_nothing_matched(mock_match):
     obj = MockObj(data, extra_data)
     eng = MockEng()
 
-    assert not article_exists(obj, eng)
-    assert 'record_matches' in obj.extra_data
+    assert not exact_match(obj, eng)
+    assert 'matches' in obj.extra_data
```
david-caro: It feels strange to me that you are checking that `'matches'` is in `extra_data` but not `'exact'`.

tests/integration/workflows/test_arxiv_workflow.py
ammirate: This is testing a new branch of the workflow: when the fuzzy matching branch is walked through, some things happen in the workflow (marks, halting).

tests/integration/workflows/test_workflows_manual_merge.py
ammirate: leftover of a previous refactor T.T

tests/unit/workflows/test_workflows_tasks_matching.py
ammirate: there is this assertion after that is checking it, isn't it ok?
`result = get_value(obj.extra_data, 'matches.exact')`

inspirehep/modules/workflows/tasks/matching.py
david-caro: Something happened here... wasn't this config moved?

tests/integration/workflows/test_workflows_manual_merge.py
ammirate: Wrong refactor :+1:

david-caro:
ci merge please

david-caro:
ci test please

ammirate:
ci test please

ammirate:
ci merge please

PR #3130, opened 2018-01-20
gitignore: ignore also symlink to the var folder

puntonim:
## Description
We currently ignore the var dir. This works well if the project folder is inside the virtual environment. But if the project folder is somewhere else, then you might want to add a symlink to the var folder. Thus, ignoring both the var folder and var symlink is the best solution.

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,7 +42,7 @@ lib/
 lib64/
 parts/
 sdist/
-var/
+var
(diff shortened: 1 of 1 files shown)


puntonim:
I've discussed with Harris: I'm the only one having this use case thus I will manage it with my global gitignore file.

PR #3134, opened 2018-01-22
orcid: push records to ORCID

### LINKED ISSUE #2798: Restore ORCID implementation

szymonlopaciuk:
**Merge on top of #2859.**

## Description
At the moment features a functionality to push single records to ORCID.
TODO: Push records coming from legacy to ORCID.

## Related Issue
Resolves  #2798.

## Checklist:


inspirehep/modules/orcid/models.py
```diff
@@ -0,0 +1,67 @@
...
+from __future__ import absolute_import, division, print_function
+
+from invenio_db import db
+
+
+class InspireOrcidPutCodes(db.Model):
+    __tablename__ = 'inspire_orcid_put_codes'
+
+    recid = db.Column(db.Integer, primary_key=True, index=True)
+    put_code = db.Column(db.Integer, nullable=False, index=True)
```
michamos: @szymonlopaciuk weren't you saying that put codes are different on different ORCID profiles, even if we have the same record for it? this would mean that you need to associate a put code not to an INSPIRE recid, but to a (recid, orcid) pair.

szymonlopaciuk: I was, thanks :+1:

PR #3136, opened 2018-01-22
global: addition of the ``cache`` module

drjova:
* Adds the ``cache`` module with a ``lock_cache`` proxy for
  easy access to the lock cache.

## Checklist:

### DIFF
--- a/inspirehep/modules/cache/__init__.py
+++ b/inspirehep/modules/cache/__init__.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""INSPIRE cache."""
+
+from __future__ import absolute_import, division, print_function
+
+from .proxies import lock_cache  # noqa: F401
--- a/inspirehep/modules/cache/providers/lock.py
+++ b/inspirehep/modules/cache/providers/lock.py
@@ -1,0 +1,83 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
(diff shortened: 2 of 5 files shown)


inspirehep/modules/cache/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+"""INSPIRE cache."""
+
+from __future__ import absolute_import, division, print_function
+
+from .proxies import lock_cache
+
+__all__ = (
+    'lock_cache',
+)
```
jacquerie: I never understood why Invenio does this. This only matters when you do `from inspirehep.modules.cache import *`, which you never want to do, so why do they do it?

tests/integration/cache/test_lock_cache.py
```diff
@@ -0,0 +1,65 @@
...
+    lock_cache.set('Defenders', 'Jessica Jones, Luke Cage')
+    # expected value
+    expected = 'Jessica Jones, Luke Cage'
+    # get the value
+    result = lock_cache.get('Defenders')
+
+    assert expected == result
+
+
+def test_lock_cache_deleteion():
```
jacquerie: `s/deleteion/deletion`

tests/integration/cache/test_lock_cache.py
```diff
@@ -0,0 +1,65 @@
...
+"""Test lock cache."""
+
+from __future__ import absolute_import, division, print_function
+
+from inspirehep.modules.cache import lock_cache
+
+
+def test_lock_cache_creation(app):
+    """Test set key."""
+    # set the key
```
jacquerie: Not a fan of this kind of comments, but not that big of a deal either.

tests/integration/cache/test_lock_cache.py
```diff
@@ -0,0 +1,65 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Test lock cache."""
```
jacquerie: Also not a fan of docstrings in tests (they tend to go stale really quickly!), but also not a big deal either.

david-caro:
> some stylistical differences

I think we are due a meeting to agree on these

inspirehep/modules/cache/__init__.py
drjova: just to make sure you're exporting what you need, but I agree it doesn't make sense in this case :)

drjova:
updated and rebased, thanks!

tests/integration/cache/test_lock_cache.py
drjova: are just for readability, I don't have a strong opinion for them I will remove them :)

inspirehep/modules/cache/providers/lock.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Args:
+            key (str): a key name.
+
+        Returns:
+            str: the value of the given key.
+        """
+        return current_cache.get(self._prefix(key))
+
+    def set(self, key, value, timeout=300):
```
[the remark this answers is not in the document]
drjova: 👍 IMHO 2h is too much, we've discussed with @StellaCh and 30mins (of inactivity) should be enough.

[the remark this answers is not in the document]
david-caro: Make sure again, I got the feedback from the catalogers directly (annete, kirsten, florian, heath...)

drjova:
ready to :shipit:

PR #3139, opened 2018-01-23
workflow: don't set `core` field if record is auto-approved

### LINKED ISSUE #3125: workflows: core should only be set to false when a human has chosen "accept"
## Expected Behavior

When a human pushes the "Accept" button in the holding pen as opposed to "Core", the record is flagged as non-core, that is `core: false` is set in the record. On the other hand, if it's decided that the record is accepted as non-core automatically based on the arXiv category it belongs to, the record should have no `core` field in the record at all.

## Current Behavior

`core: false` is written in the record in both cases.

## Context
@fschwenn reported this by email. This translates on legacy to records having `980:NONCORE` where they shouldn’t.

ammirate:
This closes #3125.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/matching.py
+++ b/inspirehep/modules/workflows/tasks/matching.py
@@ -25,7 +25,6 @@
 from __future__ import absolute_import, division, print_function
 
 import datetime
-from enum import Enum
 from functools import wraps
 
 from flask import current_app
@@ -40,12 +39,6 @@ from inspirehep.utils.record import get_arxiv_categories, get_value
 from inspirehep.modules.workflows.tasks.actions import mark
 
 from ..utils import with_debug_logging
-
-
-class Coreness(Enum):
-    core = True
-    non_core = False
-    non_relevant = None
 
 
 @with_debug_logging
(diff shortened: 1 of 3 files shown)


inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -113,45 +107,35 @@ def article_exists(obj, eng):
...
+    ) > 0
 
 
 @with_debug_logging
-def set_coreness_in_extra_data(obj, eng):
-    obj.extra_data['core'] = get_coreness(obj.data).value
+def set_core_in_extra_data(obj, eng):
+    """Set `core=True` in `obj.extra_data` if the record is core"""
+    def _is_core(record):
+        return len(
```
michamos: the `len` thing is not needed, as empty sets are falsy and non-empty sets are truthy.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -113,45 +107,60 @@ def article_exists(obj, eng):
...
+        record_categories &
+        set(
+            harvested_categories.get('core') +
+            harvested_categories.get('non-core')
+        )
+    ) > 0
+
+
+def physics_data_an_is_primary_category(record):
+    record_categories = set(get_arxiv_categories(record))
```
michamos: if you don't cast to a `set` here, you can reuse the variable later.

ammirate: copy/paste leftover :+1:

tests/unit/workflows/test_workflows_tasks_matching.py
```diff
@@ -131,12 +125,104 @@ def test_belongs_to_relevant_category_returns_true_and_is_non_core(app):
                 },
             ],
         }
+        assert validate(record['arxiv_eprints'], subschema) is None
+        assert has_fully_harvested_category(record)
+
+
+def test_has_fully_harvested_category_is_falsa_with_others_categories(app):
```
michamos: `falsa` -> `false` :smile:

PR #3143, opened 2018-01-24
procfile: adapt to new harvests queue

chris-asl:
Add harvests queue in Procfile so that our Docker configuration is in
sync with the native one (following what #3137 introduced).

## Caveats
Keep in mind that although this PR introduces the `harvests` queue, it doesn't set the [`APP_LUST_RUNS_PATH`](https://github.com/david-caro/inspire-next/blob/2ffd1faf37d837911ce29bb5935f000f374dff66/docker-compose.yml#L41) env variable.  

I thought of moving the env-vars from docker-compose to a `.env` file, which is [supported already by Docker-Compose](https://docs.docker.com/compose/environment-variables/#the-env-file). 
This will also allow us to use the same `.env` file from `honcho` which [already supports](http://honcho.readthedocs.io/en/latest/using_procfiles.html#environment-files) that, as well.

But another problem rises, in the sense that some env-vars can't be the same, like:
- `APP_SQLALCHEMY_DATABASE_URI`
- `APP_SEARCH_ELASTIC_HOSTS` 
- `APP_OAIHARVESTER_WORKDIR` 

## Checklist:

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -22,7 +22,7 @@
 
 web: gunicorn inspirehep.wsgi -c gunicorn.cfg
 cache: redis-server
-worker: celery worker -E -A inspirehep.celery --loglevel=INFO --workdir="${VIRTUAL_ENV}" --autoreload --pidfile="${VIRTUAL_ENV}/worker.pid" --purge -Q celery,migrator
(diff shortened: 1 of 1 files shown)


jacquerie:
Strictly speaking this isn't really needed, as I'm not aware of anyone who has the harvesting machinery running on a native installation (except @david-caro in production...), but let's strive to keep these files in sync as much as possible to prevent mistakes like https://github.com/inspirehep/inspire-next/pull/3093 from happening.

david-caro:
I run harvesting also locally with docker ;)

PR #3145, opened 2018-01-26
workflows: fix fix_submission_number

jacquerie:
## Description
Fixes the bug that will be explained in https://github.com/inspirehep/inspire-next/pull/2753#discussion_r164050594.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -276,7 +276,7 @@ def fix_submission_number(obj, eng):
     """
     method = get_value(obj.data, 'acquisition_source.method', default='')
     if method == 'hepcrawl':
-        obj.data['acquisition_source']['submission_number'] = obj.id
+        obj.data['acquisition_source']['submission_number'] = str(obj.id)
 
 
 @with_debug_logging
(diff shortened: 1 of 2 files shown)


david-caro:
I think that the solution to this should not be relaxing the submission_number field, but avoid overriding it with semantics instead. For example, not putting there the hecrwal job_id sometimes and the workflow id sometimes, but accommodating both in the data model, for example, having a field for the job_id (that iirc we already have), and onether field for the workflow id, or, if both should be part of the acquisition source, then having there a list of objects, something like:
```json
"acquisition_source": {
    "workfow_id": 1234,
    "crawljob_id": "osentusnth",
}
```

That way the ambiguous `submission_id` in not needed, and you keep all the info you want. If the goal of the `submission_id` was to identify a submission, then it should be another new uid that's generated when submitting (like a tracker or something). But we are not doing that, and afaik, we don't have that use case.

jacquerie:
> accommodating both in the data model

Yes, this is what should have happened. Here's the steps to make it happen:
1. Add the job id field in the schemas (@michamos).
2. Modify HEPCrawl to populate this field instead of the submission number.
3. Change the docstring of `fix_submission_number` to better reflect the new job that this function has to do anyway (we still want to have a reference to the workflow object).

Issue #3080, opened 2017-12-21
Citation count task creates new ES versions without creating DB ones for records, leaving both unsynched


david-caro:
https://sentry.inspirehep.net/inspire-sentry/prod/issues/860/

```
BulkIndexError: (u'1 document(s) failed to index.', [{u'index': {u'status': 409, u'_type': u'hep', u'_id': u'c8caca16-190a-4c57-b8fe-c54ed6bc9f1d', u'error': {u'index': u'records-hep', u'reason': u'[hep][c8caca16-190a-4c57-b8fe-c54ed6bc9f1d]: version conflict, current [7], provided [6]', u'type': u'version_conflict_engine_exception', u'shard': u'1'}, u'_index': u'records-hep'}}])
(2 additional frame(s) were not displayed)
...
  File "inspirehep/modules/migrator/tasks.py", line 231, in migrate_chunk
    request_timeout=req_timeout,

BulkIndexError: (u'1 document(s) failed to index.', [{u'index': {u'status': 409, u'_type': u'hep', u'_id': u'c8caca16-190a-4c57-b8fe-c54ed6bc9f1d', u'error': {u'index': u'records-hep', u'reason': u'[hep][c8caca16-190a-4c57-b8fe-c54ed6bc9f1d]: version conflict, current [7], provided [6]', u'type': u'version_conflict_engine_exception', u'shard': u'1'}, u'_index': u'records-hep'}}])
```

PR #3152, opened 2018-01-27
migrator: don't create another version on ES

jacquerie:
## Description
Avoid creating another version on Elasticsearch when adding the
citation counts because this creates an inconsistency between
the version in the DB and the version on ES, which means that
an exception will be raised by the indexing receiver since ES
will reject an update if its version is lower than the current.

The specific mechanism used (``_version_type: force``) has been
deprecated in ES@2 and removed in ES@5, so we will need to find
another solution when we will upgrade Elasticsearch.

## Related Issue
Closes https://github.com/inspirehep/inspire-next/issues/3080.

## Checklist:
- [ ] All new and existing tests passed

### DIFF
--- a/inspirehep/modules/migrator/tasks.py
+++ b/inspirehep/modules/migrator/tasks.py
@@ -267,6 +267,7 @@ def add_citation_counts(chunk_size=500, request_timeout=120):
                     '_index': index,
                     '_type': doc_type,
                     '_id': str(uuid),
+                    '_version_type': 'force',  # XXX: removed in ES@5
                     'doc': {'citation_count': citation_count}
                 }
 


inspirehep/modules/migrator/tasks.py
```diff
@@ -267,6 +267,7 @@ def _get_records_to_update_generator(citations_lookup):
                     '_index': index,
                     '_type': doc_type,
                     '_id': str(uuid),
+                    '_version_type': 'force',  # XXX: removed in ES@5
```
jacquerie: This is not enough: we also need to pass the current version in order for it to be preserved.

jacquerie:
I tried all possible dirty tricks I could think of to avoid prefetching the versions (like we do for UUIDs), but they all failed. I think we really have to go with @drjova's solution here, @chris-asl's one is not a quick fix after all...

chris-asl:
@jacquerie did you manage to reproduce the error that we had on Sentry? I couldn't do it on my dev-env.

Apart from that, what I tried was adding `params={'version_type': 'force'}`  [here](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/migrator/tasks.py#L305-L311).

With that solution, the equivalent test suite `tests/integration/test_citation_counts.py` passed.

jacquerie:
> Apart from that, what I tried was adding `params={'version_type': 'force'}`  [here](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/migrator/tasks.py#L305-L311).
With that solution, the equivalent test suite `tests/integration/test_citation_counts.py` passed.

I don't think this has any effect, though: a new version is created on ES anyway...

PR #3167, opened 2018-02-06
workflows: refactor fulltext download

harunurhan:
Workflow goes like this.

1. Populate `documents` based on type (submission or arxiv)
2. Download populate documents and replace url with internal url
3. Refaxtract

This also downloads documents for records coming from DESY by default,
in a more clean way.

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -280,33 +280,44 @@ def fix_submission_number(obj, eng):
 
 
 @with_debug_logging
-def submission_fulltext_download(obj, eng):
+def populate_submission_document(obj, eng):
     submission_pdf = obj.extra_data.get('submission_pdf')
     if submission_pdf and is_pdf_link(submission_pdf):
         filename = secure_filename('fulltext.pdf')
-        pdf = download_file_to_workflow(
-            workflow=obj,
-            name=filename,
+        obj.data['documents'] = [
+            document for document in obj.data.get('documents', ())
+            if document.get('key') != filename
+        ]
+        lb = LiteratureBuilder(
+            source=obj.data['acquisition_source']['source'], record=obj.data)
+        lb.add_document(
+            filename,
+            fulltext=True,
             url=submission_pdf,
+            original_url=submission_pdf,
         )
+        obj.data = lb.record
 
-        if pdf:
-            obj.data['documents'] = [
-                document for document in obj.data.get('documents', ())
-                if document.get('key') != filename
-            ]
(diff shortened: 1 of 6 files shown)


tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -35,59 +35,96 @@
...
+        'key': '1605.03844.pdf',
+        'fulltext': True,
+        'hidden': True,
+        'material': 'preprint',
+        'original_url': expected_url,
+        'url': expected_url,
+    }
+    documents = obj.data['documents']
+    assert 1 == len(documents)
+    assert expected_document['url'] == documents[0]['url']
```
drjova: It might be better to improve the assertion in these cases, something to compare the whole object WDTY?

harunurhan: @drjova aha, good catch forget about this 😄

drjova:
@harunurhan could you please also rebase? :P

drjova:
@harunurhan could you please check travis and rebase? :)

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -63,14 +63,8 @@
 
 @with_debug_logging
 @backoff.on_exception(backoff.expo, DownloadError, base=4, max_tries=5)
```
michamos: shouldn't the backoff be on the new download function (also?)?

harunurhan: `download_document` never throws `DownloadError` explicitly. That's why I thought it was only necessary for this one.

harunurhan: `download_file_to_workflow` which is used there, already retries for `ProtocolError`

drjova: @michamos shall we :shipit: ?

michamos: OK for me!

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -280,33 +280,44 @@ def fix_submission_number(obj, eng):
...
-                filename,
-                fulltext=True,
-                original_url=submission_pdf,
-                url='/api/files/{bucket}/{key}'.format(bucket=obj.files[filename].bucket_id, key=filename)
+
+@with_debug_logging
+def download_documents(obj, eng):
+    documents = obj.data.get('documents')
+    if documents:
+        for document in documents:
```
michamos: for future reference, a neater way to do this is to use `get('documents', [])`, which allows one not to have an `if`, but instead iterate directly on the (potentially empty) list.

PR #3168, opened 2018-02-06
serializers: BibTeX: fail early if no texkey

szymonlopaciuk:
## Description
Do not pass an empty texkey to the library, instead fail early if that happens.

## Related Issue
https://sentry.inspirehep.net/inspire-sentry/prod/issues/2606/

## Checklist:

### DIFF
--- a/inspirehep/modules/records/serializers/schemas/base.py
+++ b/inspirehep/modules/records/serializers/schemas/base.py
@@ -24,15 +24,15 @@
 
 from __future__ import absolute_import, division, print_function
 
-import logging
-
 from pybtex.database import Entry, Person
 from six import text_type
+
+from inspire_utils.logging import getStackTraceLogger
 
 from ..fields_export import get_authors_with_role, extractor, bibtex_type_and_fields
 
 
-LOGGER = logging.getLogger(__name__)
+LOGGER = getStackTraceLogger(__name__)
 
 
 class PybtexSchema(object):
--- a/inspirehep/modules/records/serializers/schemas/base.py
+++ b/inspirehep/modules/records/serializers/schemas/base.py
@@ -27,8 +27,6 @@ from __future__ import absolute_import, division, print_function
 from pybtex.database import Entry, Person
 from six import text_type
 
-from inspire_utils.record import get_value
-
 from ..fields_export import get_authors_with_role, extractor, bibtex_type_and_fields
 
 
@@ -49,7 +47,7 @@ class PybtexSchema(object):
             pybtex.database.Entity: Pybtex entity
         """
         doc_type, fields = bibtex_type_and_fields(record)
-        texkey = get_value(record, 'texkeys[0]')
+        texkey = record['texkeys'][0]
 
         template_data = []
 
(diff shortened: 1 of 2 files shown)


michamos:
On legacy, we were using the recid instead of the texkey if there is no texkey. That would probably be a more graceful way to fail.

inspirehep/modules/records/serializers/schemas/base.py
```diff
@@ -49,7 +47,7 @@ def load(self, record):
             pybtex.database.Entity: Pybtex entity
         """
         doc_type, fields = bibtex_type_and_fields(record)
-        texkey = get_value(record, 'texkeys[0]')
+        texkey = record['texkeys'][0]
```
puntonim: Love it: let's not use custom code for operations provided by Python itself.

michamos: `get_value` is not meant for that, it allows accessing fields deeper inside.
`val = get_value(record, 'foo.bar.baz[3]', default=42)` can be done with
```python
try:
    val = record.get('foo', {'bar': {'baz': [_, _, _, 42]}).get('bar', {'baz':  [_, _, _, 42]})[3]
except KeyError:
    val = 42
```
(or equivalently nested `try...except` blocks) which is a lot more cumbersome.

puntonim: I agree that it was a misuse of `get_value ()`.

jacquerie:
@szymonlopaciuk Interestingly, https://labs.inspirehep.net/api/literature/1652864/db now has a texkey. Since I can confirm that it wasn't there when you showed me, I guess that, on Legacy, there's an interval between which a record is created and its texkey is minted. Your testing happened right in the middle of these two events...

inspirehep/modules/records/serializers/schemas/base.py
david-caro: In the case above though, I agree with @puntonim, the expression it's simple enough to not benefit from `get_value` (with the added side effect of failing early if there's no `texkeys` key, though I'm not 100% sure that's expected).

puntonim: So let's try to have this fixed if necessary and merged tomorrow. @szymonlopaciuk does it make sense to go on and build the `Entry` returned in the `data` tuple or it makes more sense to raise a custom exception?

szymonlopaciuk: @puntonim IMO it makes more sense to carry on, following @michamos' suggestion, and replacing the texkey with the record ID. We could maybe log such case so that we know there's a problem, but I think it's better to output something over nothing. I'll make a fix tomorrow.

inspirehep/modules/records/serializers/schemas/base.py
```diff
@@ -24,14 +24,17 @@
 
 from __future__ import absolute_import, division, print_function
 
+import logging
```
michamos: shouldn't you use the logger from `inspire_utils`?

david-caro:
ci merge please

PR #3169, opened 2018-02-06
workflows: add validation at the beginning of the workflow

iulianav:
## Description

I removed the validation from the submission form and added it to the beginning of the workflow. If the record is not valid, the workflow is stopped, put in `ERROR` state, and the received `ValidationaError` is caught and its message is stored in `obj.extra_data[_error_msg]` in order to help with identifying which part of the record is not schema compliant. Then, a `WorkflowsError` is raised with a custom message (`The record contained in the workflow is not schema compliant.` ) in order to clarify what went wrong.

## Related PR

https://github.com/inspirehep/hepcrawl/pull/223/

Records harvested by DESY spider will now be validated at the beginning of the workflow.

## Motivation and Context

This is required because so far, if a record was not valid either in hepcrawl or submission form, a `ValidationError` would be thrown and the actual record would be lost. Moreover, it was hard to identify such errors because they were rather silent and required some investigation on Sentry to be brought to light.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -229,18 +229,20 @@ def is_submission(obj, eng):
     """Is this a submission?"""
     source = obj.data.get('acquisition_source')
     if source:
-        return source.get('method') == "submitter"
+        return source.get('method') == 'submitter'
     return False
 
 
 @with_debug_logging
-def is_valid(obj, *args, **kwargs):
-    """Check if the record is schema compliant."""
+def stop_in_error_if_record_not_valid(obj, eng):
+    """Check if the record is schema compliant and stop the workflow in ERROR state it it is not."""
     try:
-        if validate(obj.data, 'hep') is None:
-            return True
-    except ValidationError:
-        return False
+        validate(obj.data, 'hep')
+    except ValidationError as err:
+        obj.log.error(err.message)
+        obj.extra_data['_error_message'] = err.message
+        obj.status = ObjectStatus.ERROR
+        raise WorkflowsError('The record contained in the workflow is not schema compliant.')
 
 
 @with_debug_logging
@@ -402,7 +404,7 @@ def error_workflow(message):
         obj.log.error(message)
         obj.extra_data['_error_message'] = message
         obj.status = ObjectStatus.ERROR
-        raise WorkflowsError
+        raise WorkflowsError(message)
 
     _error_workflow.__doc__ = (
         'Force an error in the workflow object with the message "%s".'
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -240,7 +240,7 @@ def stop_in_error_if_record_not_valid(obj, eng):
         validate(obj.data, 'hep')
     except ValidationError as err:
         obj.log.error(err.message)
-        obj.extra_data['_error_message'] = err.message
+        obj.extra_data['_error_msg'] = err.message
         obj.status = ObjectStatus.ERROR
         raise WorkflowsError('The record contained in the workflow is not schema compliant.')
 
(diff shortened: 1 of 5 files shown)


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -230,6 +233,16 @@ def is_submission(obj, eng):
...
 
 
+@with_debug_logging
+def is_valid(obj, *args, **kwargs):
+    """Check if the record is schema compliant."""
+    try:
+        if validate(obj.data, 'hep') is None:
+            return True
+    except ValidationError:
+        return False
```
michamos: You are losing all information about the validation error by doing it like this. Instead, you should somehow log the info about the actual validation error.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -412,3 +414,50 @@ def test_arxiv_update_is_not_store_on_legacy_and_labs(
...
+    obj_id = obj.id
+
+    with pytest.raises(WorkflowsError):
+        start('article', invalid_record, obj_id)
+
+    obj = workflow_object_class.get(obj_id)
+
+    assert obj.status == ObjectStatus.ERROR
+    assert obj.extra_data['invalid-record'] is True
+    assert obj.extra_data['_error_message'] == 'Workflow record is not schema compliant.'
```
jacquerie: Expanding @michamos' comment above, this is where I'd expect to see the validation error message, not something that repeats the same thing that we know from the flag `invalid-record`...

iulianav: Problem is I noticed that the ValidationError message tends to be huge sometimes, but I will think up something.

bavaria95:
ci test please

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -412,3 +414,52 @@ def test_arxiv_update_is_not_store_on_legacy_and_labs(
...
+        start('article', invalid_record, obj_id)
+
+    obj = workflow_object_class.get(obj_id)
+
+    assert obj.status == ObjectStatus.ERROR
+    assert '_error_msg' in obj.extra_data
+
+
+def test_article_workflow_when_record_is_valid(workflow_app):
+    invalid_record = {
```
iulianav: s/invalid/valid/

david-caro:
ci test please

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -387,8 +402,9 @@ def error_workflow(message):
     @wraps(error_workflow)
     def _error_workflow(obj, eng):
         obj.log.error(message)
-        obj.extra_data['_error_message'] = message
+        obj.extra_data['_error_msg'] = message
```
iulianav: @jacquerie I will remove line 405 and 406, here and also in `error_workflow` action because it is already done as I raise a `WorkflowsError`. Check [this] (https://github.com/inveniosoftware-contrib/invenio-workflows/blob/1b774eb6d61ec75b4b83d06cc59b7e3f919c2db5/invenio_workflows/engine.py#L342). Please correct me if I am wrong.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -239,10 +238,9 @@ def stop_in_error_if_record_not_valid(obj, eng):
     try:
         validate(obj.data, 'hep')
     except ValidationError as err:
-        obj.log.error(err.message)
-        obj.extra_data['_error_msg'] = err.message
-        obj.status = ObjectStatus.ERROR
-        raise WorkflowsError('The record contained in the workflow is not schema compliant.')
+        message = 'The record contained in the workflow is not schema compliant: \n' + err.message
+        error_workflow_fct = error_workflow(message)
+        error_workflow_fct(obj, eng)
```
jacquerie: Calling a workflow step in a workflow step is possible, but probably not a good idea...

PR #3178, opened 2018-02-08
mappings: add page_artid to publication_info

iulianav:
## Description

Add new `publication_info.artid_and_page_start` field to enable easier queries based on `publication_info.artid` and `publication_info.page_start`. Because the data is ambiguous, the 2 fields are sometimes populated with the same value, different values or one of them is not populated at all.

## Motivation and Context

 By using `copy_to` to copy the *values* stored in each one of the 2 fields to `publication_info.artid_and_page_start`, we can now query a single field instead of all the possible combinations of the other 2 fields. This is needed for more accurate number ( keeping track of citations, references etc).

## Checklist:

### DIFF
--- a/inspirehep/modules/records/mappings/records/hep.json
+++ b/inspirehep/modules/records/mappings/records/hep.json
@@ -628,6 +628,7 @@
                 "publication_info": {
                     "properties": {
                         "artid": {
+                            "copy_to": "publication_info.page_artid",
                             "type": "string"
                         },
                         "cnum": {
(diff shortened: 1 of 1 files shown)


inspirehep/modules/records/mappings/records/hep.json
```diff
@@ -628,6 +628,10 @@
                 "publication_info": {
                     "properties": {
                         "artid": {
+                            "type": "string",
+                            "copy_to": "publication_info.artid_and_page_start"
+                        },
+                        "artid_and_page_start": {
```
jacquerie: We already have a name for this concept, can you change it to `page_artid` ?

PR #3180, opened 2018-02-09
workflows: Better document selection

michamos:
## Description

* The selection of the document used for further metadata extraction
(references, keywords) is now based on the `documents` metadata instead
of the file extension.
* Also refactors the `retrieve_uri` into a context manager, and does the minor refactor suggested in https://github.com/inspirehep/inspire-next/pull/3167/files#r166949783.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/arxiv.py
+++ b/inspirehep/modules/workflows/tasks/arxiv.py
@@ -124,8 +124,8 @@ def arxiv_plot_extract(obj, eng):
     tarball = obj.files[filename]
 
     if tarball:
-        with TemporaryDirectory(prefix='plot_extract') as scratch_space:
-            tarball_file = retrieve_uri(tarball.file.uri, outdir=scratch_space)
+        with TemporaryDirectory(prefix='plot_extract') as scratch_space, \
+                retrieve_uri(tarball.file.uri, outdir=scratch_space) as tarball_file:
             try:
                 plots = process_tarball(
                     tarball_file,
@@ -227,11 +227,8 @@ def arxiv_author_list(stylesheet="authorlist2marcxml.xsl"):
             )
             return
 
-        with TemporaryDirectory(prefix='author_list') as scratch_space:
-            tarball_file = retrieve_uri(
-                tarball.file.uri,
-                outdir=scratch_space,
-            )
+        with TemporaryDirectory(prefix='author_list') as scratch_space, \
+                retrieve_uri(tarball.file.uri, outdir=scratch_space) as tarball_file:
             try:
                 file_list = untar(tarball_file, scratch_space)
             except InvalidTarball:
--- a/inspirehep/modules/workflows/utils/__init__.py
+++ b/inspirehep/modules/workflows/utils/__init__.py
@@ -28,7 +28,7 @@ import json
 import logging
 import os
 import traceback
-from contextlib import closing
+from contextlib import closing, contextmanager
 from functools import wraps
 
 import backoff
@@ -153,14 +153,17 @@ def with_debug_logging(func):
     return _decorator
 
 
+@contextmanager
 @with_debug_logging
 def get_pdf_in_workflow(obj):
     """Return the fullpath to the PDF attached to a workflow object"""
     for filename in obj.files.keys:
         if filename.endswith('.pdf'):
-            return retrieve_uri(obj.files[filename].file.uri)
+            with retrieve_uri(obj.files[filename].file.uri) as local_file:
+                yield local_file
 
     obj.log.info('No PDF available')
+    yield None
 
 
 @backoff.on_exception(backoff.expo, requests.packages.urllib3.exceptions.ProtocolError, max_tries=5)
--- a/inspirehep/modules/workflows/utils/__init__.py
+++ b/inspirehep/modules/workflows/utils/__init__.py
@@ -155,15 +155,33 @@ def with_debug_logging(func):
 
 @contextmanager
 @with_debug_logging
-def get_pdf_in_workflow(obj):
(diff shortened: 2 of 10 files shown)


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -124,8 +124,8 @@ def arxiv_plot_extract(obj, eng):
     tarball = obj.files[filename]
 
     if tarball:
-        with TemporaryDirectory(prefix='plot_extract') as scratch_space:
-            tarball_file = retrieve_uri(tarball.file.uri, outdir=scratch_space)
+        with TemporaryDirectory(prefix='plot_extract') as scratch_space, \
+                retrieve_uri(tarball.file.uri, outdir=scratch_space) as tarball_file:
```
david-caro: Interesting, I did not know that you could use the `as` from the first element in the second element.

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -153,14 +153,35 @@ def _try_to_log(logfn, *args, **kwargs):
...
+        present, it prioritizes the fulltext. If several documents with the
+        same priority are present, it takes the first one and logs an error.
+    """
+    documents = obj.data.get('documents', [])
+    fulltexts = [document for document in documents if document.get('fulltext')]
+    documents = fulltexts or documents
+
+    if not documents:
+        obj.log.info('No document available')
+        yield None
```
david-caro: Just fyi, the `None` is the default if nothing passed.

tests/unit/helpers/mocks.py
```diff
@@ -156,3 +157,8 @@ class MockRole(object):
 
     def __init__(self, name):
         self.name = name
+
+
+@contextmanager
+def identity_context_manager(arg):
```
david-caro: I'd rename this to have the 'mock' word somewhere. I know it's in the `mocks` module, but once imported can mislead people.

tests/unit/workflows/test_workflows_tasks_actions.py
```diff
@@ -651,9 +649,10 @@ def test_refextract_from_pdf(mock_get_pdf_in_workflow):
     assert obj.data['references'][0]['raw_refs'][0]['source'] == 'arXiv'
 
 
-@patch('inspirehep.modules.workflows.tasks.actions.get_pdf_in_workflow')
-def test_refextract_from_text(mock_get_pdf_in_workflow):
-    mock_get_pdf_in_workflow.return_value = None
+@patch('inspirehep.modules.workflows.tasks.actions.get_document_in_workflow')
+def test_refextract_from_text(mock_get_document_in_workflow):
+    mock_get_document_in_workflow.return_value.__enter__.return_value = None
+    mock_get_document_in_workflow.return_value.__exit__.return_value = None
```
david-caro: is it needed to change the `__exit__` return value? Wouldn't it be just discarded?

tests/unit/workflows/test_workflows_utils.py
```diff
@@ -78,6 +81,78 @@ def test_json_api_request_retries_on_connection_error():
...
+    files['fulltext.xml'] = None
+    files['table_of_contents.pdf'] = None
+    obj = MockObj(data, {}, files=files)
+
+    with get_document_in_workflow(obj) as local_file:
+        assert local_file == files['fulltext.xml'].file.uri
+
+
+@patch('inspirehep.modules.workflows.utils.retrieve_uri', identity_context_manager)
+def test_get_document_in_workflow_takes_first_among_equals():
```
david-caro: The "prince"!

tests/unit/workflows/test_workflows_tasks_actions.py
michamos: `__exit__` is ran at the end of the context manager, and before I added this line I was getting errors because of it (don't remember the exact error though).

inspirehep/modules/workflows/utils/__init__.py
michamos: I know, but PEP8 says
>Be consistent in return statements. Either all return statements in a function should return an expression, or none of them should. If any return statement returns an expression, any return statements where no value is returned should explicitly state this as return None, and an explicit return statement should be present at the end of the function (if reachable).

inspirehep/modules/workflows/tasks/arxiv.py
michamos: Indeed, from https://docs.python.org/2/reference/compound_stmts.html#the-with-statement:
>With more than one item, the context managers are processed as if multiple with statements were nested:
>
>```
>with A() as a, B() as b:
>    suite
>```
>
>is equivalent to
>
>```
>with A() as a:
>    with B() as b:
>        suite
>```

tests/unit/helpers/mocks.py
michamos: so just `mock_context_manager`?

tests/unit/workflows/test_workflows_tasks_actions.py
david-caro: 👍

tests/unit/helpers/mocks.py
david-caro: I'd use something more specific, like `mock_retrieve_uri`.

bavaria95:
ci test please

inspirehep/modules/workflows/utils/__init__.py
david-caro: then you might be missing a `None` in the return right after this no?

michamos:
ci merge please

PR #3182, opened 2018-02-09
orcid: add celery task to push records

ammirate:
## Checklist:


inspirehep/modules/orcid/api.py
```diff
@@ -0,0 +1,236 @@
...
+        recid (string): HEP record to push
+        orcid (string): ORCID identifier to push onto
+        oauth_token (string): ORCID user OAUTH token
+        put_code (Union[string, NoneType]): put-code to push record onto,
+            if None will push as a new record
+
+    Returns:
+        string: the put-code of the updated/inserted item
+    """
+    server_name = app.config["SERVER_NAME"]
```
puntonim: If I understand correctly SERVER_NAME is a string like 'http://labs.inspirehep.net', right?
The name is very confusing, but I know this PR did not introduce this bad name.

inspirehep/modules/orcid/api.py
```diff
@@ -0,0 +1,236 @@
...
+        orcid_api.update_record(
+            orcid_id=orcid,
+            token=oauth_token,
+            request_type='work',
+            data=orcid_xml,
+            put_code=put_code,
+            content_type='application/orcid+xml',
+        )
+    else:
+        LOGGER.info(
```
puntonim: Why are we logging so much? Do we need it? Is this going o Sentry?
We probably only want to log in case of error.

david-caro: > Why are we logging so much?

I can't answer that one :)

> Do we need it?

I'd say it's useful to some extent (debugging, log archaeology after the fact...).

> Is this going o Sentry?

Nop, the info is not going to sentry, only the exception and error.

inspirehep/modules/orcid/api.py
david-caro: It comes from flask itself.

inspirehep/modules/orcid/utils.py
```diff
@@ -236,53 +78,39 @@ def _get_api_url_for_recid(server_name, api_endpoint, recid):
...
+            orcid_recid_key = get_orcid_recid_key(orcid, rec_id)
+            return r.get(orcid_recid_key)
 
-    Returns:
-        dict: BibTeX serialized record
-    """
-    bibtex_response = _get_record_by_mime(config, recid, 'application/x-bibtex')
-    return bibtex_response.text
+    except RedisLockError:
+        LOGGER.info("Push to ORCID failed for record {}".format(rec_id))
```
david-caro: Same as the ones `orcid_push.tasks`, I think that the exception should go up, and be handled by the final caller.

inspirehep/services/orcid_push/tasks.py
```diff
@@ -0,0 +1,88 @@
...
+
+    Args:
+        orcid(string): an orcid identifier.
+        rec_id(int): inspire record's id to push to ORCID.
+        oauth_token(string): orcid token.
+    """
+    put_code = get_putcode_from_redis(orcid, rec_id)
+
+    if not put_code:
+        record_put_codes, errors = get_author_putcodes(orcid, oauth_token)
```
david-caro: The errors are lost here, should they?

tests/integration/cache/test_utils.py
```diff
@@ -0,0 +1,40 @@
...
+def test_redis_locking_context_write_string_in_redis(app):
+    with redis_locking_context('my_lock') as redis:
+        redis.set('key', 'value')
+        assert redis.get('key') == 'value'
+
+
+def test_redis_locking_context_write_integer(app):
+    with redis_locking_context('my_lock') as redis:
+        redis.set('key', 1)
+        assert redis.get('key') == '1'
```
david-caro: I'm not sure these tests are useful here, they are testing if you can put stuff and retrieve stuff from redis.

I would either create unit tests, and assert that the lock is called (not very nice, though might be enough), or integration tests that actually check if the lock is put in redis.

inspirehep/modules/orcid/api.py
jacquerie: It's a core configuration variable in Flask: http://flask.pocoo.org/docs/0.12/config/

tests/integration/orcid/test_orcid_api.py
```diff
@@ -0,0 +1,135 @@
...
+    )
+    return mocked_internal_services
+
+
+def test_push_record_with_orcid(api_client, mock_orcid_api):
+    sample_orcid = '0000-0002-1825-0097'
+    expected_put_code = '895497'
+
+    result_put_code = push_record_with_orcid(
+        '4328', sample_orcid, 'fake-token', '895497'
```
david-caro: I would use the argument names in the call, it's not clear to me what is `'4328'` and what `'896597'`.

david-caro: btw. this one looks like it's just checking if push_record_with_orcid return the last of the parameters you passed to it xd

tests/integration/orcid/test_orcid_api.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def get_file(fixture_name):
+    """Get contents of fixture files"""
+    path = pkg_resources.resource_filename(
+        __name__,
+        os.path.join('fixtures', fixture_name)
+    )
+    with open(path, 'r') as file:
+        return file.read()
```
david-caro: Don't use `file` as a var name, better use something like `fixture_fd` (for file descriptor). I think though that there's somewhere in the code something that already does this fixture retrieving thingie.

Also the name of the method is misleading, I'd expect `get_file` to get me a `file`, instead is returning it's contents.

tests/integration/orcid/test_orcid_push.py
```diff
@@ -0,0 +1,104 @@
...
+import requests_mock
+import vcr
+
+from redis import StrictRedis
+
+from inspirehep.services.orcid_push.tasks import push_orcid
+
+
+@pytest.fixture
+def mocked_internal_services(api_client):
```
david-caro: This look quite similar to the other one in the api test, can we reuse them?

tests/integration/orcid/test_orcid_push.py
```diff
@@ -0,0 +1,104 @@
...
+        'ORCID_SANDBOX': True,
+        'SERVER_NAME': 'http://labs.inspirehep.net',
+    }):
+        # Push as new
+        assert not redis_setup.exists('orcid:0000-0002-2169-2152:4328')
+        push_orcid('0000-0002-2169-2152', rec_id, 'token-redacted')
+
+        # Push update
+        assert redis_setup.exists('orcid:0000-0002-2169-2152:4328')
+        push_orcid('0000-0002-2169-2152', rec_id, 'token-redacted')
```
david-caro: Shouldn't you assert something else? How do you know that the update did anything? And that the creation actually created something else than the redis entry?

inspirehep/modules/orcid/api.py
```diff
@@ -0,0 +1,236 @@
...
+        string: the put-code of the updated/inserted item
+    """
+    server_name = app.config["SERVER_NAME"]
+
+    record = _get_hep_record(app.config, recid)
+    if not record:
+        LOGGER.error('Cannot push record #{}, as metadata don\'t exist'.format(recid))
+
+    try:
+        bibtex = _get_bibtex_record(app.config, recid)
```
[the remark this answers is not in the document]
michamos: I don't think I ever said that, bibtex is another serialization of the records.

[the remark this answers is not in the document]
david-caro: I might have misheard 👍

inspirehep/modules/orcid/api.py
puntonim: Ah ok, then I love it!

inspirehep/services/orcid_push/tasks.py
```diff
@@ -0,0 +1,88 @@
...
+def store_record_in_redis(orcid, rec_id, put_code):
+    """Store the entry <orcid:recid, value> in Redis.
+
+    Args:
+        orcid(string): the author's orcid.
+        rec_id(int): inspire record's id pushed to ORCID.
+        put_code(string): the put_code used to push the record to ORCID.
+
+    Returns:
+        bool: True if the entry is set in Redis, False otherwise.
```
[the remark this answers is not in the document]
szymonlopaciuk: We are not actually using the return value anywhere, maybe we could drop this. If it's a cache then maybe we don't care enough about this to fail? @ammirate?

ammirate:
Ops, closed by mistake.

puntonim:
@szymonlopaciuk Can you check the module `modules/orcid/ext.py`? It seems not to be used anywhere. It's not part of this PR, yet we should clean it up if not used.

inspirehep/modules/orcid/api.py
```diff
@@ -0,0 +1,238 @@
...
+    user_works = api.read_record_member(
+        orcid,
+        'works',
+        oauth_token,
+        accept_type='application/orcid+json',
+    )
+
+    put_codes = [
+        str(item[0]) for item
+        in get_value(user_works, 'group.work-summary.put-code')
```
puntonim: This code breaks if `user_works` has no `group.work-summary.put-code`.
I'm not sure if this is even possible (it's ORCID's API data indeed) but maybe this would be safer for us:
`get_value(user_works, 'group.work-summary.put-code', [])`

inspirehep/modules/orcid/api.py
```diff
@@ -0,0 +1,238 @@
...
+    ]
+
+    # We can batch requests for _detailed_ records for maximum
+    # `WORKS_BULK_QUERY_LIMIT` put-codes at once:
+    detailed_works = {'bulk': []}
```
puntonim: Minor: this actually could be just a list as you are only using a single key.

inspirehep/modules/orcid/api.py
```diff
@@ -0,0 +1,238 @@
...
+
+        try:
+            url = get_value(item, 'work.url.value')
+            match = RECID_FROM_INSPIRE_URL.match(url)
+            recid = match.group(1)
+        except (TypeError, AttributeError):
+            errors.append(put_code)
+            continue
+
+        if put_code:
```
puntonim: This if clause is unnecessary

inspirehep/modules/orcid/utils.py
```diff
@@ -236,53 +78,39 @@ def _get_api_url_for_recid(server_name, api_endpoint, recid):
     return urljoin(api_url, recid)
 
 
-def _get_record_by_mime(config, recid, mime_type):
-    """
+def get_orcid_recid_key(orcid, rec_id):
+    """Return the string 'orcid:``orcid_value``:``rec_id``'"""
+    return 'orcid:{}:{}'.format(orcid, rec_id)
```
puntonim: The value in this key is a putcode, thus I would call this something like: `'orcidputcode:{}:{}'.format(orcid, rec_id)`

inspirehep/services/orcid_push/tasks.py
```diff
@@ -0,0 +1,99 @@
...
+
+    Args:
+        orcid(string): an orcid identifier.
+        rec_id(int): inspire record's id to push to ORCID.
+        oauth_token(string): orcid token.
+    """
+    put_code = get_putcode_from_redis(orcid, rec_id)
+
+    if not put_code:
+        record_put_codes, errors = get_author_putcodes(orcid, oauth_token)
```
puntonim: I don't see the point of returning `errors` only for the sake of logging. Isn't it more clear if `get_author_putcodes()` takes care of logging? Note that `push_record_with_orcid()` in the same module takes care of logging.

inspirehep/services/orcid_push/tasks.py
```diff
@@ -0,0 +1,99 @@
...
+        LOGGER.error(
+            'Failed to match putcodes {} from {} to HEP records.'.format(
+                ', '.join(errors), orcid
+            )
+        )
+
+        for rec_id, put_code in record_put_codes:
+            store_record_in_redis(orcid, rec_id, put_code)
+
+        put_code = get_putcode_from_redis(orcid, rec_id)
```
puntonim: We have just stored the putcodes in redis, why are we getting again here?

inspirehep/services/orcid_push/tasks.py
```diff
@@ -0,0 +1,99 @@
...
+
+    put_code = push_record_with_orcid(str(rec_id), orcid, oauth_token, put_code)
```
puntonim: Suppose at this point we have `put_code`. We first call `push_record_with_orcid()`, which returns the same `put_code` and then we store it in Redis in the next line. The storing is unnecessary in this case. Or am I missing something?

inspirehep/services/orcid_push/tasks.py
szymonlopaciuk: We need to get the particular one. Do you think this operation is expensive? Alternative is to get the right put-code in the preceding for-loop.

puntonim: It's a network call but not particularly expensive, thus consider this comment as minor (and leave it the way it is right now).

szymonlopaciuk:
ci merge please

PR #3194, opened 2018-02-15
arxiv: decode latex charachters in authors' full_name

ammirate:

## Checklist:


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -32,26 +32,25 @@
...
 from inspirehep.modules.workflows.errors import DownloadError
 from inspirehep.modules.workflows.utils import (
     convert,
     download_file_to_workflow,
     with_debug_logging,
 )
-
+from inspirehep.utils.latex import decode_latex
+from inspirehep.utils.record import get_arxiv_categories, get_arxiv_id
+from inspirehep.utils.url import is_pdf_link, retrieve_uri
```
jacquerie: All these import changes are not needed. You just need to add
```
from inspirehep.utils.latex import decode_latex
```

inspirehep/utils/latex.py
```diff
@@ -22,6 +22,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+import latexcodec  # noqa: F401
```
jacquerie: Yikes, is this package really modifying state at import time? That sounds like a really bad idea. Is there another package that we can use instead with a more sensible API?

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -264,6 +263,10 @@ def _author_list(obj, eng):
                     extracted_authors.extend(marcxml2record(authors_xml).get('authors', []))
 
             if extracted_authors:
+
```
jacquerie: I can't think of a reason to separate an `if` from its block: they logically go together, so they should be together in the code. So:
```python
if extracted_authors:
    for author in extracted_authors:
        author['full_name'] = decode_latex(author['full_name'])
    obj.data['authors'] = extracted_authors
```

inspirehep/utils/latex.py
jacquerie: And if instead the only option we have is this package, can we at least ensure that this import doesn't happen deep in the code, where it might happen for some code paths but not others?

inspirehep/modules/workflows/tasks/arxiv.py
ammirate: They were all already there, my editor just decided to move them around. :+1:

inspirehep/utils/latex.py
ammirate: @michamos investigated for a while and this seemed the best one. 

> can we at least ensure that this import doesn't happen deep in the code

What do you mean?

michamos: it's doing only [`codecs.register`](https://github.com/mcmtroffaes/latexcodec/blob/develop/latexcodec/codec.py#L70-L75), on import, which is needed in order to plug into `str.encode` etc. Is that OK or should it be imported differently?

inspirehep/utils/latex.py
```diff
@@ -294,3 +295,12 @@ def _get_pubnote(self):
...
             return ''
+
+
+def decode_latex(string_with_latex):
+    return string_with_latex\
+        .encode('utf-8')\
+        .decode('latex')\
+        .replace('{', '')\
+        .replace('}', '')\
+        .encode('utf-8')
```
michamos: why are you encoding at the end?

inspirehep/utils/latex.py
```diff
@@ -294,3 +295,12 @@ def _get_pubnote(self):
                     return ''
         else:
             return ''
+
+
+def decode_latex(string_with_latex):
```
michamos: maybe add a short docstring?

drjova:
@michamos updated

inspirehep/utils/latex.py
```diff
@@ -294,3 +295,19 @@ def _get_pubnote(self):
                     return ''
         else:
             return ''
+
+
+def decode_latex(latex_text):
+    """Decode letex text.
```
michamos: typo

szymonlopaciuk:
ci merge please

PR #3198, opened 2018-02-21
authors: validate at the beginning of the workflow

iulianav:
## Description

I removed the validation from the submission form and added it to the beginning of the authors workflow. If the record is not valid, a `ValidationaError` is raised, its message is stored in `obj.extra_data['_error_msg']` in order to help with identifying which part of the record is not schema compliant, the workflow is stopped, put in `ERROR` state. 

## Motivation and Context

This is required because so far, if a record was not valid either in hepcrawl or submission form, a `ValidationError` would be raised before the workflow started and the actual record was lost. Moreover, it was hard to identify such errors because they were rather silent and required some investigation on Sentry to be brought to light.

## Checklist:

### DIFF
--- a/tests/integration/workflows/test_authors_workflow.py
+++ b/tests/integration/workflows/test_authors_workflow.py
@@ -1,0 +1,81 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2018 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+import pytest
+
+from invenio_workflows import (
+    ObjectStatus,
+    start,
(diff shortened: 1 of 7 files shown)


tests/integration/workflows/test_authors_workflow.py
```diff
@@ -0,0 +1,83 @@
...
+from invenio_workflows import (
+    ObjectStatus,
+    start,
+    workflow_object_class,
+)
+
+from jsonschema import ValidationError
+
+
+def test_authors_workflow_stops_when_record_is_not_valid(workflow_app):
```
jacquerie: `s/workflow_app/isolated_app/`, because we want to achieve more test independence.

tests/integration/workflows/test_authors_workflow.py
```diff
@@ -0,0 +1,83 @@
...
+        start('author', invalid_record, obj_id)
+
+    obj = workflow_object_class.get(obj_id)
+
+    assert obj.status == ObjectStatus.ERROR
+    assert '_error_msg' in obj.extra_data
+    assert 'required' in obj.extra_data['_error_msg']
+
+
+def test_authors_workflow_continues_when_record_is_valid(workflow_app, mocked_external_services):
```
jacquerie: Same as above.

tests/integration/workflows/test_authors_workflow.py
```diff
@@ -0,0 +1,83 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Tests for authors workflows."""
```
jacquerie: Nit: this is just repeating the information contained in the filename, so I'd say it's unnecessary.

iulianav:
The current integration wf tests that I wrote are not what we need. Instead of using `workflow_app`, they should use `isolated_app`, but that results in an ERROR ( `File "/virtualenv/lib/python2.7/site-packages/flask_alembic/extension.py", line 107, in _get_cache
    return self._cache[current_app._get_current_object()]
KeyError: <Flask 'inspirehep'> ` ) due to poor design and technical debt. Something similar to [this]() should work in the future.

jacquerie:
ci merge please

PR #3204, opened 2018-02-22
Abstract legacy URL building

szymonlopaciuk:
## Description

- This introduces a `get_prod_url_for_recid` util to make record URLs from recids.
- In addition to this, API of OrcidConverter is changed to keep it flask-less and seperate from the rest of the application. We cannot use the util in there directly, so this proposes to instead pass a URL pattern and uses inspire_utils, which will allow us to seperate it from the main app in the future.

## Checklist:

### DIFF
--- a/inspirehep/modules/orcid/api.py
+++ b/inspirehep/modules/orcid/api.py
@@ -54,8 +54,6 @@ def push_record_with_orcid(recid, orcid, oauth_token, put_code=None):
     Returns:
         string: the put-code of the updated/inserted item
     """
-    server_name = app.config["SERVER_NAME"]
-
     record = _get_hep_record(app.config, recid)
 
     try:
@@ -68,7 +66,7 @@ def push_record_with_orcid(recid, orcid, oauth_token, put_code=None):
     orcid_api = _get_api()
 
     orcid_xml = OrcidConverter(
-        record, server_name, put_code=put_code, bibtex_citation=bibtex
+        record, app.config['LEGACY_RECORD_URL_PATTERN'], put_code=put_code, bibtex_citation=bibtex
     ).get_xml()
 
     if put_code:
@@ -151,7 +149,7 @@ def _get_bibtex_record(config, recid):
         recid (string): HEP record ID
 
     Returns:
-        dict: BibTeX serialized record
+        string: BibTeX serialized record
     """
     bibtex_response = _get_record_by_mime(config, recid, 'application/x-bibtex')
     return bibtex_response.text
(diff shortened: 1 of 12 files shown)


inspirehep/utils/url.py
```diff
@@ -89,3 +91,19 @@ def retrieve_uri(uri, outdir=None):
...
+        recid (Union[int, string]): record ID
+        base_url_config_var (string): where to look for the base url
+
+    Return:
+        text_type: URL, doesn't end with a slash if `recid` is empty
+    """
+    recid = text_type(recid)
+    url = os.path.join(current_app.config[base_url_config_var], 'record', recid)
+
+    return url if recid else url[:-1]
```
drjova: IMO it's safer to use ``urljoin`` :)
```python
from six.moves.urllib.parse import urljoin
```

inspirehep/modules/orcid/api.py
```diff
@@ -68,7 +67,7 @@ def push_record_with_orcid(recid, orcid, oauth_token, put_code=None):
     orcid_api = _get_api()
 
     orcid_xml = OrcidConverter(
-        record, server_name, put_code=put_code, bibtex_citation=bibtex
+        record, get_url_for_record, put_code=put_code, bibtex_citation=bibtex
```
drjova: We don't need to pass around the function :), you can parametrize it as an ``import_string``

szymonlopaciuk: Do you mean import it in-place? Otherwise, I wanted to prevent adding more dependencies inside of the converter?

drjova: Not in place but as an ``import_string`` in a ``config`` var, as a factory. This way it can easily change and you don't have a strong dependency.

szymonlopaciuk:
@drjova yes, I'm working on it, will update and push soon :+1:

inspirehep/utils/url.py
```diff
@@ -89,3 +90,17 @@ def retrieve_uri(uri, outdir=None):
 
         local_file.flush()
         yield local_file.name
+
+
+def get_prod_url_for_recid(recid, pattern_config_var='LEGACY_RECORD_URL_PATTERN'):
```
puntonim: We have to write code that is as environment-agnostic as possible. So I don't like the work `prod` in here. The word `LEGACY` is fine here instead as we can not run away from that.

puntonim: And why not having `pattern` as argument directly rather than the name of a config var?
So the call would be:  `get_prod_url_for_recid(recid, pattern=current_app.config[pattern_config_var])`

puntonim: We realized that in this case the entire function would have no sense, thus we hardcode the pattern in the function and rename it to `get_legacy_url_for_recid()`.

szymonlopaciuk:
ci merge please

szymonlopaciuk:
ci merge please

PR #3208, opened 2018-02-23
Fix rt module and duplicate tickets

drjova:


inspirehep/modules/rt/api.py
```diff
@@ -0,0 +1,47 @@
...
+from __future__ import absolute_import, division, print_function
+
+import backoff
+import rt
+
+from .tickets import create_ticket_with_template
+
+
+@backoff.on_exception(backoff.expo, rt.ConnectionError, base=4, max_tries=5)
+def submit_ticket(obj, queue, template, context, requestors, recid,
```
puntonim: Why did you move `submit_ticket()` to a new `api.py` module? What makes this different from all the other functions in the new `tasks.py` module?

drjova: the ``task`` module contains ``wraps`` which are used in the workflow

inspirehep/modules/rt/api.py
```diff
@@ -0,0 +1,47 @@
...
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""INSPIRE-HEP ticketing system API."""
+
+from __future__ import absolute_import, division, print_function
+
+import backoff
+import rt
```
puntonim: `rt` is an external service and you renamed this module `rt`. Thus now an import like `import rt` is very confusing (our module or the external service?).

david-caro:
This PR makes some design decisions that I think should be discussed, can we discuss it at some point before merging this? (it's not blocking anything right?)
We might want to discuss also what types of changes should be agreed or not, to what point can we decide stuff by ourselves or when should we bring in the rest of the team (retrospective material maybe).

inspirehep/modules/rt/tasks.py
```diff
@@ -0,0 +1,135 @@
+# -*- coding: utf-8 -*-
```
puntonim: To me, all workflow tasks should go all together in a workflow package (or layer).
And clients of external services (like `InspireRtExt`) should go to a service layer and have a similar (or even shared) interface.

And in general these kind of refactorings are actual architectural decisions that should be discussed in a refinement. Only with the input of everyone we can take the best decision and share the knowledge.

drjova:
I had a PR for moving rt outside of ``inspirehep.utils`` for two weeks now, I just moved the code so we can easily discuss were we'll move it i.e. in a separated module or so. Is not a design change it is just a way to make our life easier and help us separate it in the future

michamos:
> This PR makes some design decisions that I think should be discussed, can we discuss it at some point before merging this? (it's not blocking anything right?)

It's supposed to fix the duplicate ticket issue. If you want to take your time and discuss what to do with code architecture, can you at least split off and merge the fix for that issue?

PR #3223, opened 2018-02-28
orcid: also push to orcid in author record

michamos:
## Description

Also looks for ORCIDs to push a paper to in the author records that have claimed the paper.

## Related Issue

Depends on #3224.
Fixes https://its.cern.ch/jira/browse/INSPIR-250.

## Checklist:

### DIFF
--- a/inspirehep/modules/orcid/utils.py
+++ b/inspirehep/modules/orcid/utils.py
@@ -26,6 +26,8 @@ from __future__ import absolute_import, division, print_function
 
 import re
 
+from itertools import chain
+
 from flask import current_app
 from six.moves.urllib.parse import urljoin
 from sqlalchemy.orm.exc import NoResultFound
@@ -38,9 +40,12 @@ from invenio_oauthclient.models import (
 )
 from invenio_oauthclient.utils import oauth_link_external_id
 
+from inspire_dojson.utils import get_recid_from_ref
 from inspire_utils.logging import getStackTraceLogger
+from inspire_utils.record import get_values_for_schema
 from inspire_utils.urls import ensure_scheme
 from inspirehep.modules.cache.utils import redis_locking_context, RedisLockError
+from inspirehep.utils.record_getter import get_db_records
 
 LOGGER = getStackTraceLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


inspirehep/modules/orcid/utils.py
```diff
@@ -143,3 +148,32 @@ def get_push_access_token(orcid):
...
+
+    Returns:
+        Iterator[str]: all ORCIDs associated to these authors
+    """
+    orcids_on_record = []
+    author_recids_with_claims = []
+
+    for author in record.get('authors', []):
+        orcids_in_author = get_values_for_schema(author.get('ids', []), 'ORCID')
+        if orcids_in_author:
```
puntonim: Minor: this if is unnecessary

michamos: it's not, because of the `elif` later. I could transform the `elif [clause]` into `if not orcids_in_author and [clause]`, but I think that's less readable.

inspirehep/utils/record_getter.py
```diff
@@ -105,3 +106,19 @@ def get_db_record(pid_type, recid):
     from inspirehep.modules.records.api import InspireRecord
     pid = PersistentIdentifier.get(pid_type, recid)
     return InspireRecord.get_record(pid.object_uuid)
+
+
+def get_db_records(pid_type, recids):
```
puntonim: The name is too generic, it gets a db record by pid.

michamos: I don't disagree, but it's modelled on `get_db_record` and `get_es_records`, both of which were already there and have similar signatures.

inspirehep/utils/record_getter.py
```diff
@@ -105,3 +106,19 @@ def get_db_record(pid_type, recid):
...
+    if not recids:
+        return iter([])
+
+    query = RecordMetadata.query.join(
+        PersistentIdentifier, RecordMetadata.id == PersistentIdentifier.object_uuid
+    ).filter(
+        PersistentIdentifier.pid_value.in_(recids), PersistentIdentifier.pid_type == pid_type
+    )
+
+    return (record.json for record in query.yield_per(100))
```
puntonim: Why not returning record directly (RecordMetadata instance)? That way would be more reusable

inspirehep/modules/orcid/utils.py
```diff
@@ -143,3 +148,32 @@ def get_push_access_token(orcid):
...
+            orcids_on_record.extend(orcids_in_author)
+        elif author.get('curated_relation') is True and 'record' in author:
+            author_recids_with_claims.append(get_recid_from_ref(author['record']))
+
+    author_records = get_db_records('aut', author_recids_with_claims)
+    all_ids = (author.get('ids', []) for author in author_records)
+    orcids_in_authors = chain.from_iterable(get_values_for_schema(ids, 'ORCID') for ids in all_ids)
+
```
puntonim: You are using 4 utils coming from 3 different repos in a few lines. It was very expensive and annoying for me to follow this code. I know it is not your fault, but IMO utils modules are an anti-pattern for this reason, code should fit in better abstraction.

inspirehep/utils/record_getter.py
michamos: again, there are already functions `get_db_record` and `get_es_records`. The goal was to act in the same way as those.

michamos:
ci merge please

PR #3224, opened 2018-03-01
utils: add get_db_records util

michamos:
## Description

* Adds a `get_db_records` util to get several records at once given a PID type and a list of recids.
* Also refactors the tests for `get_es_records`
## Motivation and Context

This is useful for #3223.

## Checklist:

### DIFF
--- a/inspirehep/utils/record_getter.py
+++ b/inspirehep/utils/record_getter.py
@@ -30,6 +30,7 @@ from flask import current_app
 from werkzeug.utils import import_string
 
 from invenio_pidstore.models import PersistentIdentifier
+from invenio_records.models import RecordMetadata
 
 from inspirehep.modules.pidstore.utils import get_endpoint_from_pid_type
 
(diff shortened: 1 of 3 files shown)


inspirehep/utils/record_getter.py
```diff
@@ -105,3 +106,16 @@ def get_db_record(pid_type, recid):
...
+    """Get an iterator on record metadata from the DB."""
+    recids = [str(recid) for recid in recids]
+
+    query = RecordMetadata.query.join(
+        PersistentIdentifier, RecordMetadata.id == PersistentIdentifier.object_uuid
+    ).filter(
+        PersistentIdentifier.pid_value.in_(recids), PersistentIdentifier.pid_type == pid_type
+    )
+
+    return (record.json for record in query.all())
```
michamos: should I rather use [ `yield_per`](http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.yield_per) here? the disclaimers are a bit scary, and I don't really understand what they mean.

michamos: from that reply it looks safe, as we should have exactly one record associated to a given recid.

jacquerie: Right, that objection doesn't apply here.

michamos:
ci merge please

PR #3235, opened 2018-03-07
orcid-sync: remove Redis Lock

puntonim:


inspirehep/modules/orcid/tasks.py
```diff
@@ -48,23 +47,20 @@
 
 
 def legacy_orcid_arrays():
-    """Generator to fetch token data from redis.
+    """
+    Generator to fetch token data from redis.
+
+    Note: this function is related to the legacy tasklet:
```
michamos: `this function consumes the queue populated by the legacy tasklet`?

tests/integration/orcid/test_orcid_import_legacy_tokens.py
```diff
@@ -161,6 +163,40 @@ def test_multiple_list_generator(app, redis_setup):
...
+        self.tokens = []
+
+    def run(self):
+        with self.app.app_context():
+            self.tokens = list(legacy_orcid_arrays())
+
+
+def test_legacy_orcid_arrays_concurrency(app, redis_setup):
+    """
+    Ensure legacy_orcid_arrays() works with concurrent consumers.
```
michamos: I am not sure this test is asserting that because of the GIL. Besides we don't usually put docstrings in tests, as the function name should be self-explanatory.

puntonim: Done. You are right about threads and GIL, thus as we discussed I removed this test.

puntonim:
@szymonlopaciuk as we discussed I removed the concurrency test.

puntonim:
ci merge please

PR #3238, opened 2018-03-08
forms: add, populate and validate ORCID field

iulianav:
## Description

Add, populate and validate ORDCID field in "authors/recid/update" form.
The field behaves precisely like the one in the "authos/new" form.

The addition of the field was done by removing a simple check of the `is-update` field that hid the ORCID field in the case of an update form.

The validation - which was previously broken - has already been fixed by @szymonlopaciuk [here](https://github.com/inspirehep/inspire-next/pull/3236).

The implementation for the field population was already in place as one can see [here](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/authors/views.py#L235) where the `try ... except ...` statement is importing the record's xml from **legacy** and converting the data to the appropriate format to then populate the form fields (hint: check the `convert_for_form(data)` function). HOWEVER, I am not sure this is desirable or if the field should instead be populated from what we have in labs and NOT in legacy. **Someone should check this!**

## Motivation and Context

The user might wish to update an author's ORCID as well.

## Checklist:

### DIFF
--- a/inspirehep/modules/authors/forms.py
+++ b/inspirehep/modules/authors/forms.py
@@ -562,11 +562,7 @@ class AuthorUpdateForm(INSPIREForm):
     def __init__(self, *args, **kwargs):
         """Constructor."""
         super(AuthorUpdateForm, self).__init__(*args, **kwargs)
-        is_update = kwargs.pop('is_update', False)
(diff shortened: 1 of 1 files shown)


iulianav:
ci merge please

PR #3239, opened 2018-03-08
workflows: add validation errors callback

drjova:
* Adds a callback endpoint for the UI, this will allow the UI to run
  validation and continue the workflow if the record is valid.

* Adds in ``extra_data`` ``callback_url`` and
  ``validation_errors`` fields on callback request and when validation
  task runs.

Co-authored-by: Harris Tzovanakis <me@drjova.com>

### DIFF
--- a/inspirehep/modules/workflows/serializers/schemas/json.py
+++ b/inspirehep/modules/workflows/serializers/schemas/json.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Marshmallow JSON worfklow schema."""
+
+from __future__ import absolute_import, division, print_function
+
+from marshmallow import Schema, fields
+
+
+class WorkflowSchemaJSONV1(Schema):
+    """Schema for workflows."""
+
+    _extra_data = fields.Dict(required=True)
+    _id = fields.Integer(required=True, load_from='id', attribute='id')
+    _workflow = fields.Dict(required=False)
+
+    metadata = fields.Dict(required=True)
+
+    class Meta:
+        strict = True
--- a/inspirehep/modules/workflows/utils/__init__.py
+++ b/inspirehep/modules/workflows/utils/__init__.py
@@ -30,13 +30,15 @@ import os
 import traceback
 from contextlib import closing, contextmanager
 from functools import wraps
-
+from six import text_type
 import backoff
 import lxml.etree as ET
 import requests
-from flask import current_app
+from flask import current_app, url_for
 
 from invenio_db import db
+from inspire_schemas.utils import \
+    get_validation_errors as _get_validation_errors
 
 from inspirehep.utils.url import retrieve_uri
 from inspirehep.modules.workflows.models import (
@@ -269,3 +271,38 @@ def insert_wf_record_source(json, record_uuid, source):
     else:
         record_source.json = json
     db.session.commit()
+
+
+def get_resolve_validation_callback_url():
+    """Resolve validation callback.
+
+    Returns the callback url for resolving the validation errors.
+
+    Note:
+        It's using ``inspire_workflows.callback_resolve_validation``
+        route.
+    """
+    return url_for(
+        'inspire_workflows.callback_resolve_validation',
+        _external=True
+    )
+
+
+def get_validation_errors(data, schema):
+    """Creates a ``validation_errors`` dictionary.
+
+    Args:
+        data (dict): the object to validate.
+        schema (str): the name of the schema.
+
+    Returns:
+        dict: ``validation_errors`` formatted dict.
+    """
+    errors = _get_validation_errors(data, schema=schema)
+    error_messages = [
+        {
+            'path': map(text_type, error.absolute_path),
+            'message': text_type(error.message),
+        } for error in errors
+    ]
+    return error_messages
--- a/inspirehep/modules/workflows/views.py
+++ b/inspirehep/modules/workflows/views.py
@@ -25,14 +25,34 @@
 from __future__ import absolute_import, division, print_function
 
 from os.path import join
-from flask import Blueprint, jsonify, request, current_app
 
+from flask import (
+    Blueprint,
+    current_app,
+    jsonify,
+    request
+)
+from flask.views import MethodView
 from invenio_db import db
+from inspire_schemas.api import validate
 from invenio_workflows import workflow_object_class, ObjectStatus
 from invenio_workflows.errors import WorkflowsMissingObject
+from jsonschema.exceptions import ValidationError
 
 from inspire_utils.urls import ensure_scheme
+from inspirehep.modules.workflows.errors import (
+    CallbackError,
+    CallbackValidationError,
+    CallbackWorkflowNotFoundError,
+    CallbackWorkflowNotInValidationError
+)
+from inspirehep.modules.workflows.loaders import workflow_loader
 from inspirehep.modules.workflows.models import WorkflowsPendingRecord
+from inspirehep.modules.workflows.utils import (
+    get_resolve_validation_callback_url,
+    get_validation_errors
+)
+
 
 blueprint = Blueprint(
     'inspire_workflows',
@@ -41,6 +61,13 @@ blueprint = Blueprint(
     template_folder='templates',
     static_folder="static",
 )
+
+
+@blueprint.errorhandler(CallbackError)
+def error_handler(error):
+    """Callback error handler."""
+    response = jsonify(error.to_dict())
+    return response, error.code
 
 
 def _get_base_url():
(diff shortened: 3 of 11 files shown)


drjova:
merge after https://github.com/inspirehep/inspire-schemas/pull/311

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -269,3 +271,39 @@ def insert_wf_record_source(json, record_uuid, source):
...
+    """Creates a ``validation_errors`` dictionary from validation exception.
+
+    Args:
+        data (dict): the object to validate.
+        schema (str): the name of the schema.
+
+    Returns:
+        dict: ``validation_errors`` formatted dict.
+    """
+    validation_exceptions = validation_errors(data, schema=schema)
```
david-caro: The naming here is quite confusing :/

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -269,3 +271,39 @@ def insert_wf_record_source(json, record_uuid, source):
...
+    validation_errors = [
```
david-caro: You should probably avoid shadowing variables (or globals in this case).

inspirehep/modules/workflows/views.py
```diff
@@ -376,3 +400,184 @@ def robotupload_callback():
...
+        return inner
+    return validate_wrap
+
+
+class ResolveValidationResource(MethodView):
+    """Resolve validation error callback."""
+
+    @error_handler
+    @validate_workflow_payload(['metadata', '_extra_data', 'id'])
+    def put(self, workflow_data, **kwargs):
```
david-caro: Can you avoid using a decorator that changes the signature?

david-caro: Also, is the kwargs needed?

tests/integration/workflows/helpers/calls.py
```diff
@@ -193,3 +193,28 @@ def core_record():
...
+           If ``malformed`` is true it will make the payload
+           invalid, by removing two required ``keys``; ``id``
+           and ``_extra_data``.
+    """
+    client = app.test_client()
+    payload = {
+        "metadata": data.data,
+    }
+
+    if not malformed:
```
david-caro: Maybe do it the other way around (as you state in the comment):
```python
if malformed:
    del payload['id']
    del payload['_extra_data']
```

That also get rid of the `if not` thingie.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -615,3 +624,132 @@ def custom_continue_workflow(self, *args, **kwargs):
...
+
+    eng = WorkflowEngine.from_uuid(eng_uuid)
+    obj = eng.objects[0]
+
+    response = do_validation_callback(
+        app=workflow_app,
+        data=obj,
+    )
+
+    assert response.status_code == 200
```
david-caro: Probably you should assert the side effects of the validation acceptance here, to make sure that the workflow is continued and such.

inspirehep/modules/workflows/utils/__init__.py
drjova: hmm indeed, we made a change on the ``inspire-schema`` part

inspirehep/modules/workflows/utils/__init__.py
drjova: oups because we change the name on ``insprire-schema``

inspirehep/modules/workflows/views.py
drjova: the decorator is to help us deal with other workflow callback cases in the future (ala ``@pass_record``) indeed ``kawrgs`` is not needed

david-caro: Can we use a function instead of a decorator?

drjova: IMO is cleaner to have a decorator in such cases, you can easily change the validation for the workflow payload and you have a workflow object ready. This is how ``records-rest`` is working, it'll be a good start to keep everything tidy and reusable for the workflows  REST.

drjova:
@david-caro updated :)

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -269,3 +270,38 @@ def insert_wf_record_source(json, record_uuid, source):
...
+        It's using ``inspire_workflows.callback_resolve_validation``
+        route.
+    """
+    return url_for(
+        'inspire_workflows.callback_resolve_validation',
+        _external=True
+    )
+
+
+def validation_errors(data, schema):
```
david-caro: you are missing a verb here too xd

drjova: Hmm I didn't want to have duplicate the names (``get_validation_errors``), updated!

david-caro: you can use `make_validation_errors` or `create_validation_errors` or similar too, get is ok for me though (as we discussed, the verb is not that much of a deal, there's lots of synonyms).

drjova: with ‘’as _get_validation_errors’’, IMO makes it more consistent. WDYT?

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -269,3 +271,38 @@ def insert_wf_record_source(json, record_uuid, source):
...
+        It's using ``inspire_workflows.callback_resolve_validation``
+        route.
+    """
+    return url_for(
+        'inspire_workflows.callback_resolve_validation',
+        _external=True
+    )
+
+
+def get_validation_errors(data, schema):
```
david-caro: :+1:

inspirehep/modules/workflows/views.py
```diff
@@ -25,14 +25,28 @@
...
+from flask.views import MethodView
 from invenio_db import db
+from inspire_schemas.api import validate
 from invenio_workflows import workflow_object_class, ObjectStatus
 from invenio_workflows.errors import WorkflowsMissingObject
+from jsonschema.exceptions import ValidationError
 
 from inspire_utils.urls import ensure_scheme
+from inspirehep.modules.workflows.errors import CallbackMalformedError, \
+    CallbackValidationError, CallbackWorkflowNotFoundError
```
david-caro: We usually use parenthesis to get multiline imports (you did it for flask),

david-caro: I know that's some "styling" stuff, so we can leave it for whenever we decide on that.

inspirehep/modules/workflows/views.py
```diff
@@ -376,3 +401,180 @@ def robotupload_callback():
...
+                     message=error.description),
+                code=error.code
+            )
+        except CallbackWorkflowNotFoundError as error:
+            return make_response(
+                dict(error_code=error.error_code,
+                     message=error.description),
+                code=error.code
+            )
+        except CallbackValidationError as error:
```
david-caro: why not use simpler error handlers, per-error instead of per-view?:

    http://flask.pocoo.org/docs/0.12/patterns/apierrors/

I know that right now it works the same, but if we are going to reuse them, might be easier to extend, test and maintain.

david-caro: Incidentally it also simplifies the `make_respoonse` ;)

inspirehep/modules/workflows/views.py
drjova: for the flask it was you :P but I'll update it. 👍 we should decide and automate this kind of decisions

david-caro: I was investigating still :), I'm open to alternatives, I don't have a lot of experience on it.

inspirehep/modules/workflows/views.py
```diff
@@ -376,3 +401,180 @@ def robotupload_callback():
...
+
+        workflow.extra_data.pop('callback_url', None)
+        workflow.extra_data.pop('validation_errors', None)
+
+        workflow.data = workflow_data['metadata']
+        workflow.save()
+        db.session.commit()
+
+        if workflow.status != ObjectStatus.COMPLETED:
+            workflow.continue_workflow(delayed=True)
```
david-caro: This should change the status of the workflow to running too.

inspirehep/modules/workflows/views.py
```diff
@@ -376,3 +401,180 @@ def robotupload_callback():
...
+            raise CallbackWorkflowNotFoundError(workflow_id)
+
+        workflow.extra_data.pop('callback_url', None)
+        workflow.extra_data.pop('validation_errors', None)
+
+        workflow.data = workflow_data['metadata']
+        workflow.save()
+        db.session.commit()
+
+        if workflow.status != ObjectStatus.COMPLETED:
```
david-caro: You might want to do this check before changing the workflow (a few lines before this one).

david-caro: hmm, also even rise a different error message, like 'workflow is in completed state' or similar.

inspirehep/modules/workflows/views.py
```diff
@@ -376,3 +401,180 @@ def robotupload_callback():
...
+    Args:
+        keys (list(str)): the keys to check if they exist.
+
+    Raises:
+        CallbackMalformedError: if any of the ``keys`` is missing from
+            the request payload.
+        CallbackValidationError: if the workflow ``metadata`` is not valid
+            against ``hep`` JSONSchema.
+    """
+    workflow_data = request.get_json()
```
david-caro: it's probably better to pass the json as a parameter, so this function does not need to know about the request thingie.

inspirehep/modules/workflows/views.py
```diff
@@ -376,3 +401,180 @@ def robotupload_callback():
...
+
+    # Check for missing keys
+    for key in keys:
+        if workflow_data.get(key) is None:
+            raise CallbackMalformedError(key)
```
david-caro: This should also check that workflow_data is a dict-like object, and has a 'metadata' key.

david-caro: no, something else is off :/, looking

david-caro: The problem is that `workflow_data['_extra_data']` should be a dict, that check should be made here too and rise the `CallbackMalformedError` one.

inspirehep/modules/workflows/views.py
david-caro: The idea is to avoid using flask in the inner functions as much as possible, leaving the http interfacing for the functions that are directly called for it (that way also makes this one easier to test by itself).

inspirehep/modules/workflows/views.py
david-caro: Also, it should make sure that the workflow is halted because of validation errors :S, not sure if that's easy, but otherwise, if you might un halt a workflow that is waiting for something else. Maybe we could check the workflow's 'callback_url' to make sure that it's us, or (better but longer) finally define the halted statuses properly and check for that.

inspirehep/modules/workflows/views.py
david-caro: Passing an empty object as "id" also breaks with 500:
```
curl -vvv -H "content-type: application/json" -X PUT http://localhost:5000/callback/workflows/resolve_validation_errors -d '{"metadata": {"document_type": ["book"], "_collections": ["Literature"], "titles": [{"title": "snth"}]}, "_extra_data": {}, "id": {}}'
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 5000 (#0)
> PUT /callback/workflows/resolve_validation_errors HTTP/1.1
> Host: localhost:5000
> User-Agent: curl/7.55.1
> Accept: */*
> content-type: application/json
> Content-Length: 133
> 
* upload completely sent off: 133 out of 133 bytes
< HTTP/1.1 500 INTERNAL SERVER ERROR
< Server: gunicorn/19.7.1
< Date: Mon, 12 Mar 2018 18:59:05 GMT
< Connection: close
< Transfer-Encoding: chunked
< Content-Type: text/html; charset=utf-8
< X-XSS-Protection: 0
< 
```

inspirehep/modules/workflows/views.py
drjova: change status from ``ERROR`` -> ``HALTED``?

inspirehep/modules/workflows/views.py
drjova: The problem with adding the error handling to the view, is that all the other endpoints are not proper RESTful, it needs refactoring for having common error and message formats.

inspirehep/modules/workflows/views.py
drjova: I'm a bit confused, the status when we call the validation error is ``ERROR`` right?

inspirehep/modules/workflows/views.py
drjova: The proper way to handle this is marshmallow, which means we have to do needs refactoring to add loaders and schema. I'll do it for only this endpoint as a prototype.

david-caro: :+1:

inspirehep/modules/workflows/views.py
david-caro: I'm not sure how REST is related to this, but if we start doing it, we can refactor while we add new views as needed (that to me sounds better than doing a big plan beforehand, less risk involved, and you ensure that you are building just what you need).

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -269,3 +271,38 @@ def insert_wf_record_source(json, record_uuid, source):
     else:
         record_source.json = json
     db.session.commit()
+
+
+def resolve_validation_callback_url():
```
drjova: Shouldn't this be `get_resolve...` ?

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -269,3 +271,38 @@ def insert_wf_record_source(json, record_uuid, source):
...
+        data (dict): the object to validate.
+        schema (str): the name of the schema.
+
+    Returns:
+        dict: ``validation_errors`` formatted dict.
+    """
+    errors = _get_validation_errors(data, schema=schema)
+    error_messages = [
+        {
+            'path': map(lambda path: text_type(path), error.absolute_path),
```
drjova: Nit: `lambda x: f(x)` is equivalent to just `f`.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -633,3 +642,184 @@ def custom_continue_workflow(self, *args, **kwargs):
...
+
+    obj = workflow_object_class.create(
+        data=invalid_record,
+        data_type='hep',
+        id_user=1,
+    )
+    obj_id = obj.id
+
+    with pytest.raises(ValidationError):
+        start('article', invalid_record, obj_id)
```
drjova: Nit: maybe assert that the workflow object is in ERROR state after this? Just to be 100% clear.

inspirehep/modules/workflows/views.py
```diff
@@ -376,3 +403,140 @@ def robotupload_callback():
...
+        workflow_data (dist): the workflow dict.
+
+    Raises:
+        CallbackValidationError: if the workflow ``metadata`` is not valid
+            against ``hep`` JSONSchema.
+    """
+
+    # Check for validation errors
+    try:
+        validate(schema='hep', data=workflow_data['metadata'])
```
drjova: It can stay for now, but this machinery will also be useful to fix author submissions/updates that failed validation, and the fact that we're hardcoding the schema prevents this...

inspirehep/modules/workflows/serializers/schemas/json.py
```diff
@@ -0,0 +1,41 @@
...
+
+class WorkflowSchemaJSONV1(Schema):
+    """Schema for workflows."""
+
+    _extra_data = fields.Dict(required=True)
+    _id = fields.Integer(required=True, load_from='id',
+                         attribute='id')
+
+    metadata = fields.Dict(required=True)
+    workflow = fields.Dict(required=False)
```
drjova: _workflow

drjova:
ci merge please

PR #3242, opened 2018-03-09
workflows: fix issue with unprintable keywords

jacquerie:
## Description:
Sentry: https://sentry.inspirehep.net/inspire-sentry/prod/issues/6/

~~This doesn't yet contain a fix, just a reproduction (this PR is expected to fail).~~
Fix happened here: https://github.com/inveniosoftware-contrib/invenio-classifier/pull/38

## Checklist:

### DIFF
--- a/tests/unit/workflows/test_workflows_tasks_classifier.py
+++ b/tests/unit/workflows/test_workflows_tasks_classifier.py
@@ -22,8 +22,11 @@
 
 from __future__ import absolute_import, division, print_function
 
-from six import binary_type
+import os
+
+import pkg_resources
 from mock import patch
+from six import binary_type
 
 from inspirehep.modules.workflows.tasks.classifier import classify_paper
 from mocks import MockEng, MockObj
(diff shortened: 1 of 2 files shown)


tests/unit/workflows/test_workflows_tasks_classifier.py
```diff
@@ -128,3 +131,27 @@ def test_classify_paper_uses_keywords(get_document_in_workflow):
 
     assert obj.extra_data['classifier_results']['complete_output']['core_keywords'] == expected
     assert obj.extra_data['classifier_results']['fulltext_used'] is False
+
+
+@patch('inspirehep.modules.workflows.tasks.classifier.get_document_in_workflow')
+def test_classify_paper_handles_unprintable_keywords(get_document_in_workflow):
+    paper_with_unprintable_keywords = pkg_resources.resource_filename(
+        __name__, os.path.join('fixtures', '1802.08709.pdf'))
```
jacquerie: @puntonim this is using my crappy way of retrieving fixtures. Do you have some time to discuss how the good solution should look like?

puntonim: yes, I can show you some patterns I've seen used

puntonim: Let's focus on 2 use cases: an isolated fixture like this one and VCR fixtures.
Note: VCR stores network calls responses in files (fixtures) inside a given folder.
Note: I assume the healthy principle: 1 fixture belongs only to 1 test case. I.e. fixtures are not reused across test cases, but duplicated if necessary.

A nice pattern for the VCR case is:

```
from foo import push_to_hal  # This is what I want to test.

def test_push_to_hal_happy_flow():
    with VCRFixture():
        response = push_to_hal(records)  # Suppose it returns a sort of data model.
        assert response.records_created == 15
```

What `VCRFixture` does is setting VCR in the proper mode and using the proper folder.
The proper folder is found automatically computing the fully-qualified-name of the test case (see: https://github.com/kpn-digital/py-fqn-decorators/blob/master/fqn_decorators/decorators.py#L6).
So fixtures would be in: projectroot/fixtures/packagename/modulename/test_push_to_hal_happy_flow/.
This is to hide all this complexity: https://github.com/inspirehep/inspire-next/blob/master/tests/integration/orcid/test_orcid_push.py#L73
Add also the bool parameters `USE_MOCK` and `MOCK_OR_ERROR` in your local config file to switch VCR's recording mode (basically record the cassettes or use them).  
`VCRFixture` accepts the `ident` kwarg to force the folder path (disabling the fqn magic).

Finally the simpler case of an isolated fixture.

```
from foo import parse  # This is what I want to test.
# And I need a fixture which is a pdf file.

def test_pdf_parser_happy_flow():
    with Fixture(ident='mypaper.pdf') as pdf_file:
        result = parse(pdf_file)
        assert result.title == 'Title'
```

Also in this case `Fixture` knows where to find the fixture thanks to the fully-qualified-name of the test function.

jacquerie:
ci merge please

jacquerie:
ci merge please

PR #3243, opened 2018-03-09
forms: refactor authors update form

iulianav:
## Description

Search for the record in the db at first in order to populate the form instead of requesting XML straight from Legacy through a GET request. When the author schema will be completed and Labs will contain all author records from Legacy, we'll also be able to remove the request to Legacy altogether (as Legacy will be shut down in the future). Moreover, use `lxml` library to process the XML response from Legacy, instead of regular expressions.

Also, a wrong recid in the path will create a workflow object in ERROR state because it will fail validation.

## Motivation and Context

Previously, it was possible to submit author updates to existing literature records and tickets were being created for it.  Needless to say, this was a waste of time for the curators. `authors/0/update` redirected to the `authors/new` form, while a non-existent recid in any collection ('lit', 'aut', ...etc) would redirect to a 404 not found page.

## Checklist:

### DIFF
--- a/inspirehep/modules/authors/utils.py
+++ b/inspirehep/modules/authors/utils.py
@@ -30,6 +30,9 @@ import re
 import numpy as np
 from beard.utils.strings import asciify
 from beard.clustering import block_phonetic
+from lxml.etree import fromstring, tostring
+
+from inspire_dojson import marcxml2record
 
 
 _bai_parentheses_cleaner = \
@@ -106,6 +109,25 @@ def bai(name):
     bai = _bai_spaces.sub(".", bai)
 
     return bai
+
+
+def get_author_record_from_xml_response(xml_content):
+    """Return record contained in XML if collection is 'authors'.
+
+    Args:
+        xml_content (str): Legacy response XML content.
+
+    Returns:
+        dict: JSON author record contained in the XML.
+
+        None: if record collection is not 'authors'.
+    """
+    xml = fromstring(xml_content)
+    collections_datafields = xml.xpath("//*[local-name() = 'datafield'][@*[local-name()='tag' and .='980']]//*[local-name()='subfield']")
+    collections = [el.text for el in collections_datafields]
+    if 'HEPNAMES' in collections:
+        xml_record = xml.xpath("//*[local-name() = 'record']")[0]
+        return marcxml2record(tostring(xml_record))
 
 
 def phonetic_blocks(full_names, phonetic_algorithm='nysiis'):
--- a/inspirehep/modules/authors/views.py
+++ b/inspirehep/modules/authors/views.py
@@ -25,9 +25,8 @@
 from __future__ import absolute_import, division, print_function
 
 import copy
-import re
-
 import requests
+
 from flask import (
     abort,
     Blueprint,
@@ -35,25 +34,23 @@ from flask import (
     render_template,
     request,
     url_for,
-    redirect,
 )
 from flask_breadcrumbs import register_breadcrumb
 from flask_login import login_required, current_user
-from werkzeug.datastructures import MultiDict
-
+from inspire_utils.record import get_value
 from invenio_db import db
 from invenio_workflows import workflow_object_class, start, resume
 from invenio_workflows_ui.api import WorkflowUIRecord
+from werkzeug.datastructures import MultiDict
 
-from inspire_dojson import marcxml2record
-from inspire_utils.record import get_value
 from inspirehep.modules.forms.form import DataExporter
+from inspirehep.utils.record_getter import RecordGetterError, get_db_record
 from inspirehep.utils.url import get_legacy_url_for_recid
-
 from .forms import AuthorUpdateForm
 from .permissions import holdingpen_author_permission
 from .tasks import formdata_to_model
-
+from .core import convert_for_form
+from .utils import get_author_record_from_xml_response
 
 blueprint = Blueprint(
     'inspirehep_authors',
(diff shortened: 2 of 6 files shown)


inspirehep/modules/authors/views.py
```diff
@@ -230,21 +228,14 @@ def new():
...
-        except requests.exceptions.RequestException:
-            pass
-        data["control_number"] = recid
-    else:
-        return redirect(url_for("inspirehep_authors.new"))
+    try:
+        data = get_db_record('aut', recid)
+        convert_for_form(data)
+    except RecordGetterError:
+        abort(404)
```
jacquerie: I just realized that we can't do this yet: many author records still don't migrate successfully, so if we were to do this we would serve a 404 instead of creating a workflow in ERROR state. So the previous logic should be moved here (but it still needs a refactoring).

inspirehep/modules/authors/views.py
```diff
@@ -225,26 +227,35 @@ def new():
...
-            data = marcxml2record(xml_content)
-            convert_for_form(data)
-        except requests.exceptions.RequestException:
-            pass
-        data["control_number"] = recid
-    else:
-        return redirect(url_for("inspirehep_authors.new"))
+    data = _author_record_from_db(recid) or _author_record_from_legacy(recid) or {}
+    convert_for_form(data)
+    data["control_number"] = recid
```
Glignos: Wont you get a keyerror if the data is an empty object?

iulianav:
```
In [4]: a_dict = {}

In [5]: a_dict['a_key'] = 'a_value'

In [6]: a_dict
Out[6]: {'a_key': 'a_value'}
```

Glignos: Ah, my bad didn't notice you're actually setting it :)

inspirehep/modules/authors/views.py
```diff
@@ -225,26 +227,35 @@ def new():
...
+        return None
+
+
+def _author_record_from_legacy(recid):
+    """Get author record from Legacy or return None if it is not there."""
+    try:
+        url = get_legacy_url_for_recid(recid) + '/export/xm'
+        xml_response = requests.get(url)
+        xml = fromstring(xml_response.content)
+        xml_record = xml.xpath("//*[local-name() = 'record']")[0]
```
michamos: why this strange selector and not simply `//record`?

inspirehep/modules/authors/views.py
```diff
@@ -225,26 +227,35 @@ def new():
...
-                r"\<record\>.*\<\/record\>", re.MULTILINE + re.DOTALL)
-            xml_content = record_regex.search(xml.content).group()
-            data = marcxml2record(xml_content)
-            convert_for_form(data)
-        except requests.exceptions.RequestException:
-            pass
-        data["control_number"] = recid
-    else:
-        return redirect(url_for("inspirehep_authors.new"))
+    data = _author_record_from_db(recid) or _author_record_from_legacy(recid) or {}
```
michamos: why an empty record? if you cannot get the record from legacy, then either it doesn't exist there (-> return a 404 error) or there was some problem in communicating with legacy (-> return a 503 according to https://stackoverflow.com/a/16038053).

inspirehep/modules/authors/views.py
```diff
@@ -225,26 +227,35 @@ def new():
...
-            record_regex = re.compile(
-                r"\<record\>.*\<\/record\>", re.MULTILINE + re.DOTALL)
-            xml_content = record_regex.search(xml.content).group()
-            data = marcxml2record(xml_content)
-            convert_for_form(data)
-        except requests.exceptions.RequestException:
-            pass
-        data["control_number"] = recid
-    else:
-        return redirect(url_for("inspirehep_authors.new"))
```
michamos: you don't want to do the redirect any more?

inspirehep/modules/authors/views.py
iulianav: @jacquerie suggested that we still want an errored workflow object, rather than a 404. Right?

inspirehep/modules/authors/views.py
```diff
@@ -225,26 +227,35 @@ def new():
...
+
+def _author_record_from_legacy(recid):
+    """Get author record from Legacy or return None if it is not there."""
+    try:
+        url = get_legacy_url_for_recid(recid) + '/export/xm'
+        xml_response = requests.get(url)
+        xml = fromstring(xml_response.content)
+        xml_record = xml.xpath("//*[local-name() = 'record']")[0]
+        return marcxml2record(tostring(xml_record))
+    except requests.exceptions.RequestException:
```
tsgit: the `try` `except` conflates multiple things, but only catches `RequestException`. What happens for non-existent `recid` or a wrong `recid`, e.g. a non-integer? 
e.g.
```
$ curl http://inspirehep.net/record/9999999/export/xm
<?xml version="1.0" encoding="UTF-8"?>
<collection xmlns="http://www.loc.gov/MARC21/slim">

</collection>
```
then you get an `IndexError` on line 244

and for `recid = 9999O` you get a status `404` and HTML response which results in `XMLSyntaxError`.

I recommend to check xml_response.status_code first.

inspirehep/modules/authors/views.py
iulianav: It only worked in `recid == 0`

inspirehep/modules/authors/views.py
iulianav: "Since there is no namespace configured in the XPath expression, we don’t find any matching nodes. The XPath Expression is looking for default namespace but the document is with a different namespace. To resolve this problem of namespaces we have to use the local-name() function. The local-name() function ignores the namespace and returns the query results as if the XML did not have any namespace." Source: https://examples.javacodegeeks.com/core-java/xml/xpath/xpath-local-name-example/

inspirehep/modules/authors/views.py
```diff
@@ -230,21 +128,34 @@ def new():
...
-        data["control_number"] = recid
-    else:
-        return redirect(url_for("inspirehep_authors.new"))
+            if xml_response.status_code == 404:
+                abort(404)
+            abort(503)
+
+        xml = fromstring(xml_response.content)
+        xml_record = xml.xpath("//*[local-name() = 'record']")[0]
+        data = marcxml2record(tostring(xml_record))
```
iulianav: Sometimes we get responses like:
```
<collection xmlns="http://www.loc.gov/MARC21/slim">
<link/>
<script/>
<record>
<controlfield tag="001">4</controlfield>
<datafield tag="980" ind1=" " ind2=" ">
<subfield code="c">DELETED</subfield>
</datafield>
</record>
</collection>
```
When this happens, we still get redirected to the form. Some format check is needed to see if what we get is indeed an XML containing an author record.

inspirehep/modules/authors/views.py
```diff
@@ -230,21 +127,27 @@ def new():
...
+            xml_response = requests.get(url)
+            xml_response.raise_for_status()
         except requests.exceptions.RequestException:
-            pass
-        data["control_number"] = recid
-    else:
-        return redirect(url_for("inspirehep_authors.new"))
+            if xml_response.status_code == 404:
+                abort(404)
+            abort(503)
```
drjova: it's better to abort with ``400`` in this case a well :)

inspirehep/modules/authors/views.py
```diff
@@ -230,21 +127,27 @@ def new():
...
+
+        data = get_author_record_from_response(xml_response)
+
+    if not data:
+        abort(400)
+
+    convert_for_form(data)
+    data["control_number"] = recid
```
drjova: ``'`` to be consistent

inspirehep/modules/authors/utils.py
```diff
@@ -108,6 +111,16 @@ def bai(name):
     return bai
 
 
+def get_author_record_from_response(xml_response):
```
drjova: we can add some test cases for this as well

drjova:
@iulianav could you please also rebase?

tests/unit/authors/test_authors_utils.py
```diff
@@ -46,3 +49,66 @@ def test_that_bai_conforms_to_the_spec():
     assert bai("Müller, Andreas") == "A.Mueller"
     assert bai("Hernández-Tomé, G.") == "G.Hernandez.Tome"
     assert bai("José de Goya y Lucientes, Francisco Y H") == "F.Y.H.Jose.de.Goya.y.Lucientes"
+
+
+def test_get_author_record_from_xml_response():
+    xml = """<?xml version="1.0" encoding="UTF-8"?>\n
```
iulianav: @drjova  do you think I should use fixtures here, or leave it like this to be more readable as the XMLs are quite small?

drjova:
Supersede by https://github.com/inspirehep/inspire-next/pull/3650 🎉

PR #3244, opened 2018-03-09
orcid-sync: add whitelist INSPIR-44

puntonim:


inspirehep/modules/orcid/tasks.py
```diff
@@ -165,6 +167,12 @@ def orcid_push(self, orcid, rec_id, oauth_token):
         rec_id(int): inspire record's id to push to ORCID.
         oauth_token(string): orcid token.
     """
+    compiled = re.compile(current_app.config.get(
```
szymonlopaciuk: Wouldn't it be worth it to compile it in the config so it only happens once? Also in any case I think `re.match` can do the compilation for you if you pass the string so in this instance it doesn't seem necessary, since you only pass it once

tests/integration/orcid/test_orcid_push.py
```diff
@@ -152,3 +153,73 @@ def _push_record_with_orcid(_recid, _orcid, tk, _put_code=None):
     monkeypatch.setattr(tasks, 'push_record_with_orcid', _push_record_with_orcid)
 
     attempt_push(orcid, recid, 'fake-token')
+
+
+def test_feature_flag_orcid_push_whitelist_regex_none():
```
szymonlopaciuk: If I'm not mistaken this test test `re`, not really the code you wrote?

inspirehep/modules/orcid/tasks.py
puntonim: I'd prefer to keep the config file as dumb as possible, just a set of key = values statements and possibly json serializable such that it can evolve into a configuration service (like ETCD).
And for these 2 lines, I initially wrote the code this way because shorter statements are more readable, but actually, I am going to change it.

tests/integration/orcid/test_orcid_push.py
puntonim: True. The idea is not to test the code but the possible values for the whitelist that I suggest in the config file: `^$`, `.*`, `^(val1|val2)$`

inspirehep/modules/orcid/tasks.py
puntonim: Changed

PR #3246, opened 2018-03-12
migrator.tasks: disable migration on ORCID push

szymonlopaciuk:
## Description
Disables ORCID push during migration (however, not continuous migration) by the means of temporarily changing the ORCID push feature flag. This does not interfere with other tasks as they are run in separate processes. The flag is reset in case of errors as well.

## Checklist:

### DIFF
--- a/inspirehep/modules/migrator/tasks.py
+++ b/inspirehep/modules/migrator/tasks.py
@@ -38,6 +38,7 @@ from elasticsearch.helpers import bulk as es_bulk
 from elasticsearch.helpers import scan as es_scan
 from flask import current_app
 from flask_sqlalchemy import models_committed
+from functools import wraps
 from jsonschema import ValidationError
 from redis import StrictRedis
 from redis_lock import Lock
@@ -74,6 +75,27 @@ CHUNK_SIZE = 100
 LARGE_CHUNK_SIZE = 2000
 
 split_marc = re.compile('<record.*?>.*?</record>', re.DOTALL)
+
+
+def disable_orcid_push(task_function):
+    """Temporarily disable ORCID push
+
+    Decorator to temporarily disable ORCID push while a given task is running,
+    and only for that task. Takes care of restoring the previous state in case
+    of errors or when the task is finished. This does not interfere with other
+    tasks, firstly because of ditto, secondly because configuration is only
+    changed within the worker's process (thus doesn't affect parallel tasks).
+    """
+    @wraps(task_function)
+    def _task_function(*args, **kwargs):
+        initial_state = current_app.config['FEATURE_FLAG_ENABLE_ORCID_PUSH']
+        current_app.config['FEATURE_FLAG_ENABLE_ORCID_PUSH'] = False
+        try:
+            task_function(*args, **kwargs)
+        finally:
+            current_app.config['FEATURE_FLAG_ENABLE_ORCID_PUSH'] = initial_state
+
+    return _task_function
 
 
 def chunker(iterable, chunksize=CHUNK_SIZE):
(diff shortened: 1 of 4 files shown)


inspirehep/modules/migrator/tasks.py
```diff
@@ -76,6 +77,27 @@
...
+    Decorator to temporarily disable ORCID push while a given task is running,
+    and only for that task. Takes care of restoring the previous state in case
+    of errors or when the task is finished. This does not interfere with other
+    tasks, firstly because of ditto, secondly because configuration is only
+    changed within the worker's process (thus doesn't affect parallel tasks).
+    """
+    @wraps(task_function)
+    def _task_function(*args, **kwargs):
+        initial_state = current_app.config['FEATURE_FLAG_ENABLE_ORCID_PUSH']
+        current_app.config['FEATURE_FLAG_ENABLE_ORCID_PUSH'] = False
```
drjova: I wouldn't change the current app state, it is something that should happen ``before_first_request`` in Flask, in this stage the app is already initialised.

michamos: the idea here is to change the config dynamically during the migration. Is it a problem if the app is already initialized?

drjova: http://flask.pocoo.org/docs/0.12/appcontext/#the-application-context

> The application setup state in which the application implicitly is on the module level. It starts when the Flask object is instantiated, and it implicitly ends when the first request comes in.

I see two issues here:

* We will not have one single source of truth, because we'll modify the config on the fly
* Concurrency issues between tasks that are running under the same app

szymonlopaciuk: Related to the second point, I ran some tests, and the config change is local to the task process (so because of GIL, there is only one task within the process). As for the first, and because of the above, we don't really change the single source of truth: we don't want to pause all orcid pushes, just the ones caused by the migration.

michamos: another option that is clearly only changing the state locally, would be to use `mock.patch_dict` to change it.

drjova: I'm not sure about your local setup :) Celery has the option to run multiple tasks per process.

Treating ``current_app`` proxy like ``global`` is a bad practice for flask apps in general. @michamos's solution might be a cleaner solution. 

Do we have any other indicator that the record comes from the migration and add your test here? https://github.com/inspirehep/inspire-next/blob/0e501d0cbe84df88b9e53d49ed529f09850cddb6/inspirehep/modules/records/receivers.py#L135-L157?

szymonlopaciuk: Not to my knowledge, however this doesn't solve our problem either way: we only want to prevent manual migration from triggering the push, not e.g. prodsync.

As we are using celery with the default setting of `processes` for pool implementation we are spawning a new process, and interpreter for every task, which doesn't leak configuration changes (I checked on the same setup as QA).

Using `mock.patch` underneath does the same thing what I am, albeit implicitly. I can change it if you think it's cleaner?

david-caro: I'd prefer not using it, mainly because we already saw that it can misled people on what it does, while your solution is clearer that has side-effects (no strong opinion though).

drjova: @szymonlopaciuk thank you for your time investigating the issue :)

None of these solutions are correct, just one is slightly better than the other. It seems like a general architecture problem on how we migrate records. ``current_app`` shouldn't been used like that by its definition.

Since we are sure it'll not affect other tasks in the worker, we can leave it as is. We should avoid such hacks in the future (and hopefully soon we'll stop migrating record from legacy 🙏)

drjova:
ci merge please

PR #3267, opened 2018-03-19
migrator: rename InspireProdRecords

michamos:
This renames InspireProdRecords to LegacyRecordsMirror, which makes the
content of the table more clear. This is part of the `migrator` CLI
refactor that will use the `mirror` terminology.

### DIFF
--- a/inspirehep/modules/migrator/cli.py
+++ b/inspirehep/modules/migrator/cli.py
@@ -42,7 +42,7 @@ from inspire_schemas.api import validate
 from inspire_utils.helpers import force_list
 
 from inspirehep.utils.schema import ensure_valid_schema
-from .models import InspireProdRecords
+from .models import LegacyRecordsMirror
 from .tasks import (
     add_citation_counts,
     migrate,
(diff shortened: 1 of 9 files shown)


inspirehep/modules/migrator/cli.py
```diff
@@ -42,7 +42,7 @@
 from inspire_utils.helpers import force_list
 
 from inspirehep.utils.schema import ensure_valid_schema
-from .models import InspireProdRecords
+from .models import LegacyRecordsMirror
```
david-caro: The name looks quite strange to me, for me a mirror is a process that does mirroring (that's 100% based on my experience of course). I'd name it something like `MigrationJob`, `MigrationTask`, `MigrationCachedRecord` or similar.
The migration is there because it's only used by that set of tasks/cli commands only.

inspirehep/modules/migrator/models.py
```diff
@@ -31,7 +31,8 @@
 from sqlalchemy.ext.hybrid import hybrid_property
 
 
-class InspireProdRecords(db.Model):
+class LegacyRecordsMirror(db.Model):
+    # the model has a different name from the table for compatibility
     __tablename__ = 'inspire_prod_records'
```
david-caro: It's unfortunate to get the table names unsynced after, what would be needed to change it too?

michamos: an alembic script migrating the tables, which adds a bit of complexity.

michamos:
This fails because of alembic issues, closing for now.

david-caro:
You can push the rename only if you want ;)

PR #3268, opened 2018-03-20
workflow: retry arxiv download from non-export url INSPIR-237

puntonim:
https://its.cern.ch/jira/browse/INSPIR-237

### DIFF
--- a/inspirehep/modules/workflows/tasks/arxiv.py
+++ b/inspirehep/modules/workflows/tasks/arxiv.py
@@ -65,12 +65,18 @@ NO_PDF_ON_ARXIV = 'The author has provided no source to generate PDF, and no PDF
 @backoff.on_exception(backoff.expo, DownloadError, base=4, max_tries=5)
 def populate_arxiv_document(obj, eng):
     arxiv_id = get_arxiv_id(obj.data)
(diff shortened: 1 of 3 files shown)


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -67,7 +67,17 @@ def populate_arxiv_document(obj, eng):
...
+    if not is_valid_pdf_link:
+        # Try with the alternative arxiv url.
+        alternative_url = current_app.config['ARXIV_PDF_URL_ALTERNATIVE']\
+            .format(arxiv_id=arxiv_id)
+        is_valid_pdf_link = is_pdf_link(alternative_url)
+        if is_valid_pdf_link:
+            url = alternative_url
+
+    if not is_valid_pdf_link:
         if NO_PDF_ON_ARXIV in requests.get(url).content:
```
szymonlopaciuk: Maybe it worth to run this check also on the result of the first request? If the author didn't make the source available then it's probably not worth querying the second URL. Also then you could put this code in a for-loop or similar?

puntonim: Ok I will try with the for loop.

puntonim: Done

puntonim:
Done

PR #3275, opened 2018-03-21
add more data for fuzzy match results

harunurhan:


inspirehep/modules/workflows/serializers/json.py
```diff
@@ -0,0 +1,32 @@
...
+    def _ref_to_literature_brief(self, record_ref):
+        recid = basename(record_ref['$ref'])
+        record = get_db_record('lit', recid)
+        brief = {
+            'control_number': record['control_number'],
+            'title': get_value(record, 'titles[0].title'),
+        }
+        abstract = get_value(record, 'abstracts[0].value')
+        
+        if (abstract):
```
drjova: it's not js :P

inspirehep/config.py
```diff
@@ -50,7 +50,7 @@
 #   none -> "^$"
 #   some ORCIDs -> "^(0000-0002-7638-5686|0000-0002-7638-5687)$"
 FEATURE_FLAG_ORCID_PUSH_WHITELIST_REGEX = '.*'
-FEATURE_FLAG_ENABLE_FUZZY_MATCHER = False
+FEATURE_FLAG_ENABLE_FUZZY_MATCHER = True
```
drjova: This should stay false :)

inspirehep/modules/workflows/serializers/json.py
harunurhan: that was quick :P

tests/unit/workflows/test_workflows_tasks_matching.py
```diff
@@ -344,7 +347,6 @@ def test_fuzzy_match_returns_true_if_something_matched(mock_match, enable_fuzzy_
     assert fuzzy_match(obj, eng)
     assert 'matches' in obj.extra_data
 
-    expected = [4328]
```
drjova: hmm we need this for the test below, no?

inspirehep/modules/workflows/serializers/json.py
```diff
@@ -0,0 +1,32 @@
...
+from invenio_workflows_ui.serializers.json import JSONSerializer
+from inspire_utils.record import get_value
+from inspirehep.utils.record_getter import get_db_record
+
+
+class WorkflowJSONSerializer(JSONSerializer):
+    def serialize(self, workflow_ui_object):
+        fuzzy_matches = get_value(workflow_ui_object, '_extra_data.matches.fuzzy')
+
+        if fuzzy_matches:
```
drjova: ``if fuzzy_matches is not None:``

inspirehep/modules/workflows/serializers/json.py
```diff
@@ -0,0 +1,32 @@
...
+            workflow_ui_object['_extra_data']['matches']['fuzzy'] = \
+                map(self._ref_to_literature_brief, fuzzy_matches)
+
+        return json.dumps(workflow_ui_object.dumps(), **self._format_args())
+
+    def _ref_to_literature_brief(self, record_ref):
+        recid = basename(record_ref['$ref'])
```
drjova: we should have a test case with malformed ``$ref``, for example ``/api/record/``

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -127,9 +128,9 @@ def fuzzy_match(obj, eng):
 
     fuzzy_match_config = current_app.config['FUZZY_MATCH']
     matches = dedupe_list(match(obj.data, fuzzy_match_config))
-    record_ids = [el['_source']['control_number'] for el in matches]
-    obj.extra_data.setdefault('matches', {})['fuzzy'] = record_ids[0:5]
-    return bool(record_ids)
+    record_refs = [get_record_ref(el['_source']['control_number'], 'literature') for el in matches]
```
drjova: hmm, the default value, ``record``, should be enough https://github.com/inspirehep/inspire-dojson/blob/master/inspire_dojson/utils/__init__.py#L134

tests/integration/workflows/test_workflows_serializers.py
```diff
@@ -0,0 +1,158 @@
...
+    }]
+    with workflow_app.test_client() as client:
+        login_user_via_session(client, email='cataloger@inspirehep.net')
+        response = client.get('/api/holdingpen/' + str(obj.id))
+        result = json.loads(response.data)
+        resolved_fuzzy = result['_extra_data']['matches']['fuzzy']
+        assert resolved_fuzzy == expected_fuzzy
+
+
+def _insert_matched_hep_record_with_validation(control_number, title, abstract_value=None):
```
drjova: hmm I wouldn't add more logic to the tests, you can create fixtures for each case or, even better, run on your test case

inspirehep/modules/workflows/tasks/matching.py
jacquerie: It isn't: look where https://labs.inspirehep.net/api/record/4328 redirects to.

tests/integration/workflows/test_workflows_serializers.py
```diff
@@ -0,0 +1,158 @@
...
+
+    if abstract_value:
+        assert validate(record['abstracts'], abstracts_schema) is None
+    assert validate(record['titles'], titles_schema) is None
+
+    from inspirehep.modules.migrator.tasks import record_insert_or_replace
+    record_insert_or_replace(record)
+
+
+def _get_valid_workflow_metada():
```
drjova: ⬆️ https://github.com/inspirehep/inspire-next/pull/3275/files#r176079444 :)

tests/integration/workflows/test_workflows_serializers.py
harunurhan: maybe fixture is better, because when this goes inside tests, it makes them pretty unreadable.

inspirehep/modules/workflows/tasks/matching.py
drjova: Is it any special requirement for that? It seems to me very strange to redirect the api to a html page.

harunurhan:
@drjova how about cleanup after tests? Is it needed?

inspirehep/modules/workflows/serializers/json.py
harunurhan: so this means we have to handle it too. How do you propose to handle this case?

tests/unit/workflows/test_workflows_tasks_matching.py
harunurhan: it's now defined up

drjova:
Supersede by https://github.com/inspirehep/inspire-next/pull/3280

PR #3279, opened 2018-03-22
orcid: add get_literature_recids_by_orcid util

jacquerie:
## Description:
We record that an author of recid X has claimed a paper of recid Y by
setting ``curated_relation`` to ``True`` in the author signature of Y
that also contains a record reference to X.

In order to later retrieve this information we need to nest this field,
otherwise we would get spurious matches across different signatures.
See also c20f672 for a similar argument for ``publication_info``s.

## Related Issue:
https://its.cern.ch/jira/browse/INSPIR-343

## Checklist:

### DIFF
--- a/inspirehep/modules/authors/rest/citations.py
+++ b/inspirehep/modules/authors/rest/citations.py
@@ -24,6 +24,8 @@ from __future__ import absolute_import, division, print_function
 
 import json
 
+from elasticsearch_dsl import Q
+
 from inspirehep.modules.search import LiteratureSearch
 
 
@@ -46,17 +48,13 @@ class AuthorAPICitations(object):
         author_pid = pid.pid_value
         citations = {}
 
-        search = LiteratureSearch().query({
-            "match": {
-                "authors.recid": author_pid
-            }
-        }).params(
-            _source=[
-                "authors.recid",
-                "control_number",
-                "self",
-            ]
-        )
+        query = Q('match', authors__recid=author_pid)
+        search = LiteratureSearch().query('nested', path='authors', query=query)\
+                                   .params(_source=[
+                                       'authors.recid',
+                                       'control_number',
+                                       'self',
+                                   ])
 
         # For each publication co-authored by a given author...
         for result in search.scan():
--- a/inspirehep/modules/orcid/utils.py
+++ b/inspirehep/modules/orcid/utils.py
@@ -28,9 +28,12 @@ import hashlib
 import re
 
 from itertools import chain
+from elasticsearch_dsl import Q
 from flask import current_app
 from six.moves.urllib.parse import urljoin
 from StringIO import StringIO
+from sqlalchemy import type_coerce
+from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm.exc import NoResultFound
 
 from invenio_db import db
@@ -40,11 +43,14 @@ from invenio_oauthclient.models import (
     RemoteToken,
 )
 from invenio_oauthclient.utils import oauth_link_external_id
+from invenio_pidstore.models import PersistentIdentifier
+from invenio_records.models import RecordMetadata
 
 from inspire_dojson.utils import get_recid_from_ref
 from inspire_utils.logging import getStackTraceLogger
 from inspire_utils.record import get_values_for_schema
 from inspire_utils.urls import ensure_scheme
+from inspirehep.modules.search.api import LiteratureSearch
 from inspirehep.utils.record_getter import get_db_records
 
 LOGGER = getStackTraceLogger(__name__)
@@ -209,3 +215,36 @@ def canonicalize_xml_element(element):
         exclusive=True,
     )
     return output_stream.getvalue()
+
+
+def get_literature_recids_for_orcid(orcid):
+    """Return the Literature recids that were claimed by an ORCiD.
+
+    We record the fact that the Author record X has claimed the Literature
+    record Y by storing in Y an author object with a ``$ref`` pointing to X
+    and the key ``curated_relation`` set to ``True``. Therefore this method
+    first searches the DB for the Author records for the one containing the
+    given ORCiD, and then uses its recid to search in ES for the Literature
+    records that satisfy the above property.
+
+    Args:
+        orcid (str): the ORCiD.
+
+    Return:
+        list(int): the recids of the Literature records that were claimed
+        by that ORCiD.
+
+    """
+    orcid_object = '[{"schema": "ORCID", "value": "%s"}]' % orcid
+    author_recid = db.session.query(PersistentIdentifier.pid_value)\
+                             .filter(PersistentIdentifier.object_type == 'rec')\
+                             .filter(PersistentIdentifier.object_uuid == RecordMetadata.id)\
+                             .filter(PersistentIdentifier.pid_type == 'aut')\
+                             .filter(type_coerce(RecordMetadata.json, JSONB)['ids'].contains(orcid_object))\
+                             .one().pid_value
+
+    query = Q('match', authors__curated_relation=True) & Q('match', authors__recid=author_recid)
+    search_by_curated_author = LiteratureSearch().query('nested', path='authors', query=query)\
+                                                 .params(_source=['control_number'])
+
+    return [el['control_number'] for el in search_by_curated_author]
--- a/tests/integration/orcid/test_orcid_utils.py
+++ b/tests/integration/orcid/test_orcid_utils.py
@@ -25,11 +25,17 @@
 from __future__ import absolute_import, division, print_function
 
 import pytest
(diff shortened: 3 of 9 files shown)


inspirehep/modules/orcid/utils.py
```diff
@@ -209,3 +216,26 @@ def canonicalize_xml_element(element):
...
+                         .filter(PersistentIdentifier.object_uuid == RecordMetadata.id)\
+                         .filter(PersistentIdentifier.pid_type == 'aut')\
+                         .filter(type_coerce(RecordMetadata.json, JSONB)['ids'].contains(orcid_object))\
+                         .one().pid_value
+    except MultipleResultFound as e:
+        LOGGER.error('Multiple author records found for ORCiD {}'.format(orcid))
+        six.reraise(e)
+    except NoResultFound as e:
+        LOGGER.error('No author record found for ORCiD {}'.format(orcid))
+        six.reraise(e)
```
jacquerie: @puntonim this, or just remove the error handling from this function and let it blow?

puntonim: Remove entirely the error handling.
This code writes twice to Sentry (one with the LOGGER and one with reraise). What is the advantage of this? If it's the error message then I guess you can do six.raise_from or edit the message in e.

inspirehep/modules/orcid/utils.py
```diff
@@ -178,6 +184,7 @@ def get_orcids_for_push(record):
     return chain(orcids_on_record, orcids_in_authors)
 
 
+<<<<<<< 5755a8dc5fd82b6cba3ccdfc6c41bd1dd703f9a4
```
drjova: oups :)

inspirehep/modules/orcid/utils.py
```diff
@@ -209,3 +215,35 @@ def canonicalize_xml_element(element):
...
+        list(int): the recids of the Literature records that were claimed
+        by that ORCiD.
+
+    """
+    orcid_object = '[{"schema": "ORCID", "value": "%s"}]' % orcid
+    author_recid = db.session.query(PersistentIdentifier.pid_value)\
+                             .filter(PersistentIdentifier.object_uuid == RecordMetadata.id)\
+                             .filter(PersistentIdentifier.pid_type == 'aut')\
+                             .filter(type_coerce(RecordMetadata.json, JSONB)['ids'].contains(orcid_object))\
+                             .one().pid_value
```
puntonim: I want to share my thought about this query. I come from the Django ORM where joins are never explicit, so I often think in terms of 2 sequential queries rather than a join.

So my question is what is more performant between a join query and 2 sequential queries. Or if this question even makes sense.

Your query translates to:
```sql
SELECT pidstore_pid.pid_value AS pidstore_pid_pid_value
FROM pidstore_pid, records_metadata
WHERE pidstore_pid.object_uuid = records_metadata.id AND pidstore_pid.pid_type = %(pid_type_1)s AND ((records_metadata.json -> %(param_1)s)) @> %(param_2)s
```

While 2 sequential queries like:
```python
record_ids = db.session.query(RecordMetadata.id)\
                    .filter(type_coerce(RecordMetadata.json, JSONB)['ids'].contains(orcid_object)).all()
author_recid = db.session.query(PersistentIdentifier.pid_value)\
                         .filter(PersistentIdentifier.object_uuid.in_(record_ids))\
                         .filter(PersistentIdentifier.pid_type == 'aut').one()
```
translates to:
```sql
SELECT records_metadata.id AS records_metadata_id
FROM records_metadata
WHERE ((records_metadata.json -> %(param_1)s)) @> %(param_2)s

SELECT pidstore_pid.pid_value AS pidstore_pid_pid_value
FROM pidstore_pid
WHERE pidstore_pid.object_uuid IN (%(object_uuid_1)s) AND pidstore_pid.pid_type = %(pid_type_1)s
```

tests/integration/orcid/test_orcid_utils.py
```diff
@@ -195,3 +199,10 @@ def test_orcids_for_push_orcid_in_author_with_claim(author_in_isolated_app):
 
     assert validate(record, 'hep') is None
     assert list(get_orcids_for_push(record)) == ['0000-0002-1825-0097']
+
+
+def test_get_literature_recids_for_orcid(isolated_app):
```
puntonim: You are only testing the happy flow. What about cases like when there is no PID or there are multiple PIDs or there is no 'ids' field at all, or the 'ids' field contains no ORCID schema or the 'ids' field contains the ORCID schema but the value is empty, etc

inspirehep/modules/authors/rest/citations.py
```diff
@@ -46,17 +48,13 @@ def serialize(self, pid, record, links_factory=None):
...
-            }
-        }).params(
-            _source=[
-                "authors.recid",
-                "control_number",
-                "self",
-            ]
-        )
+        query = Q('match', authors__recid=author_pid)
+        search = LiteratureSearch().query('nested', path='authors', query=query)\
```
puntonim: You edited a number of LiteratureSearch queries. Do we have tests already covering this lines?

jacquerie: Yes, these ones: https://github.com/inspirehep/inspire-next/blob/f52318ab40727c836e49aacf23f36985d72f6134/tests/integration/test_author_api.py (which in fact were previously failing: https://travis-ci.org/inspirehep/inspire-next/jobs/357007070#L1471)

tests/integration/orcid/test_orcid_utils.py
jacquerie: Whoops, absolutely! Will add all these tests.

inspirehep/modules/orcid/utils.py
jacquerie: You just made me realize an important point: there's an index on `(object_type, object_uuid)`, which is not used when joining on just `object_uuid`:
```
inspirehep=# EXPLAIN SELECT * FROM pidstore_pid WHERE object_uuid='b70bd0bb-4ea9-4b8b-a56c-a17910e27022';
                               QUERY PLAN                               
------------------------------------------------------------------------
 Seq Scan on pidstore_pid  (cost=0.00..11.50 rows=1 width=638)
   Filter: (object_uuid = 'b70bd0bb-4ea9-4b8b-a56c-a17910e27022'::uuid)
(2 rows)

inspirehep=# EXPLAIN SELECT * FROM pidstore_pid WHERE object_type='rec' AND object_uuid='b70bd0bb-4ea9-4b8b-a56c-a17910e27022';
                                                      QUERY PLAN                                                      
----------------------------------------------------------------------------------------------------------------------
 Index Scan using idx_object on pidstore_pid  (cost=0.14..8.16 rows=1 width=638)
   Index Cond: (((object_type)::text = 'rec'::text) AND (object_uuid = 'b70bd0bb-4ea9-4b8b-a56c-a17910e27022'::uuid))
(2 rows)
```

jacquerie: Another possibility is to also filter on `object_type`, and then one can use the index. Something smarter would be to declare the index on `(object_uuid, object_type)`, so that it could be used for either!

tests/integration/orcid/test_orcid_utils.py
jacquerie: For the record, I added below tests for:
- there is no PID
- there are multiple PIDs
- there is no 'ids' field at all
- the 'ids' field contains no ORCID schema

but we agreed that don't need to write a test for 
- the 'ids' field contains the ORCID schema but the value is empty

as this is disallowed by the schema.

tests/integration/test_db_isolation.py
```diff
@@ -81,25 +81,3 @@ def test_isolated_app_fixture_has_db_isolation_step2(pids_count_in_app, isolated
...
-
-def test_isolated_app_fixture_rollback(isolated_app):
-    pids_count = PersistentIdentifier.query.count()
-
-    PersistentIdentifier.create(
-        pid_type='type1',
-        pid_value='value1',
-    )
-    db.session.rollback()
-    assert PersistentIdentifier.query.count() == pids_count
```
jacquerie: I think that both these tests are showing a wrong way of using `isolated_app`, so I deleted them: why would anyone want to commit or rollback explicitly in a test?

michamos: What can happen is that you are testing code that commits/rolls back. Don't you want to test that those kind of tests work?

puntonim: I am fine with deleting these tests.
I believe we should replace them with tests that proves what happens when we should explicitly the `db.session.commit` and `rollback`. The current behavior is that there is no isolation, so prove that with a test.
in the long run, our isolation should work also in such cases.

And I believe we should be able to explicitly call commit and rollback in tests, just because we want to test explicit commit and rollback.

PR #3280, opened 2018-03-22
workflows: add more data for fuzzy match results

harunurhan:


inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -127,11 +127,23 @@ def fuzzy_match(obj, eng):
...
 
 
+def _get_literature_summary(literature):
+    summary = {
+        'control_number': literature['control_number'],
+        'title': get_value(literature, 'titles[0].title'),
+    }
+
+    abstract = get_value(literature, 'abstracts[0].value')
+    if abstract:
```
drjova: ``is not None`` :P

tests/unit/workflows/test_workflows_tasks_matching.py
```diff
@@ -344,12 +366,53 @@ def test_fuzzy_match_returns_true_if_something_matched(mock_match, enable_fuzzy_
...
+    expected = [{
+        'control_number': 4328,
+        'title': 'title',
+    }]
     result = get_value(obj.extra_data, 'matches.fuzzy')
 
     assert expected == result
 
 
+
```
drjova: :P

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -111,7 +111,7 @@ def fuzzy_match(obj, eng):
     similar to the current workflow object's payload in the system.
 
     Also sets the ``matches.fuzzy`` property in ``extra_data`` to the list of
-    the first 5 control numbers that matched.
+    the first 5 record summary that matched.
```
drjova: record summary is a bit weird :/

harunurhan: brief is good? :)

drjova: the first 5 matched records, with the following format::

```
{
    control_number: '',
    title: '',
    abstract: '',
}
```

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -127,11 +127,23 @@ def fuzzy_match(obj, eng):
 
     fuzzy_match_config = current_app.config['FUZZY_MATCH']
     matches = dedupe_list(match(obj.data, fuzzy_match_config))
-    record_ids = [el['_source']['control_number'] for el in matches]
+    record_ids = [_get_literature_brief(el['_source']) for el in matches]
     obj.extra_data.setdefault('matches', {})['fuzzy'] = record_ids[0:5]
     return bool(record_ids)
 
 
+def _get_literature_brief(literature):
```
iulianav: I think maybe just `literature` is not a specific enough parameter name?

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -127,11 +127,23 @@ def fuzzy_match(obj, eng):
...
+    brief = {
+        'control_number': literature['control_number'],
+        'title': get_value(literature, 'titles[0].title'),
+    }
+
+    abstract = get_value(literature, 'abstracts[0].value')
+    if abstract is not None:
```
iulianav: Maybe do just `if abstract:` is enough? Or are there empty string abstract values and we care about them?

tests/unit/workflows/test_workflows_tasks_matching.py
```diff
@@ -333,7 +333,29 @@ def test_pending_in_holding_pen_returns_false_if_nothing_matched(mock_match):
...
-    mock_match.return_value = iter([{'_source': {'control_number': 4328}}])
+    schema = load_schema('hep')
+    abstracts_schema = schema['properties']['abstracts']
+    titles_schema = schema['properties']['titles']
+
+    matched_record = {
+        'control_number': 4328,
+        'abstracts': [
+            {
+                'value': 'abstract',
```
iulianav: We should strive to use test data similar to production data.

inspirehep/modules/workflows/tasks/matching.py
drjova: good point! abstracts are not required by the schema so in many cases they might be empty and is better not to send empty keys :)

inspirehep/modules/workflows/tasks/matching.py
harunurhan: what do you suggest? sth more explicit like `literature_record`?

inspirehep/modules/workflows/tasks/matching.py
harunurhan: It is very unlikely the case but do you think there could be value in knowing if abstract is present but empty?

@drjova

jacquerie: Since commit https://github.com/inspirehep/inspire-schemas/commit/470fff969b468fd68dff1e2c04ea89cfbaa75512 **every** list or string present in the schema if it's present is non-empty, so you don't have to deal with that case.

inspirehep/modules/theme/static/scss/holding-pen/holding-pen.scss
```diff
@@ -662,16 +661,39 @@ $hp_table_bg: #fff;
...
+  width: 100%;
+  
+  .panel-heading {
+    font-size: 12px;
+  }
+
+  .authors {
+    margin-bottom: 10px;
+    font-size: 14px;
+    color: #8aa2ba;
```
drjova: https://github.com/inspirehep/inspire-next/blob/a1231dee4c98fff9253f8cb205d2be7cf4371175/inspirehep/modules/theme/static/scss/base/variables.scss#L30 :)

drjova:
@harunurhan could you please also rebase and upload a screenshot?

PR #3290, opened 2018-03-26
global: centralization of minting

drjova:
* Moves the minting functionality inside the ``InspireRecord`` which from
  now on it will be handled internally.

* Removes ``record_insert_or_replace`` method from the ``migrator`` and
  tests.

* Moves the minting before the record creation, this helps us to prevent
  record creation without without ``control_number``.

* NOTE: This only moves the logic inside the ``InspireRecord``, it doesn't
  refactor any of the existing code. Regarding the minting handling for
  different actions, such as ``update``, ``delete`` etc, will be
  introduced in a different commit.

---

Note: this is part 1 of 3. This PR only removes the ``record_insert_or_replace`` function and moves the minting logic inside the ``InspireRecord``.  It doesn't refactor any existing code or handle more minting cases.


inspirehep/modules/migrator/tasks.py
```diff
@@ -379,7 +347,8 @@ def migrate_and_insert_record(raw_record, skip_files=False):
         ensure_valid_schema(json_record)
 
     try:
-        record = record_insert_or_replace(json_record, skip_files=skip_files)
+        record = InspireRecord.create_or_update(json_record, skip_files=skip_files)
+        record.commit()
```
iulianav: Is there no workaround having to call `commit()` every time outside InspireRecord? Or we want that to push several updates with just one commit, even though I believe you can commit subsequently without issues?

drjova: Good point! The idea is to move the logic of the ``commit`` outside of ``InspireRecord`` so we can call it explicitly when we need it. 

[Record](https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L137) has exactly the same behaviour, for some reason we didn't respect and we had inconsistencies such as ``delete`` was ``commit``ing but ``create`` was ``commit``ing only if ``skip_files`` was ``True``.

inspirehep/modules/records/api.py
```diff
@@ -77,24 +80,92 @@ def create(cls, *args, **kwargs):
...
+            record.download_documents_and_figures(
                 src_records=files_src_records,
             )
-            new_record.commit()
 
-        return new_record
+        return record
+
+    @classmethod
+    def create_or_update(cls, data, **kwargs):
```
michamos: conceptually, I would expect `create_or_update` to be something like:
```
try:
    update(...)
except PIDDoesNotExistError:
    create(...)
```
what is the reason it has to be much more complex than that? (I know you are not refactoring anything yet, those are hints for later).

inspirehep/modules/records/api.py
```diff
@@ -77,24 +80,92 @@ def create(cls, *args, **kwargs):
...
+        pid_type = get_pid_type_from_schema(data['$schema'])
+        control_number = data.get('control_number')
+
+        files_src_records = kwargs.pop('files_src_records', [])
+        skip_files = kwargs.pop(
+            'skip_files', current_app.config.get('RECORDS_SKIP_FILES'))
+
+        try:
+            pid = PersistentIdentifier.get(pid_type, control_number)
+            record = super(InspireRecord, cls).get_record(pid.object_uuid)
```
michamos: we should pull `get_db_record`, `get_db_records` (as `get_record_by_pid`, `get_records_by_pid`) inside the class as it's more generally useful.

inspirehep/modules/workflows/tasks/upload.py
```diff
@@ -62,11 +61,10 @@ def _get_updated_record(obj):
         # TODO: remove the skip files once labs becomes master
         record = InspireRecord.create(obj.data, id_=None, skip_files=True)
         # Create persistent identifier.
-        created_pid = inspire_recid_minter(str(record.id), record).pid_value
         # Now that we have a recid, we can properly download the documents
         record.download_documents_and_figures(src_records=[obj])
 
-        obj.data['control_number'] = created_pid
+        obj.data['control_number'] = record['control_number']
```
michamos: could this function be rewritten in terms of `InspireRecord.create_or_update`?

tests/integration/conftest.py
```diff
@@ -104,7 +104,7 @@ def isolated_app(app):
 
     """
     with db.session.begin_nested():
-        yield
+        yield app
```
michamos: :+1: I also needed this today

inspirehep/modules/workflows/tasks/upload.py
drjova: hmm this is for the workflow data, IMHO it's better to keep it separated

PR #3291, opened 2018-03-26
hal: cli script for bulk push

puntonim:


inspirehep/modules/hal/cli.py
```diff
@@ -0,0 +1,40 @@
...
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+import click
+
+from flask_cli import with_appcontext
```
drjova: ```flask.cli  import with_appcontext```, ```flask_cli``` has been deprecated :)

inspirehep/modules/hal/bulk_push.py
```diff
@@ -0,0 +1,113 @@
...
+from flask import current_app
+
+from invenio_records.models import RecordMetadata
+from inspirehep.modules.hal.core.tei import convert_to_tei
+from inspirehep.modules.hal.core.sword import create, update
+
+
+def _set_config():
+    # Set the proper configuration.
+    current_app.config['RECORDS_SKIP_FILES'] = True
```
jacquerie: This is unneeded.

inspirehep/modules/hal/cli.py
puntonim: Done

inspirehep/modules/hal/bulk_push.py
puntonim: Done

drjova:
If we'll keep the script for a while, IMO would be safer to use `` click.echo`` instead of ``print``. WDYT?

puntonim:
done

PR #3292, opened 2018-03-27
docker: add health check for the worker node

david-caro:
This prevents the worker from starting up after the tests started
and cleaning up any pending jobs.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -87,7 +87,7 @@ services:
       test-redis:
         condition: service_healthy
       test-worker:
-        condition: service_started
+        condition: service_healthy
       selenium:
         condition: service_started
       test-web:
@@ -113,7 +113,7 @@ services:
       test-redis:
         condition: service_healthy
       test-worker:
-        condition: service_started
+        condition: service_healthy
       selenium:
         condition: service_started
       test-web:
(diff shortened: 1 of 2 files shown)


david-caro:
@iulianav can you rebase your pr on top of this and see if the literature tests still fail?

docker-compose.test.yml
```diff
@@ -194,6 +194,15 @@ services:
     command: celery worker -E -A inspirehep.celery --loglevel=INFO --purge --queues celery,migrator,harvests,orcid_push
     volumes_from:
       - test-static
+    healthcheck:
+      timeout: 5s
+      interval: 5s
+      retries: 5
+      test:
+        - "CMD"
```
jacquerie: Why are we using sometimes `CMD` and sometimes `CMD-SHELL` ? I can't see the pattern.

david-caro: The difference is that CMD will run the command 'bare', with CMD-SHELL it will start the system shell, and run the given command there.
Usually if you want a specific shell (ex. bash) you should use CMD, if you don't care, but still want a shell, use CMD-SHELL, and if you don't need a shell, just CMD.

It's not a big issue to use CMD-SHELL when you could use  CMD, but the other way around might not work. Some features like `exec 3<> /dev/tcp/127.0.0.1/6379` or `[[]]` are bash specific though.

jacquerie:
@ammirate, can you check this? I would like @david-caro to merge it, but since he's not here I'd say that the decision falls back on you.

PR #3299, opened 2018-03-28
orcid: split get_literature_recids_for_orcid query

michamos:
This allows the database to use an index on the `(json -> ids)`, which
it doesn't do if the join on the PIDStore happens in the same query.

## Checklist:

### DIFF
--- a/inspirehep/modules/orcid/utils.py
+++ b/inspirehep/modules/orcid/utils.py
@@ -236,12 +236,14 @@ def get_literature_recids_for_orcid(orcid):
 
     """
     orcid_object = '[{"schema": "ORCID", "value": "%s"}]' % orcid
-    author_recid = db.session.query(PersistentIdentifier.pid_value)\
-                             .filter(PersistentIdentifier.object_type == 'rec')\
-                             .filter(PersistentIdentifier.object_uuid == RecordMetadata.id)\
-                             .filter(PersistentIdentifier.pid_type == 'aut')\
-                             .filter(type_coerce(RecordMetadata.json, JSONB)['ids'].contains(orcid_object))\
-                             .one().pid_value
+    # this first query is written in a way that can use the index on (json -> ids)
+    author_rec_uuid = db.session.query(RecordMetadata.id)\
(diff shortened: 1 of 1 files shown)


inspirehep/modules/orcid/utils.py
```diff
@@ -236,12 +236,13 @@ def get_literature_recids_for_orcid(orcid):
...
     """
     orcid_object = '[{"schema": "ORCID", "value": "%s"}]' % orcid
-    author_recid = db.session.query(PersistentIdentifier.pid_value)\
-                             .filter(PersistentIdentifier.object_type == 'rec')\
-                             .filter(PersistentIdentifier.object_uuid == RecordMetadata.id)\
-                             .filter(PersistentIdentifier.pid_type == 'aut')\
-                             .filter(type_coerce(RecordMetadata.json, JSONB)['ids'].contains(orcid_object))\
-                             .one().pid_value
+    author_rec_uuid = db.session.query(RecordMetadata.id)\
+                    .filter(type_coerce(RecordMetadata.json, JSONB)['ids'].contains(orcid_object)).one().id
```
jacquerie: Instead of `.one().id` -> `.scalar()`

inspirehep/modules/orcid/utils.py
```diff
@@ -236,12 +236,13 @@ def get_literature_recids_for_orcid(orcid):
...
+    author_recid = db.session.query(PersistentIdentifier.pid_value).filter(
+        PersistentIdentifier.object_type == 'rec',
+        PersistentIdentifier.object_uuid == author_rec_uuid,
+        PersistentIdentifier.pid_type == 'aut',
+    ).one().pid_value
```
jacquerie: Instead of `.one().pid_value` -> `.scalar()`

inspirehep/modules/orcid/utils.py
michamos: scalar has different semantics when there are no results: http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.scalar

PR #3311, opened 2018-04-03
forms: remove validation for duplicate ORCIDs

drjova:
* Removes the validation for duplicate ORCIDs.

### DIFF
--- a/inspirehep/modules/authors/forms.py
+++ b/inspirehep/modules/authors/forms.py
@@ -563,6 +563,12 @@ class AuthorUpdateForm(INSPIREForm):
         """Constructor."""
         super(AuthorUpdateForm, self).__init__(*args, **kwargs)
         is_review = kwargs.pop('is_review', False)
+        is_update = kwargs.pop('is_update', False)
+
+        if is_update:
+            # remove validation for duplicate ORCIDs on update
+            self.orcid.validators = self.orcid.validators[:-1]
+
         if is_review:
             self.bai.widget = TextInput()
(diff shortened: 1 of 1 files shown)


inspirehep/modules/authors/forms.py
```diff
@@ -563,6 +563,12 @@ def __init__(self, *args, **kwargs):
         """Constructor."""
         super(AuthorUpdateForm, self).__init__(*args, **kwargs)
         is_review = kwargs.pop('is_review', False)
+        is_update = kwargs.pop('is_update', False)
```
jacquerie: Woah, this trusts whatever is in the user request to tell us what to validate? https://github.com/inspirehep/inspire-next/blob/4d1f980894252e38459c1a9dc15762cc169cb504/inspirehep/modules/authors/views.py#L189

inspirehep/modules/authors/forms.py
```diff
@@ -563,6 +563,12 @@ def __init__(self, *args, **kwargs):
...
+
+        if is_update:
+            # remove validation for duplicate ORCIDs on update
+            self.orcid.validators = self.orcid.validators[:-1]
```
jacquerie: Can we do better than this? Something like a sanity check that the ORCID corresponds to what was there before, or that the ORCID profile has a name that is similar enough or...

michamos: we could do lots of things. Currently, we have a major regression: any author update fails if the author already has an ORCID. Let's fix that first.

PR #3313, opened 2018-04-03
forms: fix autocomplete when conference doesn't have address[0]citie…

Dinika:
Signed-off-by: Dinika Saxena <dinika.saxena@cern.ch>

## Description
Fixes autocomplete in literature submission form when the first object in address field of conferences does not contain the `cities` key

As a side note, same issue will appear in the [conferences template](https://github.com/inspirehep/inspire-next/blob/21a41c975d84bc82c81531b1d57ef63bd949f678/inspirehep/modules/theme/templates/inspirehep_theme/format/record/Conference_HTML_detailed.tpl#L97)

## Checklist:

### DIFF
--- a/inspirehep/modules/forms/static/js/forms/conferences_typeahead.js
+++ b/inspirehep/modules/forms/static/js/forms/conferences_typeahead.js
@@ -50,8 +50,9 @@ define([
           url: '/api/conferences?q=conferenceautocomplete:%QUERY*',
           filter: function(response) {
             return $.map(response.hits.hits, function(el) {
-              el.metadata.city = el.metadata.address[0].cities[0];
-              el.metadata.country = el.metadata.address[0].country_code;
+              obj = el.metadata.address.find(function (obj) { return obj.hasOwnProperty('cities') });
+              el.metadata.city = obj.cities ? obj.cities[0] : '';
+              el.metadata.country = obj.country_code ? obj.country_code : '';
               el.metadata.title = el.metadata.titles[0].title;
               return el.metadata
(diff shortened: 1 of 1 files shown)


inspirehep/modules/forms/static/js/forms/conferences_typeahead.js
```diff
@@ -50,8 +50,9 @@ define([
           url: '/api/conferences?q=conferenceautocomplete:%QUERY*',
           filter: function(response) {
             return $.map(response.hits.hits, function(el) {
-              el.metadata.city = el.metadata.address[0].cities[0];
-              el.metadata.country = el.metadata.address[0].country_code;
+              obj = el.metadata.address.find(function (obj) { return obj.hasOwnProperty('cities') });
```
Dinika: I'm using this instead of simply adding an `if(el.metadata.address[0].cities)` because in some cases (example, [this](https://labs.inspirehep.net/api/conferences/1646000) one) the object containing `cities` key is not the first element in the `address` array.

inspirehep/modules/forms/static/js/forms/conferences_typeahead.js
```diff
@@ -50,8 +50,9 @@ define([
...
+              el.metadata.city = obj.cities ? obj.cities[0] : '';
```
drjova: Hmm ``obj`` is an array ([filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter)) hence ``obj.cities`` will be always ``undefined``

PR #3316, opened 2018-04-04
db: sql snippet to create index on RecordMetadata._export_to INSPIR-343

puntonim:


scripts/index_on_jsonb_field.txt
```diff
@@ -20,3 +29,29 @@ migration. See: https://its.cern.ch/jira/browse/INSPIR-375
...
 
 -- DROP INDEX CONCURRENTLY json_ids_index;
+
+
+2. RECORDSMETADATA.JSON _export_to
+==================================
+The following SQL query adds an index to RecordMetadata.json field.
+The index can then be used in queries like:
+
+RecordMetadata.query.filter(RecordMetadata.json['_export_to'].op('@>')('[{"HAL": true}]'))
```
jacquerie: This is exactly the query above, just expressed in a syntax closer to PostgreSQL. You can also write it as
```python
RecordMetadata.query.filter(type_coerce(RecordMetadata.json, JSONB)['_export_to'].contains('[{"HAL": true}]')
```

jacquerie: I mean, it's totally not the point, but this bugged me.

puntonim: Done

michamos: You can even do `contains({'HAL': True})`

PR #3317, opened 2018-04-04
orcid: better query for push_to_orcid receiver INSPIR-334

puntonim:


inspirehep/modules/records/receivers.py
```diff
@@ -142,10 +145,11 @@ def push_to_orcid(sender, record, *args, **kwargs):
 
     task_name = current_app.config['ORCID_PUSH_TASK_ENDPOINT']
 
-    for orcid in get_orcids_for_push(record):
-        token = get_push_access_token(orcid)
-        if token is None:
-            continue
+    orcids = get_orcids_for_push(record)
```
puntonim: This code produces the same results as the old one, it just leverages the function `get_push_access_tokens` which is an optimized query.
Thus the old tests (the tests starting with `test_orcid_push*` here: https://github.com/inspirehep/inspire-next/blob/master/tests/integration/records/test_records_receivers.py#L262) exercise this new code.

puntonim:
Done. Old function deleted. But I left the new function where the old one was because the proper place would be a model manager that aggregates all the queries for a given model. But this is a new concept and it'd be a major change as currently queries are spread in every corner.

inspirehep/modules/records/receivers.py
```diff
@@ -142,10 +145,11 @@ def push_to_orcid(sender, record, *args, **kwargs):
...
+    remote_tokens = get_push_access_tokens(orcids)
+    for remote_token in remote_tokens:
+        access_token = remote_token.access_token
+        orcid = remote_token.remote_account.extra_data['orcid']
```
michamos: doesn't it do N+1 queries now?  1 to get all tokens, N to get the corresponding orcid through this line.

PR #3318, opened 2018-04-04
workflows: remove extra with_debug_logging

jacquerie:
## Description:
The ``with_debug_logging`` decorator should only be used when the
decorated function is a workflow step, so that we can trace the
workflow execution from the logs.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/beard.py
+++ b/inspirehep/modules/workflows/tasks/beard.py
@@ -33,7 +33,6 @@ from inspirehep.modules.workflows.utils import json_api_request
 from ..utils import with_debug_logging
 
 
-@with_debug_logging
 def get_beard_url():
     """Return the BEARD URL endpoint, if any."""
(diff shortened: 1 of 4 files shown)


michamos:
Your claim doesn't seem to be true. When it's not wrapping a workflow step and there is no `obj` argument, it logs but on the standard logger: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/utils/__init__.py#L128.

jacquerie:
> When it's not wrapping a workflow step and there is no obj argument, it logs but on the standard logger

It's always that logger, anyway: https://github.com/inveniosoftware-contrib/invenio-workflows/blob/617e7dbad1b06767ad23d2ddef3569443778eea8/invenio_workflows/api.py#L74-L77.

I worded it incorrectly: that decorator is meant to go on workflow steps so that we can see from the logs which step we are executing. It was added to a bunch of other functions for no good reason (who cares that we are executing some random helper function?)

PR #3332, opened 2018-04-06
orcid: get orcids and push tokens in one DB query

michamos:


michamos:
screenshot-2018-4-6 postgres explain visualizer pev
profile of the query on prod using `EXPLAIN ...` and http://tatiyants.com/pev/ to visualize it.

PR #3340, opened 2018-04-10
orcid: create a separate logger for orcid

szymonlopaciuk:
## Description

**OUTDATED, left for historical reasons:**

*We are currently losing logs for ORCID pushes, for an unknown reason. I tried to troubleshoot but after over a day of coming up blank and discussing with @puntonim we decided that this may be a good opportunity to introduce explicit logging to a file, for now for the `orcid` module. Having the logs for ORCID can enable us to learn about the statistics such as ratio of updates to creations.*

**EDIT:**

This now successfully logs to stderr for the orcid module.

Turns out our problem comes from somewhere in raven/invenio-logging where the loggers are patched. For some reason, when enabling sentry all logs below `LOGGING_SENTRY_LEVEL` are lost. Investigation shows that the propagation on the `inspirehep` logger is disabled which causes messages to not be passed to the root logger. Regardless, the root logger has only the sentry handler registered, so it's unclear if this is the source of the problem of lost logs. It seems also, that the handler is registered twice on `inspirehep`, for no apparent reason, and I'm not sure what causes it exactly.

This matter requires further investigation, but this PR aims to enable logging for ORCID ASAP, to be able to collect information on how we're doing with the push.

## Checklist:

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -29,6 +29,7 @@ import sys
 import pkg_resources
 
 from celery.schedules import crontab
+from logging.config import dictConfig
 
 from invenio_oauthclient.contrib import orcid
 from invenio_records_rest.facets import range_filter, terms_filter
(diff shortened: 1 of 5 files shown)


david-caro:
ci test please

PR #3341, opened 2018-04-11
workflows: addition of rootless merge

drjova:
* Accepts ``is-update`` workflows and passes through the merge step. If
  there are not conflicts the workflow will continue otherwise it'll
  change the status to HALTED and wait for manual merge through the
  editor.

* Adds an endpoint to resolve merge conflicts from the editor.

* Introduces two feature flags ``FEATURE_FLAG_ENABLE_MERGE_ON_UPDATE``
  to allow merges on update  and
  ``FEATURE_FLAG_ENABLE_UPLOAD_TO_LEGACY_ON_UPDATE`` to send a replace
  on legacy on update.

* Removes obsolete step ``STOP_IF_TOO_OLD``.

* closes INSPIR-304

### DIFF
--- a/inspirehep/modules/workflows/tasks/merging.py
+++ b/inspirehep/modules/workflows/tasks/merging.py
@@ -24,7 +24,17 @@
 
 from __future__ import absolute_import, division, print_function
 
+from flask import current_app
+
+from inspire_json_merger.api import merge
+from invenio_pidstore.models import PersistentIdentifier
+
+from inspirehep.modules.records.api import InspireRecord
 from inspirehep.modules.workflows.models import WorkflowsRecordSources
+from inspirehep.modules.workflows.utils import (
+    get_resolve_merge_conflicts_callback_url,
+    with_debug_logging
+)
 
 
 def get_head_source(head_uuid):
@@ -48,3 +58,53 @@ def get_head_source(head_uuid):
         return None
 
     return 'arxiv' if 'arxiv' in roots_sources else 'publisher'
+
+
+@with_debug_logging
+def has_conflicts(obj, eng):
+    """Return if the workflow has any confilicts."""
+    return obj.extra_data.get('conflicts')
+
+
+@with_debug_logging
+def merge_articles(obj, eng):
+    """Merge two articles.
+
+    The workflow payload is overwritten by the merged record, the conflicts are
+    stored in ``extra_data.conflicts``. Also, it adds a ``callback_url`` which
+    contains the endpoint which resolves the merge conflicts.
+
+    Note:
+        For the time being the ``root`` will be ignored, and we'll rely only
+        on the ``head``, hence it is a rootless implementation. Also when
+        the feature flag ``FEATURE_FLAG_ENABLE_MERGER`` is ``False`` it
+        will skip the merge.
+
+    """
+    if not current_app.config.get('FEATURE_FLAG_ENABLE_MERGER'):
+        return None
+
+    matched_control_number = obj.extra_data['matches']['approved']
+
+    head_uuid = PersistentIdentifier.get(
+        'lit', matched_control_number).object_uuid
+
+    obj.extra_data['head_uuid'] = str(head_uuid)
+
+    head = InspireRecord.get_record(head_uuid)
+    root = {}
+    update = obj.data
+
+    merged, conflicts = merge(
+        head=head,
+        root=root,
+        update=update
+    )
+
+    obj.data = merged
+
+    if conflicts:
+        obj.extra_data['conflicts'] = conflicts
+        obj.extra_data['callback_url'] = \
+            get_resolve_merge_conflicts_callback_url()
+    obj.save()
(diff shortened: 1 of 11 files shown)


inspirehep/modules/workflows/workflows/article.py
```diff
@@ -236,13 +241,18 @@
...
+        IF_ELSE(
+            # feature flag to send updates to legacy or not
+            is_upload_to_legacy_on_update_enabled,
+            [
+                send_robotupload(mode='replace'),
+            ],
+            [
+                mark('skipped-robot-upload', True),
+            ]
+        ),
```
ammirate: I would suggest to use the same style of the already existing feature flags, like
```python
        return False
````
which means inside the function itself and not in the workflow, WDYT?

File: inspirehep/modules/workflows/workflows/article.py [hunk 2]
```diff
@@ -279,9 +288,29 @@
     ),
     IF_ELSE(
         is_marked('is-update'),
-        [
-            mark('approved', True)
-        ],
+        IF_ELSE(
+            is_merge_on_update_enabled,
```
ammirate: same as before for feature flags style, I would put this check in `merge_articles`

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -296,9 +325,9 @@
 
 STORE_RECORD = [
     IF_ELSE(
-        is_marked('is-update'),
-        mark('skipped-store-record', True),
+        is_merge_on_update_enabled,
```
ammirate: same here for feature flags

tests/integration/workflows/test_arxiv_merge.py
```diff
@@ -0,0 +1,339 @@
...
+    record = InspireRecord.get_record(record_to_merge.id)
+
+    expected_titles = [
+        {'title': 'Alias Investigations'},
+        {'title': 'Jessica Jones'},
+        {'title': 'Luke Cage'},
+        {'title': 'Frank Castle'},
+    ]
+
+    assert expected_titles == record.get('titles')
```
ammirate: I would remove this assertion: if, for any reason, the `inspire-json-merger` library changes its internal merging rules, this test is going to fail, while it should still pass

inspirehep/modules/workflows/workflows/article.py
drjova: hmm IMO for ``SEND_TO_LEGACY`` and ``STORE_RECORD`` it's too much hidden logic, I'd rather state what they are doing in the workflow itself. WDYT?

drjova:
updated!

drjova:
@ammirate @david-caro updated :)

inspirehep/modules/workflows/tasks/merging.py
```diff
@@ -48,3 +59,60 @@ def get_head_source(head_uuid):
...
+    stored in ``extra_data.conflicts``. Also, it adds a ``callback_url`` which
+    contains the endpoint which resolves the merge conflicts.
+
+    Note:
+        For the time being the ``root`` will be ignored, and we'll rely only
+        on the ``head``, hence it is a rootless implementation.
+
+    Returns:
+        bool: ``True`` if the workflow merged with another record in the system.
+        ``False`` if the merger feature is not enabled.
```
ammirate: Old docstring :P

inspirehep/modules/workflows/tasks/merging.py
```diff
@@ -48,3 +59,60 @@ def get_head_source(head_uuid):
...
+
+    """
+    if not current_app.config.get('FEATURE_FLAG_ENABLE_MERGER'):
+        return None
+
+    if obj.extra_data.get('fuzzy-matched') is True:
+        matched_control_number = get_value(
+            obj.extra_data, 'fuzzy_match_approved_id')
+    else:
+        matched_control_number = get_value(obj.extra_data, 'matches.exact[0]')
```
ammirate: `matches.approved` is the place where you will find the matched record, whether it was a fuzzy or an exact match, thus you don't need the `if-else`statement, just assume it is there.
I would even access it using `obj.extra_data.['matches']['approved']`, in case it blows up if no match is set

inspirehep/modules/workflows/tasks/merging.py
```diff
@@ -48,3 +59,60 @@ def get_head_source(head_uuid):
...
+    update = obj.data
+
+    merged, conflicts = merge(
+        head=head,
+        root=root,
+        update=update
+    )
+
+    obj.data = merged
+    obj.extra_data['merged'] = True
```
ammirate: this `mark` should happen in the workflow :P

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -280,7 +273,15 @@
...
-            mark('approved', True)
+            merge_articles,
+            IF(
+                has_conflicts,
+                halt_record(
+                    action='merge_approval',
+                    message='Submission halted for merging conflicts.'
+                ),
+            ),
+            mark('approved', True),
```
ammirate: `mark('merged', True)` :muscle:

drjova: this will add the mark even when the ``merge_articles`` is disabled by the feature flag :/

drjova:
@ammirate all done and rebased :)

drjova:
Supersede by https://github.com/inspirehep/inspire-next/pull/3342

PR #3351, opened 2018-04-17
workflows: integrate reference matcher in workflows

salmanmaq:
mappings: integrate reference matcher in Labs workflows

## Description
This integrates the reference matcher in the Labs workflow so that the reference matching is performed in Labs and not in Legacy.

## Related Issue
https://its.cern.ch/jira/browse/INSPIR-447

## Motivation and Context
The new reference matcher aims to improve citation counts over the Legacy reference matcher. It incorporates data citations as well.

## Checklist:


inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -184,3 +189,60 @@ def extract_references_from_raw_ref(reference, custom_kbs_file=None):
...
+        reference: The reference metadata
+        config: The configurtaion(s) for InspireMatcher queries
+
+    Returns:
+        The record ID of the matched reference
+    """
+
+    config_default = current_app.config['WORKFLOWS_REFERENCE_MATCHER_DEFAULT_CONFIG']
+    config_jcap_and_jhep = current_app.config['WORKFLOWS_REFERENCE_MATCHER_JHEP_AND_JCAP_CONFIG']
+    config_data = current_app.config['WORKFLOWS_REFERENCE_MATCHER_DATA_CONFIG']
```
jacquerie: Let's do one thing at a time: if we were to match on data with the current code then we will produce the wrong `$ref`. So we can't have this code in here yet, unless we solve that other issue first.

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -184,3 +189,60 @@ def extract_references_from_raw_ref(reference, custom_kbs_file=None):
...
+
+    journal_title = get_value(reference, 'reference.publication_info.journal_title')
+    if journal_title in ['JCAP', 'JHEP']:
+        try:
+            if get_value(reference, 'reference.publication_info.year'):
+                reference['reference']['publication_info']['year'] = str(
+                    reference['reference']['publication_info']['year'])
```
jacquerie: Can you check if we really need this? These fields agree on the type, see https://github.com/inspirehep/inspire-next/blob/ffcd5904004a13090ea289d1380aaa2d03a056bd/inspirehep/modules/records/mappings/records/hep.json#L713-L716 and https://github.com/inspirehep/inspire-next/blob/ffcd5904004a13090ea289d1380aaa2d03a056bd/inspirehep/modules/records/mappings/records/hep.json#L893-L896.

inspirehep/modules/workflows/config.py
```diff
@@ -0,0 +1,179 @@
...
+                {
+                    'paths': [
+                        'reference.publication_info.journal_title',
+                        'reference.publication_info.journal_volume',
+                        'reference.publication_info.artid',
+                    ],
+                    'search_paths': [
+                        'publication_info.journal_title.raw',
+                        'publication_info.journal_volume',
+                        'publication_info.artid',
```
jacquerie: Can you check if we still need these searches to be distinct, given the following commit by @iulianav? https://github.com/inspirehep/inspire-next/commit/09b1255ad0146a0d124a1a93dd1b1fba8ed6d16c#diff-472ccc26aabcd94f4c20bce673a6e90d

inspirehep/modules/workflows/config.py
```diff
@@ -0,0 +1,179 @@
...
+                        'reference.publication_info.journal_title',
+                        'reference.publication_info.journal_volume',
+                        'reference.publication_info.year',
+                        'reference.publication_info.artid',
+                    ],
+                    'search_paths': [
+                        'publication_info.journal_title.raw',
+                        'publication_info.journal_volume',
+                        'publication_info.year',
+                        'publication_info.artid',
```
jacquerie: Same consideration as https://github.com/inspirehep/inspire-next/pull/3351/files#r182354877.

tests/integration/workflows/test_workflow_actions.py
```diff
@@ -329,3 +400,53 @@ def test_normalize_journal_titles_unknown_journals_no_ref(workflow_app, insert_j
...
+    }
+
+    with workflow_app.app_context():
+        with mock.patch.dict(workflow_app.config, extra_config):
+            citing_doc_workflow_uuid = start('article', [citing_record])
+
+        citing_doc_eng = WorkflowEngine.from_uuid(citing_doc_workflow_uuid)
+        citing_doc_obj = citing_doc_eng.processed_objects[0]
+
+        assert citing_doc_obj.data['references'][7]['record'] == get_record_ref(insert_cited_record['control_number'], 'literature')
```
jacquerie: I think that here we should follow the suggestion in https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html:
> Tests can avoid complexity by stating their inputs and outputs directly rather than computing them. Otherwise it's easy for tests to develop their own bugs.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -421,17 +422,19 @@ def refextract(obj, eng):
...
+        text_references = match_references(text_references)
 
     if len(pdf_references) == len(text_references) == 0:
         obj.log.info('No references extracted.')
-    elif len(pdf_references) >= len(text_references):
+    elif len(pdf_references) > len(text_references):
         obj.log.info('Extracted %d references from PDF.', len(pdf_references))
         obj.data['references'] = pdf_references
-    elif len(text_references) > len(pdf_references):
+    elif len(text_references) >= len(pdf_references):
```
jacquerie: Can you remind me why this changed behavior (preferring text references to PDF references when they are in the same amount) ?

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -184,3 +189,60 @@ def extract_references_from_raw_ref(reference, custom_kbs_file=None):
...
+        result = next(match(reference, config_data))
+        return result['_source']['control_number']
+    except StopIteration:
+        pass
+
+    try:
+        result = next(match(reference, config_default))
+        return result['_source']['control_number']
+    except StopIteration:
+        pass
```
jacquerie: Repeating the same code pattern 3 times (especially this close to each other) should make you think that a refactoring is possible: https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming).

Currently this function is an example of this other "Rule-of-three": https://en.wikipedia.org/wiki/Planescape#Rule-of-Three : )

inspirehep/modules/workflows/tasks/refextract.py
jacquerie: To be clear: this applies to the current PR, but by the end of the sprint we **are** going to handle data references!

inspirehep/modules/workflows/config.py
salmanmaq: That should be the case. However, I do not see any keys named page_artid in Labs for the few records I have inspected. Am I missing something?

inspirehep/modules/workflows/tasks/actions.py
salmanmaq: The assumption here is that the text references provided by the submitter potentially have a better quality than references extracted from PDF. So, if both of them manage to return the same amount of references, we prefer the text ones.

inspirehep/modules/workflows/tasks/refextract.py
salmanmaq: Do we handle the data issue as well in this PR, or should we try to merge this before doing that?

inspirehep/modules/workflows/tasks/refextract.py
salmanmaq: Will change.

inspirehep/modules/workflows/tasks/refextract.py
jacquerie: I'd say that we limit the scope of the current PR to Literature references, but in the very next PR we're going to handle Data references as well.

inspirehep/modules/workflows/config.py
jacquerie: This has to do with the thing I say often: "ES is not a JSON store". The `copy_to` directive **does not** enhance the JSON that you are sending as the source document, but instructs ES to index the field in another inverted index; so when you retrieve records from ES you get the source JSON back, which hasn't changed (in particular it doesn't contain `page_artid`), but searches against that field in the inverted index will work.

inspirehep/modules/workflows/tasks/actions.py
jacquerie: Ok, makes sense. Nit: let's put this behavior change in a distinct commit, so if @michamos says that we did a mistake we can easily revert it (did we ever get his opinion on this?).

inspirehep/modules/workflows/config.py
salmanmaq: Ok, nice. I'll verify and make the necessary changes.

inspirehep/modules/workflows/tasks/actions.py
salmanmaq: We didn't get his opinion on this. But that makes sense.

inspirehep/modules/workflows/config.py
salmanmaq: I checked this and the records with publication_info no longer match.

inspirehep/modules/workflows/tasks/refextract.py
salmanmaq: Verified to work without this.

inspirehep/modules/workflows/tasks/refextract.py
salmanmaq: Another comment on this, while we can do what you suggested, I think this would lead to a bit too much abstraction. What do you think about that?

jacquerie: Mh, I see what you mean, but I would still do it. Let's see what @turtle321 thinks about this?

turtle321: I woul probably try to avoid excetions for control flows if possible, thus I would have written:
```
result = next(match(reference, config_jcap_and_jhep), None)
if result:
    return result['_source']['control_number']
```

inspirehep/modules/workflows/config.py
michamos: didn't we want to search for both `artid` and `page_start` in `page_artid` instead of doing `artid -> artid`, `page_start -> page_start` only?

inspirehep/modules/workflows/config.py
michamos: Same also for my comment.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -400,19 +401,19 @@ def refextract(obj, eng):
...
         eng: a workflow engine.
 
     Returns:
         None
+
     """
     if 'references' in obj.data:
-        obj.log.info('Found references in metadata, extracting unextracted raw_refs')
-        obj.data['references'] = extract_references_from_raw_refs(obj.data['references'])
+        raw_references = extract_references_from_raw_refs(obj.data['references'])
```
michamos: why is this called `raw_references`? I would expect the result to be a list of parsed, structured references after processing by refextract. Maybe `extracted_raw_references`?

inspirehep/modules/workflows/tasks/actions.py
michamos: Seems OK to me.

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -184,3 +189,46 @@ def extract_references_from_raw_ref(reference, custom_kbs_file=None):
...
+        references: Schema compliant list of references
+
+    Returns:
+        associated_references: The same list with the associated record ids (if any)
+    """
+
+    for i, reference in enumerate(references):
+        matched_recid = match_reference(reference)
+        if matched_recid:
+            references[i]['record'] = get_record_ref(matched_recid, 'literature')
```
michamos: I would drop this function, and have a `match_reference` that does all the work for one reference, that is take a reference and spit out a reference, potentially with the `record` key set to the matched record.
You would then replace calls of `result = match_references(references)` by `result = [match_reference(ref) for ref in references]`.

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -184,3 +189,46 @@ def extract_references_from_raw_ref(reference, custom_kbs_file=None):
...
+
+    Returns:
+        The record ID of the matched reference
+    """
+
+    config_default = current_app.config['WORKFLOWS_REFERENCE_MATCHER_DEFAULT_CONFIG']
+    config_jcap_and_jhep = current_app.config['WORKFLOWS_REFERENCE_MATCHER_JHEP_AND_JCAP_CONFIG']
+
+    journal_title = get_value(reference, 'reference.publication_info.journal_title')
+    if journal_title in ['JCAP', 'JHEP']:
```
michamos: for this to work properly, the `normalize_journal_titles` has to run before `refextract` in https://github.com/inspirehep/inspire-next/blob/7aad1074b4285b4e5dbf246c7d3cfd0f1ad166af/inspirehep/modules/workflows/workflows/article.py#L157-L158.

michamos: it would be cleaner to have
```python
config = config_jcap_and_jhep if journal_title in ['JCAP', 'JHEP'] else config_default
result = next(match(reference, config_jcap_and_jhep), None)
if result:
    return result['_source']['control_number']
return None
```

tests/integration/workflows/helpers/calls.py
```diff
@@ -167,6 +167,28 @@ def generate_record():
...
+            'oai_arxiv_citing_record.json'
+        )
+    ))
+
+    categories = {'core': [], 'non-core': []}
+    for eprint in json_data.get('arxiv_eprints', []):
+        categories['core'].extend(eprint.get('categories', []))
+
+    if 'preprint_date' in json_data:
+        json_data['preprint_date'] = datetime.date.today().isoformat()
```
michamos: MINOR it would have been cleaner to use https://github.com/spulec/freezegun here instead, but it's OK for now.

tests/integration/workflows/helpers/calls.py
```diff
@@ -167,6 +167,28 @@ def generate_record():
...
+
+    assert categories
```
michamos: what is this for? you are defining `categories` to be non-empty. If you want to assert that the category is in the record, you should assert that.

tests/integration/workflows/test_workflow_actions.py
```diff
@@ -329,3 +399,53 @@ def test_normalize_journal_titles_unknown_journals_no_ref(workflow_app, insert_j
...
+
+    citing_record, categories = insert_citing_record()
+
+    extra_config = {
+        "BEARD_API_URL": "http://example.com/beard",
+        "MAGPIE_API_URL": "http://example.com/magpie",
+        'ARXIV_CATEGORIES': categories,
+    }
+
+    with workflow_app.app_context():
```
michamos: aren't you already in an application context by using the `workflow_app` fixture?

tests/unit/workflows/test_workflows_tasks_actions.py
```diff
@@ -819,77 +818,6 @@ def test_populate_journal_coverage_does_nothing_if_no_journal_is_found(mock_repl
     assert 'journal_coverage' not in obj.extra_data
 
 
-@patch('inspirehep.modules.workflows.tasks.actions.get_document_in_workflow')
```
michamos: why did you remove all these tests?

inspirehep/modules/workflows/config.py
michamos: it's not about reducing the number of queries, but having more matches, when we wrongly identify what is a `page_start` and what is an `artid`.

inspirehep/modules/workflows/tasks/refextract.py
salmanmaq: Nice catch.

tests/integration/workflows/test_workflow_actions.py
salmanmaq: I am not sure about this. Will check.

tests/unit/workflows/test_workflows_tasks_actions.py
salmanmaq: They were unit tests, and they are bound to fail due to inclusion of the inspire matcher since there's no ElasticSearch instance in Unit tests.

tests/integration/workflows/helpers/calls.py
salmanmaq: I think this is a very hacky way to get the task done in the first place (to make it auto-accepted). I am  not exactly sure why this was done in the first place in https://github.com/inspirehep/inspire-next/blob/ac4fdbbf50b03b55dd5f4983600ec65cca4879d8/tests/integration/workflows/helpers/calls.py#L170
where I took the inspiration from.

tests/integration/workflows/test_workflow_actions.py
salmanmaq: It does work without this line, but I am unfamiliar with Flask so I am not completely sure what's going on.

tests/integration/workflows/helpers/calls.py
salmanmaq: Just copied from: https://github.com/inspirehep/inspire-next/blob/ac4fdbbf50b03b55dd5f4983600ec65cca4879d8/tests/integration/workflows/helpers/calls.py#L170
Again, not sure why we need to do this.

tests/unit/workflows/test_workflows_tasks_actions.py
michamos: then they should become integration tests instead of unit, or mock the reference matching if they don't depend on it.

tests/unit/workflows/test_workflows_tasks_actions.py
```diff
@@ -1124,3 +1125,63 @@ def test_validate_record_raises_when_record_is_invalid():
 
     with pytest.raises(ValidationError):
         _validate_record(obj, eng)
+
+
+@patch('inspirehep.modules.workflows.tasks.actions.get_document_in_workflow')
+@patch(
+    'inspirehep.modules.workflows.tasks.refextract.match',
+    return_value=(_ for _ in [])
```
michamos: why not `iter([])`?

turtle321: My fault, I suggested it. But this is not really important, right? Let's try to merge this today.

michamos: I was scratching my head for a couple of minutes trying to understand this, so I would be in favor of changing it. I agree on trying to get this merged today.

turtle321: Ok

tests/integration/workflows/helpers/calls.py
```diff
@@ -167,6 +167,25 @@ def generate_record():
...
+            '../fixtures',
+            'oai_arxiv_citing_record.json'
+        )
+    ))
+
+    categories = {'core': [], 'non-core': []}
+    for eprint in json_data.get('arxiv_eprints', []):
+        categories['core'].extend(eprint.get('categories', []))
+
+    assert 'core' in categories
```
michamos: You can remove this.

salmanmaq:
ci merge please

michamos:
ci merge please

PR #3355, opened 2018-04-17
tests: rewrite the alembic tests

jacquerie:
## Description:
The previous implementation of the Alembic tests was lacking in several
regards, most notably test isolation. The approach in this PR side steps
the problem by creating two tests for `upgrade` and `downgrade`, which,
as https://github.com/inspirehep/inspire-next/commit/d578ca6b878fb66caf9e775408b4dbbd56eb3f72 shows, make writing these tests much easier.

Since commit 99aeb65 Alembic's INSPIRE branch has had two heads. This
commit fixes this situation by adding an empty merge migration.

Finally, https://github.com/inspirehep/inspire-next/pull/3267 was missing that SQLAlchemy's ``rename_table`` does not work
out of the box for PostgreSQL tables, and requires the fixes outlined in
http://petegraham.co.uk/rename-postgres-table-with-alembic/.

## Checklist:

### DIFF
--- a/inspirehep/alembic/402af3fbf68b_rename_inspire_prod_records_table.py
+++ b/inspirehep/alembic/402af3fbf68b_rename_inspire_prod_records_table.py
@@ -1,0 +1,59 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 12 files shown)


inspirehep/alembic/402af3fbf68b_rename_inspire_prod_records_table.py
```diff
@@ -0,0 +1,59 @@
...
+
+revision = '402af3fbf68b'
+down_revision = 'd99c70308006'
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    op.rename_table('inspire_prod_records', 'legacy_records_mirror')
+    op.execute('ALTER SEQUENCE inspire_prod_records_recid_seq RENAME TO legacy_records_mirror_recid_seq')
```
david-caro: The sequence should be tested too no? Is that feasible?

jacquerie: We can, but for some reason it's trickier than it looks. A query like
```SQL
SELECT sequence_name
FROM information_schema.sequences
WHERE sequence_schema='public'
```
(inspired from https://stackoverflow.com/a/26778407/374865) does not work because the test times out, while this query
```SQL
SELECT relname
FROM pg_class
WHERE relkind='S'
```
(taken from https://stackoverflow.com/a/1611680/374865) does the right thing.

turtle321:
Lesson learned: renaming a database table in this project is very expensive.

PR #3360, opened 2018-04-19
539 script: snippet to restart workflows

turtle321:


scripts/restart_workflows.py
```diff
@@ -0,0 +1,22 @@
...
+
+from invenio_workflows import workflow_object_class
+from invenio_db import db
+
+
+def restart_workflow(workflow_id, position=0):
+    wf = workflow_object_class.get(workflow_id)
+    print 'Workflow {} is currently in position {}'.format(workflow_id,
+                                                           wf.callback_pos)
+    wf.callback_pos = position
```
ammirate: `wf.callback_pos` expects a list of integers. If you want this snippet to be run only for restarting workflows from the beginning then I'd say to remove the `position` parameter. 
Otherwise, we can set `position` to `[0]`.

PR #3364, opened 2018-04-20
Implement rootful merger

iulianav:
## Checklist:


inspirehep/alembic/f5bf1193b078_recreate_workflows_record_sources_table.py
```diff
@@ -0,0 +1,96 @@
+#
```
turtle321: Was this migration AUTOgenerated?
http://alembic.zzzcomputing.com/en/latest/autogenerate.html
Please do NOT deploy this now, but align with myself or Micha before, as we are working on fixing the migration history and the schema differences between prod db and qa db.

inspirehep/modules/workflows/models.py
```diff
@@ -88,20 +96,31 @@ class WorkflowsRecordSources(db.Model):
...
-        nullable=False,
-    )
     record_id = db.Column(
         UUIDType,
         db.ForeignKey('records_metadata.id', ondelete='CASCADE'),
         nullable=False,
     )
+
+    source = db.Column(
+        db.Enum(SourceEnum),
```
turtle321: Autogenerated migrations have problems with enum types. Not sure if this is the case here though...

inspirehep/modules/workflows/models.py
```diff
@@ -88,20 +96,31 @@ class WorkflowsRecordSources(db.Model):
...
+    )
+
     json = db.Column(
-        JSONType().with_variant(
-            postgresql.JSON(none_as_null=True),
-            'postgresql',
-        ),
+        postgresql.JSONB(),
         default=lambda: dict(),
+        nullable=True
```
turtle321: Why do you need a nullable if you have a default?

inspirehep/modules/workflows/models.py
```diff
@@ -26,12 +26,20 @@
...
+import enum
+
 from sqlalchemy.dialects import postgresql
 from sqlalchemy_utils.types import JSONType, UUIDType
 
 from invenio_db import db
 
 
+class SourceEnum(enum.IntEnum):
+    arxiv = 1
```
turtle321: Using strings would be clearer, like `arxiv = 'arxiv'`. Not sure if this is possible though.

inspirehep/modules/workflows/models.py
turtle321: We could use a specific Postgres ENUM data type. Please read it up http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html#sqlalchemy.dialects.postgresql.ENUM and evaluate if it is worth.

inspirehep/modules/workflows/models.py
iulianav: This did not work because there was no total order defined on just string values and the db needed it for some backstage operations.

drjova:
Supersede by https://github.com/inspirehep/inspire-next/pull/3376

PR #3365, opened 2018-04-20
507 orcid: fetch record from db (rather than api)

turtle321:


inspirehep/modules/orcid/api.py
```diff
@@ -128,17 +129,18 @@ def calculate_hash_for_record(record):
     return hash_xml_element(orcid_rec.get_xml())
 
 
-def _get_record_by_mime(config, recid, mime_type):
+def _get_bibtex_record(config, recid):
```
michamos: I thought you also wanted to get this without using the REST API?

turtle321: I would prefer doing it in a next PR: https://its.cern.ch/jira/browse/INSPIR-548

PR #3366, opened 2018-04-20
Orcid small locks

michamos:
* Removes the global `orcid-push` lock, which was useless and a bottleneck.
* Adds a per-orcid lock when doing the actual push to ensure we don't push concurrently to the same orcid.


inspirehep/modules/orcid/api.py
```diff
@@ -101,7 +105,8 @@ def push_record_with_orcid(recid, orcid, oauth_token, put_code=None, old_hash=No
         LOGGER.info(
             "No put-code found, pushing new record #%s to ORCID %s.", recid, orcid,
         )
-        with log_time_context('Pushing new record', LOGGER):
+        with log_time_context('Pushing new record', LOGGER), \
+                redis_locking_context(lock_name, blocking=True):
```
turtle321: Use proper `expire` and `auto_renewal` args.

turtle321:
Check this: https://github.com/inspirehep/inspire-next/pull/3368

tests/integration/orcid/test_orcid_api.py
```diff
@@ -78,6 +84,56 @@ def test_push_record_with_orcid_update(mock_config, vcr_cassette):
...
+            with app.app_context():
+                push_record_with_orcid(
+                    recid='4328',
+                    orcid='0000-0002-1825-0097',
+                    oauth_token='fake-token',
+                    put_code='920107',
+                    old_hash=None,
+                )
+
+    mock_get_api.return_value.updae_record.side_effect = lambda *args, **kwargs: sleep(3)
```
turtle321: typo in `update`

tests/integration/orcid/test_orcid_api.py
```diff
@@ -78,6 +84,56 @@ def test_push_record_with_orcid_update(mock_config, vcr_cassette):
...
+    lock = Lock(redis, 'orcid:0000-0002-1825-0097')
+
+    PushThread().start()
+    sleep(1)
+
+    assert lock.acquire(blocking=False) is False
+    lock.reset()
+
+
+@mock.patch('inspirehep.modules.orcid.api._get_api')
```
turtle321: move it in the thread

tests/integration/orcid/test_orcid_api.py
```diff
@@ -78,6 +84,56 @@ def test_push_record_with_orcid_update(mock_config, vcr_cassette):
...
+                    put_code='920107',
+                    old_hash=None,
+                )
+
+    mock_get_api.return_value.updae_record.side_effect = lambda *args, **kwargs: sleep(3)
+    redis_url = app.config.get('CACHE_REDIS_URL')
+    redis = StrictRedis.from_url(redis_url)
+    lock = Lock(redis, 'orcid:0000-0002-1825-0097')
+
+    PushThread().start()
```
turtle321: use daemon threads

inspirehep/modules/orcid/api.py
turtle321: It would be nice to log this event.

tests/integration/utils/test_lock.py
```diff
@@ -0,0 +1,95 @@
...
+    assert not lock_acquired
+
+
+def test_distributed_lock_blocks(app):
+    def work(lock_acquired):
+        with app.app_context(), distributed_lock('my_lock'):
+            lock_acquired.set()
+            sleep(2)
+
+    lock_acquired = Event()
```
turtle321: Naming: `lock_acquired_event` is clearer

tests/integration/utils/test_lock.py
```diff
@@ -0,0 +1,95 @@
...
+
+    locking_thread = Thread(target=work, args=(lock_acquired,))
+    locking_thread.start()
+
+    lock_acquired.wait()
+    with distributed_lock('my_lock', blocking=True):
+        sleep(1)
```
turtle321: This is the brittle part. What about something like:
```python
for _ in range(10):
    if locking_thread.is_alive():
        sleep(1)
        continue
    assert not locking_thread.is_alive()
```

tests/integration/utils/test_lock.py
```diff
@@ -0,0 +1,95 @@
...
+def test_distributed_lock_blocks(app):
+    def work(lock_acquired):
+        with app.app_context(), distributed_lock('my_lock'):
+            lock_acquired.set()
+            sleep(2)
+
+    lock_acquired = Event()
+
+    locking_thread = Thread(target=work, args=(lock_acquired,))
+    locking_thread.start()
```
turtle321: Probably safer if this thread is a daemon, so it dies if the test fails.

tests/integration/utils/test_lock.py
```diff
@@ -0,0 +1,95 @@
...
+
+    lock_acquired.wait()
+    with distributed_lock('my_lock', blocking=True):
+        sleep(1)
+        assert not locking_thread.is_alive()
+
+
+def test_distributed_lock_auto_renews_lock(app):
+    def work(lock_acquired):
+        with app.app_context(), distributed_lock('my_lock', expire=2, auto_renewal=True):
```
turtle321: I think it is more reliable if you add a `sleep(3)` here rather than in line 92

inspirehep/modules/orcid/tasks.py
```diff
@@ -269,21 +268,24 @@ def store_record_in_redis(orcid, rec_id, put_code, hashed=None):
         put_code(string): the put_code used to push the record to ORCID.
         hashed(Optional[string]): hashed ORCID record
     """
-    with redis_locking_context('orcid_push') as r:
-        orcid_recid_key = get_orcid_recid_key(orcid, rec_id)
-        to_cache = {'putcode': put_code}
-        if hashed:
-            to_cache['hash'] = hashed
-        r.hmset(orcid_recid_key, to_cache)
+    redis_url = current_app.config.get('CACHE_REDIS_URL')
```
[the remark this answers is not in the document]
michamos: yes, we should have something like `invenio-db.db` for `redis`, but I would do it later.

tests/integration/utils/test_lock.py
```diff
@@ -0,0 +1,97 @@
...
+            sleep(1) if locking_thread.is_alive() else break
+        assert not locking_thread.is_alive()
+
+
+def test_distributed_lock_auto_renews_lock(app):
+    def work(lock_maybe_expired_event):
+        with app.app_context(), distributed_lock('my_lock', expire=2, auto_renewal=True):
+            sleep(3)
+            lock_maybe_expired_event.set()
+            sleep(30)
```
turtle321: Do you still need this?

PR #3376, opened 2018-04-24
merger: integrate rootful merger in the article workflow

iulianav:
## Description

Implement rootful merger and integrate it in the article workflow.

## Related Issue

https://its.cern.ch/jira/browse/INSPIR-306

## Motivation and Context

Up until now, the merger was rootless. Using the root will reduce the number of conflicts.

## Checklist:


inspirehep/modules/workflows/tasks/manual_merging.py
```diff
@@ -110,16 +110,29 @@ def save_roots(obj, eng):
     """
     def _merge_roots(new_uuid, head_roots, update_roots):
         """Return the new roots to link to head."""
-        head_sources = [h.source for h in head_roots]
+        head_sources = {h.source: h for h in head_roots}
         for u in update_roots:
-            if u.source not in head_sources:
+            if u.source not in head_sources.keys():
```
drjova:
```python
if u.source not in head_sources
```

is enough :)

drjova: ``u`` is a weird naming, probably something like ``updated_root`` or something similar

inspirehep/modules/workflows/tasks/manual_merging.py
```diff
@@ -110,16 +110,29 @@ def save_roots(obj, eng):
...
+                    head_roots.remove(head_sources[u.source])
+                    db.session.delete(head_sources[u.source])
+                    head_roots.append(
+                        WorkflowsRecordSources(
+                            source=u.source,
+                            record_id=new_uuid,
+                            json=u.json
+                        )
+                    )
+            db.session.delete(u)
```
drjova: why we have to delete the updated here?

iulianav: @drjova because we don't want to keep the roots of the record that got merged.

inspirehep/modules/workflows/tasks/upload.py
```diff
@@ -82,6 +69,22 @@ def _get_updated_record(obj):
     db.session.commit()
 
 
+@with_debug_logging
+def store_root(obj, eng):
+    """Insert or update the current record head's root into the ``WorkflowsRecordSources`` table."""
```
drjova: we'll need the feature flag check here

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -89,7 +90,11 @@
     has_conflicts,
     merge_articles,
 )
-from inspirehep.modules.workflows.tasks.upload import store_record, set_schema
+from inspirehep.modules.workflows.tasks.upload import (
```
drjova: sort by name :)

drjova:
@iulianav we have to add tests for the workflow itself so we make sure it works as expected. We have already multiple test cases that we can use as test stories :)

An example can be found here https://github.com/inspirehep/inspire-next/blob/master/tests/integration/workflows/test_arxiv_merge.py

inspirehep/modules/workflows/tasks/manual_merging.py
```diff
@@ -110,16 +110,28 @@ def save_roots(obj, eng):
...
+                db.session.delete(head_sources[update_root.source])
+                head_roots.append(
+                    WorkflowsRecordSources(
+                        source=update_root.source,
+                        record_id=new_uuid,
+                        json=update_root.json
+                    )
+                )
+            db.session.delete(update_root)
+            db.session.commit()
```
drjova: do we need the commit here?

inspirehep/modules/workflows/models.py
```diff
@@ -26,12 +26,20 @@
...
 from invenio_db import db
 
 
+class SourceEnum(enum.IntEnum):
+    arxiv = 1
+    submitter = 2
+    publisher = 3
+
+
 class WorkflowsAudit(db.Model):
```
michamos: don't you need an alembic migration for this?

tests/integration/workflows/test_workflows_tasks_upload.py
```diff
@@ -65,3 +71,72 @@ def test_store_record_does_not_raise_in_the_orcid_receiver(mock_attempt_push, ap
         })
 
         store_record(obj, None)  # Does not raise.
+
+
```
drjova: we should add a test case when a  with a new ``publisher`` root and an existing ``arxiv``

tests/integration/workflows/test_workflows_tasks_upload.py
```diff
@@ -65,3 +71,72 @@ def test_store_record_does_not_raise_in_the_orcid_receiver(mock_attempt_push, ap
...
+        obj = workflow_object_class.create(record)
+
+        extra_data = {
+            'head_uuid': str(head_uuid),
+            'merger_root': update_root,
+        }
+
+        obj.extra_data = extra_data
+
+        store_root(obj, None)
```
drjova: also we can check that the ``merger_root`` key exists

tests/unit/workflows/test_workflows_tasks_actions.py
```diff
@@ -322,6 +323,19 @@ def test_halt_record_accepts_custom_msg():
...
+def test_preserve_root():
+    config = {
+        'FEATURE_FLAG_ENABLE_MERGER': True
+    }
+
+    with patch.dict(current_app.config, config):
+        obj = MockObj({'foo': 'bar'}, {})
+        eng = MockEng()
+
+        assert preserve_root(obj, eng) is None
```
drjova: hmm why we need this assertion? the following one should be enough :)

inspirehep/modules/workflows/tasks/manual_merging.py
iulianav: Yes! Otherwise, when doing the bulk save we get a duplicate key value pk violation. Apparently `db.session.delete` is not enough for the staging area to track the deletions and insertions. You have to first delete the existing record that you want to re-add with the same pk. However, I believe I will try to do a `db.session.merge()` instead for each root.

michamos: I think the issue is about isolation: is it OK in principle for another process to see a partial merge? if not, it should be done in one transaction.

iulianav:
ci test please

inspirehep/modules/workflows/tasks/merging.py
```diff
@@ -50,7 +53,7 @@ def get_head_source(head_uuid):
         * None if there are no root records
     """
     roots_sources = set(
-        r.source for r in
+        r.source.name for r in
         WorkflowsRecordSources.query.filter_by(record_id=head_uuid).all()
     )
```
iulianav: @michamos shall we fix the logic here or leave it as it is?

tests/unit/workflows/test_workflows_tasks_actions.py
iulianav: I followed the style of the test above. I can remove it tho :)

tests/integration/workflows/test_workflows_tasks_upload.py
iulianav: I feel like the logic in this test is sort of duplicated in the test_arxiv_merge.py . Should I remove it there?

tests/integration/workflows/test_workflows_tasks_upload.py
iulianav: True!

inspirehep/modules/workflows/models.py
```diff
@@ -88,20 +96,31 @@ class WorkflowsRecordSources(db.Model):
...
+    )
+
     json = db.Column(
-        JSONType().with_variant(
-            postgresql.JSON(none_as_null=True),
-            'postgresql',
-        ),
+        postgresql.JSONB(),
         default=lambda: dict(),
+        nullable=True
```
drjova: do we need ``nullable`` here?

iulianav:
ci test please

inspirehep/modules/workflows/tasks/merging.py
michamos: what do you mean?

michamos: right, this is used by the manual merger only to decide which strategy to use in that case. I would say the docstring is right and the behaviour is wrong, as we are always saying that publisher is stronger than arXiv because we don't want to overwrite the work of the publisher by the arXiv version. However, this doesn't account for weird cases where you want to merge stuff from several different publishers and has never been really tested. We need to revisit this down the line when we work on manual merges.

iulianav: Then we will leave it like this for now :)

michamos: That's not what I said. You need to change it to be the same as the docstring, but I am saying that it's not 100% sure that this is the right behaviour anyway.

drjova: As we discussed IRL we can drop this function and remove it from the ``manual merger`` ([here](https://github.com/inspirehep/inspire-next/blob/8d9e7dc234ee64653113b9c001c02df1985668db/inspirehep/modules/workflows/workflows/manual_merge.py#L83) and [here](https://github.com/inspirehep/inspire-next/blob/4a123080ac1d2cdd6f445a3104f8ef06b99edb44/inspirehep/modules/workflows/tasks/manual_merging.py#L59))

inspirehep/modules/workflows/workflows/manual_merge.py
```diff
@@ -80,7 +79,7 @@ def start_merger(head_id, update_id, current_user_id=None):
     workflow_object.extra_data.update(data)
 
     # preparing identifiers in order to do less requests possible later
-    head_source = get_head_source(head.id) or merger_get_source(head)
+    head_source = get_head_source(head)
```
drjova: I'd suggest to completely remove the ``get_head_source`` from manual merge

drjova:
ci test please

inspirehep/modules/workflows/workflows/manual_merge.py
drjova: also remove it from https://github.com/inspirehep/inspire-next/blob/4a123080ac1d2cdd6f445a3104f8ef06b99edb44/inspirehep/modules/workflows/tasks/manual_merging.py#L65

inspirehep/modules/workflows/models.py
```diff
@@ -88,20 +96,30 @@ class WorkflowsRecordSources(db.Model):
         db.PrimaryKeyConstraint('record_id', 'source'),
     )
 
-    source = db.Column(
-        db.Text,
-        default='',
-        nullable=False,
-    )
     record_id = db.Column(
```
drjova: let's rename it to ``record_uuid``, it's more accurate

drjova:
could you please also squash all the commits?

inspirehep/alembic/17ff155db70d_update_workflows_record_sources_table.py
```diff
@@ -0,0 +1,95 @@
...
+
+class SourceEnum(enum.IntEnum):
+    arxiv = 1
+    submitter = 2
+    publisher = 3
+
+
+def upgrade():
+    """Upgrade database."""
+    op.add_column(
```
drjova: we have to alter the name of the ``record_uuid`` here as well

drjova:
ci test please

tests/integration/conftest.py
```diff
@@ -114,6 +114,58 @@ def isolated_app(app):
...
+        list(_es.create(ignore=[400]))
+
+        init_all_storage_paths()
+        init_users_and_permissions()
+        init_collections()
+
+        migrate('./inspirehep/demosite/data/demo-records-small.xml', wait_for_results=True)
+        es.indices.refresh('records-hep')
+
+        yield app
```
jacquerie: Watch out! This was removed in https://github.com/inspirehep/inspire-next/commit/a2a49d67aa7abd86fb39faa3023e57e1896ec64c and should stay removed.

drjova:
ci merge please

PR #3377, opened 2018-04-25
workflows: feature flag for reference removal

michamos:
This brings back `remove_references` in the workflows with a twist: a
feature flag controls whether it is active or not.


inspirehep/config.py
```diff
@@ -55,6 +55,7 @@
 FEATURE_FLAG_ENABLE_MERGER = False
 FEATURE_FLAG_ENABLE_UPDATE_TO_LEGACY = False
 """This feature flag will prevent to send a ``replace`` update to legacy."""
+FEATURE_FLAG_REMOVE_REFS_IN_WORKFLOW = True
```
drjova: shouldn't we keep the same naming for flags i.e. ``FEATURE_FLAG_ENABLE_REMOVE_REFS_IN_WORKFLOW``? It might get really confusing in the future if we have different "types"

michamos:
This is not how it should be done, will open a new PR.

PR #3378, opened 2018-04-25
workflows: feature flag for sending refs to legacy

michamos:
Unlike #3377, the optional removal of references is now done in `send_robotupload` in order not to delete references from the holdingpen and allow their inspection.

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -55,6 +55,7 @@ FEATURE_FLAG_ENABLE_FUZZY_MATCHER = False
 FEATURE_FLAG_ENABLE_MERGER = False
 FEATURE_FLAG_ENABLE_UPDATE_TO_LEGACY = False
 """This feature flag will prevent to send a ``replace`` update to legacy."""
+FEATURE_FLAG_ENABLE_SENDING_REFERENCES_TO_LEGACY = False
 
 # Default language and timezone
 # =============================
--- a/inspirehep/modules/workflows/tasks/submission.py
+++ b/inspirehep/modules/workflows/tasks/submission.py
@@ -26,6 +26,7 @@ from __future__ import absolute_import, division, print_function
 
 import os
 import logging
+from copy import copy
 from functools import wraps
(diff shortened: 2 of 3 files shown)


inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -218,6 +219,11 @@ def _send_robotupload(obj, eng):
             data = obj.extra_data.get(extra_data_key) or {}
         else:
             data = obj.data
+
+        if not current_app.config.get('FEATURE_FLAG_ENABLE_SENDING_REFERENCES_TO_LEGACY'):
+            data = copy(data)
```
drjova: ```deepcopy```?

inspirehep/config.py
```diff
@@ -55,6 +55,7 @@
 FEATURE_FLAG_ENABLE_MERGER = False
 FEATURE_FLAG_ENABLE_UPDATE_TO_LEGACY = False
 """This feature flag will prevent to send a ``replace`` update to legacy."""
```
turtle321: Does this comment refer to the next new line `FEATURE_FLAG_ENABLE_SENDING_REFERENCES_TO_LEGACY = False`? Can you make it more clear? And format it as a regular comment starting with `#`?

I'm sorry for not having seen it earlier.

inspirehep/modules/workflows/tasks/submission.py
michamos: why? I am only removing a top-level key.

drjova: is not need it, just to have a clean ``data``

inspirehep/config.py
drjova: this is module level variable documented inline ([example](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)), we might need to define the style we're gonna use in the future

michamos:
ci merge please

PR #3380, opened 2018-04-25
workflows: add data references matcher

salmanmaq:
## Description
This adds the capability to match references to data records in the current reference matcher.

## Related Issue
https://its.cern.ch/jira/browse/INSPIR-472

## Motivation and Context
This would improve citation counts.

## Checklist:

Signed-off-by: Salman Maqbool <salman.maqbool@cern.ch>

### DIFF
--- a/inspirehep/modules/workflows/config.py
+++ b/inspirehep/modules/workflows/config.py
@@ -186,3 +186,24 @@ WORKFLOWS_REFERENCE_MATCHER_JHEP_AND_JCAP_CONFIG = {
 """Configuration for matching records JCAP and JHEP records since
 they have a particular configuration and we have to look at the
 year as well for accurate matching."""
+
+WORKFLOWS_REFERENCE_MATCHER_DATA_CONFIG = {
+    'algorithm': [
+        {
+            'queries': [
+                {
+                    'path': 'reference.dois',
+                    'search_path': 'dois.value.raw',
+                    'type': 'exact',
+                },
+            ],
+        },
+    ],
+    'doc_type': 'data',
+    'index': 'records-data',
+    'source': [
+        'control_number',
+    ]
+}
+"""Configuration for matching data records. Please note that the
+index and doc_type are different for data records."""
--- a/inspirehep/modules/workflows/tasks/refextract.py
+++ b/inspirehep/modules/workflows/tasks/refextract.py
@@ -192,23 +192,41 @@ def extract_references_from_raw_ref(reference, custom_kbs_file=None):
 
 
 def match_reference(reference):
-    """Match references given a reference metadata using InspireMatcher queires.
+    """Match a reference using inspire-matcher.
 
     Args:
-        reference: The reference metadata
-        config: The configurtaion(s) for InspireMatcher queries
+        reference (dict): the metadata of a reference.
 
     Returns:
-        The record ID of the matched reference
+        dict: the matched reference.
+
     """
+    # XXX: avoid this type casting.
+    if get_value(reference, 'reference.publication_info.year'):
+        reference['reference']['publication_info']['year'] = str(reference['reference']['publication_info']['year'])
 
     config_default = current_app.config['WORKFLOWS_REFERENCE_MATCHER_DEFAULT_CONFIG']
     config_jcap_and_jhep = current_app.config['WORKFLOWS_REFERENCE_MATCHER_JHEP_AND_JCAP_CONFIG']
+    config_data = current_app.config['WORKFLOWS_REFERENCE_MATCHER_DATA_CONFIG']
 
     journal_title = get_value(reference, 'reference.publication_info.journal_title')
-    config = config_jcap_and_jhep if journal_title in ['JCAP', 'JHEP'] else config_default
+    if journal_title in ['JHEP', 'JCAP']:
+        config = config_jcap_and_jhep
+    else:
+        config = config_default
+
     result = next(match(reference, config), None)
     if result:
         matched_recid = result['_source']['control_number']
         reference['record'] = get_record_ref(matched_recid, 'literature')
+
+    result_for_data = next(match(reference, config_data), None)
+    if result_for_data:
+        matched_recid = result_for_data['_source']['control_number']
+        reference['record'] = get_record_ref(matched_recid, 'data')
+
+    # XXX: avoid this type casting.
+    if get_value(reference, 'reference.publication_info.year'):
+        reference['reference']['publication_info']['year'] = int(reference['reference']['publication_info']['year'])
+
     return reference
--- a/tests/integration/helpers/factories/db/invenio_records.py
+++ b/tests/integration/helpers/factories/db/invenio_records.py
@@ -55,7 +55,7 @@ class TestRecordMetadata(TestBaseModel):
     }
 
     @classmethod
-    def create_from_kwargs(cls, index=True, has_pid=True, **kwargs):
+    def create_from_kwargs(cls, index_name='', **kwargs):
         instance = cls()
 
         updated_kwargs = copy.deepcopy(kwargs)
@@ -81,25 +81,24 @@ class TestRecordMetadata(TestBaseModel):
         instance.record_metadata = super(TestRecordMetadata, cls)\
                 .create_from_kwargs(updated_kwargs)
 
-        if index:
+        if index_name:
             instance.es_index_result = es.index(
-                index='records-hep',
-                doc_type='hep',
+                index=index_name,
+                doc_type=index_name.split('-')[-1],
                 body=instance.record_metadata.json,
                 params={}
             )
-            instance.es_refresh_result = es.indices.refresh('records-hep')
+            instance.es_refresh_result = es.indices.refresh(index_name)
 
-        if has_pid:
-            instance.persistent_identifier = TestPersistentIdentifier\
-                    .create_from_kwargs(
-                        object_uuid=instance.record_metadata.id,
-                        pid_value=instance.record_metadata.json.get('control_number')
-                    ).persistent_identifier
+        instance.persistent_identifier = TestPersistentIdentifier\
+                .create_from_kwargs(
+                    object_uuid=instance.record_metadata.id,
+                    pid_value=instance.record_metadata.json.get('control_number'),
+                    **kwargs).persistent_identifier
         return instance
 
     @classmethod
-    def create_from_file(cls, module_name, filename, index=True, has_pid=True):
+    def create_from_file(cls, module_name, filename, index_name=''):
         """Create Record instance from file.
 
         Note:
@@ -117,4 +116,4 @@ class TestRecordMetadata(TestBaseModel):
 
         data = json.load(open(path))
         return cls.create_from_kwargs(
-            index=index, has_pid=has_pid, json=data)
+            index_name=index_name, json=data)
--- a/tests/integration/workflows/test_workflows_tasks_actions.py
+++ b/tests/integration/workflows/test_workflows_tasks_actions.py
@@ -86,7 +86,7 @@ def insert_journals_in_db(workflow_app):
 @pytest.fixture
 def insert_cited_record(workflow_app):
     """Adds a record with arXiv: 1308.0815 which can be cited by another record.
-    See test_refextract_from_pdf for reference and context.
+    See test_refextract_from_pdf_for_default_config for reference and context.
     Only to be used in conjunction with arXiv: 1407.7587 as it cites this
     record."""
 
@@ -432,7 +432,8 @@ def test_refextract_from_pdf(
     workflow_app,
     mocked_external_services
 ):
-    """Test refextract from PDF by going through the entire workflow."""
+    """Test refextract from PDF and reference matching for default Configuration
+     by going through the entire workflow."""
 
     citing_record, categories = insert_citing_record()
 
--- a/tests/integration/workflows/test_workflows_tasks_actions.py
+++ b/tests/integration/workflows/test_workflows_tasks_actions.py
@@ -32,8 +32,6 @@ import pkg_resources
 import pytest
 
 from invenio_db import db
-from invenio_pidstore.models import PersistentIdentifier
-from invenio_records.models import RecordMetadata
 from inspire_schemas.api import load_schema, validate
 from invenio_search.api import current_search_client as es
 from invenio_workflows import (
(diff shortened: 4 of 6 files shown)


inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -204,11 +204,25 @@ def match_reference(reference):
 
     config_default = current_app.config['WORKFLOWS_REFERENCE_MATCHER_DEFAULT_CONFIG']
     config_jcap_and_jhep = current_app.config['WORKFLOWS_REFERENCE_MATCHER_JHEP_AND_JCAP_CONFIG']
+    config_data = current_app.config['WORKFLOWS_REFERENCE_MATCHER_DATA_CONFIG']
 
     journal_title = get_value(reference, 'reference.publication_info.journal_title')
-    config = config_jcap_and_jhep if journal_title in ['JCAP', 'JHEP'] else config_default
+    if journal_title in ['JHEP', 'JCAP']:
+        config = config_jcap_and_jhep
+        if get_value(reference, 'reference.publication_info.year'):
```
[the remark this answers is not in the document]
salmanmaq: Oh, right. But I already refactored it while working for erratum citations, so the next PR should automatically fix this anyway.

tests/integration/workflows/test_workflows_tasks_refextract.py
```diff
@@ -0,0 +1,163 @@
...
+            }
+        ],
+        'titles': [
+            {
+                'title': 'The Strongly-Interacting Light Higgs'
+            }
+        ],
+    }
+
+    cited_record = InspireRecord.create(cited_record_json, id_=None)
```
turtle321: What about using https://github.com/inspirehep/inspire-next/blob/master/tests/integration/helpers/factories/db/invenio_records.py#L39

tests/integration/helpers/factories/db/invenio_records.py
```diff
@@ -81,25 +81,34 @@ def create_from_kwargs(cls, index=True, has_pid=True, **kwargs):
         instance.record_metadata = super(TestRecordMetadata, cls)\
                 .create_from_kwargs(updated_kwargs)
 
-        if index:
+        if index == 'records-hep':
```
drjova: I'd also pass the ``doc_type`` as a keyword, so we don't have any index restriction since we want to make it as abstract as possible

tests/integration/helpers/factories/db/invenio_records.py
```diff
@@ -55,7 +55,7 @@ class TestRecordMetadata(TestBaseModel):
     }
 
     @classmethod
-    def create_from_kwargs(cls, index=True, has_pid=True, **kwargs):
+    def create_from_kwargs(cls, index=None, pid_type=None, **kwargs):
```
turtle321: An index is a data structure. This variable does not hold an index, but the name of an index. Thus `index_name` or something similar is a more appropriate naming.

tests/integration/helpers/factories/db/invenio_records.py
drjova: we could also add ``@lassmethods`` for our two main cases, ``create_literature`` and ``create_data`` so we can use them easily

salmanmaq: class methods seems like a nice idea.

tests/integration/helpers/factories/db/invenio_records.py
```diff
@@ -81,25 +81,34 @@ def create_from_kwargs(cls, index=True, has_pid=True, **kwargs):
...
+            instance.es_index_result = es.index(
+                index='records-data',
+                doc_type='data',
+                body=instance.record_metadata.json,
+                params={}
+            )
+            instance.es_refresh_result = es.indices.refresh('records-data')
 
-        if has_pid:
+        if pid_type:
```
turtle321: You have to pass the kwargs around like: https://github.com/inspirehep/inspire-next/blob/master/tests/integration/helpers/factories/db/invenio_oauthclient.py#L147
So:
- remove the `pid_type` from the signature of this method
- remove the if clause at this line
- change the next line to
```
... = TestPersistentIdentifier.create_from_kwargs(
    object_uuid=instance.record_metadata.id,
    pid_value=instance.record_metadata.json.get('control_number'),
    **kwargs).persistent_identifier
```

turtle321: This way we do not need to specify the `pid_type` when you want a `lit` because `TestPersistentIdentifier` defaults to that.

tests/integration/helpers/factories/db/invenio_records.py
turtle321: Also if index is a string, default it to `''`

inspirehep/modules/workflows/config.py
```diff
@@ -186,3 +186,24 @@
 """Configuration for matching records JCAP and JHEP records since
 they have a particular configuration and we have to look at the
 year as well for accurate matching."""
+
+WORKFLOWS_REFERENCE_MATCHER_DATA_CONFIG = {
```
ammirate: Probably `inspirehep/modules/workflows/config.py` is not the most appropriate place for this guy :) 
There are already similar configurations in https://github.com/inspirehep/inspire-next/blob/master/inspirehep/config.py, like `EXACT_MATCH` and `FUZZY_MATCH`

salmanmaq: I think module specific configurations would be a better way to do this, no?

tests/integration/helpers/factories/db/invenio_records.py
salmanmaq: Coming to think of this again, do we really need separate methods for that? With the current configuration (latest commit), we can essentially use 1 required(if we want to index the record in ES) and 1 optional arguments to create any type of record: index_name and pid_type.

inspirehep/modules/workflows/config.py
turtle321: We discussed this in the team and we believe that having specific config files per module is a better solution than having a big centralized config file.

ammirate: I prefer this way as well! My suggestion was just to keep using the same style around the peoject, but if we agree on this we can start moving configs in related modules.

tests/integration/helpers/factories/db/invenio_records.py
```diff
@@ -81,25 +81,24 @@ def create_from_kwargs(cls, index=True, has_pid=True, **kwargs):
...
-
-        if has_pid:
-            instance.persistent_identifier = TestPersistentIdentifier\
-                    .create_from_kwargs(
-                        object_uuid=instance.record_metadata.id,
-                        pid_value=instance.record_metadata.json.get('control_number')
-                    ).persistent_identifier
+            instance.es_refresh_result = es.indices.refresh(index_name)
+
+        instance.persistent_identifier = TestPersistentIdentifier\
```
drjova: heads-up: we will always create a ``pid``, do we really want this?

tests/integration/helpers/factories/db/invenio_records.py
drjova: it'd be cleaner for the tests, just to call what you need without having "hard coded" strings, if you think that it is out of the US's scope, we can leave it as is.

tests/integration/helpers/factories/db/invenio_records.py
turtle321: We asked ourself the same question. I believe it makes sense to have a PIS every time a record is created. Also, this is cheap with a properly isolated app (PR on it's way)

tests/integration/helpers/factories/db/invenio_records.py
```diff
@@ -81,25 +81,24 @@ def create_from_kwargs(cls, index=True, has_pid=True, **kwargs):
...
+                .create_from_kwargs(
+                    object_uuid=instance.record_metadata.id,
+                    pid_value=instance.record_metadata.json.get('control_number'),
```
michamos: don't you need a `pid_type` here?

tests/integration/helpers/factories/db/invenio_records.py
```diff
@@ -55,7 +55,7 @@ class TestRecordMetadata(TestBaseModel):
     }
 
     @classmethod
-    def create_from_kwargs(cls, index=True, has_pid=True, **kwargs):
+    def create_from_kwargs(cls, index_name='', **kwargs):
```
jacquerie: Mh, probably you've already discussed this, but why isn't the default `records-hep` ?

tests/integration/workflows/test_workflows_tasks_actions.py
```diff
@@ -428,12 +373,47 @@ def test_refextract_from_pdf(
...
+        ],
+        'titles': [
+            {
+                'source': 'arXiv',
+                'title': 'Solving a two-electron quantum dot model in terms of polynomial solutions of a Biconfluent Heun equation'
+            }
+        ],
+    }
 
+    TestRecordMetadata.create_from_kwargs(
```
jacquerie: Here I would `create_from_file`, so that that big record above doesn't interrupt the flow of the test.

tests/integration/helpers/factories/db/invenio_records.py
salmanmaq: The default is None for the case where we don't want to index the record in ES.

tests/integration/workflows/test_workflows_tasks_actions.py
```diff
@@ -453,5 +433,5 @@ def test_refextract_from_pdf(
     citing_doc_eng = WorkflowEngine.from_uuid(citing_doc_workflow_uuid)
     citing_doc_obj = citing_doc_eng.processed_objects[0]
 
-    assert citing_doc_obj.data['references'][7]['record']['$ref'] == 'http://localhost:5000/api/literature/1'
+    assert citing_doc_obj.data['references'][7]['record']['$ref'] == 'http://localhost:5000/api/literature/1000'
```
jacquerie: Nit: I don't understand this change... is there some state leaking from a previous workflow test?

tests/integration/helpers/factories/db/invenio_records.py
salmanmaq: I suppose I'll leave it like this for now.

tests/integration/helpers/factories/db/invenio_records.py
jacquerie: Ah, yes, makes sense. But then why isn't the default `None`, instead of empty string? : )

tests/integration/helpers/factories/db/invenio_records.py
salmanmaq: Not really, since it gets assigned 'lit' by default in TestPersistentIdentifier.

tests/integration/workflows/test_workflows_tasks_actions.py
salmanmaq: I doubt that but there does seem to be some issue with multiple records being inserted in a single test. So if I insert the cited record, and then try to run the workflow for the citing record, I get some error that the PID already exists. This is possibly some artefact of how we insert records using TestRecordMetadata.

tests/integration/helpers/factories/db/invenio_records.py
michamos: sure, but what if you're creating a data record? then it should have `dat` instead of `lit`.

tests/integration/workflows/test_workflows_tasks_actions.py
salmanmaq: In fact, I am sure that it's not a leak from other test since this happens even if I just run onlt this particular test.

tests/integration/helpers/factories/db/invenio_records.py
salmanmaq: In that case, we'll pass the pid_type='dat' argument and it will pick it from kwargs.

tests/integration/workflows/test_workflows_tasks_actions.py
jacquerie: I see. Maybe the workflow will try to insert record 1 because that's the state of the PID counter, but the test fixture already put there a record 1 without updating the PID counter.

Let's keep the code like this for now, but this is a little too brittle for my taste...

tests/integration/helpers/factories/db/invenio_records.py
salmanmaq: Empty string since it will contain string arguments (as suggested by Paolo).

tests/integration/workflows/test_workflows_tasks_actions.py
jacquerie: Yes, my recommendation would be to either have short records with strictly the necessary information in the body of the test, or long records loaded as fixtures (for example, for regressions that require complicated sets of data to be triggered).

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -204,11 +204,25 @@ def match_reference(reference):
...
+            reference['reference']['publication_info']['year'] = str(reference['reference']['publication_info']['year'])
```
michamos: this seems very wrong to me. If ES expects a string, either the mapping or the config needs to be changed. The record schema says that [`year` is an integer](https://github.com/inspirehep/inspire-schemas/blob/963237a736db12324136760c06479d41d6abed9c/inspire_schemas/records/elements/reference.yml#L247-L252), if you change this the reference will fail validation.

tests/integration/workflows/test_workflows_tasks_refextract.py
```diff
@@ -0,0 +1,105 @@
...
+                'page_start': '045',
+                'year': 2007
+            }
+        }
+    }
+
+    reference = match_reference(reference)
+
+    assert reference['record']['$ref'] == 'http://localhost:5000/api/literature/1'
+    assert isinstance(reference['reference']['publication_info']['year'], int)
```
michamos: I meant actually call the `validate` function, like here https://github.com/inspirehep/inspire-next/blob/16504109805e9b93b034ef5cf4f50683ea55b579/tests/unit/workflows/test_workflows_tasks_refextract.py#L130

salmanmaq:
ci merge please

salmanmaq:
ci merge please

david-caro:
ci merge please

david-caro:
ci merge please

david-caro:
ci merge please

david-caro:
ci merge please

david-caro:
ci merge please

david-caro:
ci merge please

PR #3383, opened 2018-04-26
549 orcid: add timeout to python-orcid calls

turtle321:


requirements.txt
```diff
@@ -27,7 +27,9 @@
 -e git+https://github.com/inspirehep/python-client-sword2.git#egg=sword2
 
 # python-orcid with support for v2.0 API and XML
--e git+https://github.com/ORCID/python-orcid.git@2.0update#egg=orcid
+# -e git+https://github.com/ORCID/python-orcid.git@2.0update#egg=orcid
```
michamos: you can remove this line, they released the new API upstream so we won't need it again.

turtle321:
@michamos we also need to fix this: https://github.com/ORCID/python-orcid/pull/24/files#r184513248

turtle321:
And I will add a test here.

inspirehep/modules/orcid/api.py
```diff
@@ -101,6 +101,7 @@ def push_record_with_orcid(recid, orcid, oauth_token, put_code=None, old_hash=No
                 data=orcid_xml,
                 put_code=put_code,
                 content_type='application/orcid+xml',
+                timeout=30,
```
jacquerie: @turtle321 this is the value that we agreed on, right?

turtle321: It looks reasonable.
I am also fine with 60 secs for non "user facing" tasks.

michamos:
https://github.com/ORCID/python-orcid/pull/24 was merged and released, so we can use the released package.

turtle321:
ci merge please

PR #3400, opened 2018-05-04
global: bump celery to version 4

jacquerie:
## Description:
The inline comments that I made in https://github.com/inspirehep/inspire-next/pull/2138 still apply here. ~~On the other hand, this is expected to fail because we recently introduced code that uses a pattern that is disallowed by default in Celery 4, see: http://docs.celeryq.org/en/latest/userguide/tasks.html#task-synchronous-subtasks.~~ This has been fixed by https://github.com/inspirehep/inspire-next/pull/3401, as shown by https://travis-ci.org/inspirehep/inspire-next/jobs/374936514#L1500.

## Related Issue:
Jira: https://its.cern.ch/jira/browse/INSPIR-632, also closes https://github.com/inspirehep/inspire-next/issues/2037 but ~~depends on https://github.com/inspirehep/inspire-crawler/pull/38~~ and ~~depends on https://github.com/inspirehep/hepcrawl/pull/241~~.

## Checklist:


docker-compose.test.yml
```diff
@@ -30,8 +30,8 @@ services:
       service: base
     environment:
       - APP_SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://inspirehep:dbpass123@test-database:5432/inspirehep
-      - APP_BROKER_URL=amqp://guest:guest@test-rabbitmq:5672//
-      - APP_CELERY_RESULT_BACKEND=amqp://guest:guest@test-rabbitmq:5672//
+      - APP_CELERY_BROKER_URL=pyamqp://guest:guest@test-rabbitmq:5672//
+      - APP_CELERY_RESULT_BACKEND=rpc://guest:guest@test-rabbitmq:5672//
```
michamos: why not redis?

docker-compose.test.yml
```diff
@@ -30,8 +30,8 @@ services:
       service: base
     environment:
       - APP_SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://inspirehep:dbpass123@test-database:5432/inspirehep
-      - APP_BROKER_URL=amqp://guest:guest@test-rabbitmq:5672//
-      - APP_CELERY_RESULT_BACKEND=amqp://guest:guest@test-rabbitmq:5672//
+      - APP_CELERY_BROKER_URL=pyamqp://guest:guest@test-rabbitmq:5672//
```
michamos: is this really needed? `amqp` is still in the installation docs: http://docs.celeryproject.org/en/latest/getting-started/brokers/rabbitmq.html#installation-configuration

setup.py
```diff
@@ -95,7 +94,6 @@
     'invenio-workflows~=7.0,>=7.0.1',
     'langdetect~=1.0,>=1.0.7',
     'latexcodec~=1.0,>=1.0.5',
-    'librabbitmq~=1.0,>=1.6.1',
```
michamos: This should still be supported (quoting http://docs.celeryproject.org/en/latest/userguide/configuration.html#broker-url):

> The transport part is the broker implementation to use, and the default is amqp, (uses librabbitmq if installed or falls back to pyamqp). There are also other choices available, including; redis://, sqs://, and qpid://.

jacquerie: Mh, uninstalling `librabbitmq` is recommended by Celery's creator: https://github.com/celery/celery/issues/3675#issuecomment-266596324.

docker-compose.test.yml
michamos: > The old legacy “amqp” **result** backend has been deprecated, and will be removed in Celery 5.0.

(emphasis mine), I am talking about the broker, not the result backend.

setup.py
michamos: I am reading that differently: the person is explicitly asking for `pyamqp`, with `librabbitmq` installed. From the docs, I am gathering that you can use the `amqp` protocol to use either, but if you are explicitly asking for `pyamqp`, you shouldn't have `librabbitmq` installed. Otherwise, it would completely contradict the official docs in several places.

docker-compose.test.yml
jacquerie: Ah, I missed that! Then I don't know the answer off the top of my head. *Probably* I did it because of some deprecation notice or something like that. Let me revert that change and see what happens.

docker-compose.test.yml
jacquerie: Redis is a much more sensible choice for a result backend, but I'm thinking that that migration requires some care to be performed without data loss on PROD. So I'd propose to skip that for now, and just bump Celery in this PR.

setup.py
jacquerie: You were right here and above about `pyamqp` ! Amended the PR.

michamos: Indeed, despite what the docs say, `pyamqp` is recommended by the celery people: https://github.com/celery/librabbitmq/issues/118#issuecomment-385461136

michamos:
This requires changing the puppet config too. I will open a PR for that.

docker-compose.test.yml
jacquerie: ...it turns out that PROD has been using Redis as a result backend since immemorable times, so let me add a commit that does the same for DEV.

jacquerie:
> This requires changing the puppet config too. I will open a PR for that.

AFAIK all the required changes have either happened in https://gitlab.cern.ch/ai/it-puppet-hostgroup-inspire/commit/07204d101736d11721bb9afca82a38e06f4eab33 or will happen in https://gitlab.cern.ch/ai/it-puppet-hostgroup-inspire/merge_requests/61.

jacquerie:
As this PR requires extensive testing on QA (I'm thinking at least one night's harvest) and the `v0.8.1` release contains things that are more urgent, this PR is blocked by https://github.com/inspirehep/inspire-next/pull/3402.

michamos:
ci merge please

jacquerie:
> ci merge please

The first one didn't work because there's yet another place where configuration is declared, but this commit by @michamos should have updated that as well: https://gitlab.cern.ch/inspire/jenkins/commit/f0e70cd436292766da13c112fc1348efdb1f6977.

PR #3401, opened 2018-05-04
668 migrator: remove unneeded tasks

michamos:
This module used to have the `@shared_task` decorator applied very
liberally, even for functions that were always called as regular
functions and never scheduled through celery. Now, only functions that
are actually used as tasks have the decorator. This incidentally removes
all tasks synchronously calling other tasks, which is bad practice and
forbidden in Celery 4.*.


michamos:
ci merge please

PR #3403, opened 2018-05-08
workflows: fix workflow type check

drjova:
It has some common code boilerplate commonly needed to start a
workflow.

### DIFF
--- a/tests/integration/workflows/test_workflows_tasks_upload.py
+++ b/tests/integration/workflows/test_workflows_tasks_upload.py
@@ -22,12 +22,10 @@
 
 from __future__ import absolute_import, division, print_function
 
-from mock import patch
+from mock import MagicMock, patch
 from flask import current_app
 
 from invenio_workflows import workflow_object_class
-from invenio_workflows.models import Workflow
-
 
 # FIXME: otherwise this task is not found by Celery.
 from inspirehep.modules.orcid.tasks import orcid_push  # noqa: F401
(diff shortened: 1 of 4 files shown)


david-caro:
It would make sense to use the start_workflow on inspire-crawler too.

drjova:
wouldn't be a weird dependency though? probably we should move it in the invenio-workflows or smth

david-caro:
> It would make sense to use the start_workflow on inspire-crawler too.

Actually... not as it is though xd, as it has to be able to start the `start` task in the special `harvests` queue (it's a parameter to the `apply_async`).

drjova:
Supersede by #3405

drjova:
ci test please

tests/integration/workflows/test_workflows_tasks_upload.py
```diff
@@ -26,14 +26,24 @@
...
 
 # FIXME: otherwise this task is not found by Celery.
 from inspirehep.modules.orcid.tasks import orcid_push  # noqa: F401
 from inspirehep.modules.workflows.tasks.upload import store_record
 
 
+class AttrDict(dict):
+    def __init__(self, *args, **kwargs):
+        super(AttrDict, self).__init__(*args, **kwargs)
+        self.__dict__ = self
```
david-caro: This looks very ugly to me :/, as it will have the actual methods of the dict (for example, update).

david-caro: I would use another mock object, like `MockEng` or similar (even a MagickMock from the mock module).

drjova: This is exactly how we do it on the unit tests https://github.com/inspirehep/inspire-next/blob/master/tests/unit/helpers/mocks.py#L30-L33

david-caro:
ci merge please

PR #3404, opened 2018-05-08
676 script: remove refs from workflows

turtle321:


scripts/workflows/remove_refs.py
```diff
@@ -0,0 +1,25 @@
...
+def remove_references(workflow_id):
+    wf = workflow_object_class.get(workflow_id)
+    print 'Workflow {} is currently in position {}'.format(workflow_id,
+                                                           wf.callback_pos)
+    # Note that an empty list is not schema compliant.
+    del wf.data['references']
+    wf.save()
+    db.session.commit()
+
+    res = wf.continue_workflow(delayed=True)
```
ammirate: If we don't put the `continue_next` param, you are going to have a robot upload error after, right?

ammirate: Check this guy https://github.com/inveniosoftware-contrib/invenio-workflows/blob/master/invenio_workflows/api.py#L324

turtle321: "continue_next" seems to be the default, see: https://github.com/inveniosoftware-contrib/invenio-workflows/blob/master/invenio_workflows/api.py#L346
Anyway I will add it to be more explicit

ammirate:
So, after discussing IRL, `restart_task` is the right thing to do for continuing the workflow :+1:

PR #3414, opened 2018-05-22
766 alembic: create index on signature_blocks

michamos:
This creates an index on the `signature_block` field of `authors` in `RecordMetadata.json`, which can be used to efficiently get all records for authors with a given signature block. This is useful for author disambiguation. Note that this requires using a custom PostgreSQL function because `authors` is a json array containing objects, so the naive `(json->'authors'->>'signature_block')` doesn't work.

### DIFF
--- a/inspirehep/alembic/cb9f81e8251c_add_record_metadata_indices.py
+++ b/inspirehep/alembic/cb9f81e8251c_add_record_metadata_indices.py
@@ -30,7 +30,7 @@ from alembic import op
 revision = 'cb9f81e8251c'
 down_revision = 'fddb3cfe7a9c'
 branch_labels = ()
-depends_on = None
+depends_on = '07fb52561c5c'
 
 
 def upgrade():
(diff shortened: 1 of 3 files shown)


inspirehep/alembic/cb9f81e8251c_add_record_metadata_indices.py
```diff
@@ -30,7 +30,7 @@
 revision = 'cb9f81e8251c'
 down_revision = 'fddb3cfe7a9c'
 branch_labels = ()
-depends_on = None
+depends_on = '07fb52561c5c'
```
michamos: this is needed because otherwise locally a migration trying to create an index on the `records_metadata.json` column using `jsonb` operators runs before the migration transforming that column to `jsonb`.

jacquerie:
Can we close this as superseded by https://github.com/inspirehep/inspire-next/pull/3420 ?

michamos:
We are going with ElasticSearch for this feature in #3420, so closing.

PR #3418, opened 2018-05-23
disambiguation: first commit

jacquerie:
## Description:
The attempt to spin this module out as an independent blueprint (https://github.com/inspirehep/inspire-disambiguation) ran into the core issue of how to read/write the records independently of `inspire-next`, and the related issue of how to test code that interacts with the database.

One solution could have been to set up an independent Docker environment, similar to how `hepcrawl` does it. A problem with that approach is the undesirable repetition; the US to bump Celery to version 4 required modifying the same configuration variables in 4 different repositories. One could use the `docker-compose.yml` and `docker-compose.override.yml` pattern documented in https://docs.docker.com/compose/extends/#multiple-compose-files, which instead has the issue of having developers deal with git submodules, a slightly advanced topic. If I had enough time, I probably would have started a refactor to spin out things like the `isolated_app` and the test record factory.

The simplest solution (the one that has the highest chance of achieving something in the next month) is to succumb once again to the powerful allure of the monolith, and just make `disambiguation` a module in `inspire-next`.

Oh, and I added a commit to fix the message that rounded the number of migrated records to the nearest hundred during migration because it annoyed me.

## Related Issue:
Jira: https://its.cern.ch/jira/browse/INSPIR-746

## Checklist:

### DIFF
--- a/inspirehep/modules/disambiguation/__init__.py
+++ b/inspirehep/modules/disambiguation/__init__.py
@@ -1,0 +1,27 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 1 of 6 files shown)


jacquerie:
ci test please

jacquerie:
ci merge please

PR #3422, opened 2018-05-28
disambiguation: add method to save training data

jacquerie:
## Related Issue:
Jira: https://its.cern.ch/jira/browse/INSPIR-744

## Checklist:


tests/integration/disambiguation/test_disambiguation_core_db_readers.py
```diff
@@ -34,32 +34,34 @@ def test_get_all_curated_signatures(isolated_app):
...
+                'Standard Model',
+            ],
+            'publication_id': 792017,
+            'title': 'The quest for elementary particles',
+            'topics': ['Phenomenology-HEP'],
         },
+        'publication_id': 792017,
         'signature_block': 'ELj',
-        'uuid': '94f560d2-6791-43ec-a379-d3dc4ad0ceb7',
+        'signature_id': '94f560d2-6791-43ec-a379-d3dc4ad0ceb7',
```
jacquerie: I'm really tempted to deviate from the paper here and call it a `signature_uuid`...

salmanmaq: That sounds reasonable. Will require a few changes on my side, but this would be more consistent with INSPIRE naming conventions.

inspirehep/modules/disambiguation/config.py
```diff
@@ -25,5 +25,11 @@
 from __future__ import absolute_import, division, print_function
 
 
+DISAMBIGUATION_CLUSTERS_PATH = '/tmp/clusters.json'
+"""The path to the file that will store the curated clusters."""
+
+DISAMBIGUATION_SIGNATURES_PATH = '/tmp/signatures.jl'
```
salmanmaq: Is the file extension correct here?

inspirehep/modules/disambiguation/config.py
```diff
@@ -25,5 +25,11 @@
...
+"""The path to the file that will store the curated signatures."""
+
 DISAMBIGUATION_MODEL_PATH = '/tmp/inspire-disambiguation-model.pkl'
```
salmanmaq: I think we need to save the linkage function here which is supposed to have a .dat extension.

inspirehep/modules/disambiguation/config.py
jacquerie: Yes, it's a less used extension to mean this format: http://jsonlines.org/

Actually that's another breaking change from the paper: in order to write a big JSON array to disk one would need to first have it in memory, and I believe that no INSPIRE machine has enough RAM to do that.

If instead we save each signature as a JSON line we can write to disk as soon as we read a signature from the DB.

inspirehep/modules/disambiguation/config.py
jacquerie: Ah, don't worry too much about this guy, I introduced it in https://github.com/inspirehep/inspire-next/commit/6d7a5444419618705dd3edbfc6db61e2eb5282c4 to prove that we are able to load configuration from disk, but the filename will probably change as needed.

inspirehep/modules/disambiguation/config.py
jacquerie: All it means is that code like
```python
with open('signatures.json') as fd:
    signatures = json.load(fd)
    for signature in signatures:
        ...
```
becomes
```python
with open('signatures.jl') as fd:
    for line in fd:
        signature = json.loads(line)
        ...
```

salmanmaq: I checked the size of the data in memory (using sys.getsizeof ) provided by the beard paper repo and it's barely a few MBs.

jacquerie: > I checked the size of the data in memory (using sys.getsizeof ) provided by the beard paper repo and it's barely a few MBs.

Can you detail exactly what you are doing, so we know that we're comparing apples to apples? The signatures that we produce are comparable to `signatures.json` + `records.json`, and we can expect a lot more signatures and higher record metadata quality (in particular more abstracts, since they couldn't release the publisher ones, but we are free to use them while training!)

salmanmaq: htop and ctop give a different picture though. At least 3 GB in memory. I suppose you can go ahead with your proposal. It's better anyway.

salmanmaq: > Can you detail exactly what you are doing

So I loaded the signatures.json and records.json in a Jupyter notebook using json.load. Then I checked using sys.getsizeof on both records and signatures and they return ~ 9 MB and ~ 3 MB respectively. However, when I checked the memory usage in htop, it increased by ~ 3 GB. I might not be doing something correct using getsizeof.

michamos: `sys.getsizeof` only looks at the size of the top-level object in case of collections:
```
In [1]: import sys

In [2]: d = {'foo': {}}

In [3]: sys.getsizeof(d)
Out[3]: 280

In [4]: d['foo'] = {i: i for i in xrange(100000)}

In [5]: sys.getsizeof(d)
Out[5]: 280

In [6]: sys.getsizeof(d['foo'])
Out[6]: 6291736

```

salmanmaq: Thanks for the explanation. That makes sense.

tests/integration/disambiguation/test_disambiguation_api.py
```diff
@@ -0,0 +1,96 @@
...
+
+from factories.db.invenio_records import TestRecordMetadata
+
+from inspirehep.modules.disambiguation.api import save_training_data
+
+
+def test_save_training_data(isolated_app):
+    TestRecordMetadata.create_from_file(__name__, '792017.json')
+
+    clusters_temporary_fd, clusters_path = mkstemp()
```
michamos: why not use [`tmpdir`](https://docs.pytest.org/en/latest/tmpdir.html)? It's used e.g. [here](https://docs.pytest.org/en/latest/tmpdir.html) and handles all the mess for you.

jacquerie: Excellent idea. Using it simplified the test a great deal.

jacquerie: I liked it so much that I added https://github.com/inspirehep/inspire-next/pull/3422/commits/4df1c3bba93b102e299b2617745b956a64fc218c to rewrite all tests that were using `mkdtemp` / `mkstemp` so that we only have correct examples in the code base!

jacquerie:
ci merge please

PR #3433, opened 2018-05-31
serializers: addition of ``ui`` serializer

drjova:

### DIFF
--- a/inspirehep/modules/records/serializers/json_literature.py
+++ b/inspirehep/modules/records/serializers/json_literature.py
@@ -30,27 +30,16 @@ from inspire_utils.date import format_date
 from inspirehep.modules.records.wrappers import LiteratureRecord
 
 
-def process_es_hit(record):
+def _get_ui_metadata(record):
+    """Record extra metadata for the UI.
+
+    Args:
+        record(dict): the record.
+
+    Returns:
+        dict: the extra metadata.
     """
-    Process record coming from Elasticsearch.
 
-    Allows to remove unnecessary fields to reduce bandwidth and speed
-    up the client application.
-    """
-    if 'authors' in record:
-        record['authors'] = record['authors'][:10]
-    if 'references' in record:
-        del record['references']
-
-    return record
-
-
-def get_display_fields(record):
-    """
-    Add extra fields used for display by client application.
-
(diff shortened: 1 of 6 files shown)


inspirehep/modules/records/serializers/json_literature.py
```diff
@@ -67,27 +56,37 @@ def get_display_fields(record):
...
+
+        ui_metadata = _get_ui_metadata(record)
+
+        if 'authors' in record:
+            record['authors'] = record['authors'][:10]
+        if 'references' in record:
+            del record['references']
+
+        record.update(ui_metadata)
+        result['metadata'] = record
```
david-caro: This should not be needed, though I agree that is more readable like this...

Maybe short-circuiting the if would make it more readable :/, not sure:
```python
def _preprocess_result(result):
    if 'metadata' not in result:
        return result

    record = result['metadata']
    ...
    return result
```

btw. what does it mean that result has no 'metadata'? is that expected? should we throw an exception if not there?

inspirehep/modules/records/serializers/json_literature.py
```diff
@@ -67,27 +56,37 @@ def get_display_fields(record):
...
+
+        # FIXME: Depricated, must be removed once the new UI is released
```
david-caro: s/Depricated/Deprecated/

inspirehep/modules/records/serializers/json_literature.py
drjova: Actually ``metadata`` should always be there, is part of the ``preprocess`` https://github.com/inveniosoftware/invenio-records-rest/blob/3d63611e7e85ee5a981ff1faa67d11bc166bdfbe/invenio_records_rest/serializers/base.py#L140-L173

david-caro: Then maybe better throw an exception if it's not there (just removing the check might be enough though, a key error at that point is kind-of understandable too).

drjova: yep, I'll remove it

drjova:
@david-caro updated

david-caro:
:+1:

drjova:
ci merge please

PR #3434, opened 2018-05-31
74 script: delete unusable orcid tokens

turtle321:


scripts/delete_orcid_tokens.py
```diff
@@ -0,0 +1,30 @@
...
+
+import re
+
+from invenio_db import db
+from invenio_oauthclient.models import RemoteToken
+
+
+def find_unusable_orcid_tokens():
+    # Access tokens are in the form:
+    #   "7559f1gt-a65g-49s4-9cf4-35b27989289f"
```
turtle321: These 2 examples are not real token

turtle321:
I tested that and it works

PR #3441, opened 2018-06-04
disambiguation: add ethnicity estimator

jacquerie:
## Description:
Integrates https://github.com/inspirehep/beard/blob/87b37a3028825b5199616b195663fee526c26be3/examples/applications/author-disambiguation/ethnicity.py.

## Checklist:

### DIFF
--- a/inspirehep/modules/disambiguation/core/ml/models.py
+++ b/inspirehep/modules/disambiguation/core/ml/models.py
@@ -1,0 +1,73 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2017 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


inspirehep/modules/disambiguation/core/ml/models.py
```diff
@@ -0,0 +1,72 @@
...
+        self.C = C
+
+    def load(self, input_filename):
+        ethnicities, lasts, firsts = [], [], []
+        with open(input_filename, 'r') as fd:
+            reader = csv.DictReader(fd)
+            for row in reader:
+                ethnicities.append(row['RACE'])
+                lasts.append(row['NAMELAST'])
+                firsts.append(row['NAMEFRST'])
```
jacquerie: This is a little more convoluted and slow than the original https://github.com/inspirehep/beard/blob/87b37a3028825b5199616b195663fee526c26be3/examples/applications/author-disambiguation/ethnicity.py#L39, but it has the benefit of avoiding a big dependency such as `pandas` for exactly one line of code.

setup.py
```diff
@@ -103,6 +104,7 @@
     'raven[flask]~=6.0,>=6.2.1',
     'refextract~=0.0,>=0.2.2',
     'requests~=2.0,>=2.18.4',
+    'scikit-learn~=0.0,>=0.19.1',
```
jacquerie: Strictly speaking these are not necessary, since `beard` already requires them. But the general principle I tried to follow in `setup.py` has always been "if it is explicitly imported, it must be part of `setup.py`".

jacquerie:
ci merge please

PR #3443, opened 2018-06-04
disambiguation: add distance estimator

jacquerie:
## Description:
Does not yet do what it says on the tin.

## Checklist:


inspirehep/modules/disambiguation/core/ml/models.py
```diff
@@ -69,5 +72,6 @@ def fit(self):
         ])
         self.estimator.fit(self.X, self.y)
 
+    @ensure_attributes(['estimator'], 'Run "fit" before "predict".')
```
jacquerie: @michamos, do you think this decorator is a stupid idea?

jacquerie: I thought about it, and I think it's a stupid idea.

PR #3444, opened 2018-06-04
records: add functionality check unlinked references

vbalbp:
Signed-off-by: Victor Balbuena <vbalbp@gmail.com>

## Checklist:

### DIFF
--- a/inspirehep/modules/records/checkers.py
+++ b/inspirehep/modules/records/checkers.py
@@ -1,0 +1,100 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2018 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+"""Records checkers."""
+
+from __future__ import absolute_import, division, print_function
+
+from collections import defaultdict
+
+from sqlalchemy import type_coerce
+from sqlalchemy.dialects.postgresql import JSONB
+
+from invenio_records.models import RecordMetadata
+from inspire_utils.record import get_value
+
+
+def increase_cited_count(result, identifier, core):
+    """Increases the number of times a reference with the same identifier has appeared"""
+    if core:
+        result[identifier] = (result[identifier][0] + 1, result[identifier][1])
+    else:
+        result[identifier] = (result[identifier][0], result[identifier][1] + 1)
+
+
+def calculate_score_of_reference(counted_reference):
+    """Given a tuple of the number of times cited by a core record and a non core record,
+    calculate a score associated with a reference.
+
+    The score is calculated giving five times more importance to core records"""
+    _, (core_count, non_core_count) = counted_reference
+    return core_count * 5 + non_core_count
+
+
+def order_dictionary_into_list(result_dict):
+    """Return ``result_dict`` as an ordered list of tuples"""
+    sorted_list = sorted(result_dict.items(), key=calculate_score_of_reference, reverse=True)
+
+    return sorted_list
+
+
+def get_all_unlinked_references():
+    """Return a list of dict, in which each dictionary corresponds to one reference object
+    and the status of core or non core"""
+    query = (
+        RecordMetadata.query
+        .filter(
+            type_coerce(RecordMetadata.json, JSONB)['_collections']
+            .contains(['Literature'])
+        )
+        .with_entities(RecordMetadata.json)
+    )
+
+    for record in query.yield_per(1000):
+        core = record.json.get('core')
+        for reference in record.json.get('references', []):
+            if 'record' not in reference:
+                yield {'core': core, 'reference': reference}
+
+
+def check_unlinked_references():
+    """Return two lists with the unlinked references that have a doi or an arxiv id.
+
+    If the reference read has a doi or an arxiv id, it is stored in the data structure.
+    Once all the data is read, it is ordered by most relevant to less relevant."""
+
+    result_doi, result_arxiv = defaultdict(lambda: (0, 0)), defaultdict(lambda: (0, 0))
+
+    data = get_all_unlinked_references()
+
+    for reference in data:
+        dois = get_value(reference, 'reference.reference.dois', [])
+        for doi in dois:
+            increase_cited_count(result_doi, doi, reference["core"])
+        arxiv_id = get_value(reference, 'reference.reference.arxiv_eprint', '')
+        if arxiv_id:
+            increase_cited_count(result_arxiv, arxiv_id, reference["core"])
+
+    result_doi = order_dictionary_into_list(result_doi)
+    result_arxiv = order_dictionary_into_list(result_arxiv)
+
+    return result_doi, result_arxiv
--- a/inspirehep/modules/records/cli.py
+++ b/inspirehep/modules/records/cli.py
@@ -1,0 +1,61 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2014-2018 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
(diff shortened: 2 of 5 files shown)


inspirehep/modules/records/cli.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@ref_wo_rec.command()
+@ref_wo_rec.argument('file_name',type=str)
+@ref_wo_rec.option('--doi','-d',isflag=True, default=False,
+				   help='Write a file only with references with DOI')
+@ref_wo_rec.option('--arxiv','-a',isflag=True, default=False,
+				   help='Write a file only with references with arxiv id')
+def write_references_on_file():
+	reference_list = obtain_references_without_record
```
michamos: is this supposed to be a function call? you're missing `()` then. You would have caught this with tests.

inspirehep/modules/records/cli.py
```diff
@@ -0,0 +1,33 @@
...
+	result_doi, result_arxiv = create_table_of_references_missing_record(data)
+
+	if doi:
+		with open('{0}_doi.txt'.format(file_name),'w') as f:
+			for item in result_doi:
+				f.write('{0}: {1}\n'.format(item[0].encode('utf-8'),str(item[1]).encode('utf-8')))
```
michamos: look at the `click` documentation for file handling, I am sure it can handle this kind of things in a better way.

inspirehep/modules/records/checkers.py
```diff
@@ -0,0 +1,135 @@
...
+	and writes two different files from it.
+
+	If the reference read has a doi or an arxiv id, it is stored in the data structure.
+	Once all the data is read, it is ordered by most relevant to less relevant
+	and written into two different files, one for dois and another for arxiv ids."""
+
+	result_doi, result_arxiv = {}, {}
+
+	for reference in data:
+		ref = json.loads(reference)
```
michamos: leftover?

inspirehep/modules/records/checkers.py
```diff
@@ -0,0 +1,135 @@
...
+		if core:
+			result[identifier] = (result[identifier][0] + 1, result[identifier][1])
+		else:
+			result[identifier] = (result[identifier][0], result[identifier][1] + 1)
+	else:
+		if core:
+			result[identifier] = (1, 0)
+		else:
+			result[identifier] = (0, 1)
+	return result
```
michamos: Don't `return` if you are mutating the `dict` in-place.

inspirehep/modules/records/checkers.py
```diff
@@ -0,0 +1,135 @@
...
+
+	If the reference read has a doi or an arxiv id, it is stored in the data structure.
+	Once all the data is read, it is ordered by most relevant to less relevant
+	and written into two different files, one for dois and another for arxiv ids."""
+
+	result_doi, result_arxiv = {}, {}
+
+	for reference in data:
+		ref = json.loads(reference)
+		if has_doi(ref):
```
michamos: it's better to do it like this:
```
dois = get_value(ref, 'reference.reference.dois', [])`
for doi in dois:
    increase_cited_count(result_doi, doi, ref['core'])
```

inspirehep/modules/records/checkers.py
```diff
@@ -0,0 +1,135 @@
...
+			doi = ref["reference"]["reference"]["dois"][0]
+			result_doi = increase_cited_count(result_doi, doi, ref["core"])
+		if has_arxiv(ref):
```
michamos: same.

inspirehep/modules/records/checkers.py
```diff
@@ -0,0 +1,92 @@
...
+    else:
+        result[identifier] = (result[identifier][0], result[identifier][1] + 1)
+
+
+def calculate_score_of_reference(counted_reference):
+    """Given a tuple of the number of times cited by a core record and a non core record,
+    calculate a score associated with a reference.
+
+    The score is calculated giving five times more importance to core records"""
+    return counted_reference[1][0] * 5 + counted_reference[1][1]
```
michamos: you can do for clarity
```python
_, (core_count, non_core_count) = counted_reference
return core_count*5 + non_core_count
```

inspirehep/modules/records/checkers.py
```diff
@@ -0,0 +1,92 @@
...
+
+    result_doi, result_arxiv = defaultdict(lambda: (0, 0)), defaultdict(lambda: (0, 0))
+
+    data = get_all_unlinked_references()
+
+    for reference in data:
+        dois = get_value(reference, 'reference.reference.dois', [])
+        for doi in dois:
+            increase_cited_count(result_doi, doi, reference["core"])
+        arxiv_id = get_value(reference, 'reference.reference.arxiv_eprint', [])
```
michamos: if the field is not a list but a string, it's better to use `''` as a default (if you need to call string operations on it later on), or not pass a default (so it will use `None`) if you only need to test for truthiness like here.

tests/integration/records/test_cli.py
```diff
@@ -0,0 +1,61 @@
...
+from __future__ import absolute_import, division, print_function
+
+from inspirehep.modules.records.cli import check
+
+
+def test_check_unlinked_references_generate_files(app_cli_runner, isolated_app, tmpdir):
+    doi_fd = tmpdir.join('doi_file')
+    arxiv_fd = tmpdir.join('arxiv_file')
+    result = app_cli_runner.invoke(check, ['unlinked_references', str(doi_fd), str(arxiv_fd)])
+    doi_fd_open = doi_fd.open('r')
```
michamos: you don't need to do this, you can call `readlines(cr=False)` directly on `doi_fd`, which gives you a list of all lines.

inspirehep/modules/records/checkers.py
```diff
@@ -0,0 +1,93 @@
...
+    """Return ``result_dict`` as an ordered list of tuples"""
+    sorted_list = sorted(result_dict.items(), key=calculate_score_of_reference, reverse=True)
+
+    return sorted_list
+
+
+def get_all_unlinked_references():
+    """Return a list of dict, in which each dictionary corresponds to one reference object
+    and the status of core or non core"""
+    query = RecordMetadata.query.filter(type_coerce(RecordMetadata.json, JSONB)['_collections'].contains(['Literature'])).with_entities(RecordMetadata.json)
```
michamos: can you split this line, it's a bit too long? you should try not to go too much above 100 as it becomes hard to read afterwards. One nice way to split it (that we don't currently use in inspire, but should IMHO) is to surround it with `()` as here: https://github.com/ambv/black#call-chains.

tests/integration/records/test_cli.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def test_check_unlinked_references_generate_files(app_cli_runner, isolated_app, tmpdir):
+    doi_fd = tmpdir.join('doi_file')
+    arxiv_fd = tmpdir.join('arxiv_file')
+    result = app_cli_runner.invoke(check, ['unlinked_references', str(doi_fd), str(arxiv_fd)])
+
+    assert result.exit_code == 0
+
+    result = doi_fd.readlines(cr=False)[0]
```
michamos: you should store the result of the file read in a variable, so you don't have to read the whole file every time.

inspirehep/modules/records/cli.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@click.group()
+def check():
+    """Commands to check different types of data"""
+
+
+@check.command()
+@click.argument('doi_file_name', type=click.File('w'), default='doi_file')
+@click.argument('arxiv_file_name', type=click.File('w'), default='arxiv_file')
```
michamos: can you make the defaults more explicit? something like `missing_cited_dois.txt` and `missing_cited_arxiv_eprints.txt`.

inspirehep/modules/records/cli.py
```diff
@@ -0,0 +1,48 @@
...
+@with_appcontext
+def unlinked_references(doi_file_name, arxiv_file_name):
```
michamos: Can you provide a meaningful docstring here? this will be shown to the user when doing `inspirehep check unlinked_references --help`.

inspirehep/modules/records/cli.py
```diff
@@ -0,0 +1,48 @@
...
+import click
+
+from flask.cli import with_appcontext
+
+from .checkers import check_unlinked_references
+
+
+@click.group()
+def check():
+    """Commands to check different types of data"""
```
michamos: `Commands to perform checks on records`

inspirehep/modules/records/cli.py
```diff
@@ -0,0 +1,55 @@
...
+@with_appcontext
+def unlinked_references(doi_file_name, arxiv_file_name):
+    """Find often cited literature that is not on INSPIRE.
+
+    It generates two files with a list of DOI/arxiv ids respectively,
+    in which each line has the respective identifier, folowed by two numbers,
+    representing the amount of times the literature has been cited
+    by a core and a non-core article respectively.
+    The lists are ordered by an internal measure of relevance."""
+    result_doi, result_arxiv = check_unlinked_references()
```
michamos: This is very slow because of the (almost) full table scan, so you could use a [spinner](https://github.com/click-contrib/click-spinner) (you need to add it to setup.py) and do:
```ipython
with spinner():
    click.echo('Looking up unlinked references...')
    result_doi, result_arxiv = check_unlinked_references()

click.echo('Done!')
click.echo(u'Output written to "{}" and "{}"'.format('doi_file_name', 'arxiv_file_name'))
```

vbalbp:
The task takes around 18 minutes to complete

PR #3449, opened 2018-06-06
disambiguation: add open_file_in_folder util

jacquerie:
## Description:
Implements the helper that was discussed in chat (https://inspirehep.zulipchat.com/#narrow/stream/122202-devel/subject/mkdir.20-p.20helper), and refactors part of the `disambiguation` module to use it.

## Checklist:


inspirehep/modules/disambiguation/config.py
```diff
@@ -25,16 +25,19 @@
 from __future__ import absolute_import, division, print_function
 
 
-DISAMBIGUATION_CLUSTERS_PATH = '/tmp/clusters.json'
+DISAMBIGUATION_BASE_PATH = '/var/data/disambiguation/'
```
david-caro: This will break during deployment, unless properly configured (overridden) or setup properly in puppet/similar. Not that this pr should do anything about it, just pointing it out.

david-caro: It should break in docker too imo :/ (test user should not be able to create a dir there by itself).

jacquerie: What I meant to do here is to use `var/data/disambiguation` like `invenio-records-files` uses `var/data/records` or `invenio-workflows-files` uses `var/data/workflows`. As for why this doesn't fail in the tests, I mock everything with `tmpdir`.

inspirehep/modules/disambiguation/config.py
```diff
@@ -25,16 +25,19 @@
...
+"""Base path to all files used in this module."""
+
+DISAMBIGUATION_CLUSTERS_PATH = DISAMBIGUATION_BASE_PATH + 'clusters.json'
```
salmanmaq: Trivial but, we might want to save the predicted clusters later as well. So we might want to rename this to DISAMBIGUATION_INPUT_CLUSTERS_PATH.

jacquerie: I don't disagree with this comment, but let's wait to do it until we absolutely need it. Please remind me if I forget!

jacquerie:
ci merge please

PR #3454, opened 2018-06-08
disambiguation: refactor to use JSON Lines

jacquerie:
## Description:
The main benefit of this refactor is that now `author_id` and `cluster_id` are independent, which means that we can represent in a uniform way
1. An input cluster that refers to an author.
2. An input cluster that does not refer to an author.
3. An output cluster.

This also prevents some needless copies that happen in the various disambiguation steps (for example: https://github.com/inspirehep/beard/blob/87b37a3028825b5199616b195663fee526c26be3/examples/applications/author-disambiguation/sampling.py#L177-L179).

## Checklist:

### DIFF
--- a/inspirehep/modules/disambiguation/api.py
+++ b/inspirehep/modules/disambiguation/api.py
@@ -27,6 +27,7 @@ from __future__ import absolute_import, division, print_function
 import json
 from collections import defaultdict
 
+import six
 from flask import current_app
 
 from inspirehep.modules.disambiguation.core.db.readers import (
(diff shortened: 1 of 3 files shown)


inspirehep/modules/disambiguation/ext.py
```diff
@@ -42,12 +42,12 @@ def init_config(self, app):
         disambiguation_base_path = os.path.join(app.instance_path, 'disambiguation')
 
         app.config['DISAMBIGUATION_BASE_PATH'] = disambiguation_base_path
-        app.config['DISAMBIGUATION_CLUSTERS_PATH'] = os.path.join(
-            disambiguation_base_path, 'clusters.json')
+        app.config['DISAMBIGUATION_CURATED_SIGNATURES_PATH'] = os.path.join(
+            disambiguation_base_path, 'signatures.jsonl')
+        app.config['DISAMBIGUATION_INITIAL_CLUSTERS_PATH'] = os.path.join(
```
jacquerie: Now that I think about it, these should be the `INPUT_CLUSTERS`, since we are later going to add the `OUTPUT_CLUSTERS`...

jacquerie: This also addresses this comment: https://github.com/inspirehep/inspire-next/pull/3449#discussion_r193377680.

jacquerie:
Blocked by https://github.com/inspirehep/inspire-next/pull/3456, as I don't want to add more stuff to the release.

inspirehep/modules/disambiguation/api.py
```diff
@@ -37,38 +38,50 @@
...
-    by ``BEARD`` as ground truth, while the latter contains one line per curated signature
-    present in INSPIRE.
+    Saves two files to disk called (by default) ``input_clusters.jsonl`` and
+    ``curated_signatures.jsonl``. The former contains one line per each cluster
+    initially present in INSPIRE, while the latter contains one line per each
+    curated signature that will be used as ground truth by ``BEARD``.
     """
-    clusters = defaultdict(list)
+    signatures_with_author = defaultdict(list)
+    signatures_without_author = []
```
michamos: stupid question: do you really need these two different structures, that then get serialized in the same way? can't you simply assign to the key `None` when there is no author?

jacquerie: But won't this create a single cluster with all signatures without an author smooshed together, while we want each to form a cluster on their own?

michamos: good point. So it was stupid after all :)

jacquerie: I guess we can move the `if` later, and iterate while writing to the file if `author_id` is `None`...

inspirehep/modules/disambiguation/api.py
```diff
@@ -37,38 +38,50 @@
...
+                signatures_without_author.append(signature['signature_uuid'])
+
+    with open_file_in_folder(current_app.config['DISAMBIGUATION_INPUT_CLUSTERS_PATH'], 'w') as fd:
+        for cluster_id, (author_id, signature_uuids) in enumerate(six.iteritems(signatures_with_author)):
+            fd.write(json.dumps({
+                'author_id': author_id,
+                'cluster_id': cluster_id,
+                'signature_uuids': signature_uuids,
+            }) + '\n')
+        for cluster_id, signature_uuid in enumerate(signatures_without_author, cluster_id + 1):
```
jacquerie: ...because I agree with you that it's unpleasant that I leverage the fact that `cluster_id` escapes the scope of the `for` loop, and I dislike that I'm using the not-so-well-known two argument form of `enumerate` !

tests/integration/disambiguation/test_disambiguation_api.py
```diff
@@ -43,26 +43,34 @@
...
+            'signature_uuids': input_cluster['signature_uuids'],
+        } for input_cluster in input_clusters
+    }  # XXX: so that the assertion doesn't depend on cluster_id.
 
-    assert '1010819' in clusters
-    assert ['94f560d2-6791-43ec-a379-d3dc4ad0ceb7'] == clusters['1010819']
+    assert {
+        'author_id': 1010819,
+        'signature_uuids': ['94f560d2-6791-43ec-a379-d3dc4ad0ceb7'],
+    } in reversed_input_clusters.values()
```
michamos: why not simply `assert any(cluster for cluster in input_cluster if 'author_id' == 1010819 and 'signature_uuids == ['94f560d2-6791-43ec-a379-d3dc4ad0ceb7'])`?

jacquerie: I actually prefer my assertion here because I don't think `py.test` would be able to provide any useful error message when presented with an `any`... Also, if you `--pdb` in my assertion when it fails you have some state (the `reversed_input_clusters` variable) to inspect.

jacquerie:
ci merge please

tests/integration/disambiguation/test_disambiguation_api.py
michamos: if you `--pdb` you can inspect `input_clusters` in any case, but OK.

PR #3455, opened 2018-06-08
[WIP] disambiguation: add distance estimator

salmanmaq:
This module should allow training of the distance estimator using the trained ethnicity estimator and sampled signature pairs.

Signed-off-by: Salman Maqbool salman.maqbool@cern.ch

## Checklist:


inspirehep/modules/disambiguation/api.py
```diff
@@ -34,6 +34,7 @@
     get_all_publications,
 )
 from inspirehep.modules.disambiguation.core.ml.models import EthnicityEstimator
+from inspirehep.modules.disambiguation.core.ml.models import DistanceEstimator
```
jacquerie: The `inspire-next` project does not follow `beard`'s convention for imports, here we prefer to group them by module and we sort them in alphabetical order:
```python
from inspirehep.modules.disambiguation.core.ml.models import (
    DistanceEstimator,
    EthnicityEstimator,
)
```

inspirehep/modules/disambiguation/core/ml/models.py
```diff
@@ -25,14 +25,40 @@
...
+from inspirehep.modules.disambiguation.core.ml.utils import get_second_given_name
+from inspirehep.modules.disambiguation.core.ml.utils import get_second_initial
+from inspirehep.modules.disambiguation.core.ml.utils import get_title
+from inspirehep.modules.disambiguation.core.ml.utils import get_abstract
+from inspirehep.modules.disambiguation.core.ml.utils import get_coauthors_from_range
+from inspirehep.modules.disambiguation.core.ml.utils import get_keywords
+from inspirehep.modules.disambiguation.core.ml.utils import get_collaborations
+from inspirehep.modules.disambiguation.core.ml.utils import get_topics
+from inspirehep.modules.disambiguation.core.ml.utils import group_by_signature
+from inspirehep.modules.disambiguation.core.ml.utils import load_signatures
```
jacquerie: https://github.com/inspirehep/inspire-next/pull/3455/files#r194333002 applies as well in this file.

inspirehep/modules/disambiguation/ext.py
```diff
@@ -52,6 +52,10 @@ def init_config(self, app):
             disambiguation_base_path, 'ethnicity.csv')
         app.config['DISAMBIGUATION_ETHNICITY_MODEL_PATH'] = os.path.join(
             disambiguation_base_path, 'ethnicity.pkl')
+        app.config['DISAMBIGUATION_SAMPLED_PAIRS_PATH'] = os.path.join(
+            disambiguation_base_path, 'pairs.json')
+        app.config['DISAMBIGUATION_DISTANCE_MODEL_PATH'] = os.path.join(
+            disambiguation_base_path, 'linkage.dat')
```
jacquerie: Let's be consistent with the naming (and deviate from the paper) and call it `distance.pkl`.

inspirehep/modules/disambiguation/core/ml/utils.py
```diff
@@ -0,0 +1,420 @@
...
+        Signature
+
+    Returns
+    -------
+    :returns: int
+        Year of publication if present on the signature, -1 otherwise
+    """
+    v = s["publication"]["year"]
+    v = int(v) if v else -1
+    return v
```
jacquerie: ...and neither this one.

inspirehep/modules/disambiguation/api.py
```diff
@@ -77,3 +78,13 @@ def train_and_save_ethnicity_model():
     estimator.load_data(current_app.config['DISAMBIGUATION_ETHNICITY_DATA_PATH'])
     estimator.fit()
     estimator.save_model(current_app.config['DISAMBIGUATION_ETHNICITY_MODEL_PATH'])
+
+
+def train_and_save_distance_estimator():
+    """Train the distance estimator model and save it to disk."""
+    estimator = DistanceEstimator(current_app.config['DISAMBIGUATION_DISTANCE_MODEL_PATH'])
```
jacquerie: This doesn't sound correct, given https://github.com/inspirehep/inspire-next/pull/3455/files#diff-4164879ce92c4a6f3a961ac28e1b084aR110 !

tests/integration/disambiguation/test_disambiguation_api.py
```diff
@@ -33,15 +34,59 @@
...
+    },
+    {
+        'signature_uuid': '2',
+        'publication_id': 2,
+        'author_name': 'Litov, Leander',
+        'author_id': 2,
+        'author_affiliation': 'Another Made Up Affiliation',
+        'signature_block': 'LATAVl'
+    }
+]
```
jacquerie: It won't hurt to use real data fixtures. In particular what worries me is that what is a `signature_uuid` here is simply a string, which is not a problem now, but might be.

For the ethnicity data I used fake data because we don't want to leak that data set which isn't ours, but for this one there's no problem.

inspirehep/modules/disambiguation/api.py
salmanmaq: Yup, silly mistake there.

inspirehep/modules/disambiguation/core/ml/utils.py
```diff
@@ -0,0 +1,420 @@
...
+    if isinstance(signatures, list):
+        signatures = {s["signature_uuid"]: s for s in signatures}
+
+    if isinstance(records, list):
+        records = {r["publication_id"]: r for r in records}
+
+    for signature_uuid, signature in signatures.items():
+        signature["publication"] = records[str(signature["publication_id"])]
+
+    return signatures, records
```
salmanmaq: I think we can also return just the signatures here. I don't see the records being used after this point.

inspirehep/modules/disambiguation/core/ml/models.py
```diff
@@ -25,14 +25,40 @@
...
 from sklearn.feature_extraction.text import TfidfVectorizer
+from sklearn.pipeline import FeatureUnion
 from sklearn.pipeline import Pipeline
 from sklearn.svm import LinearSVC
 
+from beard.similarity import CosineSimilarity
+from beard.similarity import ElementMultiplication
+from beard.similarity import EstimatorTransformer
+from beard.similarity import PairTransformer
+from beard.similarity import StringDistance
```
michamos: you can also do
```
from beard.similarity import (CosineSimilarity,
                              ElementMultiplication,
                              #...
                              StringDistance)
```
and so on to avoid the wall of text.

inspirehep/modules/disambiguation/core/ml/models.py
```diff
@@ -76,3 +102,161 @@ def fit(self):
...
+
+
+class DistanceEstimator(object):
+    def __init__(self, ethnicity_estimator_filename):
+        self.ethnicity_estimator = EthnicityEstimator()
+        self.ethnicity_estimator.load_model(ethnicity_estimator_filename)
+
+    def load_data(self, signatures_filename, records_filename, pairs_filename):
+        signatures, records = load_signatures(signatures_filename, records_filename)
+        pairs = json.load(open(pairs_filename, 'r'))
```
jacquerie: This is no longer correct, because the sampled pairs file is now a `.jsonl` file.

inspirehep/modules/disambiguation/core/ml/models.py
```diff
@@ -76,3 +102,161 @@ def fit(self):
...
+
+        self.X = np.empty((len(pairs), 2), dtype=np.object)
+        self.y = np.empty(len(pairs), dtype=np.int)
+
+        for k, (i, j, target) in enumerate(pairs):
+            self.X[k, 0] = signatures[i]
+            self.X[k, 1] = signatures[j]
+            self.y[k] = target
```
jacquerie: This is no longer correct, because each line of the `.jsonl` file now looks like this:
```json
{"same_cluster": True, "signature_uuids": ["cbf081db-fcb7-4386-baaf-f30636debfa7", "d08e1eb7-fa1b-4ea0-8917-5b3de969c582"]}
```

inspirehep/modules/disambiguation/core/ml/utils.py
```diff
@@ -0,0 +1,369 @@
...
+    records = json.load(open(records_filename, "r"))
+
+    if isinstance(signatures, list):
+        signatures = {s["signature_uuid"]: s for s in signatures}
+
+    if isinstance(records, list):
+        records = {r["publication_id"]: r for r in records}
+
+    for signature_uuid, signature in signatures.items():
+        signature["publication"] = records[str(signature["publication_id"])]
```
jacquerie: This eager join requires a lot of memory: the same `publication` is repeated for each signature, and each `publication` contains a list of all the author names, so a record with lots of signatures explodes quadratically in memory use!

I think that here we should prefer a lazy join.

inspirehep/modules/disambiguation/core/ml/utils.py
jacquerie: This is true, but we can avoid the problem altogether by pulling the util inside of the `load_data` of the model class.

inspirehep/modules/disambiguation/core/ml/utils.py
```diff
@@ -0,0 +1,369 @@
...
+        Signature
+
+    Returns
+    -------
+    :returns: string
+        Title of the publication
+    """
+    v = s["publication"]["title"]
+    v = v if v else ""
+    return v
```
jacquerie: Logic like this can be rewritten as `get_value(signature, 'publication.title', default='')`, but this assumes the eager join that I'd like to eliminate for the reason explained in https://github.com/inspirehep/inspire-next/pull/3455/files#r195340445.

inspirehep/modules/disambiguation/core/ml/utils.py
```diff
@@ -0,0 +1,369 @@
...
+    Returns
+    -------
+    :returns: tuple
+        Signatures and records. Both are lists of dictionaries.
+    """
+
+    signatures = []
+    for line in open(signatures_filename, "r"):
+        signatures.append(json.loads(line))
+    records = json.load(open(records_filename, "r"))
```
jacquerie: This is no longer correct, because the publications file is now a `.jsonl` file.

inspirehep/modules/disambiguation/core/ml/utils.py
```diff
@@ -0,0 +1,369 @@
...
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction
+
+"""Helpers for author disambiguation.
+
+.. codeauthor:: Gilles Louppe <g.louppe@cern.ch>
+.. codeauthor:: Mateusz Susik <mateusz.susik@cern.ch>
```
jacquerie: I'm going to remove this, but let's find a way to credit the original authors: even if the code has been substantially rewritten, most of the ideas belong to them.

inspirehep/modules/disambiguation/core/ml/models.py
```diff
@@ -76,3 +101,174 @@ def fit(self):
 
     def predict(self, X):
         return self.estimator.predict(X)
+
+
+class DistanceEstimator(object):
+    def __init__(self, ethnicity_model_path):
+        self.ethnicity_estimator = EthnicityEstimator()
+        self.ethnicity_estimator.load_model(ethnicity_model_path)
```
jacquerie: Come to think of it, why are we doing this? The constructor of `DistanceEstimator` should get an instance of an `EthnicityEstimator`, not know how to build one! It's the responsibility of the caller to know how to instantiate `EthnicityEstimator`.

inspirehep/modules/disambiguation/core/ml/sampling.py
```diff
@@ -113,12 +114,14 @@ def sample_signature_pairs(signatures_path, clusters_path, pairs_size):
...
+        category for category in [
+            same_cluster_same_name,
+            same_cluster_different_name,
+            different_cluster_same_name,
+            different_cluster_different_name,
+        ] if category
+    ]
+    for _, category in itertools.izip(xrange(pairs_size), itertools.cycle(non_empty_categories)):
+        for pair in np.random.choice(category, size=1):
+            yield pair
```
michamos: why not simply?
```python
for category in non_empty_categories:
    for pair in np.random.choice(category, replace=True, size=(pairs_size // len(non_empty_categories))):
        yield pair
```
This seems reasonable, but less efficient (`pairs_size` calls to the `np.random.choice` function vs `4`) and might be a bit harder to understand.

inspirehep/modules/disambiguation/core/ml/models.py
```diff
@@ -76,3 +101,174 @@ def fit(self):
...
+
+    def load_data(self, signatures_path, pairs_path, pairs_size, publications_path):
+        reversed_publications = {}
```
michamos: I know @jacquerie has used this terminology earlier, but it's not a priori clear in which sense this is reversed. I would call this `publications_by_id` and similarly `signatures_by_uuid`, etc.

inspirehep/modules/disambiguation/core/ml/models.py
```diff
@@ -76,3 +101,174 @@ def fit(self):
...
+                ('combiner', ElementMultiplication())
+            ]))
+        ])
+        classifier = RandomForestClassifier(n_estimators=500, n_jobs=8)
+
+        self.distance_estimator = Pipeline([('transformer', transformer), ('classifier', classifier)])
+        self.distance_estimator.fit(self.X, self.y)
+
+    def predict(self, X):
+        return self.distance_estimator.predict(X)
```
michamos: General comment: I would be more explicit about the state of the model here.
On general grounds, you want to either `fit` or `predict`. In the first case, you need to `load_data` before and `save_model` after. In the second case, you need to `load_model` and call `predict`. However, none of this is enforced by the class, you can happily do nonsense combinations like `predict` without `load_model`. It might make sense to enforce this doesn't happen.

inspirehep/modules/disambiguation/core/ml/utils.py
```diff
@@ -0,0 +1,330 @@
...
+
+    Parameters
+    ----------
+    :param s: dict
+        Signature
+
+    Returns
+    -------
+    :returns: string
+        Normalized author name
```
michamos: Use Google-style napoleon docstrings

inspirehep/modules/disambiguation/core/ml/utils.py
```diff
@@ -0,0 +1,330 @@
...
+    """
+    v = s["author_name"]
```
michamos: Is it clear that there's always a full name?

inspirehep/modules/disambiguation/core/ml/utils.py
```diff
@@ -0,0 +1,330 @@
...
+    v = normalize_name(v) if v else ""
```
michamos: What's the difference between this and the one from `inspire-utils` (not very urgent now, but something to think about for the future).

inspirehep/modules/disambiguation/core/ml/utils.py
```diff
@@ -0,0 +1,330 @@
...
+
+from beard.utils import (
+    given_name,
+    given_name_initial,
+    name_initials,
+    normalize_name,
+)
+
+
+def get_author_full_name(s):
```
michamos: one-letter parameters are bad. What is `s`?

inspirehep/modules/disambiguation/core/ml/sampling.py
jacquerie: That's an excellent idea, although it requires `pairs_size` to be a multiple of 12 (the LCM of 1, 2, 3, and 4, the number of possible non-empty categories) to make sure that we will always have that many samples.

inspirehep/modules/disambiguation/core/ml/utils.py
salmanmaq: The name could, for example, be Maqbool S., but in that case, that would be considered the full name. The algorithm works with that.

inspirehep/modules/disambiguation/core/ml/utils.py
salmanmaq: Should be signature. But yes, utils needs an overall refactoring.

inspirehep/modules/disambiguation/core/ml/models.py
jacquerie: Good idea. Will amend the current PR and add a commit to fix the previous uses.

inspirehep/modules/disambiguation/core/ml/models.py
jacquerie: This is the idea I discarded in https://github.com/inspirehep/inspire-next/pull/3443... I preferred going with an API method such as https://github.com/inspirehep/inspire-next/blob/87a00d2503efb490b16c992671eeb92ac4410938/inspirehep/modules/disambiguation/api.py#L103-L108 that enforces a certain order rather than enforce it at the decorator level. If you think it was a good idea after all I can resurrect it!

inspirehep/modules/disambiguation/core/ml/models.py
jacquerie: This happened in https://github.com/inspirehep/inspire-next/pull/3455/commits/c9c6e5917c7412c56eeadc55d28159bb65cd977c.

inspirehep/modules/disambiguation/core/ml/sampling.py
jacquerie: This happened in https://github.com/inspirehep/inspire-next/pull/3455/commits/62548f927675f0fd357d00612a42caf12b8790ed.

inspirehep/modules/disambiguation/core/ml/utils.py
jacquerie: To answer @michamos' original question: yes, because the signature's `author_name` comes from the authors' `full_name` in the record, whose presence is enforced by the schema: https://github.com/inspirehep/inspire-schemas/blob/a542b6efb21a361813ccc80d8a335407672767fd/inspire_schemas/records/hep.yml#L412

jacquerie: In particular this mapping is declared here: https://github.com/inspirehep/inspire-next/blob/87a00d2503efb490b16c992671eeb92ac4410938/inspirehep/modules/disambiguation/core/db/readers.py#L150.

jacquerie:
ci test please

jacquerie:
ci test please

jacquerie:
```
In [1]: from inspirehep.modules.disambiguation.api import train_and_save_ethnicity_model

In [2]: from inspirehep.modules.disambiguation.api import save_curated_signatures_and_input_clusters

In [3]: from inspirehep.modules.disambiguation.api import save_publications

In [4]: from inspirehep.modules.disambiguation.api import save_sampled_pairs

In [5]: from inspirehep.modules.disambiguation.api import train_and_save_distance_model

In [6]: %time train_and_save_ethnicity_model()
CPU times: user 3min 31s, sys: 1.24 s, total: 3min 32s
Wall time: 3min 34s

In [7]: %time save_curated_signatures_and_input_clusters()
CPU times: user 57.8 s, sys: 2.1 s, total: 59.9 s
Wall time: 1min 12s

In [8]: %time save_publications()
CPU times: user 47.1 s, sys: 2.79 s, total: 49.9 s
Wall time: 57.5 s

In [9]: %time save_sampled_pairs()
CPU times: user 47.4 s, sys: 2.04 s, total: 49.4 s
Wall time: 49.5 s
```

inspirehep/modules/disambiguation/core/ml/models.py
```diff
@@ -76,3 +93,286 @@ def fit(self):
...
+            ]))
+        ])
+        classifier = RandomForestClassifier(n_estimators=500, n_jobs=8)
+
+        self.distance_estimator = Pipeline([('transformer', transformer), ('classifier', classifier)])
+        self.distance_estimator.fit(self.X, self.y)
+
+
+def get_author_full_name(signature):
+    return normalize_name(signature['author_name'])
```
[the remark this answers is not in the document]
michamos: why would that be needed? you're pickling `self.distance_estimator`, not the whole instance.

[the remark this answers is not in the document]
jacquerie: Because the model would contain a reference to, say, the `self.group_by_signature` method, which `pickle` would not be able to serialize.

inspirehep/modules/disambiguation/config.py
```diff
@@ -25,4 +25,16 @@
 from __future__ import absolute_import, division, print_function
 
 
-DISAMBIGUATION_SAMPLED_PAIRS_SIZE = 1000000
+DISAMBIGUATION_SAMPLED_PAIRS_SIZE = 12 * 100000
```
salmanmaq: I am not sure why this is in a separate config file, and not ext.py.

jacquerie: Because it can be: it's actually bad that configuration has to live in `ext.py` just because they need to have access to an application variable in order to know their value.

inspirehep/modules/disambiguation/core/ml/models.py
```diff
@@ -65,14 +82,300 @@ def save_model(self, output_filename):
...
+                        ('name', FuncTransformer(func=get_author_full_name)),
+                        ('shaper', Shaper(newshape=(-1,))),
+                        ('classifier', EstimatorTransformer(self.ethnicity_estimator.estimator)),
+                    ]),
+                    groupby=group_by_signature,
+                )),
+                ('sigmoid', FuncTransformer(func=expit)),
+                ('shaper', Shaper(newshape=(-1, 2))),
+                ('combiner', ElementMultiplication()),
+                ('reshaper', Shaper(newshape=(-1, 7))),
```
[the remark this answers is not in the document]
jacquerie: This change did in fact become necessary because of the introduction of the previous `shaper`. I just tested locally that we can in fact get rid of both.

jacquerie:
ci merge please

jacquerie:
```
In [10]: %time train_and_save_distance_model()
CPU times: user 51min 3s, sys: 1min 49s, total: 52min 53s
Wall time: 27min 5s
```

inspirehep/modules/disambiguation/config.py
jacquerie: In fact, defaults for future configuration options that you are going to add (like the number of jobs to use during training or prediction) should probably be added here.

PR #3458, opened 2018-06-11
migrator view: filter out 'DELETED' invalid records

iulianav:
## Description

Do not return invalid records with collection 'DELETED' to the view.
Also, add a composite index on `valid` and `collection` columns of the
LegacyRecordsMirror table model and create an alembic recipe for it.

## Related Issue

https://its.cern.ch/jira/browse/INSPIR-867
https://its.cern.ch/jira/browse/INSPIR-868

## Motivation and Context

- We are not interested in checking errors for records that have been deleted.
- Indexing on 'collection' and 'valid' will increase the performance of the query.

## Checklist:

### DIFF
--- a/inspirehep/alembic/2f5368ff6d20_update_indices_on_legacy_records_mirror_table.py
+++ b/inspirehep/alembic/2f5368ff6d20_update_indices_on_legacy_records_mirror_table.py
@@ -1,0 +1,58 @@
+#
+# This file is part of Invenio.
+# Copyright (C) 2016-2018 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Add index on legacy_records_mirror"""
+
+from __future__ import absolute_import, division, print_function
+
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '2f5368ff6d20'
+down_revision = 'eaab22c59b89'
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    op.create_index(
+        'ix_legacy_records_mirror_valid_collection',
+        'legacy_records_mirror',
+        ['valid', 'collection'],
+    )
+
+    op.drop_index(
+        'ix_legacy_records_mirror_valid',
+        'legacy_records_mirror',
+    )
+
+    op.drop_index(
+        'ix_legacy_records_mirror_recid',
+        'legacy_records_mirror',
+    )
+
+
+def downgrade():
+    """Downgrade database."""
+    op.drop_index(
+        'ix_legacy_records_mirror_valid_collection',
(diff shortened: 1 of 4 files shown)


inspirehep/alembic/2f5368ff6d20_add_index_on_legacy_records_mirror_table.py
```diff
@@ -0,0 +1,36 @@
...
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    op.create_index(
+        'ix_legacy_records_mirror_collection_valid',
+        'legacy_records_mirror',
+        ['collection', 'valid'],
```
michamos: it's probably more useful to have the index in the opposite direction, that is on `('valid', 'collection')`, so that it can be used instead of the current index on `valid` (that you can remove, as it would do double duty with this one).

You can think about an index as something like a phone book (if you know what a phonebook is). That is an index on `(last name, first name)`, and can be used to obtain quickly the phone number of a person if you know either:
* the last name
* the last name and the first name
but it's of no use if you know only the first name.

Here, we want something that can be used to get quickly invalid records, or invalid records and their collection, but we don't need the collection of a record without its validity. That's why the other order is better.

iulianav: okay! Thanks!

inspirehep/modules/migrator/models.py
```diff
@@ -38,6 +38,10 @@
 class LegacyRecordsMirror(db.Model):
     __tablename__ = 'legacy_records_mirror'
 
+    __table_args__ = (
+        db.Index('ix_legacy_records_mirror_valid_collection', 'valid', 'collection'),
+    )
+
     recid = db.Column(db.Integer, primary_key=True, index=True)
     last_updated = db.Column(db.DateTime, default=datetime.utcnow, nullable=False, index=True)
     _marcxml = db.Column('marcxml', db.LargeBinary, nullable=False)
```
michamos: as I was saying, the whole point of this is that you can now drop the index in `valid` in the next line, because it performs part of the same function as the new index you're adding. You also need to add a migration to remove/add that index on upgrade/downgrade.

drjova:
ci merge please

inspirehep/alembic/2f5368ff6d20_update_indices_on_legacy_records_mirror_table.py
```diff
@@ -0,0 +1,58 @@
...
+
+    op.drop_index(
+        'ix_legacy_records_mirror_valid',
+        'legacy_records_mirror',
+    )
+
+    op.drop_index(
+        'ix_legacy_records_mirror_recid',
+        'legacy_records_mirror',
+    )
```
michamos: wait, why did you remove this?

PR #3459, opened 2018-06-12
792 testlib: add mitmproxy scenario recording client

szymonlopaciuk:
## Description
Adds a client function for recording new interactions and a helper in the form of a decorator to abstract E2E test scenario management.

## Motivation and Context
Makes it easier to use the mitmproxy for E2E tests in inspire-next. To record interactions during a test:

```python
from inspirehep.testlib.api.mitm_client import with_mitmproxy

@with_mitmproxy(should_record=True)
def test_get_record_from_legacy():
    response = requests.get('http://inspirehep.net/record/123456')
    response.raise_for_status()
```

Bear in mind that the service has to still be registered in the mitmproxy to allow for recording (functionality to do it dynamically is coming soon, for now we have Arxiv, Legacy and RT). After the interactions are recorded you can run your test like so, to avoid recording (`should_record` is `False` by default).

```python
from inspirehep.testlib.api.mitm_client import with_mitmproxy

@with_mitmproxy
def test_get_record_from_legacy():
    response = requests.get('http://inspirehep.net/record/123456')
    response.raise_for_status()
```

Default name for a scenario is the test name without leading `test_`, so in the above case the recording will be saved in `$SCENARIOS_PATH/get_record_from_legacy/LegacyService/interaction_0.yaml`. When using default setting in `docker-compose.yaml`, `"$SCENARIOS_PATH" == "/inspire-next/tests/e2e/scenarios/"`.

## Checklist:

### DIFF
--- a/inspirehep/testlib/api/mitm_client.py
+++ b/inspirehep/testlib/api/mitm_client.py
@@ -24,7 +24,63 @@
 
 from __future__ import absolute_import, division, print_function
 
+from decorator import decorate
+from os import environ
+
 from . import Session
+
+
+def with_mitmproxy(*args, **kwargs):
+    '''Decorator to abstract fixture recording and scenario setup for the E2E tests with mitmproxy.
+
+    Args:
+        scenario_name (Optional[str]): scenario name, by default test name without 'test_' prefix
+
+        should_record (Optional[bool]): is recording new interactions allowed during test run,
+            by default `False`
+
+        *args (List[Callable]): list of length of either zero or one: decorated function.
+            This is to allow the decorator to function both with and without calling it
+            with parameters: if args is present, we can deduce that the decorator was used
+            without parameters.
+
+    Returns:
+        Callable: a decorator the can be used both with and without calling brackets
+            (if all params should be default)
+    '''
+    scenario_name = kwargs.pop('scenario_name', None)
+    should_record = kwargs.pop('should_record', False)
+
+    if not args:
+        assert not kwargs, 'Parameters %s not expected' % kwargs
+
+    def _with_mitmproxy(func, *args, **kwargs):
+        def ommit_test_in_name(name):
+            if name.startswith('test_'):
+                return name[5:]
+
+        mitmproxy_url = environ.get('MITMPROXY_HOST', 'http://mitm-manager.local')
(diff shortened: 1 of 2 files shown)


szymonlopaciuk:
ci test please

inspirehep/testlib/api/mitm_client.py
```diff
@@ -24,18 +24,72 @@
...
+
 from . import Session
 
 
+def with_mitmproxy(*args, **kwargs):
+    '''Decorator to abstract fixture recording and scenario setup for the E2E tests with mitmproxy.
+
+    Args:
+        scenario_name (Optional[str]): scenario name, by default test name without 'test_' prefix
+        should_record (Optional[bool]): is recording new interactions allowed during test run?
```
david-caro: describe default

inspirehep/testlib/api/mitm_client.py
```diff
@@ -24,18 +24,72 @@
...
+        *args (List[Callable]): list of length of either zero or one: decorated function
```
david-caro: Add a hint on why

inspirehep/testlib/api/mitm_client.py
```diff
@@ -24,18 +24,72 @@
...
+
+
 class MITMClient(object):
     def __init__(self, proxy_host='http://mitm-manager.local'):
         self._client = Session(base_url=proxy_host)
 
     def set_scenario(self, scenario_name):
-        self._client.put('/config', json={'active_scenario': scenario_name})
+        resp = self._client.put('/config', json={'active_scenario': scenario_name})
+        resp.raise_for_status()
```
david-caro: check if we want to return the resp.json here too

szymonlopaciuk: `PUT /config` or `POST /config` don't return anything yet, and in fact PUT-ting triggers a `204 No Content` reply, so we might need to change that first (aand make the whole thing actual REST). In fact the response codes for `/record` endpoint are similar (despite returning data) so that an actual mistake in mitm-proxy which we should fix first.

szymonlopaciuk:
ci test please

szymonlopaciuk:
ci merge please

PR #3464, opened 2018-06-13
workflows: addition of copy local files

drjova:

### DIFF
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -32,6 +32,7 @@ import arrow
 from elasticsearch.exceptions import NotFoundError
 from flask import current_app
 from fs.opener import fsopen
+from six.moves.urllib.parse import urlparse, unquote
 
 from inspire_dojson.utils import get_recid_from_ref, strip_empty_values
 from inspire_schemas.api import validate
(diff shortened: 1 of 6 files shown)


inspirehep/modules/records/api.py
```diff
@@ -459,7 +460,8 @@ def _download_to_docs_or_figs(
         if key not in self.files:
             key = self._get_unique_files_key(base_file_name=key)
 
-        stream = fsopen(doc_or_fig_obj['url'], mode='rb')
+        url = unquote(doc_or_fig_obj['url'])
```
michamos: is it clear that you're dealing with a local file here? otherwise unquoting is wrong.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -381,11 +383,15 @@ def download_documents(obj, eng):
...
         filename = document['key']
         url = document['url']
-        downloaded = download_file_to_workflow(
-            workflow=obj,
-            name=filename,
-            url=url,
-        )
+        scheme = urlparse(url).scheme
+        if scheme == 'file':
+            downloaded = copy_file_to_workflow(obj, filename, url)
```
michamos: do we really need two different methods here? cannot you replace `requests` by `fsopen` in `download_file_to_workflow`?

drjova: I didn't want to touch any existing functions, the file handling should be refactor asap

david-caro: I was thinking the same too, that we don't need both anymore with fsopen. Not 100% sure if it's completely compatible though.

drjova: we have already logic for handling errors
https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/utils/__init__.py#L210-L230

IMO modifying that might have other side effects which we don't know, we don't reliable tests at the moment

david-caro:
ci merge please

PR #3465, opened 2018-06-13
records: fix transaction issue causing zomie pids

iulianav:
## Description

Encapsulate the record creation and minting statements found in `InspireRecord.create()` in a `with db.session.negin_nested():` SAVEPOINT. In case of an exception, the state of the DB will be rolled back to the savepoint, regardless of whether a PID or a record has already been created. Thus, now a PID cannot be created without an associated record. The opposite is also true.

## Related Issue

https://its.cern.ch/jira/browse/INSPIR-732

## Motivation and Context

At the moment, if the record creation fails because of errors like `ValidationError` , a PID is still created and stored for that record. Similarly, we had the opposite problem in the past: inspire records were saved with no associated PIDs when the minting process would fail because of errors like `PIDAlreadyExists`.

## Checklist:

### DIFF
--- a/tests/integration/records/test_records_api.py
+++ b/tests/integration/records/test_records_api.py
@@ -26,6 +26,7 @@ import os
 from tempfile import NamedTemporaryFile
 import pytest
 
+from invenio_pidstore.models import PersistentIdentifier, RecordIdentifier
 from jsonschema import ValidationError
 from six.moves.urllib.parse import quote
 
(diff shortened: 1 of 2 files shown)


tests/integration/records/test_records_api.py
```diff
@@ -72,3 +76,52 @@ def test_download_local_file(isolated_app):
...
+        InspireRecord.create(invalid_record)
+
+    record_identifier = RecordIdentifier.query.filter_by(recid=1936477).one_or_none()
+    persistent_identifier = PersistentIdentifier.query.filter_by(pid_value='1936477').one_or_none()
+
+    assert not record_identifier
+    assert not persistent_identifier
+
+
+def test_create_does_not_save_zombie_records_if_record_minting_fails(isolated_app):
```
drjova: actually we don't need this case because it's already covered by calling ``mint`` before ``.create``. If creation fails in ``PersistentIdentifier`` or ``RecordIdentifier`` will never create the record. WDYT?

iulianav: I was thinking of it more like a regression test just in case someone touches that part of the code again. What do you think?

drjova: hmm it's bit arbitrary, when you have a ``pid``  you should have a ``RecordIdentifier``. In this case, we assume that we don't have one which is a wrong assumption. IMO this test is not necessary since we are not testing any real case.

iulianav: ok!

drjova:
@iulianav could you please remove the test and rebase? thanks!

drjova:
ci merge please

PR #3498, opened 2018-06-19
workflows: save after article is halted

drjova:

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -504,7 +504,6 @@ def preserve_root(obj, eng):
         return
 
     obj.extra_data['merger_root'] = deepcopy(obj.data)
-    obj.save()
 
 
 @with_debug_logging


david-caro:
Why is this needed?

drjova:
submissions are not indexed after they are marked as "HALTED" hence are not displayed in the corresponding box in HP.

drjova:
so workflows will not finish running the rest of the current block? ```HALT_FOR_APPROVAL_IF_NEW_OR_STOP_IF_NOT_RELEVANT``` so this should be moved actually inside the ``halt_record``

drjova:
ok now I see https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/tasks/actions.py#L164 the engine is halted so will stop there

david-caro:
yep :)
But the wokrflows should be indexed on halt anyhow. Are you sure that it's not that they fail to index? (like with the wrong mapping and such, you can try to open a shell, get the workflow with `workflow_object_class.get(1234)` and hit `wflw.save()`)

drjova:
yep I double checked it, all the cases had the same issue and all have been resolved just by ``wflw.save()``

drjova:
it's easy to reproduce, just by submitting a new paper

david-caro:
That feels strange, should have been caught long time age :/, where do you run it? can you reproduce locally?

drjova:
The problem is older I think, so the verbose explanation from my side is:

When the submission starts, it doesn't have the engine (the issue we were investigation some time ago), when you halted if the engine doesn't exist and it adds it. In this stage the workflow has been indexed **without** ``"workflow_name: "HEP"`` that's why even if https://qa.inspirehep.net/api/holdingpen/1077005 is HALTED and contains:

```python
_workflow: {
  data_type: "hep",
  id_parent: null,
  id_user: 58736,
  id_workflow: "f92da654-73ad-11e8-8b04-fa163ea56ae2",
  status: "HALTED",
  workflow_class: "article",
  workflow_name: "HEP",
}
```

it doesn't appear in the results https://qa.inspirehep.net/holdingpen/list/?page=1&size=10&source=submitter&workflow_name=HEP&status=HALTED because we have in the filters: ```&workflow_name=HEP```

if I deselect the HEP it appears:
https://qa.inspirehep.net/holdingpen/list/?page=1&size=10&source=submitter&status=HALTED

david-caro:
How come that we have other submission workflows in halted? And this has not been seen anywhere else? That is very suspicious to me :/ points to an underlying issue. Also, I can't reproduce locally, can you?

drjova:
Yes, I can reproduce it locally. Do you have the merger enabled ```FEATURE_FLAG_ENABLE_MERGER = True```? 

The only difference is that we save the workflow before we halt it (in the ``preserve_root``). https://github.com/inspirehep/inspire-next/commit/f5f6818b05fd640a2d6999a42c8a32868eb91012#diff-16f31546a6d82e399e344031e67a88c6R276
It seems that it's unnecessary to save it there, since the workflow is handled by the engine, so I removed the save from there and let the workflow handle the object.

In general, it makes total sense for me, we don't have consistency how we start a workflow and the smallest change in the workflow order might have different results, for example on submission we don't even attaching the engine https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/literaturesuggest/views.py#L77-L101.

The underlying problem here is the engine, if the engine is attached from the beginning then we'll not have this issue, but as we tried the last time was too hard to fix (due to the failing e2e tests).

david-caro:
I still don't find any sense for it :/, I had the flag enabled I think yes, I'll retry

david-caro:
Actually, I'm seeing this also when running the ui tests for https://github.com/inspirehep/inspire-next/pull/3501

david-caro:
Locally, I see clearly the different info on the indexer and db, on the index:
```
"_workflow": {

    "status": "HALTED",
    "id_user": 1,
    "data_type": "hep",
    "workflow_position": [
        27
        ,
        3
        ,
        3
    ],
    "id_parent": null,
    "workflow_class": null,
    "id_workflow": null,
    "workflow_name": null

},
```

from db:
```
"_workflow": {
"data_type": "hep",
"id_parent": null,
"id_user": 1,
"id_workflow": "ca4bd592-73d8-11e8-8e20-0242ac180008",
"status": "HALTED",
"workflow_class": "article",
"workflow_name": "HEP",
"workflow_position": [
27,
3,
3
]
},
```

The history list is the same, so that happened in the same task, that is:
```
{
"doc": "Halt the workflow object, action=hep_approval, message=Submission halted for curator approval.",
"name": "halt_record",
"nicename": "\"Submission halted for curator approval.\"",
"parameters": [],
"time": "2018-06-19 15:52:46.092167"
}
```

drjova:
exactly that's the problem, ``preserve_root`` saves and has ``"workflow_name": null`` and then the engine is been attached and saves in the ``halt_record`` but it doesn't index. Do we have any debounce on the index requests?

david-caro:
The halt_record ends up triggering a reindex too, I added a tracedump in the reindex function:
```
    exception_handler(obj, self, callbacks, exc_info)                                
  File "/virtualenv/lib/python2.7/site-packages/invenio_workflows/engine.py", line 401, in HaltProcessing                               
    id_workflow=eng.uuid)                                        
  File "/virtualenv/lib/python2.7/site-packages/invenio_workflows/api.py", line 136, in save                                                                          
    workflow_object_after_save.send(self)                                                       
  File "/virtualenv/lib/python2.7/site-packages/blinker/base.py", line 267, in send                           
    for receiver in self.receivers_for(sender)]                                                     
  File "/virtualenv/lib/python2.7/site-packages/invenio_workflows_ui/receivers.py", line 44, in index_workflow_object
    workflow_api_class.create(sender)                                      
  File "/virtualenv/lib/python2.7/site-packages/invenio_workflows_ui/api.py", line 69, in wrapper
    self_or_cls.indexer.index(result)                             
  File "/virtualenv/lib/python2.7/site-packages/invenio_workflows_ui/indexer.py", line 61, in index
    import traceback; logging.getLogger(__name__).error(traceback.format_stack())  
```
and in that reindex, the workflow has no 'workflow_name' on it:
```
Indexing record {'_workflow': {'status': 'HALTED', 'id_user': 1, 'data_type': u'hep', 'workflow_position': [27, 3, 3], 'id_parent': None, 'workflow_class': None, '
id_workflow': None, 'workflow_name': None}, ...
```

david-caro:
Interestingly enough though, the save call that triggered those had a `wokrflow_id` passed to it:
```
Saving workflow <WorkflowObject(model = <WorkflowObjectModel(id = 1, id_workflow = None, status = ObjectStatus.INITIAL, id_parent = None, created = 2018-06-19 16:5
3:35.909980, )), params status:ObjectStatus.HALTED, callback_pos:[27, 3, 3], id_workflow:4eb1c848-73e1-11e8-935d-0242ac180008
```

david-caro:
A `db.session.commit()` right before the index works...

drjova:
which happens here https://github.com/inveniosoftware-contrib/invenio-workflows/blob/617e7dbad1b06767ad23d2ddef3569443778eea8/invenio_workflows/engine.py#L407

drjova:
did you try my pr? It works for me by removing the extra ``save()``

david-caro:
> which happens here https://github.com/inveniosoftware-contrib/invenio-workflows/blob/617e7dbad1b06767ad23d2ddef3569443778eea8/invenio_workflows/engine.py#L407

Yep, that's too late though, as the indexing happens before the `save` finishes (through a signal):
which happens here https://github.com/inveniosoftware-contrib/invenio-workflows/blob/617e7dbad1b06767ad23d2ddef3569443778eea8/invenio_workflows/engine.py#L407

> did you try my pr? It works for me by removing the extra save()

I'm not sure what code I'm testing anymore xd, I'll test with your pr tomorrow.

david-caro:
I think that there's some kind of unsynchronization between the model `workflow` property and the `id_workflow`...
https://github.com/inveniosoftware-contrib/invenio-workflows/blob/617e7dbad1b06767ad23d2ddef3569443778eea8/invenio_workflows/models.py#L208

as imo when the id_workflow is populated it should populate the workflow one too... maybe some misconfiguration of sqlalchemy relation...

drjova:
maybe the ``post_update`` thingy?

david-caro:
This works for me: https://github.com/inveniosoftware-contrib/invenio-workflows/pull/52

david-caro:
> maybe the post_update thingy?

Sounds unrelated to me, it seems to be about how to translate to db instructions, but not changing the overall behavior.

post_update¶ –

this indicates that the relationship should be handled by a second UPDATE statement after an INSERT or before a DELETE. Currently, it also will issue an UPDATE after the instance was UPDATEd as well, although this technically should be improved. This flag is used to handle saving bi-directional dependencies between two individual rows (i.e. each row references the other), where it would otherwise be impossible to INSERT or DELETE both rows fully since one row exists before the other. Use this flag when a particular mapping arrangement will incur two rows that are dependent on each other, such as a table that has a one-to-many relationship to a set of child rows, and also has a column that references a single child row within that list (i.e. both tables contain a foreign key to each other). If a flush operation returns an error that a “cyclical dependency” was detected, this is a cue that you might want to use post_update to “break” the cycle.

drjova:
> This works for me: inveniosoftware-contrib/invenio-workflows#52

is it without my pr? I'll test it locally now :)

The thing is, it doesn't solve the actual issue, which is workflows without workflow assigned.

david-caro:
Yep, that's by itself.

In this case, it's not that the workflow object has no workflow assigned, as it's being assigned right there.

Agree though that the issue that the submissions are created without a workflow is still there, but this was not it.

david-caro:
btw. just tested your pr and works too.

drjova:
> This works for me: inveniosoftware-contrib/invenio-workflows#52

Works for me too!

> Agree though that the issue that the submissions are created without a workflow is still there, but this was not it.

for me it's the same problem, if we were correctly assigning workflows from the beginning we wouldn't have this issue in the first place, even if the workflow relationship was hairy.

david-caro:
> it's the same problem, if we were correctly assigning workflows from the beginning we wouldn't have this issue in the first place, even if the workflow relationship was hairy.

This might as well happen on that first step, meaning, that with this issue, even if we create the workflow at the same time as the workflow object, it might be that it's not properly indexed until the next `save`. Unlikely, but this is what happens here.

drjova:
Supersede by https://github.com/inveniosoftware-contrib/invenio-workflows/pull/52

PR #3503, opened 2018-06-20
workflows: put the refextract timeout on subtasks

michamos:
Increasing the timeout to 10 min in
92d3166d84652520cf294aef430ae9c8296ea9de didn't solve the problem. This
might be because when there are lots of references, also reference
matching takes a long time, exceeding the allowed time. As we really
want to protect from refextract going wild and never finishing, rather
than ensuring the task runs in a fixed amount of time, it makes sense
not to include reference matching in the timeout.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -38,7 +38,6 @@ from sqlalchemy import (
     cast,
     type_coerce,
 )
-from timeout_decorator import timeout
 from werkzeug import secure_filename
 
 from invenio_db import db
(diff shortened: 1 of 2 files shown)


david-caro:
Shouldn't we still keep some upper timeout on the refextract task itself?

michamos:
Why? as I explained in the commit message, what we really want to prevent is `refextract` taking a very long time. If reference matching is slow, so be it, we know it runs in `O(len(references))` by design, and will finish eventually. We don't put timeouts on every step in the workflow, only on `refextract` and `plotextract`.

michamos:
> Shouldn't we still keep some upper timeout on the refextract task itself?

After private discussion,this was split off in a separate issue: https://its.cern.ch/jira/browse/INSPIR-927.

michamos:
ci merge please

PR #3506, opened 2018-06-21
846 orcid: do not re-cache all putcodes on new records 

turtle321:


turtle321:
ci test please

turtle321:
ci test please

turtle321:
ci test please

turtle321:
ci test please

turtle321:
ci test please

turtle321:
ci test please

inspirehep/modules/orcid/api.py
```diff
@@ -63,12 +70,19 @@ def push_record_with_orcid(recid, orcid, oauth_token, put_code=None, old_hash=No
...
 
-    new_hash = calculate_hash_for_record(record)
+    orcid_cache = OrcidCache(orcid)
+
+    if not putcode or not old_hash:
+        cached_putcode, cached_old_hash = orcid_cache.read_record_data(recid)
+        putcode = putcode or cached_putcode
+        old_hash = old_hash or cached_old_hash
+
+    new_hash = OrcidHasher().compute_hash(record)
```
szymonlopaciuk: Why not make it a class/static method?

inspirehep/modules/orcid/cache.py
```diff
@@ -0,0 +1,113 @@
...
+        Args:
+            recid(int): inspire record's id.
+        """
+        key = self._get_key(recid)
+        value = self.redis.hgetall(key)
+        return value.get('putcode'), value.get('hash')
+
+
+class OrcidHasher(object):
+    def compute_hash(self, record):
```
szymonlopaciuk: As mentioned above, why not make it a classmethod, `OrcidHasher` doesn't seem to have any state, no?

inspirehep/modules/orcid/cache.py
```diff
@@ -0,0 +1,113 @@
...
+        Args:
+            record (dict): HEP record
+
+        Returns:
+            string: hash of the record
+        """
+        orcid_rec = OrcidConverter(record, app.config['LEGACY_RECORD_URL_PATTERN'])
+        return self._hash_xml_element(orcid_rec.get_xml())
+
+    def _hash_xml_element(self, element):
```
szymonlopaciuk: And same here

inspirehep/modules/orcid/exceptions.py
```diff
@@ -0,0 +1,42 @@
...
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+
+class DuplicatedExternalIdentifiersError(Exception):
+    http_status_code = 409
+    orcid_error_code = 9021
```
szymonlopaciuk: Apologies for being the devil's advocate, but are we sure this will be constant? I don't remember any docs in regards to this, so maybe worth to add a source or something, so we know where it comes from? Although, I don't mind that much, I just remember the errors from ORCID sometimes seemed random...

tests/integration/helpers_test/test_helpers_utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_override_config(app):
+    with override_config(foo1='bar1', foo2='bar2'):
+        from flask import current_app
+        assert current_app.config['foo1'] == 'bar1'
+        assert current_app.config['foo2'] == 'bar2'
+
+    assert not hasattr(current_app.config, 'foo1')
+    assert not hasattr(current_app.config, 'foo2')
```
szymonlopaciuk: If there was a `leak` this doesn't detect it, I think you meant to use `not in`:
```python
In [1]: config = {'foo1'}

In [2]: not hasattr(config, 'foo1')
Out[2]: True

In [3]: not hasattr(config, 'not there')
Out[3]: True

In [4]: 'foo1' not in config
Out[4]: False

In [5]: 'not there' not in config
Out[5]: True
```
Also maybe it would be worth to check if the stuff that was in the config before is also available when overriding?

tests/integration/orcid/test_orcid_api.py
```diff
@@ -27,7 +27,175 @@
...
+                old_hash=None,
+            )
+
+        assert putcode == self.putcode
+        assert hash_value == self.hash_value
+        mock_recache_author_putcodes.assert_not_called()
+        assert vcr_cassette.all_played
+
+    @pytest.mark.vcr()
+    def test_new_record_alredy_existent_error(self, vcr_cassette):
```
szymonlopaciuk: Nit: 'already'

tests/integration/orcid/test_orcid_api.py
```diff
@@ -27,7 +27,175 @@
...
+                old_hash=None,
+            )
+
+        assert putcode == self.putcode
+        assert hash_value == self.hash_value
+        mock_recache_author_putcodes.assert_called_with(self.orcid, self.oauth_token)
+        assert vcr_cassette.all_played
+
+    @pytest.mark.vcr()
+    def test_new_record_alredy_existent_error_and_putcode_not_found(self, vcr_cassette):
```
szymonlopaciuk: Same here :smiley:

inspirehep/modules/orcid/exceptions.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @classmethod
+    def match(cls, exception):
```
michamos: maybe `__eq__`, so you can do `Exception == DuplicatedExternalIdentifiersError`?

inspirehep/modules/orcid/cache.py
michamos: it needs `self` to find the other method it's calling. I agree on your distate of this style though :)

inspirehep/modules/orcid/api.py
michamos: I think I would make the hashing part of the `OrcidCache`, which you initialize with `orcid` and `record`, and acts like an oracle that tells you whether to `post`, `put` or do nothing.

tests/integration/helpers/utils.py
```diff
@@ -77,3 +78,21 @@ def _create_record(record_json):
     es.indices.refresh()
 
     return record_json
+
+
+def override_config(**kwargs):
```
michamos: nice!

inspirehep/modules/orcid/cache.py
turtle321: It is a class as a mean of grouping relate methods. In particular it is a simple template design pattern, often implemented in Python the way I did, see this f.e.:
https://sourcemaking.com/design_patterns/template_method/python/1

`compute_hash()` can be a class method yes, but I prefer to keep it as instance method with a classic signature

tests/integration/helpers_test/test_helpers_utils.py
turtle321: Oh yes, nice points both of them, I'll do that. I overlooked `current_app.config` and thought it was an object instead of a dict.

tests/integration/orcid/test_orcid_api.py
turtle321: 👍

tests/integration/orcid/test_orcid_api.py
turtle321: 👍

inspirehep/modules/orcid/api.py
```diff
@@ -36,23 +36,30 @@
...
-def push_record_with_orcid(recid, orcid, oauth_token, put_code=None, old_hash=None):
+def push_record_with_orcid(recid, orcid, oauth_token, putcode=None, old_hash=None):
     """Push record to ORCID with a specific ORCID ID.
 
     Args:
         recid (string): HEP record to push
         orcid (string): ORCID identifier to push onto
         oauth_token (string): ORCID user OAUTH token
-        put_code (Union[string, NoneType]): put-code to push record onto,
+        putcode (Union[string, NoneType]): put-code to push record onto,
```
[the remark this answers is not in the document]
turtle321: 👍

inspirehep/modules/orcid/api.py
```diff
@@ -63,12 +70,19 @@ def push_record_with_orcid(recid, orcid, oauth_token, put_code=None, old_hash=No
     """
     record = get_db_record('lit', recid)
 
-    new_hash = calculate_hash_for_record(record)
+    orcid_cache = OrcidCache(orcid)
```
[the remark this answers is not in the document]
turtle321: Nice idea, I'll move this to module level.

inspirehep/modules/orcid/api.py
turtle321: I like your idea of hiding the hasher inside the OrcidCache either by encapsulation or association. This way the code that uses OrcidCache only knows about the concept of caching and not hashing. I will try this.

inspirehep/modules/orcid/api.py
```diff
@@ -36,23 +36,30 @@
...
-    hash_xml_element,
     log_time_context,
 )
 from inspirehep.modules.records.serializers import bibtex_v1
 from inspirehep.utils.record_getter import get_db_record
 
+from .cache import OrcidCache, OrcidHasher
+from .exceptions import (
+    DuplicatedExternalIdentifiersError,
+    PutcodeNotFoundInCacheException,
```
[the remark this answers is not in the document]
turtle321: Naming convention:
suffix Exception: if the exc is used as control flow
suffix Error: if it is an actual error
https://www.python.org/dev/peps/pep-0008/#exception-names

inspirehep/modules/orcid/api.py
turtle321: What's the benefit of that? I don't have a strong opinion here, but the fact that this can be a class method does not mean that it has to. And for regular use cases, I tend to avoid class/static methods such that the class has a very common signature.
By very common signature I mean:
```
ferrari = Car()
ferrari.start()
```

By non regular use cases, I mean f.i. this: https://github.com/inspirehep/inspire-next/pull/3506/commits/ac2963af9a641b5f6e843fd27c91ccb23aa75f77#r197426458
where I want to be able to do a `DuplicatedExternalIdentifiersError.match(myexc)` because I am interested in matching with the exception class and I do not want to create an instance.

inspirehep/modules/orcid/exceptions.py
turtle321: I asked myself the same question and hesitated a bit. But I asked ORCID engineers and they confirmed that error codes should not change over time in the same API version.

szymonlopaciuk: :+1:

turtle321:
ci test please

tests/integration/helpers/factories/db/invenio_records.py
```diff
@@ -95,6 +99,10 @@ def create_from_kwargs(cls, index_name='', **kwargs):
                     object_uuid=instance.record_metadata.id,
                     pid_value=instance.record_metadata.json.get('control_number'),
                     **kwargs).persistent_identifier
+
+        instance.inspire_record = InspireRecord(instance.record_metadata.json,
```
turtle321: Now you can can get the `InspireRecord` model instance directly from the `TestRecordMetadata` factory. @drjova

turtle321:
ci test please

turtle321:
ci test please

turtle321:
ci test please

inspirehep/modules/orcid/api.py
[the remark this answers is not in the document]
turtle321: Done (actually in last version I removed the putcode from the signature)

inspirehep/modules/orcid/api.py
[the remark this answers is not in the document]
turtle321: Done, with a trick

inspirehep/modules/orcid/exceptions.py
turtle321: We discussed this

tests/integration/helpers_test/test_helpers_utils.py
turtle321: Done

turtle321:
@michamos @szymonlopaciuk all comments addressed.
I am sorry about the size of this PR, I started small but eventually it got huge.
For a smooth review I suggest to focus on 2 abstractions: `OrcidCache` and `_OrcidHasher`.

You can skip the other code, especially the one in `modules/orcid.api.py` as I am going to touch that in my next PR: the plan is to split the business logic from the plain orcid-api-client interaction. So you will get a second chance to review that part.

turtle321:
ci test please

inspirehep/modules/orcid/cache.py
```diff
@@ -31,69 +31,114 @@
...
+            _init_redis_client()
+        self.redis = _redis_client
 
-    def _get_key(self, recid):
+    def get_key(self, recid):
         """Return the string 'orcidcache:``orcid_value``:``recid``'"""
         return 'orcidcache:{}:{}'.format(self.orcid, recid)
 
-    def write_record_data(self, recid, putcode, hash_value=None):
+    def write_work_putcode(self, recid, putcode, inspire_record=None):
```
michamos: this doesn't only store the putcode. What about `write_work_metadata`?

inspirehep/modules/orcid/cache.py
```diff
@@ -31,69 +31,114 @@
...
+            self.read_work_putcode(recid)
+        if not self._new_hash_value:
+            self._new_hash_value = _OrcidHasher(inspire_record).compute_hash()
+        return self._cached_hash_value != self._new_hash_value
 
 
-class OrcidHasher(object):
-    def compute_hash(self, record):
-        """Generate hash for an ORCID-serialised HEP record
+class _OrcidHasher(object):
```
michamos: I know we've already discussed this, but a class with an `__init__` and one public method is not a class, it's a function. In the meantime I remembered this good talk https://www.youtube.com/watch?reload=9&v=o9pEzgHorH0 with a more detailed explanation.

inspirehep/modules/orcid/cache.py
```diff
@@ -31,69 +31,114 @@
...
+        if inspire_record:
+            if not self._new_hash_value:
+                self._new_hash_value = _OrcidHasher(inspire_record).compute_hash()
+            data['hash'] = self._new_hash_value
 
-    def read_record_data(self, recid):
+        key = self.get_key(recid)
+        self.redis.hmset(key, data)
+
+    def read_work_putcode(self, recid):
```
michamos: this also gets the hash from the cache, which is not obvious from the name.

inspirehep/modules/orcid/api.py
```diff
@@ -97,39 +105,58 @@ def push_record_with_orcid(recid, orcid, oauth_token, put_code=None, old_hash=No
...
+            if DuplicatedExternalIdentifiersError.match(exc):
+                recache_author_putcodes(orcid, oauth_token)
+                putcode = orcid_cache.read_work_putcode(recid)
+                if not putcode:
+                    raise PutcodeNotFoundInCacheException(
+                        'Putcode not found in cache for recid {} after having'
+                        ' recached all putcodes for the orcid {}'.format(recid, orcid))
+                return push_record_with_orcid(recid, orcid, oauth_token)
+            else:
+                raise exc
```
szymonlopaciuk: I think it's better to use raise without `exc`. In Python 2 `raise exc` will lose the original traceback: https://stackoverflow.com/a/4825279. Unless this is what you want

inspirehep/modules/orcid/api.py
```diff
@@ -53,50 +53,42 @@
...
-
-    new_hash = OrcidHasher().compute_hash(record)
-    if new_hash == old_hash:
-        LOGGER.info(
-            'Hash unchanged: not pushing #%s as not a meaningful update', recid
-        )
-        return putcode, new_hash
+    putcode = orcid_cache.read_work_putcode(recid)
+    if not orcid_cache.has_work_content_changed(recid, inspire_record):
+        LOGGER.info('Hash unchanged: not pushing #%s as not a meaningful update', recid)
```
michamos: add the orcid to the log?

inspirehep/modules/orcid/cache.py
```diff
@@ -31,69 +31,114 @@
...
+
+def _init_redis_client():
+    """
+    Note: the `_redis_client` module-level var cannot be initialized at module-load
+    time because accessing flask.current_app raises the exception:
+    "RuntimeError: Working outside of application context"
+    """
+    url = app.config.get('CACHE_REDIS_URL')
+    global _redis_client
+    _redis_client = StrictRedis.from_url(url)
```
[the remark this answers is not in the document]
turtle321: Ok, I didn't know this

inspirehep/modules/orcid/api.py
turtle321: Good point, what about using the `raise_from` from six? https://pythonhosted.org/six/#six.raise_from

inspirehep/modules/orcid/api.py
turtle321: All log statements will be re-written in my next changes when I am tackling log aggregation and monitoring. Anyway, I'll add it.

inspirehep/modules/orcid/cache.py
```diff
@@ -31,69 +31,114 @@
...
-        self.redis = StrictRedis.from_url(redis_url)
+        """
+        Orcid cached data.
+
+        Args:
+            orcid (string): orcid identifier.
+        """
         self.orcid = orcid
+        self._cached_hash_value = None
+        self._new_hash_value = None
```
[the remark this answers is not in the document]
turtle321: Yes, it makes more sense. I will move `recid` to the constructor's signature.

inspirehep/modules/orcid/cache.py
turtle321: I don't have a strong preference, but I want to keep the concept of hash unknown by the user of this class. Thus what this method does, is just writing the putcode.

inspirehep/modules/orcid/cache.py
turtle321: See prev comment about the concept of hash being an internal implementation detail of the cache object. Thus this methods "reads and returns a putcode for a given (in the constructor in the last version) orcid, recid"

inspirehep/modules/orcid/api.py
turtle321: Done

inspirehep/modules/orcid/cache.py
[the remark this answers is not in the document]
turtle321: Done

inspirehep/modules/orcid/cache.py
turtle321: I explained the reasons to group 3 related functions in a single object. Class/static methods are pretty popular all over, even in CPython lib source code

turtle321: I can add another thing: our codebase is a jungle of code without architecture. Grouping code into abstractions that goes beyond the `utils` module is what we need.
A class goes into this direction, even in this case which I recognize being not a perfect use case.

As for the over use of classes, I recognize the problems coming with over use of inheritance. For the rest in my experience, OOP was always beneficial to code architecture

turtle321:
ci test please

inspirehep/modules/orcid/cache.py
turtle321: Also, I watched the video and he talks about this use case in the last question: https://youtu.be/o9pEzgHorH0?t=1594

michamos: I don't think that question is exactly your use case, but I agree that anything that goes in the direction of better code organization is positive. So let's not waste more time on bikeshedding and agree to disagree.

inspirehep/modules/orcid/cache.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        Write the putcode and the hash for the given (orcid, recid).
+
+        Args:
+            putcode (string): the putcode used to push the record to ORCID.
+            inspire_record (InspireRecord): InspireRecord instance. If provided,
+             the hash for the record content is re-computed.
+        """
+        if not putcode:
+            raise EmptyPutcodeError
```
michamos: why not simply `ValueError('putcode must not be empty')`?

turtle321: Valid point, I will do that

inspirehep/modules/orcid/cache.py
```diff
@@ -0,0 +1,153 @@
...
+        return 'orcidcache:{}:{}'.format(self.orcid, self.recid)
+
+    def write_work_putcode(self, putcode, inspire_record=None):
+        """
+        Write the putcode and the hash for the given (orcid, recid).
+
+        Args:
+            putcode (string): the putcode used to push the record to ORCID.
+            inspire_record (InspireRecord): InspireRecord instance. If provided,
+             the hash for the record content is re-computed.
```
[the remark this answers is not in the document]
turtle321: In a short, concise, highly-specialized object documenting exceptions is unnecessary as finding them in the source code is more clear (and does not require docs to be maintained). 
But I'll add it as it seems to be the style in this codebase - not really actually...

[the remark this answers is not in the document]
turtle321: done

turtle321:
I'll merge when the build is green

turtle321:
ci merge please

PR #3507, opened 2018-06-21
holdingpen: fix holdingpen-hep mapping for facets

szymonlopaciuk:
## Description
**See: https://its.cern.ch/jira/browse/INSPIR-907**

This fixes:
- 'Filter by action' facet

## Related Issue
Some facets in holdingpen don't have explicit mapping.

## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/mappings/holdingpen/hep.json
+++ b/inspirehep/modules/workflows/mappings/holdingpen/hep.json
@@ -10,6 +10,10 @@
             "properties": {
                 "_extra_data": {
                     "properties": {
+                        "_action": {
+                            "index": "not_analyzed",
+                            "type": "string"
+                        },
                         "is-update": {
                             "type": "boolean"
                         },


michamos:
have you checked all facets, or only that one?

szymonlopaciuk:
@michamos From the code side the other ones are already there. On the UI side filters for status, source, subject are all working, as for collections I don't really know how to get anything other that `HEP` to show.

szymonlopaciuk:
There seems to be a problem with sorting by relevance prediction actually, so looking into it...

michamos:
:+1: collection seems to be the workflow name

szymonlopaciuk:
In regards to the sorting for relevance prediction, it just doesn't work at all (always displays previous sorting). Also doesn't work on labs. I checked the mapping and there is a field for it, so makes me think a problem could be somewhere else. Any tips @drjova ?

david-caro:
> +1 collection seems to be the workflow name

yep, just checked :)

david-caro:
> as for collections I don't really know how to get anything other that HEP to show.

You can just search without restrictions, you should see, at least, Author, HEP, MERGE and on qa edit_article (yay for consistenci xd)

szymonlopaciuk:
I created a dummy 'author' workflow on local to check and that facet also works. However I don't know if it should also appear in the holdingpen 'New Authors' card, cause if so, I just found another bug :/

david-caro:
If you created with the author form it should.

david-caro:
(it expects the `is-update` to have `false` or `true`)

szymonlopaciuk:
From the form it works. From the code it threw an exception on no email of the submitter end ended up in ERROR state in harvests.

szymonlopaciuk:
...I'm discovering so many things wrong I'll just keep updating the JIRA ticket for now...

david-caro:
That's not related to this one right? this one actually fixes one of those issues no? (and as such can go in)

szymonlopaciuk:
@david-caro It can on its own, you're right

david-caro:
ci merge please

Issue #3490, opened 2018-06-18
HoldingPen: GUI: Dashboard


ksachs:
Big search slot on dashboard.

At some point we might want to re-organize the boxes on the dashboard.
Less distinction between arXiv / Journal.

ksachs:
##  Add search-bar to dashboard
Currently there is none, and you always have to go to a random list view to start searching. 

![dashboard](https://user-images.githubusercontent.com/7591594/42314823-2b3f3f96-8046-11e8-9103-8d0f73dbb628.png)


It's the same empty search as on the detailed page, so just cut and paste 

    <div class="pull-right">
      <form action="/holdingpen/list" class="hp-search-form">
        <input type="text" name="q" placeholder="Search the holding pen">
      </form>
    </div>

from  `inspirehep/modules/workflows/templates/inspire_workflows/details.html` to `inspirehep/modules/workflows/templates/inspire_workflows/index.html`.

No deeper knowledge of front-end programming required.

## Better visibility

try to find the search bar (yes - there is one) on:
![visibility_old](https://user-images.githubusercontent.com/7591594/42314967-8c1d7ba2-8046-11e8-9ab5-41e993838e2b.png)

and compare to
![visibility_new](https://user-images.githubusercontent.com/7591594/42314982-941155b8-8046-11e8-9e8e-40b6f8979d30.png)

in addition black on white gives a higher contrast and is easier to work with.

Change colors in `inspirehep/modules/theme/static/scss/holding-pen/holding-pen.scss`:

    -      background-color: $clouds;
    +      background-color: white;
    -     color: $grey;
    +      color: black;
    
There is no variable for black and white in `inspirehep/modules/theme/static/scss/base/variables.scss`.

## Longer search-bar on focus
The search term in the holdingpen can be very long. Therefore the search-bar should be as wide as possible on focus. It would be nice if it could stay in the banner on the list page even if the window is not full-screen:

![max_width](https://user-images.githubusercontent.com/7591594/42317000-8d133a7e-804b-11e8-9136-7c7affa08920.png)

Change width in `inspirehep/modules/theme/static/scss/holding-pen/holding-pen.scss`:

    -        width: 500px;
    +        width: 700px;

Again, limited programming skills are sufficient.

With 700px the beginning of the search bar is not visible in narrow windows, but the rest of the page is not very usable either:
[...]

PR #3518, opened 2018-06-27
HP GUI: search-bar

ksachs:
## Description
* add search-bar to dashboard
* increase focus width
* change color to black on white

## Related Issue
HoldingPen: GUI: Dashboard #3490

## Motivation and Context
quick win

## Checklist:

### DIFF
--- a/inspirehep/modules/theme/static/scss/holding-pen/holding-pen.scss
+++ b/inspirehep/modules/theme/static/scss/holding-pen/holding-pen.scss
@@ -187,15 +187,15 @@ $hp_table_bg: #fff;
       -webkit-transition: all .5s;
       -moz-transition: all .5s;
       transition: all .5s;
-      background-color: $clouds;
+      background-color: white;
       border: none;
       padding: 7px;
       margin-top: -8px;
       border-radius: 3px;
       font-size: .9em;
-      color: $grey;
+      color: black;
       &:focus {
-        width: 500px;
+        width: 700px;
       }
     }
   }
--- a/inspirehep/modules/workflows/templates/inspire_workflows/index.html
+++ b/inspirehep/modules/workflows/templates/inspire_workflows/index.html
@@ -22,6 +22,11 @@
     <li class="active">Dashboard
     </li>
   </ul>
+  <div class="pull-right">
+    <form action="/holdingpen/list" class="hp-search-form">
+      <input type="text" name="q" placeholder="Search the holding pen">
+    </form>
+  </div>
 </div>
 {% endblock page_banner %}
 


tsgit:
the base palette for inspirehep is https://flatuicolors.com/palette/defo and it is parametrized throughout the individual theme files. 
instead of piecemeal changing some settings by hardcoding personal preference in a particular file, maybe the choice of palette should be reconsidered (it's not my first choice), or at least the customizations should be parameterized.

there are 12 other palettes at https://flatuicolors.com/ (and there certainly are many other options elsewhere)

I would challenge the notions of a "quick win" and "fast review", because ill-considered quick fixes are what got Inspire into the legacy mess and associate patchwork

that being said, I do agree that the current theme is lacking contrast and is not optimal for reading. these things heavily depend on choice of font, color temperature of the monitor, ambient light, font hinting and anti-aliasing, type of screen (i.e. IPS is widely considered best for text), inherent contrast of the theme, etc. . by keeping things parameterized one could offer different preference themes to different curators

I think currently there is no dedicated UI team, so I don't know who would be looking into that, though.

ksachs:
thanks for the feedback.

black and white are used in other places in addition to the palette, usually as default. So I thought that's OK. We definitely need a palette that allows to use black and white. Currently all colors are sort-of hard-coded in the scss file by using names associated to the color (e.g. orange) not to the usage. 

We had some discussion here at DESY. The visibility issue came up when our catalogers first saw the interface. For this minor issue I didn't want to involve all catalogers. Up to now no one asked us which colors we prefer. Maybe we should start to involve all catalogers in such decisions.

Something that needs discussion with developers is how to define a style for 'CORE', 'Accept', 'Reject'. Currently they are defined in classes

     .accept_core_keyword {
     color: $green-sea;
    }
    .accept_keyword {
     color: $orange;
    }
 
But we want to re-use them for number-of-core-references and others.
Personally I would prefer to define colors e.g. `$accept: $green-sea` to be used in different classes.
I really like the idea of using parameters that can easily be changed later.
Obviously we have to limit the number of colors used, i.e. define all colors in one place.


So there are several points:
1) Discussion which issues to improve. Some things are easy enough so I can prepare screen-shots and demo-code similar to #3490 . For the back-end I could make proposals together with the catalogers at DESY. And then discuss with inspire-curators or a smaller group what we want. I would not do full testing of the code at this step.
1b) For front-end issues we might need user testing.
2) Discussion with the developers (whom / how/ when) how to structure the code. I.e. whether to use parameters for colors.
3) Decide which issues are important enough to spend developer time on review and merge. I guess this can be done in the normal sprint planning.

michamos:
@ksachs we're closing this as part of some house-cleaning. Holdingpen UI improvements have not been among the priorities raised by curators in recent times, but let me know if you still believe we should do something there.

PR #3523, opened 2018-06-29
add login_success template

harunurhan:
* Adds login_success template that passes loggedin user details to the
parent tab which runs the SPA that attempted the login.

### DIFF
--- a/inspirehep/modules/theme/views.py
+++ b/inspirehep/modules/theme/views.py
@@ -794,6 +794,19 @@ def linkedaccounts():
     return redirect('/')
 
 
+@blueprint.route('/login_success', methods=['GET'])
+def login_success():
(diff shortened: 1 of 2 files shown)


inspirehep/modules/theme/views.py
```diff
@@ -794,6 +794,19 @@ def linkedaccounts():
     return redirect('/')
 
 
+@blueprint.route('/login_success', methods=['GET'])
+def login_success():
+    """Injects current user to the template and passes it to the parent tab."""
+    return render_template(
+        'inspirehep_theme/login_success.html',
+        user={
+            'data': {
```
harunurhan: body to be decided

drjova:
Supersede by https://github.com/inspirehep/inspire-next/pull/3563

PR #3549, opened 2018-07-18
WIP: global: change the static url path

drjova:
* Changes the ``static_url_path`` from ``/static`` to ``/assets`` to
  help us move to the new UI infrastructure.

### DIFF
--- a/inspirehep/factory.py
+++ b/inspirehep/factory.py
@@ -66,6 +66,7 @@ create_api = create_app_factory(
     extension_entry_points=['invenio_base.api_apps'],
     converter_entry_points=['invenio_base.api_converters'],
     instance_path=instance_path,
+    static_url_path='/assets',
 )
 
 create_app = create_app_factory(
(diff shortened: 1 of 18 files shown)


drjova:
First merge: https://github.com/inspirehep/inspirehep-search-js/pull/91

david-caro:
Hmm... couldn't we separate the new ui/old ui by a more meaningful url? (something like `/newui/static`, `/react/static` or something similar).
Changing it only for the new ui part minimizes the risk on the current ui too.

harunurhan:
@david-caro 

Somehow we don't have a control over the path of static files that are created after building the react app. Unfortunately it's not configurable and it has a lot of references to it from other compiled, minified js files and html. So your solution only possible if we serve whole react app under a different url like `newui` or `react`, then all links inside the app will have to be prefixed too which is not good.

david-caro:
okok :/, ugly stuff. So maybe then changing `assets` for `oldui` or `flask`. The idea is to make it clear by the name that you are going to a different backend (one is flask, the other react).

drjova:
hmm ``oldui`` or ``flask`` for static files is a bit weird though :/ I'd go for ``flask``

david-caro:
naming is one of the oldest computer science unsolved problem xd

drjova:
true 📟

drjova:
Supersede by https://github.com/inspirehep/inspire-next/pull/3563

PR #3552, opened 2018-07-19
records: add authors prop to literature wrapper

iulianav:
## Description

Add `authors` and `supervisors` properties to LiteratureRecord wrapper. Two new fields are added in order to be used by the UI: first_name, last_name.

Signed-off-by: Iuliana Voinea <iulianavoinea96@gmail.com>

## Related Issue

https://its.cern.ch/jira/browse/INSPIR-1044

## Checklist:

### DIFF
--- a/inspirehep/modules/records/serializers/schemas/json.py
+++ b/inspirehep/modules/records/serializers/schemas/json.py
@@ -26,11 +26,12 @@ from __future__ import absolute_import, division, print_function
 
 from inspire_dojson.utils import get_recid_from_ref, strip_empty_values
 from inspire_utils.helpers import force_list
-from marshmallow import Schema, fields, missing, pre_dump, post_dump
+from marshmallow import Schema, fields, post_dump, pre_dump
 
+from inspirehep.modules.records.serializers.fields.list_with_limit import ListWithLimit
 from inspirehep.modules.records.utils import get_linked_records_in_field
 
-from .common import IsbnSchemaV1, ThesisInfoSchemaV1, ConferenceInfoItemSchemaV1
+from .common import AuthorSchemaV1, IsbnSchemaV1, ThesisInfoSchemaV1, SupervisorSchemaV1, ConferenceInfoItemSchemaV1
 
 
 class RecordMetadataSchemaV1(Schema):
@@ -39,6 +40,7 @@ class RecordMetadataSchemaV1(Schema):
     accelerator_experiments = fields.Raw()
     acquisition_source = fields.Raw()
     arxiv_eprints = fields.Raw()
+    authors = ListWithLimit(fields.Nested(AuthorSchemaV1, dump_only=True), limit=10)
     book_series = fields.Raw()
     # citeable = fields.Raw()
     collaborations = fields.Raw()
@@ -84,6 +86,7 @@ class RecordMetadataSchemaV1(Schema):
     # related_records = fields.Raw()
     report_numbers = fields.Raw()
     # self = fields.Raw()
+    supervisors = ListWithLimit(fields.Nested(SupervisorSchemaV1, dump_only=True), attribute='authors', limit=10)
     texkeys = fields.Raw()
     thesis_info = fields.Nested(ThesisInfoSchemaV1, dump_only=True)
     # title_translations = fields.Raw()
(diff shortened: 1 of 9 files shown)


inspirehep/modules/records/serializers/schemas/json.py
```diff
@@ -44,8 +45,17 @@ class RecordSchemaJSONUIV1(Schema):
...
 class MetadataAuthorsSchemaV1(Schema):
 
-    authors = fields.Raw(dump_only=True)
+    authors = fields.Method('get_authors')
     collaborations = fields.Raw(dump_only=True)
+    supervisors = fields.Method('get_supervisors')
+
+    def get_authors(self, data):
+        record = LiteratureRecord(data)
+        return record.authors or missing
```
drjova: it's better to remove the missing cause we are always expecting authors key on the UI

inspirehep/modules/records/serializers/schemas/json.py
```diff
@@ -44,8 +45,17 @@ class RecordSchemaJSONUIV1(Schema):
...
+
+    def get_supervisors(self, data):
+        record = LiteratureRecord(data)
+        return record.supervisors or missing
```
drjova: it's better to remove the missing for the same reason ☝️

inspirehep/modules/records/wrappers.py
```diff
@@ -68,6 +69,46 @@ def get_link_info_for_external_sys_identifiers(self, extid, ext_sys_id_info):
             return {'url_name': ext_sys_id_info[0],
                     'url_link': ext_sys_id_info[1] + extid}
 
+    @staticmethod
+    def get_enhanced_author(author):
```
drjova: what about the ``Jr.`` case?

inspirehep/modules/records/serializers/schemas/json.py
```diff
@@ -44,8 +45,17 @@ class RecordSchemaJSONUIV1(Schema):
 
 class MetadataAuthorsSchemaV1(Schema):
 
-    authors = fields.Raw(dump_only=True)
+    authors = fields.Method('get_authors')
```
drjova: ``default=[]``

inspirehep/modules/records/serializers/schemas/json.py
michamos: `authors` is not required, so there isn't always an author. Sometimes there's a `corporate_author` instead.

inspirehep/modules/records/wrappers.py
```diff
@@ -68,6 +69,46 @@ def get_link_info_for_external_sys_identifiers(self, extid, ext_sys_id_info):
...
+        """Enhances the author object with name fields that are needed for the UI."""
+        item = copy(author)
+        names = author['full_name'].split(',', 1)
+        if len(names) > 1:
+            item.update({
+                'first_name': names[1].replace(',', '').strip(),
```
iulianav: @drjova check `replace` for dealing qith `Jr.` case. There is also a test for it!

inspirehep/modules/records/serializers/schemas/json.py
iulianav: @drjova then I have to always return empty list from the prperties, right?

inspirehep/modules/records/serializers/schemas/json.py
```diff
@@ -44,8 +45,17 @@ class RecordSchemaJSONUIV1(Schema):
 
 class MetadataAuthorsSchemaV1(Schema):
 
-    authors = fields.Raw(dump_only=True)
+    authors = fields.Method('get_authors', default=[])
     collaborations = fields.Raw(dump_only=True)
+    supervisors = fields.Method('get_supervisors', default=[])
+
+    def get_authors(self, data):
+        record = LiteratureRecord(data)
```
drjova: I just realised ``LiteratureRecord`` does an ES call each time, can we move it this somewhere else?

inspirehep/modules/records/wrappers.py
drjova: we should remove this function as well

drjova:
@iulianav could you please remove the unused code from the wrappers, squash the commits and update the commit message

PR #3569, opened 2018-07-30
INSPIR-1056 workflows: don't repeat ticket creation

ammirate:
## Checklist:

### DIFF
--- a/inspirehep/modules/workflows/tasks/submission.py
+++ b/inspirehep/modules/workflows/tasks/submission.py
@@ -67,7 +67,7 @@ def submit_rt_ticket(obj,
                                                         recid)
     obj.extra_data[ticket_id_key] = new_ticket_id
     obj.log.info(u'Ticket {0} created'.format(new_ticket_id))
-    return True
+    return new_ticket_id
 
 
 def create_ticket(template,
@@ -97,17 +97,20 @@ def create_ticket(template,
                     requestors=user.email if user else '',
                 )
             )
-            return
+            return {}
 
         recid = obj.extra_data.get("recid") or obj.data.get("control_number")
 
-        submit_rt_ticket(obj,
-                         queue,
-                         template,
-                         context,
-                         user.email if user else '',
-                         recid,
-                         ticket_id_key)
+        new_ticket_id = submit_rt_ticket(
+            obj,
+            queue,
+            template,
+            context,
+            user.email if user else '',
+            recid,
+            ticket_id_key
+        )
+        return {ticket_id_key: new_ticket_id}
 
     return _create_ticket
 
@@ -119,7 +122,8 @@ def reply_ticket(template=None,
     @with_debug_logging
     @wraps(reply_ticket)
     def _reply_ticket(obj, eng):
-        ticket_id = obj.extra_data.get("ticket_id", "")
+        ticket_id_key = "ticket_id"
+        ticket_id = obj.extra_data.get(ticket_id_key)
 
         if not rt_instance:
             obj.log.error("No RT instance available. Skipping!")
@@ -128,17 +132,17 @@ def reply_ticket(template=None,
                     ticket_id=ticket_id,
                 )
             )
-            return
+            return {}
 
         if not ticket_id:
             obj.log.error("No ticket ID found!")
-            return
+            return {}
 
         user = User.query.get(obj.id_user)
         if not user:
             obj.log.error(
                 "No user found for object %s, skipping ticket creation", obj.id)
-            return
+            return {}
 
         if template:
             context = {}
@@ -155,7 +159,8 @@ def reply_ticket(template=None,
                 tickets.reply_ticket(ticket_id, body, keep_new)
             else:
                 obj.log.error("No body for ticket reply. Skipping reply.")
-                return
+
+        return {}
 
     return _reply_ticket
 
@@ -168,7 +173,7 @@ def close_ticket(ticket_id_key="ticket_id"):
         ticket_id = obj.extra_data.get(ticket_id_key, "")
         if not ticket_id:
             obj.log.error("No ticket ID found!")
-            return
+            return {}
 
         if not rt_instance:
             obj.log.error("No RT instance available. Skipping!")
@@ -177,9 +182,10 @@ def close_ticket(ticket_id_key="ticket_id"):
                     ticket_id=ticket_id,
                 )
             )
-            return
+            return {}
 
         tickets.resolve_ticket(ticket_id)
+        return {}
 
     return _close_ticket
 
--- a/inspirehep/modules/workflows/utils/__init__.py
+++ b/inspirehep/modules/workflows/utils/__init__.py
@@ -158,6 +158,53 @@ def with_debug_logging(func):
     return _decorator
 
 
+def do_not_repeat(step_id):
+    """Decorator used to skip workflow steps when a workflow is re-run.
+
+
+    Will store the result of running the workflow step in source_data.persistent_data
+    after running the first time, and skip the step on the following runs, also applying
+    previously recorded 'changes' to extra_data.
+
+    The decorated function has to conform to the following signature:
+
+        def decorated_step(obj: WorkflowObject, eng: WorkflowEngine) -> Dict[str, Any]: ...
+
+    Where obj and eng are usual arguments following the protocol of all workflow steps.
+    The returned value of the decorated_step will be used as a patch to be applied on the
+    workflow object's source data (which 'replays' changes made by the workflow step).
+
+    Args:
+        step_id (str): name of the workflow step, to be used as key in persistent_data
+
+    Returns:
+        callable: the decorator
+    """
+    def decorator(func):
+        @wraps(func)
+        def _do_not_repeat(obj, eng):
+            source_data = obj.extra_data['source_data']
+            is_task_repeated = step_id in obj.extra_data['source_data'].setdefault('persistent_data', {})
+            if is_task_repeated:
+                extra_data_update = source_data['persistent_data'][step_id]
+                obj.extra_data.update(extra_data_update)
+                obj.save()
+                return
+
+            return_value = func(obj, eng)
+            if not isinstance(return_value, dict):
+                raise TypeError(
+                    "Functions decorated by 'do_not_repeat' must return a "
+                    "dictionary compliant to extra_data info"
+                )
+            source_data['persistent_data'][step_id] = return_value
+            obj.save()
+            return return_value
+
+        return _do_not_repeat
+    return decorator
+
+
 def ignore_timeout_error(return_value=None):
     """Ignore the TimeoutError, returning return_value when it happens.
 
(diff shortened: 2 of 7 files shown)


ammirate:
ci test please

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -158,6 +158,40 @@ def _try_to_log(logfn, *args, **kwargs):
...
+                'Step "%s" already executed in a previous workflow run' %
+                step_id
+            )
+            obj.extra_data['skipped_steps'][step_id] = True
+            obj.save()
+            return
+        else:
+            func(*args, **kwargs)
+            obj.extra_data['skipped_steps'][step_id] = False
+            obj.save()
```
david-caro: nono, this is not good, you should use a "parametrized decorator", like:
```python

def my_parameterized_decorator(param1):
    def _real_decorator(func):
        @wraps(func)
        def _wrapper_func(*args, **kwargs):
            print('param1 was %s' % param1)
            if _already_run(param1):
                return

            return func(*args, **kwargs)

    return _real_decorator

## then you can use it like
@prametrized_decorator(param1="something')
def myfunc(some, args, and_kwargs=None):
    pass
```

david-caro: That way you don't have to mangle the args or anything, and it's totally agnostic of the function inside.
You should also return whatever the function returns (just in case, very recommended practice).

david-caro: the api would be nicer if you could use it both ways, but we don't want that for this one, and it makes the code a bit uglier (a big if to check `callable` for example).

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -158,6 +158,40 @@ def _try_to_log(logfn, *args, **kwargs):
...
+    value is False if the step has never been skipped, otherwise True.
+
+    Args:
+        func (object): the workflow's step to skip from the second run on
+        step_id (string): an identifier for the step to be skipped
+
+    Return:
+        a decorator function
+    """
+    @with_debug_logging
```
david-caro: you probably don't need this there.

ammirate:
This is more or less my previous approach, then after talking IRL we agreed to have the decorator directly in the workflow definition, since don't repeat a step is a business decisions that should be hidden in the code but explicitly written in the workflow

david-caro:
From my previous reply:
```
## or as you were using it
myfunc = parametrized_decorator(param1='something')(oldfunc)
```

ammirate:
ci test please

ammirate:
ci test please

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -158,6 +158,21 @@ def _try_to_log(logfn, *args, **kwargs):
     return _decorator
 
 
+def do_not_repeat(step_id):
```
michamos: maybe add a docstring explaining what this is for?

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -317,7 +331,9 @@
 
 
 STORE_RECORD = [
-    store_record,
+    do_not_repeat('store_record')(
```
michamos: I am not sure this is correct. What you really want to avoid is minting a new recid, but here you avoid saving it altogether. This doesn't matter currently as we still have legacy that will overwrite the record anyway, but is not really robust.

The better solution is probably to write the minted recid also somewhere to `extra_data` (wrapping https://github.com/inspirehep/inspire-next/blob/cd0ef5d5309f63ef4ce4f5f69604638ed55a7cf4/inspirehep/modules/workflows/tasks/upload.py in a `try...finally` to make sure that always happens, even if the record creation raises) and have a mechanism to make that survive restarting, generalizing `skipped_steps` to be able to store arbitrary data.

ammirate: After some discussions IRL, we have decided to not skip `store_data` since it should be harmless to run it again in case of failures or restarts.

ammirate:
ci test please

ammirate:
ci test please

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -119,7 +122,8 @@ def reply_ticket(template=None,
     @with_debug_logging
     @wraps(reply_ticket)
     def _reply_ticket(obj, eng):
-        ticket_id = obj.extra_data.get("ticket_id", "")
+        ticket_id_key = "ticket_id"
+        ticket_id = obj.extra_data.get(ticket_id_key, "")
```
michamos: couldn't you return an empty `dict`in this case, as this function is not writing anything to `extra_data`?

david-caro:
ci test please

szymonlopaciuk:
ci test please

szymonlopaciuk:
ci test please

szymonlopaciuk:
ci test please

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -148,14 +152,16 @@ def _reply_ticket(obj, eng):
                                                template,
                                                context,
                                                keep_new)
+            return {ticket_id_key: ticket_id}
```
michamos: my point was a bit different, maybe I was not clear enough. In this step, you're not writing to the `ticket_id` key, only reading. This would mean that the decorator has to do only the skipping here if restarting, not the population of the key in `extra_data` from `persistent_data`, so it should never return a non-empty `dict`.

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -158,6 +158,35 @@ def _try_to_log(logfn, *args, **kwargs):
...
+    """Decorator used to skip workflow steps when a workflow is re-run.
+
+    Will store the result of running the workflow step in source_data.persistent_data
+    after running the first time, and skip the step on the following runs.
+    """
+    def decorator(func):
+        @wraps(func)
+        def _do_not_repeat(obj, eng):
+            source_data = obj.extra_data['source_data']
+            if step_id in obj.extra_data['source_data'].setdefault('persistent_data', {}):
```
michamos: I would do
```
is_task_repeated = step_id in obj.extra_data['source_data'].setdefault('persistent_data', {})
if is_task_repeated:
   # ...
```
for clarity.

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -158,6 +158,35 @@ def _try_to_log(logfn, *args, **kwargs):
...
+                extra_data_update = source_data['persistent_data'].get(step_id, {})
```
michamos: you cannot be inside this branch of the `if` if `step_id` is not there, so you don't need to do a `.get` but can use `[step_id]`.

inspirehep/modules/workflows/tasks/submission.py
szymonlopaciuk: I see, I was considering that myself, it makes sense

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -158,6 +158,35 @@ def _try_to_log(logfn, *args, **kwargs):
     return _decorator
 
 
+def do_not_repeat(step_id):
+    """Decorator used to skip workflow steps when a workflow is re-run.
+
+    Will store the result of running the workflow step in source_data.persistent_data
+    after running the first time, and skip the step on the following runs.
```
michamos: maybe add a word about the protocol that the wrapped function should implement to have its writes persisted?

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -982,3 +984,102 @@ def test_start_wf_with_no_source_data_fails(workflow_app):
...
+    mocked_arxiv_download,
+    workflow_app,
+    mocked_external_services,
+):
+    def return_value(val):
+        def _return_value(obj, eng):
+            return {'id': val}
+        return _return_value
+
+    custom_wf_steps = [
```
michamos: I would add here the step that is used in the real workflows to restore from `source_data`, so you're testing the interactions between the two.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -982,3 +984,102 @@ def test_start_wf_with_no_source_data_fails(workflow_app):
...
+        do_not_repeat('one')(
+            return_value(1),
+        ),
+        do_not_repeat('two')(
+            return_value(2),
+        ),
+    ]
+
+    custom_wf_steps_to_repeat = [
```
michamos: here too.

szymonlopaciuk:
ci test please

szymonlopaciuk:
ci test please

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -97,17 +97,20 @@ def _create_ticket(obj, eng):
                     requestors=user.email if user else '',
                 )
             )
-            return
+            return {ticket_id_key: ""}
```
david-caro: {}

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -177,7 +182,7 @@ def _close_ticket(obj, eng):
                     ticket_id=ticket_id,
                 )
             )
-            return
+            return {ticket_id_key: ticket_id}
```
david-caro: not needed

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -982,3 +985,104 @@ def test_start_wf_with_no_source_data_fails(workflow_app):
...
+        eng = WorkflowEngine.from_uuid(workflow_uuid)
+        obj = eng.processed_objects[0]
+
+        assert obj.status == ObjectStatus.COMPLETED
+        assert obj.extra_data['source_data']['persistent_data']
+        assert obj.extra_data['source_data']['persistent_data'] == expected_persistent_data_first_run
+
+        eng = WorkflowEngine.from_uuid(obj.id_workflow)
+        eng.callbacks.replace(custom_wf_steps_to_repeat)
+        eng.process([obj])
```
david-caro: You need to reset the callback_pos to [0]

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -982,3 +985,104 @@ def test_start_wf_with_no_source_data_fails(workflow_app):
...
+
+        eng = WorkflowEngine.from_uuid(workflow_uuid)
+        obj = eng.processed_objects[0]
+        assert obj.extra_data['source_data']['persistent_data'] == expected_persistent_data_second_run
```
david-caro: Check also that the extra_data was changed according to the first run of the workflow.

szymonlopaciuk:
ci test please

szymonlopaciuk:
ci test please

szymonlopaciuk:
ci test please workflows

szymonlopaciuk:
ci merge please

szymonlopaciuk:
ci merge please

PR #3581, opened 2018-08-07
global: remove invenio-collections dependency

szymonlopaciuk:
Signed-off-by: Szymon Łopaciuk <szymon@lopaciuk.eu>

## Checklist:


szymonlopaciuk:
ci test please

michamos:
Will this kind of searches still work? https://labs.inspirehep.net/search?page=1&size=25&cc=HAL%20Hidden

michamos:
We seem to have (non-functioning) templates for it also: https://sentry.inspirehep.net/inspire-sentry/prod/issues/210464/

szymonlopaciuk:
ci test please

szymonlopaciuk:
@michamos I tired out the `/search?cc=HAL%20Hidden` type search, and they seem to be working. I assume it's because we store the collections as part of the record itself.

szymonlopaciuk:
As for the [sentry error](https://sentry.inspirehep.net/inspire-sentry/prod/issues/210464/) the problem is that the page that request was trying to access, `/collections/`, is handled by invenio-collections, and their template param `collection` (https://github.com/inveniosoftware/invenio-collections/blob/master/invenio_collections/views.py#L63) collides with out param template. Some templates in inspire (e.g. https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/theme/templates/inspirehep_theme/header.html#L29) use the collection param (of type string) to render the correctly 'themed' page. However when rendering `/collections/` this parameter is instead provided by `invenio-collections` (and is of type of type Collection). This type mismatch is what triggers the exception. This page, however, would be removed when we remove `invenio-collections` falling back to the good old 404.

szymonlopaciuk:
Regarding the permissions, the restricted collections are stored in the DB as collection names using `invenio-access`, and the actual query is built in inspire, against the record metadata itself, here: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/search/api.py#L114. So it seems, as if `invenio-collections` are not used for enforcing permissions either.

david-caro:
ci merge please

PR #3585, opened 2018-08-09
workflows: fix for ``author_list``

drjova:
* Fixes a bug with ``author_list`` when the tarball is not part of the
  ``files``.

### DIFF
--- a/inspirehep/modules/workflows/tasks/arxiv.py
+++ b/inspirehep/modules/workflows/tasks/arxiv.py
@@ -233,9 +233,9 @@ def arxiv_author_list(stylesheet="authorlist2marcxml.xsl"):
     def _author_list(obj, eng):
         arxiv_id = get_arxiv_id(obj.data)
         filename = secure_filename('{0}.tar.gz'.format(arxiv_id))
-        tarball = obj.files[filename]
-
-        if not tarball:
+        try:
+            tarball = obj.files[filename]
+        except KeyError:
             obj.log.info(
                 'Skipping author list extraction, no tarball with name "%s" found' % filename
             )
--- a/tests/unit/workflows/test_workflows_tasks_arxiv.py
+++ b/tests/unit/workflows/test_workflows_tasks_arxiv.py
@@ -728,6 +728,43 @@ def test_arxiv_derive_inspire_categories_does_nothing_with_existing_categories()
     assert expected == result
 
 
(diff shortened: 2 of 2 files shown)


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -233,9 +233,9 @@ def arxiv_author_list(stylesheet="authorlist2marcxml.xsl"):
     def _author_list(obj, eng):
         arxiv_id = get_arxiv_id(obj.data)
         filename = secure_filename('{0}.tar.gz'.format(arxiv_id))
-        tarball = obj.files[filename]
-
-        if not tarball:
+        try:
+            tarball = obj.files[filename]
+        except KeyError:
```
david-caro: Just curious, was this an api change? Can `tarball` still be `None`? Should we check for that too?

drjova: AFAIK ``obj.files`` item it doesn't make sense to be ``None``, it should be there or not. I didn't dig in more on this but I can check how we get the files into the workflow object

drjova: so you cannot have ``None`` it reads directly from the bucket

david-caro: Not the `obj.files` item itself, but whatever it's `__getitem__` returns.

drjova: Hmm it'll give you a ``KeyError`` actually https://github.com/inveniosoftware/invenio-records-files/blob/master/invenio_records_files/api.py#L159-L164

drjova: and I just saw, this is how we handle the plots actually https://github.com/inspirehep/inspire-next/blob/b40a21e9bfc92922ed88edef801cd7ef07187e05/inspirehep/modules/workflows/tasks/arxiv.py#L126-L139

david-caro: :+1:

drjova: if it's ok, could you please also approve it 🤓

drjova:
ci merge please

drjova:
ci merge please

Issue #3586, opened 2018-08-09
HP: Halt gently to resolve DOI conflicts


ksachs:
## Expected Behavior
If there is more than one exact match the workflow should halt and wait for a curator.
Curator can resolve the problem on legacy and continue the workflow.
Two possibilities to continue the workflow:
- re-start the matcher (requires that labs is sync with legacy)
- curator selects the correct record (similar to fuzzy matcher)

I guess the second solution is easier to implement.

## Current Behavior
unresolved multiple exact matches lead to upload failures and workflows in error state that have to be resolved manually.

## Minimal solution - temporary fix
Demo code in 
https://github.com/inspirehep/inspirehep-search-js/pull/95
https://github.com/inspirehep/inspire-next/pull/3587

* Accept exact match only if there is exactly one exact match.
* Otherwise proceed to fuzzy match.
* Since the problem arises in general for arXiv updates we should have the arXiv record with a very similar title, which should be found by fuzzy matcher.
* Therefore it is usually enough to raise a flag when halted to confirm fuzzy match.
* In case the information is wrong only 1 record will be found by fuzzy matcher. So the recids found by exact match should be displayed.
* A link to legacy with a search for the recids would be nice
* A link to BibMerge would be nice

## Screenshots (if appropriate):
![doiconflict](https://user-images.githubusercontent.com/7591594/43906764-5b528e6c-9bf4-11e8-91cf-661fd3c17a19.png)

ksachs:
If you don't want `obj.extra_data['matches']['conflicts']` there is also the possibility to store the conflicts in `obj.extra_data['matches']['fuzzy']` and add a field (e.g. `"comment":"Conflict"`).
Fuzzy matcher would add additional matches and set status (`fuzzy match` or a new `DOI Conflicts`).

The display would work essentially as now for fuzzy matches, just display prominently the additional field if it is there. The conflicting matches would be shown first, followed by the real fuzzy matches.
With a `Best match` button, as it is now, the cataloger doesn't need to wait for the synchronization with legacy. If you prefer a `Re-do matching`, that might work too.

StellaCh:
closing ticket as this is implemented, but feel free to reopen it if something's missing

PR #3587, opened 2018-08-09
DEMO: Halt gently to resolve DOI conflicts

ksachs:
**untested!**

* exact match requires exactly one exact match
* otherwise it proceeds to search for fuzzy matches

This is a minimal solution!
It assumes that at least one exact match is also found as fuzzy match.
If would be better if the workflow would halt in case of more than one exact matches,
present the result to a cataloger to resolve.

## Related Issue
https://github.com/inspirehep/inspire-next/issues/3586

## Motivation and Context
unresolved multiple exact matches lead to upload failures and workflows in error state that have to be resolved manually.

## Checklist:


ammirate:
Superseded by #3607.

PR #3589, opened 2018-08-10
INSPIR-1126 and INSPIR-1149

pazembrz:
INSPIR-1126 alembic: Allow querying for citing records
INSPIR-1149 serializers: Extend the json sent to elasticsearch to include the citation count
Also fixes problem with testing alembic migrations added after "eaab22c59b89".

Elastic search now will receive citation counts
Also Serializer for search page will receive citation count
Also Serializer for single Literature entry will now receive citation count

Also fixed few Indentation to be consistent with PEP8

## Checklist:

- [] All new and existing tests passed.

### DIFF
--- a/inspirehep/alembic/0bc0a6ee1bc0_create_index_on_referenced_records.py
+++ b/inspirehep/alembic/0bc0a6ee1bc0_create_index_on_referenced_records.py
@@ -1,0 +1,42 @@
+#
+# This file is part of Invenio.
+# Copyright (C) 2016-2018 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""create index on referenced_records"""
+from __future__ import absolute_import, division, print_function
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = '0bc0a6ee1bc0'
+down_revision = '2f5368ff6d20'
+branch_labels = ()
+depends_on = '07fb52561c5c'
+
+
+def upgrade():
+    """Upgrade database."""
+    op.execute('''
+        create or replace function referenced_records(json jsonb) returns text[] as  $$
+        select
+            array_agg(
+                references_arr->'record'->>'$ref'
+            )  as result
+         from
+            jsonb_array_elements(json->'references') as references_arr; $$
+        language sql immutable
+    ''')
+    op.execute('''
+        create index ix_records_metadata_json_referenced_records
+            on records_metadata
+            using gin(referenced_records(json))
+    ''')
+
+
+def downgrade():
+    """Downgrade database."""
+    op.execute('drop index if exists ix_records_metadata_json_referenced_records')
+    op.execute('drop function if exists referenced_records(json jsonb)')
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -33,8 +33,9 @@ from elasticsearch.exceptions import NotFoundError
 from flask import current_app
 from fs.opener import fsopen
 from six.moves.urllib.parse import urlparse, unquote
+from invenio_records.api import RecordMetadata
 
-from inspire_dojson.utils import get_recid_from_ref, strip_empty_values
+from inspire_dojson.utils import get_recid_from_ref, strip_empty_values, absolute_url
 from inspire_schemas.api import validate
 from inspire_schemas.builders import LiteratureBuilder
 from invenio_files_rest.models import Bucket
@@ -42,20 +43,25 @@ from invenio_pidstore.errors import PIDDoesNotExistError
 from invenio_pidstore.models import PersistentIdentifier
 from invenio_records_files.api import Record
 from invenio_db import db
+from sqlalchemy import Text
+from sqlalchemy.sql.functions import GenericFunction
+from sqlalchemy.dialects import postgresql
 
 from inspirehep.modules.pidstore.minters import inspire_recid_minter
-from inspirehep.modules.pidstore.utils import get_pid_type_from_schema
+from inspirehep.modules.pidstore.utils import get_pid_type_from_schema, get_endpoint_from_pid_type
 from inspirehep.utils.record_getter import (
     RecordGetterError,
     get_es_record_by_uuid
 )
 
-
 MAX_UNIQUE_KEY_COUNT = 50000
 
 
-class InspireRecord(Record):
+class referenced_records(GenericFunction):
+    type = postgresql.ARRAY(Text)
 
+
+class InspireRecord(Record):
     """Record class that fetches records from DataBase."""
 
     @classmethod
@@ -586,9 +592,30 @@ class InspireRecord(Record):
 
         return new_key
 
+    def _get_ref(self):
+        """Returns full url to this object (as in $ref)"""
+        pid_value = self.get('control_number')
+        pid_type = get_pid_type_from_schema(self.get('$schema'))
+        endpoint = get_endpoint_from_pid_type(pid_type)
+        return absolute_url(u'/api/{endpoint}/{control_number}'.format(endpoint=endpoint,
+                                                                       control_number=pid_value))
+
+    def _query_citing_records(self):
+        """Returns records which cites this one"""
+        ref = self._get_ref()
+        if not ref:
+            raise Exception("There is no $ref for this object")
+        citations = RecordMetadata.query.with_entities(RecordMetadata.id).filter(
+            referenced_records(RecordMetadata.json).contains([ref]))
+        return citations
+
+    def get_citations_count(self):
+        """Returns citations count for this record"""
+        count = self._query_citing_records().count()
+        return count
+
 
 class ESRecord(InspireRecord):
-
     """Record class that fetches records from ElasticSearch."""
 
     @classmethod
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -59,6 +59,10 @@ def is_hep(record):
     return 'hep.json' in record.get('$schema')
 
 
+def is_data(record):
+    return 'data.json' in record.get('$schema')
+
+
 #
 # before_record_insert & before_record_update
 #
@@ -183,6 +187,15 @@ def enhance_after_index(sender, json, *args, **kwargs):
     populate_inspire_document_type(sender, json, *args, **kwargs)
     populate_name_variations(sender, json, *args, **kwargs)
     populate_title_suggest(sender, json, *args, **kwargs)
+    populate_citations_count(sender, json, *args, **kwargs)
+
+
+def populate_citations_count(sender, json, *args, **kwargs):
+    """Populate citations_count in ES from"""
+    if is_hep(json) and is_data(json) and hasattr(sender, 'get_citations_count'):
+        # Make sure that sender has method get_citations_count
+        citation_count = sender.get_citations_count()
+        json.update({'citation_count': citation_count})
 
 
 def populate_bookautocomplete(sender, json, *args, **kwargs):
@@ -287,14 +300,13 @@ def populate_recid_from_ref(sender, json, *args, **kwargs):
             items = []
 
         for key, value in items:
-            if (isinstance(json_root, dict) and isinstance(value, dict) and
-                    '$ref' in value):
+            if (isinstance(json_root, dict) and isinstance(value, dict) and '$ref' in value):
                 # Append '_recid' and remove 'record' from the key name.
                 key_basename = key.replace('record', '').rstrip('_')
                 new_key = '{}_recid'.format(key_basename).lstrip('_')
                 json_root[new_key] = get_recid_from_ref(value)
             elif (isinstance(json_root, dict) and isinstance(value, list) and
-                    key in list_ref_fields_translations):
+                  key in list_ref_fields_translations):
                 new_list = [get_recid_from_ref(v) for v in value]
                 new_key = list_ref_fields_translations[key]
                 json_root[new_key] = new_list
--- a/tests/integration/records/test_records_api.py
+++ b/tests/integration/records/test_records_api.py
@@ -23,6 +23,7 @@
 from __future__ import absolute_import, division, print_function
 
 import os
+
 from tempfile import NamedTemporaryFile
 import pytest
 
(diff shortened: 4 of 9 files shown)


inspirehep/alembic/0bc0a6ee1bc0_create_index_on_referenced_records.py
```diff
@@ -0,0 +1,56 @@
...
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""create index on referenced_records"""
+
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = '0bc0a6ee1bc0'
+down_revision = '2f5368ff6d20'
```
drjova: hmm down revision should be the HEAD which I think is: ``53e8594bc789``

inspirehep/modules/records/api.py
```diff
@@ -586,9 +585,18 @@ def _append_current_control_number(control_number, base_name):
 
         return new_key
 
+    def _get_citations_count(self):
+        """Returns citations count for this record"""
+        count = 0
+        with db.session.begin_nested():
+            count = db.session.query(func.count_citations("%d" % self.id)).first()
+            if not count:
+                raise Exception("DB did not returned any data for citations!")
```
drjova: shall we create a custom one?

inspirehep/modules/records/receivers.py
```diff
@@ -182,6 +182,16 @@ def enhance_after_index(sender, json, *args, **kwargs):
...
     populate_name_variations(sender, json, *args, **kwargs)
     populate_title_suggest(sender, json, *args, **kwargs)
+    populate_citations_count(sender, json, *args, **kwargs)
+
+
+def populate_citations_count(sender, json, *args, **kwargs):
+    """Populate citations_count in ES from postgres function count_citations"""
+    if hasattr(sender, '_get_citations_count'):
+        # Make sure that sender has method _get_citations_count
+        citation_count=sender._get_citations_count()
```
drjova: spaces before - after ``=`` 🤓

inspirehep/modules/records/api.py
```diff
@@ -586,9 +585,18 @@ def _append_current_control_number(control_number, base_name):
 
         return new_key
 
+    def _get_citations_count(self):
+        """Returns citations count for this record"""
+        count = 0
+        with db.session.begin_nested():
+            count = db.session.query(func.count_citations("%d" % self.id)).first()
```
david-caro: Instead of `first` you can use `one`, that will raise an exception, but in this case I'd perefer the more explanatory exception :)

inspirehep/modules/records/receivers.py
david-caro: This will fail the pep8 check I think (missing space around operator)

tests/integration/test_alembic.py
```diff
@@ -137,6 +145,12 @@ def test_upgrade(alembic_app):
     assert 'legacy_records_mirror' in _get_table_names()
     assert 'legacy_records_mirror_recid_seq' in _get_sequences()
 
+    # 0bc0a6ee1bc0
```
david-caro: I think you should add the upgrade step that was missing, that way, if the one that was missing fails, it will not misguide anyone into thinking that was your upgrade.

inspirehep/alembic/0bc0a6ee1bc0_create_index_on_referenced_records.py
david-caro: I think it's ok:
```
dbdbc1b19cf2 -> 35c1075e6360 (default) (effective head), Force naming convention.
720ddf51e24b -> a26f133d42a9 (invenio_workflows) (effective head), Create invenio_workflows tables.
71634726ec7e (9848d0149abd) -> c25ef2c50ffa (invenio_userprofiles) (head), Create userprofiles tables.
862037093962 -> 07fb52561c5c (invenio_records) (head), Alter column from json to jsonb.
2da9a03b0833 (2e97565eba72, 862037093962) -> 1ba76da94103 (default, invenio_records_files) (head), Create records files tables.
f615cee99600 -> 999c62899c20 (invenio_pidstore) (head), Create pidstore tables.
97bbc733896c -> bff1f190b9bd (default, invenio_oauthclient) (head), Add Timestamp to oauthclient tables.
2e97565eba72 -> f741aa746a7d (invenio_files_rest) (head), Alter files_files.last_check column to nullable.
9848d0149abd -> e12419831262 (invenio_accounts) (head), Add new columns on SessionActivity.
dbe499bfda43 -> 04480be1593e (invenio_access) (head), Add ActionSystemRoles.
eaab22c59b89 -> 2f5368ff6d20 (inspirehep) (head), Add index on legacy_records_mirror
```

inspirehep/modules/records/api.py
pazembrz: I think yes.

inspirehep/modules/records/api.py
pazembrz: I do not think it is possible to receive more or less than one result here. And still when using `one` it would require to catch exception and raise ours as the default one won't give us any useful information.

pazembrz:
ci merge please

inspirehep/alembic/0bc0a6ee1bc0_create_index_on_referenced_records.py
```diff
@@ -0,0 +1,56 @@
...
+
+def upgrade():
+    """Upgrade database."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute('''
+        create or replace function referenced_records(json jsonb) returns text[] as  $$
+        select
+            array_agg(
+              split_part(
+                reference_arr->'record'->>'$ref',E'literature/',2
```
michamos: this is not correct, not only `literature` records can be cited, also `data`.

inspirehep/alembic/0bc0a6ee1bc0_create_index_on_referenced_records.py
```diff
@@ -0,0 +1,56 @@
...
+              )
+            )  as result
+         from
+            jsonb_array_elements(json->'references') as reference_arr; $$
+        language sql IMMUTABLE
```
michamos: why the strange mix of upper and lower case?

inspirehep/alembic/0bc0a6ee1bc0_create_index_on_referenced_records.py
```diff
@@ -0,0 +1,56 @@
...
+    op.execute('''
+        create index ix_records_metadata_json_referenced_records
+            on records_metadata
+            using gin(referenced_records(json))
+    ''')
+
+    op.execute('''
+        create or replace function count_citations(ids text) returns bigint as $$
+        select count(id) as citations_count from records_metadata
+            where referenced_records(json) @> ARRAY[ids]; $$
```
michamos: this is incorrect, as it's not [`immutable`](https://www.postgresql.org/docs/9.4/static/xfunc-volatility.html). Besides, I am not sure hardcoding this in the DB is the best way to go, as we will need to remove some specific records from the counts in the future, based on metadata of the citing/cited record. If you do this in Python, you can also drop the string filtering in the DB and have the `referenced_records` return the full URI.

inspirehep/alembic/0bc0a6ee1bc0_create_index_on_referenced_records.py
```diff
@@ -0,0 +1,47 @@
...
+    """Upgrade database."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.execute('''
+        create or replace function referenced_records(json jsonb) returns text[] as  $$
+        select
+            array_agg(
+                reference_arr->'record'->>'$ref'
+            )  as result
+         from
+            jsonb_array_elements(json->'references') as reference_arr; $$
```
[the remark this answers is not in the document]
pazembrz: It is not an single reference but an array/set of them so it could also be confusing.

inspirehep/modules/records/api.py
```diff
@@ -586,9 +584,28 @@ def _append_current_control_number(control_number, base_name):
...
 
+    def _get_citations(self):
+        """Returns UUID of records which cites this one"""
+        with db.session.begin_nested():
+            _ref = self.get('self', {}).get('$ref')
+            if not _ref:
+                raise Exception("There is no $ref for this object")
 
-class ESRecord(InspireRecord):
+            citations = db.session.execute(
```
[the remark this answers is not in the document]
pazembrz: I thought that pure SQL would be better/faster here, but you may be right that is premature optimization and basically it obfuscates the code. Also it requires to get all the results before counting references so it is not optimal. I will change it tomorrow.

inspirehep/alembic/0bc0a6ee1bc0_create_index_on_referenced_records.py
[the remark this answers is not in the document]
michamos: `references`?

inspirehep/modules/records/api.py
```diff
@@ -586,9 +584,28 @@ def _append_current_control_number(control_number, base_name):
 
         return new_key
 
+    def _get_citations(self):
+        """Returns UUID of records which cites this one"""
+        with db.session.begin_nested():
+            _ref = self.get('self', {}).get('$ref')
```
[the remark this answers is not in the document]
drjova: invenio provided function I'm not aware of any, the flask way is to use ``url_for('invenio_records_rest.literature_item', pid_value=12321)`` but you have to now the ``pid_type`` of the record (example: https://github.com/inveniosoftware/invenio-records-rest/blob/97c320c79e7c609b333a2a73fdfec3a26834dff3/invenio_records_rest/links.py#L26-L27)

inspirehep/modules/records/api.py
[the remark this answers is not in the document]
drjova: That's actually a great idea and will get rid all of these side utils. Let's plan for it ASAP 👍

inspirehep/alembic/0bc0a6ee1bc0_create_index_on_referenced_records.py
```diff
@@ -0,0 +1,47 @@
...
+        language sql immutable
+    ''')
+    op.execute('''
+        create index ix_records_metadata_json_referenced_records
+            on records_metadata
+            using gin(referenced_records(json))
```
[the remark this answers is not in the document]
pazembrz: I'm not sure if we have any recommendations for coding style in SQL? But for correctness of the SQL itself it doesn't matter

inspirehep/modules/records/api.py
```diff
@@ -586,9 +592,30 @@ def _append_current_control_number(control_number, base_name):
 
         return new_key
 
+    def _get_ref(self):
+        """Returns full url to this object (as in $ref)"""
+        pid = PersistentIdentifier.query.filter_by(object_uuid=self.id).one()
```
michamos: Several remarks here:

1. the relation between PIDs and records is not one-to-one, as:
- in the future we will add non-INSPIRE PIDs like DOIs or arXiv IDs to the pidstore;
- already now, records can be merged, and then the PIDs of the two previous records both point to surviving one.
However, you don't need to ask the pidstore for the PID, as it's already in the record: `control_number` is the `pid_value` and the `$schema` encodes the `pid_type`.

2. **Not for this PR** because of the issue with merges, we need to be able to track citations across redirects. The current design is to have a task that periodically rewrites all redirected pids in the records: https://github.com/inspirehep/inspire-next/blob/dd0493e4f3fe9e6d1ddc17156b1ac3b604bbc8ac/inspirehep/modules/records/tasks.py#L48-L131 (this has never been run). Maybe it would be better to deal with it at search time instead, by returning all INSPIRE pids for a given record instead of only the canonical one?

inspirehep/modules/records/api.py
```diff
@@ -586,9 +592,30 @@ def _append_current_control_number(control_number, base_name):
...
+        endpoint = get_endpoint_from_pid_type(pid.pid_type)
+        return absolute_url(u'/api/{endpoint}/{control_number}'.format(endpoint=endpoint,
+                                                                       control_number=self.get('control_number')))
 
-class ESRecord(InspireRecord):
+    def _query_citing_records(self):
+        """Returns records which cites this one"""
+        ref = self._get_ref()
+        with db.session.begin_nested():
```
michamos: do you need the a nested transaction if you're never writing to the DB?

tests/integration/records/test_records_api.py
```diff
@@ -92,3 +93,16 @@ def test_create_does_not_save_zombie_identifiers_if_record_creation_fails(isolat
...
+
+
+def test_citations_count_equals_zero(isolated_app):
+    record = get_db_record('lit', 1497201)
+    assert record.get_citations_count() == 0L
+
+
+def test_citations_count_non_zero(isolated_app):
+    record_1 = get_db_record('lit', 1477591)
+    assert record_1.get_citations_count() == 1L
```
michamos: you don't know who's populating the citation counts for demo records: your code or https://github.com/inspirehep/inspire-next/blob/4b538c9529715e7ebdcb9ee3a6eb92f4ec778410/scripts/recreate_records#L43 ? If you remove the old code in the `migrator` module then you're sure it's yours (as an alternative, you can also insert a pair of citing/cited records in the DB and check the count is correct, but the old code will have to die anyway).

pazembrz: I know that data in `InspireHep.get_citations_count()` are not taken from citation_count field in ES which is created by the old code but they are coming from index in DB `ix_records_metadata_json_referenced_records` and this index is created by me.

pazembrz:
ci test please

pazembrz:
ci test please

drjova:
@pazembrz could you please squash the commits?

inspirehep/modules/records/receivers.py
```diff
@@ -183,6 +183,15 @@ def enhance_after_index(sender, json, *args, **kwargs):
     populate_inspire_document_type(sender, json, *args, **kwargs)
     populate_name_variations(sender, json, *args, **kwargs)
     populate_title_suggest(sender, json, *args, **kwargs)
+    populate_citations_count(sender, json, *args, **kwargs)
+
+
+def populate_citations_count(sender, json, *args, **kwargs):
+    """Populate citations_count in ES from"""
+    if hasattr(sender, 'get_citations_count'):
```
drjova: this will apply to all records, is it what we want?

pazembrz: I'm not sure, Its like that from Monday when I changed it to work with data also. I think that it won't hurt if all entries would have info about citations even if they won't have any.

pazembrz:
ci merge please

drjova:
ci merge please

drjova:
@pazembrz I just squash the commits into 1

Issue #2647, opened 2017-08-17
holdingpen: enrichment: run bibclassify with anti-HEP-ontology on abstract


fschwenn:
One should check for keywords explicitly indicating non-relevance for HEP and display the numbers.

## Expected Behavior
bibclassify should run on the abstract+title+author keywords with the antiHEP-ontology (see #2282). The number of anti-CORE-keywords should be displayed. I have no experience with running it on the fulltext but I would expect too many false positives.
These numbers should also be handed over to the function calculating the automatic decision whether to take or reject a record.

## Current Behavior
None in the actual holdingpen. In the DESY-workflow, we do it.

## Context
For many instrumental or other papers rather at the surrounding research fields finding no core keywords does not necessarily mean that the paper is not interesting for us. But if the abstract contains "biological", "cardiac" and "in vivo" it's clear that we can reject it.

PR #3610, opened 2018-08-22
update HEP Ontology

fschwenn:

### DIFF
--- a/inspirehep/taxonomies/HEPontCore.rdf
+++ b/inspirehep/taxonomies/HEPontCore.rdf
@@ -1,0 +1,16042 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- HEP ontology v.1.7 - April 2015 -->
+<!-- <Kirsten.Sachs@desy.de> <Florian.Schwennsen@desy.de> -->
+
+<rdf:RDF xmlns="http://www.w3.org/2004/02/skos/core#"
+    xmlns:dc="http://purl.org/dc/elements/1.1/"
+    xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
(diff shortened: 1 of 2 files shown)


ammirate:
@fschwenn can you add a bit of context, please?

david-caro:
btw. I don't think invenio-classifier is able to handle multiple ontology files.

fschwenn:
@ammirate The context is that DESY takes care of the HEP ontology. From time to time we update it. And that updated ontology is then also updated in INSPIRE.

fschwenn:
@david-caro It's enough if classifier can handle on at a atime.

david-caro:
I mean that it does not handle multiple files, not at the same time, not one after the other, so if we wan to use more than one ontology file, we will need to change it.

fschwenn:
bibclassify had the ontology as an argument. Do you mean that in the classifier it is hard-coded?!

david-caro:
yep, it's hardcoded in the configuration:
https://github.com/inspirehep/inspire-next/blob/master/inspirehep/config.py#L1162

fschwenn:
but classifier accepts arguments, doesnt it?
https://github.com/inspirehep/inspire-next/blob/b0b9d015a494add095f03b95688b3228fceecbe6/inspirehep/modules/workflows/tasks/classifier.py#L67

david-caro:
yep, but accepts only one, not a list

fschwenn:
but that's fine - as I said, we need to call it with different onotologies (at different times), but only one per call

david-caro:
Then we should change the classifier task to run multiple times, one for each file no?

fschwenn:
see #2647

david-caro:
:+1:

PR #3621, opened 2018-08-30
serializers: fix "missing" not serializable bug

harunurhan:
* Never returns "missing" from NestedWithoutEmptyObject.

* Adds cleanup @post_dump to CitationItemSchema in order to clean all
empty stuff, since now list with single None item can occur after the
change in NestedWithoutEmptyObject

Sentry:

https://sentry.inspirehep.net/inspire-sentry/inspire-qa/issues/229577


ammirate:
ci merge please

PR #3623, opened 2018-09-05
Specify which fields should be returned from ES for LiteratureSearch

pazembrz:
INSPIR-1268 - In progress


ammirate:
ci merge please

PR #3630, opened 2018-09-10
"Recreate record" script fix

pazembrz:
Add alembic to recreate_records script so it would be able to create index for citations.


david-caro:
ci merge please

PR #3636, opened 2018-09-12
Remove deleted records from search

monaawi:
added a check to see if a record has deleted flag true before indexing it to es. If it is deleted, then the index is deleted from es as well
added integration tests

### DIFF
--- a/inspirehep/modules/records/cli.py
+++ b/inspirehep/modules/records/cli.py
@@ -36,6 +36,18 @@ from flask.cli import with_appcontext
 from .checkers import check_unlinked_references
 from .tasks import batch_reindex
 
+from invenio_records.models import RecordMetadata
+
+from sqlalchemy import (
+    String,
+    cast,
+    type_coerce,
+    or_,
+    not_
+)
+
+from sqlalchemy.dialects.postgresql import JSONB
+
 
 @click.group()
 def check():
@@ -109,13 +121,15 @@ def simpleindex(yes_i_know, pid_type, batch_size, queue_name):
     click.secho('Sending record UUIDs to the indexing queue...', fg='green')
 
     query = (
-        db.session.query(PersistentIdentifier.object_uuid)
+        db.session.query(PersistentIdentifier.object_uuid).join(RecordMetadata, type_coerce(PersistentIdentifier.object_uuid, String) == type_coerce(RecordMetadata.id, String))
         .filter(
             PersistentIdentifier.pid_type.in_(pid_type),
             PersistentIdentifier.object_type == 'rec',
             PersistentIdentifier.status == PIDStatus.REGISTERED,
+            or_(not_(type_coerce(RecordMetadata.json, JSONB).has_key('deleted')), RecordMetadata.json["deleted"] == cast(False, JSONB))  # noqa: F401
         )
     )
+
     request_timeout = current_app.config.get('INDEXER_BULK_REQUEST_TIMEOUT')
     all_tasks = []
     records_per_tasks = {}
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -25,10 +25,12 @@
 from __future__ import absolute_import, division, print_function
 
 import uuid
+import logging
 
 from celery import Task
 from flask import current_app
 from flask_sqlalchemy import models_committed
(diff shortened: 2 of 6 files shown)


inspirehep/modules/records/receivers.py
```diff
@@ -54,6 +54,7 @@
     get_push_access_tokens,
     get_orcids_for_push,
 )
+from elasticsearch import NotFoundError
```
ammirate: In python there is a convention for import statements order: 
```
Standard library imports.
Related third party imports.
Local application/library specific imports.
You should put a blank line between each group of imports.
```
Sooo, you should move this guy upper :stuck_out_tongue: 

https://www.python.org/dev/peps/pep-0008/#imports

inspirehep/modules/records/receivers.py
```diff
@@ -157,19 +158,21 @@ def index_after_commit(sender, changes):
...
-            if change in ('insert', 'update'):
+            if change in ('insert', 'update') and not model_instance.json.get("deleted"):
                 indexer.index(Record(model_instance.json, model_instance))
             else:
-                indexer.delete(Record(model_instance.json, model_instance))
-
+                try:
+                    indexer.delete(Record(model_instance.json, model_instance))
+                except NotFoundError:
+                    pass
```
ammirate: I would add a comment at least here, probably we might also want to log

tests/integration/workflows/test_edit_article.py
```diff
@@ -36,6 +36,8 @@
 
 from calls import do_robotupload_callback
 from factories.db.invenio_records import TestRecordMetadata
+from inspirehep.modules.search import LiteratureSearch
+from elasticsearch import NotFoundError
```
ammirate: same here :point_up_2:

monaawi:
ci test please

inspirehep/modules/records/cli.py
```diff
@@ -109,13 +120,15 @@ def simpleindex(yes_i_know, pid_type, batch_size, queue_name):
     click.secho('Sending record UUIDs to the indexing queue...', fg='green')
 
     query = (
-        db.session.query(PersistentIdentifier.object_uuid)
+        db.session.query(PersistentIdentifier.object_uuid).join(RecordMetadata, type_coerce(PersistentIdentifier.object_uuid, String) == type_coerce(RecordMetadata.id, String))
         .filter(
             PersistentIdentifier.pid_type.in_(pid_type),
             PersistentIdentifier.object_type == 'rec',
             PersistentIdentifier.status == PIDStatus.REGISTERED,
+            or_(RecordMetadata.json["deleted"].is_(None), RecordMetadata.json["deleted"] == cast(False, JSONB))
```
michamos: * I don't think that's the right query. We never have `deleted=None`. You probably want something like `not_(type_coerce(RecordMetadata.json, JSONB).has_key('deleted'))` instead.
* Another issue is that this query will be very expensive, as we currently don't have an index on `deleted`, so it's a full table scan. As only a small fraction of records is deleted, wouldn't it be better to filter them out in Python instead?

monaawi: I added deleted None because I had checked previously and we do have deleted None and I also tested out the query and it works. 
Filtering them out in python sounds good though, I will look into it

monaawi: yes, you are right. If the deleted flag doesn't exist, then it returns None

michamos: It's strange that there is no distinction between a `{"foo": null}` and `{}` in JSONB, but that seems to be the behavior here. I would still use the my suggestion as it expresses intent more clearly.

michamos: >     * And I was thinking coercing whole record could take time, but no idea how these `JSON` blobs work.

`type_coercing` doesn't do anything with the data. It just tells SQLAlchemy that the thing you have in your hands actually has type `JSONB` contrarily to what it would have inferred (due to invenio also supporting different databases that don't have `JSONB`), so you can use `JSONB` methods on it.

monaawi:
ci test please

monaawi:
ci test please

monaawi:
ci test please

monaawi:
ci test please

ammirate:
ci merge please

PR #3642, opened 2018-09-18
Inspir-1338 Cli function for handling duplicates

pazembrz:
Cli function for searching, listing and removing broken records

`inspirehep handle_duplicates --help`


pazembrz:
ci test please

pazembrz:
ci test please

inspirehep/modules/records/cli.py
```diff
@@ -228,3 +229,61 @@ def _finished_tasks_count():
...
+
+@click.command()
+@click.option('--remove-no-control-number', is_flag=True)
+@click.option('--remove-duplicates', is_flag=True)
+@click.option('--remove-not-in-pidstore', is_flag=True)
+@click.option('-c', '--print-without-control-number', is_flag=True)
+@click.option('-p', '--print-pid-not-in-pidstore', is_flag=True)
+@click.option('-d', '--print-duplicates', is_flag=True)
+@with_appcontext
+def handle_duplicates(remove_no_control_number, remove_duplicates,
```
szymonlopaciuk: Seems to me like this command is doing two related (but sort of separate) things, either printing records that satisfy certain criteria (w/o control number, not in pidstore, duplicates) or deleting them. Maybe instead of having those flags, it would make sense to introduce two separate commands like `show` and `remove` with the same parameters instead?

Think `handle_duplicates show -c -p` and `handle_duplicates remove -d`. Otherwise, in theory you can do `handle_duplicates --remove-no-control-number --print-without-control-number`?

At the end of the day I don't really know the user story that much, so this is just a suggestion.

pazembrz: In theory removing of any record from DB should be hard and should require from the user additional work so she/he have time to think what he is doing before pressing enter :)

pazembrz:
ci test please

pazembrz:
ci merge please

pazembrz:
ci merge please

PR #3643, opened 2018-09-18
INSPIR-1299 Fix 0 citations bug in search on some records 

pazembrz:
Fixed issue where not all records were properly re-indexed because of wrong data passed

### DIFF
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -600,12 +600,14 @@ class InspireRecord(Record):
         return absolute_url(u'/api/{endpoint}/{control_number}'.format(endpoint=endpoint,
                                                                        control_number=pid_value))
 
-    def _query_citing_records(self):
+    def _query_citing_records(self, session=None):
         """Returns records which cites this one."""
+        if not session:
+            session = db.session
         ref = self._get_ref()
         if not ref:
             raise Exception("There is no $ref for this object")
-        citations = RecordMetadata.query.with_entities(RecordMetadata.id).filter(
+        citations = session.query(RecordMetadata).with_entities(RecordMetadata.id).filter(
             referenced_records(RecordMetadata.json).contains([ref]))
         return citations
 
@@ -613,9 +615,10 @@ class InspireRecord(Record):
     def get_citing_records_query(self):
         return self._query_citing_records()
 
-    def get_citations_count(self):
+    def get_citations_count(self, session=None):
         """Returns citations count for this record."""
-        count = self.get_citing_records_query.count()
+
+        count = self._query_citing_records(session).count()
         return count
 
     def dumps(self):
--- a/inspirehep/modules/records/exceptions.py
+++ b/inspirehep/modules/records/exceptions.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of INSPIRE.
+# Copyright (C) 2018 CERN.
+#
+# INSPIRE is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# INSPIRE is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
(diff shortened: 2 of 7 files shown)


pazembrz:
ci test please

inspirehep/modules/records/receivers.py
```diff
@@ -186,7 +185,7 @@ def enhance_after_index(sender, json, *args, **kwargs):
         populate_authors_full_name_unicode_normalized(json)
         populate_inspire_document_type(json)
         populate_name_variations(json)
-        populate_citations_count(sender, json)
+        populate_citations_count(record=record, json=json)
 
     elif is_author(json):
         populate_authors_name_variations(json)
```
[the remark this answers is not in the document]
pazembrz: I won't touch that as I do not want to make a mess. Create a new task for this issue to make this fix the proper way (with the tests, and checking if it is not breaking anything etc.)

pazembrz:
ci test please

pazembrz:
ci test please

pazembrz:
ci test please

pazembrz:
ci test please

pazembrz:
ci test please

inspirehep/modules/records/api.py
```diff
@@ -600,22 +600,25 @@ def _get_ref(self):
...
+            session = db.session
         ref = self._get_ref()
         if not ref:
             raise Exception("There is no $ref for this object")
-        citations = RecordMetadata.query.with_entities(RecordMetadata.id).filter(
+        citations = session.query(RecordMetadata).with_entities(RecordMetadata.id).filter(
             referenced_records(RecordMetadata.json).contains([ref]))
         return citations
 
     @property
```
drjova: this can be removed

inspirehep/modules/records/api.py
```diff
@@ -600,22 +600,25 @@ def _get_ref(self):
...
     def get_citing_records_query(self):
         return self._query_citing_records()
```
drjova: we can remove this method

inspirehep/modules/records/exceptions.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+#
```
drjova: usually we use ``errors.py`` to be consistent with invenio :)

inspirehep/modules/records/exceptions.py
```diff
@@ -0,0 +1,31 @@
...
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+
+class BasicRecordException(Exception):
```
drjova: what about using this what about using https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/errors.py#L14

inspirehep/modules/records/api.py
pazembrz: Which one? Both `_query_citing_records` and `get_citing_records_query` are used in the code.

drjova: the property ``get_citing_records_query`` we don't handle the session

inspirehep/modules/records/exceptions.py
pazembrz: Except my exceptions.py in records there are only two files with Exceptions, one errors.py for Workflows and one exceptions.py for orcid. So for current code we can say that `usually` we use exceptions.py ;) But to be honest I do not see any written information where we should declare exception types/classes.

drjova: That's true we don't have any written information, I'll stick with the invenio naming just because we are extending its functionality especially in this case.

inspirehep/modules/records/api.py
pazembrz: But we use it in the `LiteratureCitationsResource` view, and in there we do not need to handle the session as it always should be correct. Replacing it for the `_query_citing_records` in this view would also require to change `_query_citing_records` method name to `query_citing_records` which would also require to change call to it in `get_citations_count`. Do we really want to do that?

drjova: I missed ``LiteratureCitationsResource`` 👍 ignore my comment

inspirehep/modules/records/exceptions.py
pazembrz: You are right that as we are extending the functionality of Invenio we should with the invenio naming...
BUT actually I do not like to create exceptions in errors.py file as it's simple example of code obfuscation. (Who is looking for exceptions in file called errors?)
Also naming exceptions as errors looks like reducing the importance of it. We shouldn't repeat Invenio mistakes/problems IMHO.

pazembrz:
ci test please

pazembrz:
ci merge please

PR #3644, opened 2018-09-19
citation_count: index new citations records - INSPIR-1127

ammirate:
## Checklist:

### DIFF
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -48,7 +48,7 @@ from sqlalchemy.sql.functions import GenericFunction
 
 from inspirehep.modules.pidstore.minters import inspire_recid_minter
 from inspirehep.modules.pidstore.utils import get_pid_type_from_schema, get_endpoint_from_pid_type
-from inspirehep.modules.records.utils import populate_earliest_date
+from inspirehep.modules.records.utils import get_pid_from_record_uri, populate_earliest_date
 from inspirehep.utils.record_getter import (
     RecordGetterError,
     get_es_record_by_uuid
@@ -647,6 +647,49 @@ class InspireRecord(Record):
     def to_dict(self):
         """Gets a deep copy of the record's json."""
         return deepcopy(dict(self))
+
+    def get_modified_references(self):
+        """Return the ids of the references diff between the latest and the
+        previous version.
+
+        The diff includes references added or deleted. Changes in a
+        reference's content won't be detected.
+
+        Also, it detects if record was deleted/un-deleted compared to the
+        previous version and, in such cases, returns the full list of
+        references.
+
+        References not linked to any record will be ignored.
+
+        Note: record should be committed to DB in order to correctly get the
+        previous version.
+
+
+        Returns:
+            Set[Tuple[str, int]]: pids of references changed from the previous
+            version.
+        """
+        def _get_ids_from_refs(references):
+            return set([
+                get_pid_from_record_uri(ref['record']['$ref'])
+                for ref in references
+                if 'record' in ref
+            ])
+
+        try:
+            prev_version = self.model.versions[-2].json
+        except IndexError:
+            prev_version = {}
+
+        changed_deleted_status = self.get('deleted', False) ^ prev_version.get('deleted', False)
+
+        if changed_deleted_status:
+            return _get_ids_from_refs(self.get('references', []))
+
+        ids_latest = _get_ids_from_refs(self.get('references', []))
+        ids_oldest = _get_ids_from_refs(prev_version.get('references', []))
+
+        return set.symmetric_difference(ids_latest, ids_oldest)
 
 
 class ESRecord(InspireRecord):
--- a/inspirehep/modules/records/receivers.py
+++ b/inspirehep/modules/records/receivers.py
@@ -31,10 +31,15 @@ from celery import Task
 from flask import current_app
 from flask_sqlalchemy import models_committed
 from elasticsearch import NotFoundError
+from sqlalchemy import tuple_
 
+
+from invenio_db import db
 from invenio_indexer.signals import before_record_index
+from invenio_pidstore.models import PersistentIdentifier
 from invenio_records.models import RecordMetadata
 from invenio_records.signals import (
+    after_record_insert,
     after_record_update,
     before_record_insert,
     before_record_update,
@@ -48,6 +53,8 @@ from inspirehep.modules.orcid.utils import (
     get_orcids_for_push,
 )
 from inspirehep.modules.records.api import InspireRecord
+from inspirehep.modules.records.indexer import InspireRecordIndexer
+from inspirehep.modules.records.tasks import batch_reindex
 from inspirehep.modules.records.utils import (
     is_author,
     is_book,
@@ -71,7 +78,7 @@ from inspirehep.modules.records.utils import (
     populate_recid_from_ref,
     populate_title_suggest,
 )
-from inspirehep.modules.records.indexer import InspireRecordIndexer
+
 
 LOGGER = logging.getLogger(__name__)
 
@@ -153,6 +160,7 @@ def index_after_commit(sender, changes):
     has been really committed to the DB.
     """
     indexer = InspireRecordIndexer()
+
     for model_instance, change in changes:
         if isinstance(model_instance, RecordMetadata):
             if change in ('insert', 'update') and not model_instance.json.get("deleted"):
@@ -206,3 +214,27 @@ def enhance_after_index(sender, json, record, *args, **kwargs):
 
     elif is_data(json):
         populate_citations_count(record=record, json=json)
+
+
+@after_record_insert.connect
+@after_record_update.connect
+def index_new_cited_records_after_record_update(sender, record, *args, **kwargs):
+    """Index records whose reference has been added or deleted"""
+    pids = record.get_modified_references()
+
+    if not pids:
+        return
+
+    uuids = [
+        pid.object_uuid for pid in
+        db.session.query(PersistentIdentifier.object_uuid).filter(
+            PersistentIdentifier.object_type == 'rec',
+            tuple_(PersistentIdentifier.pid_type, PersistentIdentifier.pid_value).in_(pids)
+        )
+    ]
+
+    if uuids:
+        return batch_reindex.apply_async(
+            kwargs={'uuids': uuids},
+            queue='bulk_index',
+        )
--- a/inspirehep/modules/records/tasks.py
+++ b/inspirehep/modules/records/tasks.py
@@ -34,8 +34,8 @@ from sqlalchemy.orm.exc import NoResultFound
 from invenio_db import db
 from invenio_pidstore.models import PersistentIdentifier
 from invenio_search import current_search_client as es
-
 from inspire_dojson.utils import get_recid_from_ref
+
 from inspirehep.modules.records.api import InspireRecord
 from inspirehep.modules.records.utils import get_endpoint_from_record
 from inspirehep.modules.pidstore.utils import get_pid_type_from_schema
@@ -182,7 +182,7 @@ def get_merged_records():
 
 
 @shared_task(ignore_result=False, max_retries=0)
-def batch_reindex(uuids, request_timeout):
+def batch_reindex(uuids, request_timeout=None):
     """Task for bulk reindexing records."""
     def actions():
         for uuid in uuids:
@@ -191,6 +191,9 @@ def batch_reindex(uuids, request_timeout):
                 yield create_index_op(record, version_type='force')
             except NoResultFound as e:
                 logger.warn('Record %s failed to load: %s', uuid, e)
+
+    if not request_timeout:
+        request_timeout = current_app.config['INDEXER_BULK_REQUEST_TIMEOUT']
 
     success, failures = bulk(
         es,
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -24,13 +24,12 @@ from __future__ import absolute_import, division, print_function
 
 import sys
 import os
-
+import mock
 import pytest
 import sqlalchemy
+
 from flask_alembic import Alembic
-
 from functools import partial
-
 from click.testing import CliRunner
 from flask import current_app
 from flask.cli import ScriptInfo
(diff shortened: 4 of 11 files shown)


ammirate:
ci test please

ammirate:
ci test please

ammirate:
ci test please

ammirate:
ci test please

inspirehep/modules/records/utils.py
```diff
@@ -430,3 +430,43 @@ def populate_author_suggest(json, *args, **kwargs):
...
+            diff.append(list1.pop(0))
+        elif item1 > item2:
+            diff.append(list2.pop(0))
+        else:
+            list1.pop(0)
+            list2.pop(0)
+
+    diff.extend(list1)
+    diff.extend(list2)
+    return diff
```
drjova: maybe something like ``set(a) - set(b)`` for the differences is enough?

tests/integration/records/test_records_api.py
```diff
@@ -180,3 +183,157 @@ def test_not_count_citations_from_other_collections(isolated_app):
...
+        'document_type': ['article'],
+        'titles': [{'title': 'Such a cool title'}],
+        '_collections': ['Literature'],
+    }
+
+    record = InspireRecord.create(json, skip_files=True)
+    record.commit()
+    db.session.commit()
+
+    metadata = deepcopy(dict(record))
```
drjova: ``record.dumps()`` :)

tests/integration/records/test_records_receivers.py
```diff
@@ -36,17 +37,30 @@
...
 
+def remove_record_from_db(record):
+    pid = PersistentIdentifier.get('lit', record['control_number'])
+    recid = RecordIdentifier.query.filter_by(recid=record['control_number']).one()
+    db.session.delete(record.model)
+    db.session.delete(pid)
+    db.session.delete(recid)
+    db.session.commit()
+
+
```
drjova: https://github.com/inspirehep/inspire-next/blob/bf85efbbdb57d94e987d1714e383420a82510c19/tests/integration/helpers/utils.py#L39

inspirehep/modules/records/receivers.py
```diff
@@ -206,3 +215,45 @@ def enhance_after_index(sender, json, record, *args, **kwargs):
 
     elif is_data(json):
         populate_citations_count(record=record, json=json)
+
+
+@after_record_insert.connect
+@after_record_update.connect
+def after_update(sender, *args, **kwargs):
```
turtle321: I'd name this function after what it does rather than when it does it.
Or maybe follow the patter: what_when, like `def reindex_referenced_records_after_record_update(...)`

inspirehep/modules/records/receivers.py
```diff
@@ -206,3 +215,45 @@ def enhance_after_index(sender, json, record, *args, **kwargs):
...
+    json = kwargs.get('record')
```
turtle321: `json_data` (avoid the name clash with the `json` module)

inspirehep/modules/records/receivers.py
```diff
@@ -206,3 +215,45 @@ def enhance_after_index(sender, json, record, *args, **kwargs):
...
+        t[0] for t in db.session.query(PersistentIdentifier.object_uuid)
+        .filter(PersistentIdentifier.pid_type == 'lit')
+        .filter(PersistentIdentifier.pid_value.in_(pids))
+        .all()
+    ]
+
+    timeout = current_app.config.get('INDEXER_BULK_REQUEST_TIMEOUT')
+
+    if uuids:
+        return batch_reindex.apply_async(
```
turtle321: I'm thinking about the overall impact of this. Most cited papers might end up being reindex too often. I suggest to monitor the performance of this and in case use debouncing (bulk more reindex requests within a short timeframe in a single reindex)

inspirehep/modules/records/receivers.py
```diff
@@ -206,3 +215,45 @@ def enhance_after_index(sender, json, record, *args, **kwargs):
...
+
+    if json.get('deleted'):
+        # in this way the references diff will be the list of references itself
+        record = InspireRecord(json)
+    else:
+        pid_type = get_pid_type_from_schema(json['$schema'])
+        pid_value = json['control_number']
+        try:
+            rec_uuid = PersistentIdentifier.get(pid_type, pid_value).object_uuid
+            record = InspireRecord.get_record(rec_uuid)
```
michamos: isn't this basically `get_db_record`?

inspirehep/modules/records/receivers.py
michamos: more explicit name?

tests/integration/records/test_records_api.py
ammirate: `record.dumps` adds extra fields. Maybe `record.to_dict()` best suits :+1:

tests/integration/records/test_records_receivers.py
ammirate: yeah, it was not here before :P

inspirehep/modules/records/receivers.py
ammirate: Good point :+1:

inspirehep/modules/records/receivers.py
```diff
@@ -206,3 +215,45 @@ def enhance_after_index(sender, json, record, *args, **kwargs):
...
+        except NoResultFound:
+            # deleted record
+            record = InspireRecord(json)
+
+    modified_references = record.get_modified_references()
+
+    pids = [
+        str(get_recid_from_ref(ref['record'])) for ref in modified_references
```
michamos: this should really use `get_pid_from_record_uri` so it also works with citations to data: https://github.com/inspirehep/inspire-next/blob/b4e9b80149dbf7d6a922742753fd9b71e845a856/inspirehep/modules/records/utils.py#L84-L90.

inspirehep/modules/records/utils.py
ammirate: Nope, Set difference is different from symmetric difference, which is what we want.

https://en.wikipedia.org/wiki/Symmetric_difference

turtle321: I think you want to use `set.symmetric_difference` but maybe you'll have a problem with non hashable types.  In that case you might convert to string.
Also keep in mind the bad performance of deleting from the head with Python lists (they are dynamic arrays that need to be shifted in that case).

inspirehep/modules/records/receivers.py
```diff
@@ -206,3 +215,45 @@ def enhance_after_index(sender, json, record, *args, **kwargs):
...
+    ]
+    uuids = [
+        t[0] for t in db.session.query(PersistentIdentifier.object_uuid)
+        .filter(PersistentIdentifier.pid_type == 'lit')
+        .filter(PersistentIdentifier.pid_value.in_(pids))
+        .all()
```
michamos: * for `dat` to work too, you need to do something like https://github.com/inspirehep/inspire-next/blob/4d116e13e625dc4ea4a8b9a3ab2bcf6261ecec77/inspirehep/utils/record_getter.py#L132-L135
* `t.object_uuid` is probably clearer than `t[0]`.
* **For later** we should really think hard about how to organize DB queries, currently they happen everywhere and they are often partially duplicated.

inspirehep/modules/records/receivers.py
```diff
@@ -206,3 +215,45 @@ def enhance_after_index(sender, json, record, *args, **kwargs):
...
+    ]
+
+    timeout = current_app.config.get('INDEXER_BULK_REQUEST_TIMEOUT')
```
michamos: is there a good reason for this timeout not to be the default of `batch_reindex` in case no explicit timeout is passed?

inspirehep/modules/records/receivers.py
```diff
@@ -206,3 +215,45 @@ def enhance_after_index(sender, json, record, *args, **kwargs):
...
+
+    if uuids:
+        return batch_reindex.apply_async(
+            kwargs={
+                'uuids': uuids,
+                'request_timeout': timeout,
+            },
+            queue='indexer_task',
```
michamos: **Not for this PR** similar question, why does the caller need to know the queue? we should use [routers](http://docs.celeryproject.org/en/latest/userguide/routing.html#routers) in the config instead.

tests/integration/records/test_records_api.py
```diff
@@ -180,3 +183,157 @@ def test_not_count_citations_from_other_collections(isolated_app):
     TestRecordMetadata.create_from_kwargs(json=ref)
 
     assert record_1.get_citations_count() == 1
+
+
+def test_get_modified_references_when_prev_version_has_zero(app):
+    json = {
```
turtle321: Maybe I'm missing something (like the proper signals are not triggered when using factories??), but to me it looks like you could have used the factories.

json_data = []
json_data['references'] = [...]
factory = TestRecordMetadata.create_from_kwargs(json=json_data)
# factory.inspire_record  # if you need to access the InspireRecord instance

inspirehep/modules/records/receivers.py
michamos: an alternative would be to send increments/decrements to ES instead of full reindexing: https://www.elastic.co/guide/en/elasticsearch/reference/5.4/docs-update.html#_scripted_updates

inspirehep/modules/records/api.py
```diff
@@ -648,6 +651,28 @@ def to_dict(self):
...
+        Args:
+            rec_uuid (int): the record identifier.
+
+        Returns:
+            list: references changed from the previous version.
+        """
+        latest = self.get('references', [])
+
+        try:
+            oldest = self.model.versions[-2].json.get('references', [])
```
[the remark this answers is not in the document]
ammirate: The receiver is called after commit, do we really need it?
However, there are tests that are testing this behavior, do you think we should add more or add a better doc?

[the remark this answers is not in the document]
michamos: Maybe an explanation in the docstring is sufficient then.

inspirehep/modules/records/utils.py
ammirate: done

tests/integration/records/test_records_api.py
ammirate: done

inspirehep/modules/records/receivers.py
ammirate: done

inspirehep/modules/records/api.py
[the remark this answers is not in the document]
ammirate: done

inspirehep/modules/records/api.py
```diff
@@ -648,6 +651,28 @@ def to_dict(self):
...
+        except (AttributeError, IndexError):
```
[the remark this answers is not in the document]
ammirate: done

inspirehep/modules/records/receivers.py
```diff
@@ -206,3 +215,45 @@ def enhance_after_index(sender, json, record, *args, **kwargs):
...
     elif is_data(json):
         populate_citations_count(record=record, json=json)
+
+
+@after_record_insert.connect
+@after_record_update.connect
+def after_update(sender, *args, **kwargs):
+    json = kwargs.get('record')
+
+    if json.get('deleted'):
```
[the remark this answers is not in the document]
ammirate: done

inspirehep/modules/records/receivers.py
ammirate: :+1:

inspirehep/modules/records/receivers.py
ammirate: done

inspirehep/modules/records/receivers.py
ammirate: done

inspirehep/modules/records/receivers.py
ammirate: :+1:

inspirehep/modules/records/receivers.py
ammirate: done

ammirate:
ci test please

inspirehep/modules/records/receivers.py
```diff
@@ -206,3 +214,27 @@ def enhance_after_index(sender, json, record, *args, **kwargs):
...
+@after_record_update.connect
+def index_new_cited_records_after_record_update(sender, record, *args, **kwargs):
+    """Index records whose reference has been added or deleted"""
+    pids = record.get_modified_references()
+
+    if not pids:
+        return
+
+    uuids = [
+        pid[0] for pid in
```
michamos: `pid.object_uuid` is probably prettier.

inspirehep/modules/records/receivers.py
```diff
@@ -206,3 +214,27 @@ def enhance_after_index(sender, json, record, *args, **kwargs):
...
+        db.session.query(PersistentIdentifier.object_uuid).filter(
+            PersistentIdentifier.object_type == 'rec',
+            tuple_(PersistentIdentifier.pid_type, PersistentIdentifier.pid_value).in_(pids)
+        ).all()
```
michamos: minor: I think results of queries are iterable, so you probably don't need `.all()` here.

inspirehep/modules/records/tasks.py
```diff
@@ -182,7 +182,7 @@ def _get_uuids_to_merge():
 
 
 @shared_task(ignore_result=False, max_retries=0)
-def batch_reindex(uuids, request_timeout):
+def batch_reindex(uuids, request_timeout=900):
```
michamos: why 900? shouldn't this be read from config?

tests/integration/conftest.py
```diff
@@ -53,6 +52,7 @@
 )  # noqa
 
 
+@mock.patch('inspirehep.modules.records.receivers.batch_reindex.apply_async')
```
michamos: maybe it's better to localize this to the migration of records, otherwise you're patching this for all tests using the `app` fixture.

tests/integration/records/test_records_receivers.py
```diff
@@ -445,3 +448,344 @@ def test_check_enhance_after_index_receiver_when_record_not_provided(isolated_ap
...
+    db.session.commit()
+
+    expected_kwargs = {'uuids': [cited.id]}
+
+    mocked_indexing_task.assert_called_with(
+        kwargs=expected_kwargs,
+        queue='indexer_task'
+    )
+    # execute mocked task
+    batch_reindex(**expected_kwargs)
```
michamos: you could also add this as a `side_effect` to the mock (as you wish).

ammirate:
ci merge please

ammirate:
ci merge please

PR #3647, opened 2018-09-21
Sort citations by earliest_date

monaawi:
Replaced the db query with an elastic search query which contains pagination and sorting. Added earliest_date to the serializer.
Removed citations_count from the literature endpoint as it was not used.

## Checklist:

### DIFF
--- a/inspirehep/modules/records/views.py
+++ b/inspirehep/modules/records/views.py
@@ -25,13 +25,13 @@
 from __future__ import absolute_import, division, print_function
 
 from flask import Blueprint, request, abort
-from inspirehep.modules.records.api import InspireRecord
 from invenio_rest.views import ContentNegotiatedMethodView
 from invenio_records_rest.views import pass_record
 
 from inspirehep.modules.search.search_factory import inspire_facets_factory
 from .serializers import json_literature_citations_v1_response, \
     json_literature_search_aggregations_ui_v1
+from inspirehep.modules.records.utils import get_citations_from_es
 
 blueprint = Blueprint(
     'inspirehep_records',
@@ -62,14 +62,13 @@ class LiteratureCitationsResource(ContentNegotiatedMethodView):
         if page < 1 or size < 1:
             abort(400)
 
-        citing_records_results = record.get_citing_records_query.paginate(
-            page, size, False)
+        citing_records_results = get_citations_from_es(record, page, size)
         citing_records_count = citing_records_results.total
-        citing_records_uuids = [result[0] for result in
-                                citing_records_results.items]
-        citing_records = InspireRecord.get_records(citing_records_uuids)
+        citing_records = [citation.to_dict() for citation in citing_records_results]
+
         data = {'citations': citing_records,
                 'citation_count': citing_records_count}
+
         return self.make_response(pid, data)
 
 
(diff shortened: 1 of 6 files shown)


inspirehep/modules/records/views.py
```diff
@@ -40,6 +40,22 @@
...
+            'control_number',
+            'earliest_date',
+            'titles',
+            'publication_info'
+        ],
+        from_=(page - 1) * size,
+        size=size,
+    ).sort('-earliest_date').execute().hits
+
+
```
drjova: I'd move this in utils

inspirehep/modules/records/views.py
```diff
@@ -40,6 +40,22 @@
 )
 
 
+def get_citations_from_es(record, page, size):
```
drjova: it's better to have default page=1, size=10 for example

inspirehep/modules/records/views.py
```diff
@@ -40,6 +40,22 @@
...
+    return LiteratureSearch().query(
```
drjova: what will happen if record is ``None`` or ``{}``?

monaawi:
ci test please

inspirehep/utils/citations.py
```diff
@@ -61,3 +61,22 @@ def get_and_format_citations(record):
         result.append(row)
 
     return result
+
+
+def get_citations_from_es(record, page=1, size=10):
```
drjova: it's a bit cleaner to use: 
```python
if 'control_number' not in record:
    return {}
return LiteratureSearch().query( ....
```

inspirehep/utils/citations.py
```diff
@@ -61,3 +61,23 @@ def get_and_format_citations(record):
...
+    if 'control_number' not in record:
+        return []
```
drjova: my bad, it's better to return ``None``. It doesn't affect the api because we handle the NotExistingRecord cases.

inspirehep/utils/citations.py
```diff
@@ -61,3 +61,23 @@ def get_and_format_citations(record):
         result.append(row)
 
     return result
+
+
+def get_citations_from_es(record, page=1, size=10):
```
drjova: I just noticed! We already have a ``utils`` module here: https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/utils.py which are specific for the records

tests/integration/records/test_records_views.py
```diff
@@ -24,22 +24,33 @@
...
         'titles': [
             {
                 'title': 'Jessica Jones',
             },
         ],
+        '_collections': ['Literature']
     }
-    record = TestRecordMetadata.create_from_kwargs(
-        json=record_json).inspire_record
+    record = InspireRecord.create(record_json)
```
drjova: by using ``TestRecordMetadata`` the record is indexed, is not enough for our test cases?

monaawi: I wanted to use TestRecordMetadata at first but then I looked through the other tests that we have and that use elastic search and they did it like this so I wanted to keep it consistent with what we had so far.

drjova: The other tests are might be old, we are trying to use ``TestRecordMetadata`` in the new tests and while we refactoring.

monaawi:
ci test please

monaawi:
ci test please

drjova:
ci merge please

PR #3649, opened 2018-09-24
suggestors: specify fiels to include in _source

harunurhan:
For now, it only includes fields we need on author submission form.


drjova:
ci merge please

PR #3666, opened 2018-10-02
config: addition of ``_source`` in authors suggest

drjova:


drjova:
@michamos at the moment we use ``name.preferred_name`` for advisors autocomplete, is that ok?

michamos:
I think `value` is better

PR #3686, opened 2018-10-11
1140 app metrics: add timeexecution and health endpoints

turtle321:


turtle321:
Related PR in Puppet:
https://gitlab.cern.ch/ai/it-puppet-hostgroup-inspire/merge_requests/80

inspirehep/config.py
```diff
@@ -1623,3 +1623,10 @@
                'arxiv_eprints', 'public_notes', 'number_of_pages',
                'publication_info', 'earliest_date']
 }
+
+
+# App metrics
+# ===========
+APPMETRICS_ENABLED = False
```
michamos: is this a feature flag? then it should follow the pattern and be something like `FEATURE_FLAG_ENABLE_APPMETRICS` and be defined close to the others.

inspirehep/modules/theme/views.py
```diff
@@ -344,6 +345,24 @@ def ping():
     return 'OK'
 
 
+@blueprint.route('/health')
+@time_execution
+def health():
```
michamos: not sure this is the right place for it, as `theme` is the old ui that is being replaced by the new react one. @drjova ?

ammirate: Same for the above `ping`, which is used from nagios to check nginx status.
Shell we move them both somewhere else?

inspirehep/config.py
turtle321: True that. Done.

inspirehep/modules/theme/views.py
turtle321: And this is also the place where we serve the root '/'. So we will move them together, but not in this PR.

ammirate:
ci merge please

turtle321:
ci merge please

PR #3687, opened 2018-10-11
citation count: index new citation's records - INSPIR-1127

ammirate:
Changes in the current receiver `models_committed` to trigger a celery task, which computes the references diff between the current and previous version of the record and indexes the cited records.
However, before computing the references diff, the task checks that the current version of the record in the DB is the one it expects to find (passed before by the receiver). If it is not, it retries itself with exponential time.
A new celery queue called `bulk_index` is used for reindexing the cited records.

### DIFF
--- a/inspirehep/modules/records/tasks.py
+++ b/inspirehep/modules/records/tasks.py
@@ -29,7 +29,8 @@ from celery.utils.log import get_task_logger
 from elasticsearch.helpers import bulk, scan
 from flask import current_app
 from six import iteritems
-from sqlalchemy.orm.exc import NoResultFound
+from sqlalchemy import tuple_
+from sqlalchemy.orm.exc import NoResultFound, StaleDataError
 
 from invenio_db import db
 from invenio_pidstore.models import PersistentIdentifier
@@ -40,6 +41,7 @@ from inspirehep.modules.records.api import InspireRecord
 from inspirehep.modules.records.utils import get_endpoint_from_record
 from inspirehep.modules.pidstore.utils import get_pid_type_from_schema
 from inspirehep.utils.record import create_index_op
+from inspirehep.utils.record_getter import get_db_record
 
 
 logger = get_task_logger(__name__)
@@ -182,7 +184,7 @@ def get_merged_records():
 
 
 @shared_task(ignore_result=False, max_retries=0)
-def batch_reindex(uuids, request_timeout):
+def batch_reindex(uuids, request_timeout=None):
     """Task for bulk reindexing records."""
     def actions():
         for uuid in uuids:
@@ -191,6 +193,9 @@ def batch_reindex(uuids, request_timeout):
                 yield create_index_op(record, version_type='force')
             except NoResultFound as e:
                 logger.warn('Record %s failed to load: %s', uuid, e)
+
+    if not request_timeout:
+        request_timeout = current_app.config['INDEXER_BULK_REQUEST_TIMEOUT']
 
     success, failures = bulk(
         es,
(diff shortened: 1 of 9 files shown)


inspirehep/modules/records/tasks.py
```diff
@@ -204,3 +210,38 @@ def actions():
         'success': success,
         'failures': [failure for failure in failures or []],
     }
+
+
+@shared_task(ignore_result=False, bind=True)
+def index_new_cited_records(self, pid_type, pid_value, db_version):
```
michamos: I hate to say this, but I think the name is not very clear, as you also index records that were cited but are not anymore. Maybe `index_modified_citations_from_record`? I would also add a docstring for good measure :)

inspirehep/modules/records/tasks.py
```diff
@@ -204,3 +210,38 @@ def actions():
...
+    record = get_db_record(pid_type, pid_value)
+
+    if record.model.version_id < db_version:
+        backoff = 2 ** (self.request.retries + 1)
+        eta = datetime.datetime.now() + datetime.timedelta(seconds=backoff)
```
michamos: Celery 4.2 has this builtin, but 4.1 doesn't seem to unfortunately.

inspirehep/modules/records/tasks.py
```diff
@@ -204,3 +210,38 @@ def actions():
...
+        logger.warn(
+            'Record {} not yet updated to version {} on DB'.format((pid_type, pid_value), db_version)
+        )
+        raise self.retry(eta=eta)
```
michamos: Here you can retry forever. Do you want to put a limit on the number of retries?

inspirehep/modules/records/tasks.py
```diff
@@ -204,3 +210,38 @@ def actions():
...
+    uuids = [
+        str(pid.object_uuid) for pid in
+        db.session.query(PersistentIdentifier.object_uuid).filter(
+            PersistentIdentifier.object_type == 'rec',
+            tuple_(PersistentIdentifier.pid_type, PersistentIdentifier.pid_value).in_(pids)
+        )
+    ]
+
+    if uuids:
+        return batch_reindex.apply_async(
```
michamos: can you really schedule a task from another task? I remember vaguely that celery was complaining about this kind of things during the 3->4 migration, but don't remember the specifics.

tests/integration/conftest.py
```diff
@@ -84,28 +83,30 @@ def app():
     )
 
     with app.app_context():
-        # Celery task imports must be local, otherwise their
-        # configuration would use the default pickle serializer.
-        from inspirehep.modules.migrator.tasks import migrate_from_file
+        with mock.patch('inspirehep.modules.records.receivers.index_new_cited_records.apply_async'):
```
michamos: you can also do `with context_manager1, context_manager2:`, so you don't have to reindent everything.

inspirehep/modules/records/tasks.py
ammirate: Leftover from the previous pr :+1:

inspirehep/modules/records/tasks.py
ammirate: yes, how many retries in your opinion? 
With 3 it would wait at least 2 + 4 + 8 = 14 seconds, which sounds reasonable for normal use cases.

inspirehep/modules/records/tasks.py
ammirate: Doh

ammirate:
ci test please

ammirate:
ci test please

inspirehep/modules/records/tasks.py
```diff
@@ -204,3 +209,41 @@ def actions():
         'success': success,
         'failures': [failure for failure in failures or []],
     }
+
+
+class MismatchDBVersionError(Exception):
```
michamos: maybe use https://docs.sqlalchemy.org/en/latest/orm/exceptions.html#sqlalchemy.orm.exc.StaleDataError instead?

ammirate:
ci merge please

PR #3691, opened 2018-10-16
author_count: use properly elasticsearch range INSPIR-1468

ammirate:
Related issue: https://its.cern.ch/jira/browse/INSPIR-1468

### DIFF
--- a/tests/integration/test_search_views.py
+++ b/tests/integration/test_search_views.py
@@ -24,7 +24,11 @@ from __future__ import absolute_import, division, print_function
 
 import json
 
+from invenio_db import db
+from invenio_search import current_search_client as es
 from mock import patch
+
+from inspirehep.modules.records.api import InspireRecord
 
 
 def test_search_conferences_is_there(app_client):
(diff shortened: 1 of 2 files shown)


tests/integration/test_search_views.py
```diff
@@ -379,3 +383,25 @@ def _debug(log_output):
...
+        '$schema': 'http://localhost:5000/schemas/records/hep.json',
+        'document_type': ['article'],
+        'titles': [{'title': 'Article with 10 authors'}],
+        '_collections': ['Literature'],
+        'authors': []
+    }
+    for i in range(10):
+        record_json['authors'].append({'full_name': 'Pincopallino' + str(i)})
+
+    rec = InspireRecord.create(data=record_json)
```
drjova: I'd use the factory for creating a record :)

monaawi: I think the factory wouldn't work here if he needs the record to be in ES as well. I am not sure though

drjova: hmm, It sends the record on ES if you pass ``index_name`` https://github.com/inspirehep/inspire-next/blob/master/tests/integration/helpers/factories/db/invenio_records.py#L120

monaawi: yes, we looked into that once and it didn't work (when I was working on the tests in test_records_views). And then we decided it was not in the scope of the story and that we will look into it

ammirate: I confirm, my first attempt used the factory and it did not work.

PR #3703, opened 2018-10-25
workflows: use new classifier and add vcr to tests

vbalbp:


vbalbp:
ci test please

inspirehep/modules/workflows/tasks/classifier.py
```diff
@@ -38,6 +41,77 @@
...
+    predictor_url = get_classifier_url()
+    if not predictor_url:
+        return
+
+    payload = prepare_payload(obj.data)
+
+    try:
+        results = json_api_request(predictor_url, payload)
+    except requests.exceptions.RequestException:
+        results = {}
```
michamos: you shouldn't ignore classifier errors so we can fix issues.

michamos:
ci test please

ammirate:
ci test please

vbalbp:
ci test please

drjova:
ci test please

michamos:
ci test please

drjova:
ci test please

drjova:
cit test please

drjova:
ci test please

michamos:
ci test please

inspirehep/config.py
```diff
@@ -1168,7 +1168,7 @@
 
 # Web services and APIs
 # =====================
-BEARD_API_URL = None  # e.g. "http://beard.inspirehep.net/api"
+CLASSIFIER_API_URL = None  # e.g. "http://inspire-classifier.web.cern.ch/api"
```
michamos: this will require a change in puppet.

inspirehep/modules/workflows/tasks/classifier.py
```diff
@@ -38,6 +40,74 @@
...
+    }
+    return payload
+
+
+@with_debug_logging
+def guess_coreness(obj, eng):
+    """Workflow task to ask inspire classifier for a coreness assessment."""
+    predictor_url = get_classifier_url()
+    if not predictor_url:
+        return None
```
michamos: just `return` as this is used for control flow only, the function never returns anything.

inspirehep/modules/workflows/tasks/classifier.py
```diff
@@ -38,6 +40,74 @@
...
+
+    payload = prepare_payload(obj.data)
+
+    results = json_api_request(predictor_url, payload)
+
```
michamos: I would remove some whitespace here.

inspirehep/modules/workflows/tasks/classifier.py
```diff
@@ -38,6 +40,74 @@
...
+    if results:
```
michamos: this is wrong as discussed IRL some time ago. You want to fail hard if the classifier doesn't work. So replace https://github.com/inspirehep/inspire-next/blob/855251f0d40436b12b3bd9c2ed5920315468f722/inspirehep/modules/workflows/utils/__init__.py#L77 with `response.raise_for_status()` and remove the `if results:` here.

tests/unit/workflows/test_workflows_tasks_classifier.py
```diff
@@ -51,6 +57,182 @@
...
+
+    obj = MockObj({}, {})
+    eng = MockEng()
+
+    assert guess_coreness(obj, eng) is None
+    assert 'relevance_prediction' not in obj.extra_data
+
+
+@patch('inspirehep.modules.workflows.tasks.classifier.get_classifier_url')
+def test_guess_coreness_does_not_fail_when_request_fails(mocked_get_classifier_url):
```
michamos: you'll need to change the test to assert that we get an exception when the request fails.

tests/unit/workflows/test_workflows_tasks_classifier.py
```diff
@@ -51,6 +57,182 @@
...
+
+@patch('inspirehep.modules.workflows.tasks.classifier.get_classifier_url')
+def test_guess_coreness_fails_when_request_fails(mocked_get_classifier_url):
+    mocked_get_classifier_url.return_value = 'https://inspire-classifier.web.cern.ch/api/predict/does-not-exist'
+
+    obj = MockObj({}, {})
+    eng = MockEng()
+
+    with pytest.raises(Exception):
+        assert guess_coreness(obj, eng) is None
```
michamos: you should remove the `assert` and just call `guess_coreness` here, as you're in fact asserting that calling `guess_coreness` raises here, not that the assertion you write is correct. It's also better to use the more specific `requests.RequestException` instead of `Exception`.

tests/unit/workflows/test_workflows_tasks_classifier.py
```diff
@@ -51,6 +57,182 @@
...
+        'title': 'Effects of top compositeness',
+        'abstract': 'We investigate the effects of (...)',
+    }
+    result = prepare_payload(record)
+
+    assert expected == result
+
+
+@patch('inspirehep.modules.workflows.tasks.classifier.get_classifier_url')
+def test_guess_coreness_fails_without_a_classifier_url(g_b_u):
```
michamos: `def test_guess_coreness_returns_early_without_a_classifier_url(mocked_get_classifier_url):`

tests/integration/workflows/conftest.py
```diff
@@ -187,7 +187,7 @@ def mocked_external_services(workflow_app):
             requests_mock.ANY,
             re.compile(
                 '.*' +
-                workflow_app.config['BEARD_API_URL'] +
+                workflow_app.config['CLASSIFIER_API_URL'] +
```
michamos: this is wrong, as the new classifier doesn't provide the `phonetic_blocks` service. On the other hand, I don't think it's used any longer, so you can just remove this mock.

PR #3713, opened 2018-11-01
Skip wrong entries in date fields, to not break whole batches when re…

pazembrz:
…indexing


michamos:
How can that happen? As far as I see, the solution to this is to align the mappings with the schemas:
- if a field is declared as a date by the schema, it *has* to be a valid date, so it's safe to map it as a date;
- if not, then it's just an opaque string, and it should not be mapped as a date.
Or is there anything I am overlooking here?

PR #3716, opened 2018-11-02
celery: bump to 4.2.0

ammirate:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,7 @@ install_requires = [
     'backoff~=1.0,>=1.4.3',
     'backports.tempfile>=1.0rc1',
     'beard~=0.0,>=0.2.0',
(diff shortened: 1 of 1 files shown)


michamos:
what about https://github.com/inveniosoftware/invenio-celery/issues/53 ?

ammirate:
That's what I wanted to check :point_up:

ammirate:
Maybe we can downgrade to 4.0.2? The timezone bug is really annoying

michamos:
> Maybe we can downgrade to 4.0.2? The timezone bug is really annoying

I know, we investigated it back when we upgraded from celery 3, but there were real blockers in 4.0.*, so we chose the lesser evil.

ammirate:
Close this for now.

PR #3718, opened 2018-11-05
accounts: add module and local login

harunurhan:
Adds noew account module with POST login view to be used only locally
and for testing


inspirehep/config.py
```diff
@@ -687,6 +689,12 @@
     },
     'suggesters': {
         'journal_title': {
+            '_source': [
+                'short_title',
+                'title',
+                'control_number',
+                'self',
+            ],
```
drjova: do we really need this here?

tests/integration/accounts/test_views.py
```diff
@@ -0,0 +1,121 @@
...
+        'email': 'test_guy@inspirehep.net',
+        'password': '123123'
+    }
+    user = User(
+        email=user_json['email'],
+        password=hash_password(user_json['password']),
+        active=True,
+    )
+    db.session.add(user)
+    db.session.commit()
```
drjova: probably you don't need this

harunurhan:
ci merge please

PR #3719, opened 2018-11-05
workflow: add new endpoint to get linked references INSPIR-1529

ammirate:


inspirehep/utils/record_getter.py
```diff
@@ -66,6 +66,10 @@ def wrapper(*args, **kwargs):
 
 @raise_record_getter_error_and_log
 def get_es_record(pid_type, recid, **kwargs):
+
+    import pytest
+    pytest.set_trace()
```
drjova: oups

inspirehep/modules/workflows/api.py
```diff
@@ -0,0 +1,39 @@
...
+
+from inspire_utils.dedupers import dedupe_list
+
+from inspirehep.modules.workflows.tasks.refextract import (
+    extract_references_from_raw_refs,
+    match_references,
+)
+
+
+def get_linked_references(references):
```
drjova: hmm I'd put this on utils

ammirate: mmm ok, it can be a util :+1:

michamos: why not move the whole reference matcher to a different module (`refextract`?)? it's not really related to the workflows (just happens to be used by the workflows only currently).

drjova: @michamos makes total sense, I'm not sure how easy it is though

ammirate: @michamos @drjova I have moved everything not related to workflows from `workflows/refextract` to `refextract/tasks`. Give it a look and let me know how it looks like :mag_right:

michamos: I don't think that's the right place for it. `tasks.py` should be reserved for celery tasks (or workflow tasks). This is neither.

ammirate: I agree, but sadly this is not the case. I would put it in `api`, wdyt?

drjova: ``utils.py`` is more suitable IMO ☺️

inspirehep/modules/editor/api.py
```diff
@@ -42,6 +42,7 @@
     editor_use_api_permission,
```
michamos: Cannot comment there, but you created a file called `%[` at the root of the repository.

inspirehep/modules/refextract/config.py
```diff
@@ -0,0 +1,231 @@
...
+
+"""Refextract config."""
+
+#
+# Configuration used when matching references.
+#
+
+"""Configuration for matching all HEP records (including JHEP and JCAP records)
+using unique identifiers."""
+
```
michamos: These are docstrings for the configs, so they should be *after* the constants, not before (https://stackoverflow.com/a/20230473).

tests/integration/workflows/test_workflows_tasks_refextract.py
```diff
@@ -407,21 +407,21 @@ def test_match_reference_doesnt_touch_curated():
```
michamos: you should probably move the tests for the matcher too.

tests/unit/workflows/test_workflows_tasks_refextract.py
```diff
@@ -250,48 +252,3 @@ def test_extract_references_from_raw_ref_wrong_schema():
     assert len(result) == 1
     assert 'reference' not in result[0]
     assert result[0]['raw_refs'][0] == reference['raw_refs'][0]
-
-
-def test_extract_references_from_raw_ref_reference_exists():
```
michamos: why remove this test?

inspirehep/modules/editor/api.py
ammirate: ops

tests/unit/workflows/test_workflows_tasks_refextract.py
ammirate: doh

inspirehep/modules/editor/api.py
```diff
@@ -287,3 +288,10 @@ def upload_files():
         full_url = fs.getsyspath(filename)
 
     return jsonify({'path': full_url})
+
+
+@blueprint_api.route('/linked_references', methods=['POST'])
+def get_linked_refs():
+    references = request.json
+    matched_refs = match_references(references)
```
harunurhan: this throws when `source` is missing in any item of `raw_refs`. I think it'd be much better if you modify `match_references` so that it skips those ones and `links` the ones with `source`, rather than throwing an exception.

inspirehep/modules/editor/api.py
```diff
@@ -287,3 +288,10 @@ def upload_files():
...
+    return jsonify(matched_refs)
```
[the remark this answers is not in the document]
ammirate:
```python
{ 'references': [] }
```
:point_up:

ammirate:
@michamos everything addressed

PR #3725, opened 2018-11-08
Added paralelization to checks

pazembrz:
Fixed parameters for CLI record checks

### DIFF
--- a/inspirehep/modules/records/cli.py
+++ b/inspirehep/modules/records/cli.py
@@ -33,6 +33,8 @@ import pprint
 from os import path, makedirs
 from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


michamos:
don't we have celery to run tasks in parallel? why implement yet another way to achieve the same thing?

inspirehep/modules/records/cli.py
```diff
@@ -374,6 +378,26 @@ def _gen_query(query, page_start=1, page_end=-1, window_size=100):
             query = None
 
 
+class MyThreadPool(ThreadPool):
```
pazembrz: I set this name to be sure that no one uses it anywhere else. As this is a dummy ThreadPool only for this CLI task.

PR #3736, opened 2018-11-15
migrator: make exceptions dashboard fast

michamos:
The previous query was filtering out `DELETED` records in a naive way
that doesn't benefit from the DB index. This is fixed by providing the
explicit list of collections that we do want to return.


michamos:
Don't know whether we should add tests for this kind of performance issues. One way would be to use "explain" on the query and make sure that it uses an index for everything, but it would require some work to parse the results.

michamos:
redis got bumped to 3.0 and became more strict: https://github.com/andymccurdy/redis-py/#encoding-of-user-input

PR #3738, opened 2018-11-15
1638 orcid tests: redis==3.0.0.post1 only allows accept basic data types

turtle321:
The test fails because it tries to cache a Mock object, now replaced
with the real string.


turtle321:
ci merge please

PR #3740, opened 2018-11-19
tests: non-eager testing for celery tasks and simpleindex refactor 

ammirate:
User Story INSPIR-1323 https://its.cern.ch/jira/browse/INSPIR-1323
This PR is mainly composed of two parts:

- It introduces a new test suite, called `integration_async`, where it is possible to test Celery tasks in `EAGER_MODE = False`, in a clean environment.

- It refactors the CLI command `simpleindex`, adding also some tests for the `integration_async` test suite.

Note: `integration_async` has been created for three main reasons:
1. Performance: testing async Celery tasks is slower then eager tasks, and adding more and more tests would have affected the already long execution time of the `integration` test suire.
2. Isolation: the new test suite does not iclude any demo-site record and cleans up the environment in every test.
3. the `celery_session_app` and `celery_session_worker` have scope `session`, interfering with the other `app` used in the `integration` tests. I have tried to use the `celery_app` and `celery_worker`, but they are affected by some issues: https://github.com/celery/celery/issues/5105


tests/integration_async/conftest.py
```diff
@@ -0,0 +1,142 @@
...
+# Use the helpers folder to store test helpers.
+# See: http://stackoverflow.com/a/33515264/374865
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), '../integration/helpers'))
+
+
+from factories.db.invenio_records import TestRecordMetadata  # noqa
+
+
+@pytest.fixture(scope='session')
+def super_app():
```
pazembrz: :1st_place_medal:  :+1:

inspirehep/modules/records/cli.py
```diff
@@ -112,6 +112,65 @@ def next_batch(iterator, batch_size):
...
+    Return a query for retrieving all non deleted records by pid_type
+
+    Args:
+        pid_types(List[str]): a list of pid types
+
+    Return:
+        SQLAlchemy query for non deleted record with pid type in `pid_types`
+    """
+    query = (
+        db.session.query(PersistentIdentifier.object_uuid).join(RecordMetadata, type_coerce(PersistentIdentifier.object_uuid, String) == type_coerce(RecordMetadata.id, String))
```
drjova: why we need ``db.session.query`` and not ``Model.query``?

inspirehep/modules/records/cli.py
```diff
@@ -112,6 +112,65 @@ def next_batch(iterator, batch_size):
...
+        for failure in errors:
+            try:
+                # batch failed
+                task_id = failure['task_id']
+                failed_uuids = tasks_uuids[task_id]
+                failures_json.append({
+                    'ids': failed_uuids,
+                    'error': repr(failure['error']),
+                })
+            except Exception:
```
drjova: maybe more specific than Exception? it's not very readable

inspirehep/modules/records/cli.py
```diff
@@ -221,36 +273,8 @@ def _finished_tasks_count():
     _prepare_logdir(failures_log_path)
     _prepare_logdir(errors_log_path)
```
drjova: I'd move these two in _dump_errors_ :)

inspirehep/modules/records/cli.py
ammirate: Well, that's how the SQLAlchemy documents the `join` https://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.join
The `query` object is not callable, thus you can not do something like `Model.query(Model.id).join(...`

tests/integration_async/conftest.py
```diff
@@ -0,0 +1,142 @@
...
+    with app.app_context():
+        yield app
+
+
+@pytest.fixture(scope='function', autouse=True)
+def clear_environment(app):
+    with app.app_context():
+        db.session.close()
+        db.drop_all()
+        drop_alembic_version_table()
```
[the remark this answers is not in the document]
ammirate: Do you mean `isolated_app`? If yes, we can not use it because it has the demo-site records and the goal is to have an empty env here.

tests/integration_async/test_records_cli.py
```diff
@@ -0,0 +1,301 @@
...
+    assert result.exit_code == 0
+    assert '0 succeeded' in result.output_bytes
+    assert '1 failed' in result.output_bytes
+    assert '0 batches errored' in result.output_bytes
+
+    with open(os.path.join(log_path, 'records_index_failures.log')) as log_file:
+        content = json.loads(log_file.read())
+        assert len(content) == 1
+        assert content[0]['id']  # the task failed
+        assert 'Invalid format: ""2013-01-14_final"' in content[0]['error']['caused_by']['reason']
```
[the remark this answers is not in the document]
ammirate: If you check `broken_field` it has a `"` in front. This makes the indexing fail. Actually, this is a real case that I took from the logs of the previous reindexing from production :)

ammirate:
ci test please

tests/integration_async/conftest.py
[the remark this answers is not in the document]
drjova: true we have the database implementation inside 👍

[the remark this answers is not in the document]
drjova: It'd be nice to split it in the future

ammirate:
ci test please e2e

ammirate:
ci test please

tests/integration_async/conftest.py
```diff
@@ -0,0 +1,142 @@
...
+    def _records_factory(n=1, additional_props={}, pid_type='lit'):
+        records = []
+        for i in range(n):
+            rec = TestRecordMetadata.create_from_kwargs(
+                json=additional_props,
+                pid_type=pid_type
+            ).record_metadata
+            records.append(rec)
+
+        db.session.commit()
```
drjova: IMO we don't need the ``commit``

PR #3742, opened 2018-11-19
Update mappinjg for external_system_identifiers.schema to be case insensitive

pazembrz:


drjova:
👍 what about the ``curated``?

PR #3752, opened 2018-11-27
serializers: add LaTeX serializer

vbalbp:
Signed-off-by: Victor Balbuena <vbalbp@gmail.com>


vbalbp:
ci test please

inspirehep/modules/records/serializers/schemas/latex/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+
+        return publication_info
+
+    def get_date(self, data):
+        if not data.get_citations_count():
+            return fields.missing
+        now = datetime.datetime.now()
+        return now.strftime("%d %b %Y")
+
+    def get_count(self, data):
```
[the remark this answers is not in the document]
vbalbp: Nope, if the number of citations is 0, we have in the dictionary ``citations_count: 0`` which leads the Jinja template to show ``%0 citations counted in INSPIRE as of $today``. I figured that was not the desired behaviour.

inspirehep/modules/records/serializers/templates/latex_template.tex
```diff
@@ -0,0 +1,54 @@
+%\cite{\VAR{ data.texkeys }}
```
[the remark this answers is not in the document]
vbalbp: The problem is, whitespace control doesn't only control end of lines, but also spaces, meaning that if we want a new string to be printed with a whitespace at the beginning but no end of line, you have to put the whole ``if`` to ``endif`` in the same line. If you do:
```
first text
{%- if condition -%}
 some more text
{%- endif %}
```
(which I understand would be the correct way to write it AND use the whitespaces), if the condition is met, the result would be: ``first textsome more text``
By letting it print the whitespaces, it will print the initial space before ``some other text`` but also the start of a new line. By putting everything in the same line, there is no new line to be printed.

inspirehep/modules/records/serializers/templates/latex_template.tex
```diff
@@ -0,0 +1,54 @@
...
+\bibitem{\VAR{ data.texkeys }}
+{%- if 'authors' in data or 'collaborations' in data %}
+{% endif %}
```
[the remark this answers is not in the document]
vbalbp: Yes :)

inspirehep/modules/records/serializers/templates/latex_template.tex
```diff
@@ -0,0 +1,54 @@
...
+{%- if 'collaborations' in data -%}
+{%- if 'authors' in data -%}\VAR{ data.authors.0 }{% if data.authors|length > 1 %} \textit{et al.}{% endif %}{% endif %} [
+{%- for collab in data.collaborations -%}
+\VAR{ collab.value }
+{%- if collab == data.collaborations[data.collaborations|length - 2] and data.collaborations|length > 1 %} and {% elif not loop.last %}, {% endif %}
+{%- endfor %}]
+{%- else -%}
+{%- for author in data.authors -%}
+\VAR{ author }
+{%- if author == data.authors[data.authors|length - 2] and data.authors|length > 1 %} and {% elif not loop.last %}, {% endif %}
```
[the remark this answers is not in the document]
vbalbp: I tried to search online but I could only find ``loop.last``, I'll check that out

inspirehep/modules/records/serializers/schemas/latex/__init__.py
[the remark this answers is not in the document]
michamos: why not? then you can just check in the jinja template whether citations_count is non-zero.

inspirehep/modules/records/serializers/jinja.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def latex_template(self):
+        latex_jinja_env = jinja2.Environment(
+            variable_start_string='\VAR{',
+            variable_end_string='}',
+            loader=jinja2.FileSystemLoader(os.path.abspath('/'))
+        )
+        template_path = pkg_resources.resource_filename('inspirehep', 'modules/records/serializers/templates/latex_template.tex')
+
+        template = latex_jinja_env.get_template(template_path)
```
drjova: I'd suggest to move this logic in https://github.com/inspirehep/inspire-next/blob/8aefa86d3d9be0d4f1f08c852c84a2c8ff73db7c/inspirehep/modules/records/ext.py 

even better registered them in the blueprint and use them as usual:

```python
from flask import render_template

 return render_template('inspire_records/template_name', foo='bar', bar='foo')

```

vbalbp:
ci test please

vbalbp:
ci test please

setup.py
```diff
@@ -102,6 +102,7 @@
     'invenio-workflows-ui~=2.0,>=2.0.11',
     'invenio-workflows~=7.0,>=7.0.1',
     # Pin the version of jsonschema, as not to catch alpha 3.0 release
+    'jinja2~=2.0,>=2.10',
```
michamos: this should be one line higher, because the comment is about `jsonschema`.

tests/unit/records/serializers/literature/test_schemas_latex.py
```diff
@@ -0,0 +1,263 @@
...
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+import datetime
+import json
+
+from inspirehep.modules.records.serializers.schemas.latex import LatexSchema
+
+TODAY = datetime.datetime.now().strftime("%d %b %Y")
```
michamos: the test might fail if run around midnight. You should use [`freezegun`](https://github.com/spulec/freezegun) for this kind of things (add it to the test deps).

tests/unit/records/serializers/literature/test_schemas_latex.py
```diff
@@ -0,0 +1,263 @@
...
+        'titles': [
+            {
+                'title': 'Jessica Jones',
+            },
+        ],
+        'authors': [
+            'F.~Castle',
+            'J.~Smith',
+            'J.~Black, Jr.',
+            'J.~'
```
michamos: this is not expected, if there's only one name it should be treated as a last name, not a first name.

vbalbp:
ci test please

tests/integration/records/test_records_serializers_latex.py
```diff
@@ -0,0 +1,567 @@
...
+        }]
+    }
+    TestRecordMetadata.create_from_kwargs(
+        json=record_json, index_name='records-hep')
+    response = isolated_api_client.get(
+        '/literature/64209',
+        headers={'Accept': 'application/x-latex-eu'}
+    )
+
+    expected = "%\\cite{a123bx}\n\\bibitem{a123bx}\nS.~Lee \\textit{et al.} [LHCb, Marvel and Comics]\n%``The Amazing Spiderman,''\nPhys.\\ Rev.\\ A.\\  \\textbf{58} (2014) no.120, 500-593\ndoi:10.1088/1361-6633/aa5514\n[arxiv:1607.06746 [hep-th]]."
```
michamos: can you make this a multiline string (in all tests)? [`textwrap.dedent`](https://docs.python.org/2/library/textwrap.html#textwrap.dedent) might be useful.

tests/unit/records/serializers/literature/test_schemas_latex.py
```diff
@@ -0,0 +1,265 @@
...
+# or submit itself to any jurisdiction.
+
+from __future__ import absolute_import, division, print_function
+
+import json
+
+from freezegun import freeze_time
+from inspirehep.modules.records.serializers.schemas.latex import LatexSchema
+
+TODAY = "19 Dec 1994"
```
michamos: I would put this inside the test, as `freeze_time` is a test decorator.

vbalbp:
ci test please

inspirehep/modules/records/serializers/templates/latex_template.tex
```diff
@@ -0,0 +1,49 @@
...
+    \VAR{ ' [' }
+    \VAR{- format_list(data.collaborations) }
+    \VAR{- ']' }
+{%- else -%}
+    \VAR{- format_list(data.authors) }
+{%- endif -%}
+\VAR{ '\n%``' ~ data.titles.0.title ~ ",''" }
+\VAR{- '\n' if 'journal_title' in data.publication_info }
+{%- if 'publication_info' in data -%}
+    {%- if 'journal_title' in data.publication_info -%}
```
michamos: can't you combine both conditions with an `and`?

tests/integration/records/test_records_serializers_latex.py
```diff
@@ -0,0 +1,649 @@
...
+
+    assert response.status_code == 200
+
+    result = response.data
+    expected = """\
+    %\\cite{1373790}
+    \\bibitem{1373790}
+    T.~Sch\xc3\xb6rner-Sadenius
+    %``The Large Hadron Collider,''
+    doi:10.1007/978-3-319-15001-7
```
michamos: I would expect an extra newline after this line to separate the two entries.

vbalbp:
ci test please

inspirehep/modules/records/utils.py
```diff
@@ -96,7 +96,7 @@ def get_pid_from_record_uri(record_uri):
 def get_author_display_name(name):
     """Returns the display name in format Firstnames Lastnames"""
     parsed_name = ParsedName.loads(name)
-    return " ".join(parsed_name.first_list + parsed_name.middle_list + parsed_name.last_list)
+    return " ".join(parsed_name.first_list + parsed_name.last_list)
```
drjova: this might break https://github.com/inspirehep/inspire-next/blob/a9e47eab306a26df7ffa429765da8d95f95b9d23/inspirehep/modules/records/utils.py#L409

michamos: :+1: this should have broken a test though :cry: @vbalbp can you add a test for it and fix it?

vbalbp: ``ParsedName.first_list`` has been changed on inspire_utils to return was previously was ``ParsedName.first_list + ParsedName.middle_list``, since we don't make a distinction between first and middle names in INSPIRE.

vbalbp:
ci test please

PR #3753, opened 2018-11-28
Decrease number of shards in ES for records-hep to 3

pazembrz:


pazembrz:
ci test please

drjova:
will it work for dev? IMO this kind of configuration should be part of the production

drjova:
do we still need this?

PR #3754, opened 2018-11-29
QueryParser Fixes and improvements

pazembrz:
Remove Query Parser from search factory. Use query_from_iq method from propper SearchClass (LiteratureSearch, AuthorSearch etc.)
Remove debug log test
Add tests for query and _source


inspirehep/modules/search/api.py
```diff
@@ -96,6 +98,15 @@ class Meta:
         doc_types = 'hep'
         default_filter = DefaultFilter(Q('match', _collections="Literature"))
 
+    def query_from_iq(self, query_string):
```
drjova: isn't this defined here? https://github.com/inspirehep/inspire-next/pull/3754/files#diff-58aeee5e0130cba3f67716bf65f1dec5R42

pazembrz: `query_from_iq` is redefined in LiteratureSearch to use QueryParser here, `query_from_iq` which is in SearcMixin is not using query parser at all.

PR #3758, opened 2018-11-30
1685 orcid: hook to force cache miss

turtle321:


drjova:
rebase for fixing the ui-lieterature tests

PR #3759, opened 2018-12-03
Remove session management from citations methods

pazembrz:


drjova:
odules/records/api.py:618:62: E127 continuation line over-indented for visual indent

PR #3760, opened 2018-12-03
records: remove custom indexer

harunurhan:
* Uses default, invenio's RecordIndexer instead


drjova:
could you please rebase?

PR #3764, opened 2018-12-05
Check if bai exists during enhancement for facet_author_name

monaawi:
In case bai doesn't exist, replace it with 'BAI'


monaawi:
ci test please

PR #3778, opened 2018-12-10
serializers: add should_display_positions

harunurhan:


inspirehep/modules/records/serializers/schemas/json/authors/__init__.py
```diff
@@ -61,6 +62,18 @@ def get_facet_author_name(data):
         else:
             return data['facet_author_name']
 
+    @staticmethod
+    def get_should_display_positions(data):
+        positions = data.get('positions', [])
+
+        if len(positions) is 1:
+            position = positions[0]
```
drjova:
```
if positions:
```

or 

```
get_value
```

PR #3789, opened 2018-12-14
Added include_in_all option to mapping in literature.

pazembrz:
Added include_in_all option to mapping in literature. Added tests to check if it behaves properly

### DIFF
--- a/inspirehep/modules/records/mappings/v5/records/hep.json
+++ b/inspirehep/modules/records/mappings/v5/records/hep.json
@@ -1,6 +1,7 @@
 {
     "mappings": {
         "hep": {
+            "include_in_all": false,
             "_all": {
                 "enabled": true
             },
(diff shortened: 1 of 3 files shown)


michamos:
aren't you including too much here? for example refs to other records will also be included. In the issue, I detailed every subfield that should be included.

inspirehep/modules/records/mappings/v5/records/hep.json
```diff
@@ -1005,6 +1047,7 @@
                     "type": "text"
                 },
                 "thesis_info": {
+                    "include_in_all": true,
```
michamos: you should exclude the linked record here.

inspirehep/modules/records/mappings/v5/records/hep.json
```diff
@@ -1051,15 +1094,18 @@
...
                         },
                         "title": {
+                            "include_in_all": true,
                             "type": "text"
                         }
                     },
                     "type": "object"
                 },
                 "titles": {
+                    "include_in_all": true,
```
michamos: we don't need the source here.

ammirate: good spot :+1:

PR #3793, opened 2019-01-09
Remove empty documents list from workflow data

pazembrz:
Simple try to bypass validation error for documents key.


pazembrz:
Not required anymore

PR #3796, opened 2019-01-10
Dedup urls in references

monaawi:
INSPIR-866


michamos:
Is this really the right way to do this? if refextract does this, it's probably better to add this inside the refextract code, instead of an extra workflow step, so it also works in the editor.

PR #3797, opened 2019-01-10
cli: add find_arxiv_duplicates command INSPIR-1835

ammirate:

### DIFF
--- a/inspirehep/modules/records/cli.py
+++ b/inspirehep/modules/records/cli.py
@@ -41,9 +41,15 @@ from invenio_pidstore.models import PersistentIdentifier, PIDStatus
 from flask import current_app
 from flask.cli import with_appcontext
 from invenio_records_files.models import RecordsBuckets
+from invenio_workflows import workflow_object_class, ObjectStatus
+from invenio_workflows.models import WorkflowObjectModel
 
-from inspirehep.utils.record_getter import get_db_record, get_es_record, \
-    RecordGetterError
+from inspirehep.utils.record_getter import (
+    get_db_record,
+    get_db_records,
+    get_es_record,
+    RecordGetterError,
+)
 from inspirehep.modules.records.checkers import check_unlinked_references
 from inspirehep.modules.records.tasks import batch_reindex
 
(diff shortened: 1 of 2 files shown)


inspirehep/modules/records/cli.py
```diff
@@ -354,6 +356,67 @@ def handle_duplicates(remove_no_control_number, remove_duplicates,
...
+    workflows_ids = [
+        wf_id[0] for wf_id in WorkflowObjectModel.query.
+            with_entities(WorkflowObjectModel.id).
+            filter(WorkflowObjectModel.created > starting_date).
+            filter(WorkflowObjectModel.status == ObjectStatus.COMPLETED).
+            all()
+    ]
+    duplicates = dict()
+
+    with click.progressbar(length=len(workflows_ids), label='Processing workflows') as bar:
```
michamos: that's a strange way to do it. Why not simply
```suggestion
    with click.progressbar(workflow_ids, label='Processing workflows') as bar:
```
followed by
```
    for wf_id in bar:
```
then you don't have to update the bar by hand.

inspirehep/modules/records/cli.py
```diff
@@ -354,6 +356,67 @@ def handle_duplicates(remove_no_control_number, remove_duplicates,
...
+        for wf_id in workflows_ids:
+            wf = workflow_object_class.get(wf_id)
+            recid = wf.data.get('control_number')
+
+            if not recid:
```
michamos: you should be able to do this filtering in the DB by adding something like `.filter(WorkflowObjectModel.data).has_key('control_number')` (might require a `type_coerce` to `JSON` to work). That should be way faster.

inspirehep/modules/records/cli.py
```diff
@@ -354,6 +356,67 @@ def handle_duplicates(remove_no_control_number, remove_duplicates,
...
+                bar.update(1)
+                continue  # rejected paper
+
+            if wf.data.get('arxiv_eprints'):
```
michamos: you could add a similar test for the presence of `arxiv_eprints` to the query.

inspirehep/modules/records/cli.py
```diff
@@ -354,6 +360,72 @@ def handle_duplicates(remove_no_control_number, remove_duplicates,
     db.session.commit()
 
 
+@check.command()
+@with_appcontext
+def find_arxiv_duplicates(starting_date=None):
```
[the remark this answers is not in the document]
pazembrz: IMHO it's not necessary to call this CLI command with parameter at all. Basically it should swipe whole db. Specifying a date here is very specific use-case which shouldn't be available from CLI. It still can be used when calling this as a function from inspire shell.

PR #3801, opened 2019-01-15
Add bai to author submission serializer

monaawi:
INSPIR-1817


inspirehep/modules/submissions/serializers/schemas/author.py
```diff
@@ -80,6 +81,8 @@ def before_dump(self, data):
             'twitter': self.get_first_or_missing(
                 get_values_for_schema(data.get('ids', []), 'TWITTER')),
             'websites': get_value(data, 'urls.value', default=missing),
+            'bai': self.get_first_or_missing(
```
harunurhan: Actually, we won't need it for author updates, since there is already a create profile, right @michamos ?

PR #3802, opened 2019-01-15
workflows: attach user to edit wf transaction

harunurhan:


inspirehep/modules/workflows/workflows/edit_article.py
```diff
@@ -40,6 +46,20 @@ def update_record(obj, eng):
     record = get_db_record('lit', control_number)
     record.update(obj.data)
     record.commit()
+    db.session.commit()
```
drjova: actually there is https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L272, it should be ok to remove it.

inspirehep/modules/workflows/workflows/edit_article.py
```diff
@@ -41,6 +47,19 @@ def update_record(obj, eng):
     record.update(obj.data)
     record.commit()
 
+    transaction_id = None
+    record = get_db_record('lit', control_number)
+    for revision in record.revisions:
+        if revision.revision_id == record.revision_id:
+            transaction_id = revision.model.transaction_id
+            break
```
drjova: maybe something like:

```python
try:
  revision = record.revision[record.revision_id]
  transaction_id = revision.model.transaction_id
expect IndexError:
  pass # do something
```

tests/integration/workflows/test_edit_article.py
```diff
@@ -66,6 +70,12 @@ def edit_workflow(workflow_app):
     eng_uuid = start('edit_article', data=factory.record_metadata.json)
     obj = WorkflowEngine.from_uuid(eng_uuid).objects[0]
 
+    # user id is required for edit article workflows
+    wf = workflow_object_class.get(obj.id)
+    wf.id_user = user.get_id()
```
drjova: how it was working before?

inspirehep/modules/workflows/workflows/edit_article.py
```diff
@@ -41,6 +47,19 @@ def update_record(obj, eng):
...
+
+    Transaction = transaction_class(RecordMetadata)
+    transaction = Transaction.query.filter(Transaction.id == transaction_id).one()
+    transaction.user_id = obj.id_user
+    db.session.add(transaction)
+    db.session.commit()
```
drjova: I'd move the transaction logic to separate function so you can easily test it, now you have to write the code https://github.com/inspirehep/inspire-next/pull/3802/files#diff-e05ba1f919990d666d23d8a40efd95e2R196

drjova:
@harunurhan what will happen if the revision is not found? will it still display ``System``?

tests/integration/workflows/test_edit_article.py
harunurhan: because user_id wasn't being used anywhere in the code. now it is.

inspirehep/modules/workflows/workflows/edit_article.py
harunurhan: well this is just an early thing, I'll separate things later 👍

harunurhan:
You mean when user is not found? @drjova 
Because when there is no revision, there is nothing to show anyway. But this also wouldn't happen as long as record is modified.

For the missing user, it assumes that all edit article workflows started by a user.

inspirehep/modules/workflows/workflows/edit_article.py
harunurhan: I'll need to verify, that thing is not an array index or not.

inspirehep/modules/workflows/workflows/edit_article.py
```diff
@@ -41,6 +47,24 @@ def update_record(obj, eng):
     record.update(obj.data)
     record.commit()
 
+    user_id = obj.id_user
```
drjova: is ``obj.id_user`` always set?

drjova: because I see on the tests we force add it

inspirehep/modules/workflows/workflows/edit_article.py
```diff
@@ -41,6 +47,24 @@ def update_record(obj, eng):
...
+    if user_id:
+        _set_transaction_user_id_for_last_record_update(control_number, obj.id_user)
+
+    db.session.commit()
+
+
+def _set_transaction_user_id_for_last_record_update(control_number, user_id):
+    record = get_db_record('lit', control_number)
+    revision = record.revisions[record.revision_id]
```
drjova: I'll still put it inside
```python
try:
except IndexError:
```
and log if the revision not found, and just for readability will move the function before ``update_record``

inspirehep/modules/workflows/workflows/edit_article.py
```diff
@@ -41,6 +47,24 @@ def update_record(obj, eng):
     record.update(obj.data)
     record.commit()
 
+    user_id = obj.id_user
+    if user_id:
+        _set_transaction_user_id_for_last_record_update(control_number, obj.id_user)
```
drjova:
```suggestion
        _set_transaction_user_id_for_last_record_update(control_number, user_id)
```

inspirehep/modules/workflows/workflows/edit_article.py
michamos: you should never index into `record.revisions`. It's completely broken, as `record == record.revisions[record.revision_id]` does not hold (because a new `revision_id` might be generated but not committed, so it can jump).

PR #3806, opened 2019-01-16
workflows: restart matched workflows in error INSPIR-1855

pazembrz:
When workflow finishes and finds that there is another workflow which matches this one,
it restarts the waiting one.
When workflow is restarted and there is another workflow in error state because of current workflow, current workflow won't stop anymore but will continue.
Adds `delay` workflow step which allows to delay workflow start.
From now on Workflows won't permanently stop older workflows matching this one,
they will create a chains of workflows which will run automatically one after another.

WARNING: Introduces big changes to workflows!

for file tests/integration/workflows/test_arxiv_workflow.py diff is sooooo big because I had to run black on it. There are only changes in two tests and one test is removed as it was dead after changes.

I need as much reviews as possible for this PR, as this really might be dangerous for workflows.


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -629,3 +632,50 @@ def load_from_source_data(obj, eng):
...
+    Args:
+        wf: Workflow to restart
+        original_workflow: Workflow which restarts
+        position: To which position wf should be restarted
+    """
+    wf.callback_pos = position
+    wf.extra_data['source_data']['extra_data']['delay'] = 10
+    restarted_by_wf_list = wf.extra_data.get('restarted-by-wf', [])
+    restarted_by_wf_list.append(original_workflow.id)
+    wf.extra_data['source_data']['extra_data']['restarted-by-wf'] = restarted_by_wf_list
```
michamos: these 3 lines are:
```python
wf.extra_data['source_data']['extra_data'].setdefault('restarted-by-wf', []).append(original_workflow.id)
```

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -257,20 +268,35 @@ def raise_if_match_wf_in_error_or_initial(obj, eng):
...
-        obj.extra_data['error_workflows_matched'] = matched_ids
         raise WorkflowsError(
-            'Cannot continue processing. Found workflows in ERROR or INITIAL '
+            'Cannot continue processing.'
+            'Found not completed workflows in holdingpen '
             'state: {}'.format(matched_ids)
         )
 
 
+def remove_blocked_by_me(obj, matched_ids):
```
michamos: I would do this differently: define a predicate that can then be used to filter the list of matched wfs
```python
def is_workflow_id_blocked_by_workflow(obj, workflow_id):
    wf = workflow_object_class.get(wf_id)
    return obj.id in wf.extra_data.get('holdingpen_matches', [])
```

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -212,6 +212,28 @@ def _is_core(record):
...
+            if False then return all matched workflows.
+    """
+
+    def _non_completed(base_record, match_result):
+        return not get_value(match_result,
+                             '_source._workflow.status') == 'COMPLETED'
+
+    matched_ids = pending_in_holding_pen(obj, _non_completed)
+    if skip_blocked:
+        matched_ids = remove_blocked_by_me(obj, matched_ids)
```
michamos: `matched_ids = [id_ for id_ in matched_ids if not is_workflow_id_blocked_by_workflow(obj, id_)]` (see further comment).

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -257,20 +268,35 @@ def raise_if_match_wf_in_error_or_initial(obj, eng):
...
-        ) in ('ERROR', 'INITIAL')
-
-    matched_ids = pending_in_holding_pen(obj, _filter)
+    matched_ids = get_wfs_not_completed(obj)
     if bool(matched_ids):
-        obj.extra_data['error_workflows_matched'] = matched_ids
         raise WorkflowsError(
-            'Cannot continue processing. Found workflows in ERROR or INITIAL '
+            'Cannot continue processing.'
+            'Found not completed workflows in holdingpen '
```
michamos: `not completed` -> `non-completed` or `incomplete`.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -257,20 +268,35 @@ def raise_if_match_wf_in_error_or_initial(obj, eng):
...
             'state: {}'.format(matched_ids)
```
michamos: ~~`state`~~

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -212,6 +212,28 @@ def _is_core(record):
         obj.extra_data['core'] = True
 
 
+def get_wfs_not_completed(obj, skip_blocked=True):
```
michamos: `get_incomplete_matching_workflows`

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -212,6 +212,28 @@ def _is_core(record):
...
+    """Return list of all matched workflows which are not completed
```
michamos: Not true, it also writes in `extra_data`. That's surprising for a function called `get_...`.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -231,23 +253,12 @@ def match_non_completed_wf_in_holdingpen(obj, eng):
...
-    """Raise if a matching wf is in ERROR or INITIAL state in the HoldingPen.
-
-    Uses a custom configuration of the ``inspire-matcher`` to find duplicates
-    of the current workflow object in the Holding Pen not in the
-    that are in ERROR or INITIAL state.
-
-    If any match is found, it sets ``error_workflows_matched`` in
-    ``extra_data`` to the list of ids that matched and raise an error.
+def raise_if_match_wf(obj, eng):
+    """Raise if a matching wf is not in compleated state in the HoldingPen.
```
michamos: `completed`

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -417,6 +443,7 @@ def stop_matched_holdingpen_wfs(obj, eng):
     Returns:
         None
     """
+    from inspirehep.modules.workflows.tasks.actions import mark
```
michamos: why a local import?

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -257,20 +268,35 @@ def raise_if_match_wf_in_error_or_initial(obj, eng):
...
 
     """
-    def _filter(base_record, match_result):
-        return get_value(
-            match_result, '_source._workflow.status'
-        ) in ('ERROR', 'INITIAL')
-
-    matched_ids = pending_in_holding_pen(obj, _filter)
+    matched_ids = get_wfs_not_completed(obj)
     if bool(matched_ids):
```
michamos: what''s the `bool` for?

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -629,3 +632,50 @@ def load_from_source_data(obj, eng):
...
+def restart_workflow(wf, original_workflow, position=[0]):
+    """Restarts workflow
+
+    Args:
+        wf: Workflow to restart
+        original_workflow: Workflow which restarts
+        position: To which position wf should be restarted
+    """
+    wf.callback_pos = position
+    wf.extra_data['source_data']['extra_data']['delay'] = 10
```
michamos: what's the delay for?

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -342,16 +344,7 @@
 
 
 MARK_IF_MATCH_IN_HOLDINGPEN = [
```
michamos: `STOP_IF_MATCH_IN_HOLDINGPEN`?

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -390,45 +383,23 @@
...
     ),
 
-    IF_ELSE(
-        is_marked('already-in-holding-pen'),
-        IF_ELSE(
-            has_same_source('holdingpen_matches'),
-            # stop the matched wf and continue this one
-            [
-                stop_matched_holdingpen_wfs,
-                mark('stopped-matched-holdingpen-wf', True),
```
michamos: I don't think this logic should disappear: if there are two updates that need to be processed, it's better to skip the earliest one, otherwise the curator has to work twice.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -629,3 +632,50 @@ def load_from_source_data(obj, eng):
...
+        restart_workflow(next_wf, obj)
+
+
+@with_debug_logging
+def delay_if_necessary(obj, eng):
+    """Delays workflow if necessary,
+    delay timeout should be in `extra_data['source_data']['persistant_data']['delay']` key """
+    delay = get_value(obj.extra_data, 'source_data.extra_data.delay')
+    if delay:
+        time.sleep(float(delay))
```
michamos: why the `float`?

inspirehep/modules/workflows/tasks/matching.py
pazembrz: as actions.py already imports matching.py

inspirehep/modules/workflows/tasks/matching.py
pazembrz: I have no idea. It's not my code ;)
In this situation list would be converted to bool so it looks like forcing cast to bool is unnecessary.

inspirehep/modules/workflows/tasks/actions.py
pazembrz: It's important to delay workflow restart. It may happen that workflow restarted(R) by previous workflow(P) will fail as workflow "P" still has status "RUNNING" as restarting workflow "R" is actually last step in the workflow "P". Delay happens as first function in first step of workflow "R".

michamos: but you can't know how long it will take before the other one changes from `RUNNING` to `COMPLETED`. What happens if the other guy is still `RUNNING` after 10s?

inspirehep/modules/workflows/workflows/article.py
pazembrz: I didn't want to change name of the workflows, but you may be right to change this to different name. I think that stop is not a good word as it's used for workflows which are stopped permanently and they status is set to COMPLETED.

inspirehep/modules/workflows/workflows/article.py
pazembrz: No, as I understand from what Antonio explained me, if curator will accept/reject it once Inspire will remember what to do for all other workflows matched by this one. If not, then this code really has to be restored.

inspirehep/modules/workflows/tasks/actions.py
pazembrz: as sleep accepts 1 and 1.5 and 1.9432 but it won't accept 'an hour' etc. so it's good to cast value to float (as we won't reduce accuracy), but if there will be something not castable to float it will produce more readable exception in wf _error_msg ("ValueError: could not convert string to float: xxxx) than just `TypeError: a float is required`

inspirehep/modules/workflows/tasks/actions.py
pazembrz: We have to be reasonable here. 10s is really long time for workflow to change state from `RUNNING` to `COMPLETED` as this is really last part of last step.

inspirehep/modules/workflows/tasks/actions.py
drjova: I'd just ``time.sleep(delay)`` ☺️

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -629,3 +632,50 @@ def load_from_source_data(obj, eng):
         obj.save()
     except KeyError:
         raise ValueError("Can't start/restart workflow as 'source_data' is either missing or corrupted")
+
+
+def restart_workflow(wf, original_workflow, position=[0]):
```
drjova: it's better to keep ``obj`` as we have everywhere

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -231,23 +253,12 @@ def match_non_completed_wf_in_holdingpen(obj, eng):
...
-def raise_if_match_wf_in_error_or_initial(obj, eng):
-    """Raise if a matching wf is in ERROR or INITIAL state in the HoldingPen.
-
-    Uses a custom configuration of the ``inspire-matcher`` to find duplicates
-    of the current workflow object in the Holding Pen not in the
-    that are in ERROR or INITIAL state.
-
-    If any match is found, it sets ``error_workflows_matched`` in
-    ``extra_data`` to the list of ids that matched and raise an error.
+def raise_if_match_wf(obj, eng):
```
drjova:
```suggestion
def raise_if_match_workflow(obj, eng):
```

inspirehep/modules/workflows/tasks/actions.py
pazembrz: Is it? `WAIT_FOR_LEGACY_WEBCOLL` step is few steps before `FIND_NEXT_AND_RUN_IF_NECESSARY` step

inspirehep/modules/workflows/tasks/actions.py
pazembrz: Which wf `_error_msg` is easier to debug : `TypeError: a float is required` or `ValueError: could not convert string to float: xxxx`?

inspirehep/modules/workflows/workflows/article.py
ammirate: @pazembrz that is true only if we are talking about Rejected workflows: if a curator rejects a paper, the next update for that paper will be auto-rejected. Instead, when a curator edits the metadata of the first workflow and then a new update comes (with the same incorrect data), probably the curator should correct again the same data.
That's why we should keep the logic of 'stop the current and process the newest'.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -212,6 +212,34 @@ def _is_core(record):
         obj.extra_data['core'] = True
 
 
+def set_wf_not_completed_ids_to_wf(obj, skip_blocked=True):
+    """Return list of all matched workflows ids which are not completed
```
ammirate: Nit: you forgot to prettify the docstring

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -212,6 +212,34 @@ def _is_core(record):
...
+    def is_workflow_blocked_by_another_workflow(workflow_id):
+        workflow = workflow_object_class.get(workflow_id)
+        return obj.id in workflow.extra_data.get('holdingpen_matches', [])
+
+    matched_ids = pending_in_holding_pen(obj, _non_completed)
+    if skip_blocked:
+        matched_ids = [_id for _id in matched_ids if
+                       not is_workflow_blocked_by_another_workflow(_id)]
+    obj.extra_data['holdingpen_matches'] = matched_ids
+    obj.save()
```
ammirate: I was thinking that we can use the `save_workflow` function, which also commits to the DB, this one just indexes to ES (and we try also to unify the way we do it).

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -257,20 +274,36 @@ def raise_if_match_wf_in_error_or_initial(obj, eng):
...
-            'Cannot continue processing. Found workflows in ERROR or INITIAL '
-            'state: {}'.format(matched_ids)
+            'Cannot continue processing workflow {wf_id}.'
+            'Found not-completed workflows in holdingpen '
+            ': {blocking_ids}'.format(wf_id=obj.id,
+                                      blocking_ids=matched_ids)
         )
 
 
+def remove_blocked_by_me(obj, matched_ids):
```
ammirate: Don't you think this should be named something like `get_blocking_workflows_ids`, specifying in the docstring that cycles are avoided?
It does not remove anything from anywhere :reminder_ribbon:

inspirehep/modules/workflows/tasks/matching.py
ammirate: +1

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -622,10 +625,58 @@ def load_from_source_data(obj, eng):
...
+    obj.extra_data['source_data']['extra_data']['delay'] = 10
+    obj.extra_data['source_data']['extra_data'].setdefault(
+        'restarted-by-wf', []).append(original_workflow.id)
+    obj.save()
+    db.session.commit()
+    obj.continue_workflow('restart_task', delayed=True)
+
+
+@with_debug_logging
+def run_next_if_necessary(obj, eng):
```
ammirate: Why not `restart_blocked_workflow`? :)

pazembrz: because it's not only restarting, its checking if restart if necessary, then it finds the proper workflow to restart and then restarts it.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -257,17 +285,16 @@ def raise_if_match_wf_in_error_or_initial(obj, eng):
...
-    def _filter(base_record, match_result):
-        return get_value(
-            match_result, '_source._workflow.status'
-        ) in ('ERROR', 'INITIAL')
-
-    matched_ids = pending_in_holding_pen(obj, _filter)
+    matched_ids = set_wf_not_completed_ids_to_wf(obj,
+                                                 skip_blocked=True,
+                                                 skip_halted=True
+                                                 )
```
ammirate: This is so niiice... :P

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -435,3 +464,40 @@ def has_more_than_one_exact_match(obj, eng):
     """Does the record have more than one exact match."""
     exact_matches = obj.extra_data['matches']['exact']
     return len(set(exact_matches)) > 1
+
+
+@with_debug_logging
+def run_next_if_necessary(obj, eng):
+    """Remove current wf id from matched workflows and run next one"""
+    blocked_wfs = set_wf_not_completed_ids_to_wf(obj, False)
```
ammirate: I would use kwargs, it's not very intuitive which flag you are setting to False from this call:
`set_wf_not_completed_ids_to_wf(obj, skip_blocked=False, skip_halted=False)`

tests/integration_async/test_workflows.py
```diff
@@ -0,0 +1,384 @@
...
+from invenio_search import current_search_client as es
+from invenio_workflows import ObjectStatus, start, workflow_object_class
+
+
+def generate_record():
+    """Provide record fixture."""
+    json_data = json.loads(pkg_resources.resource_string(
+        __name__,
+        os.path.join(
+            '../integration/workflows/fixtures',
```
ammirate: I think having shared fixtures is not really a good idea, even more shared between different test suits. If that fixture fits your use case, just copy it in the `integration_async/workflows/fixture` folder

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -212,6 +214,44 @@ def _is_core(record):
...
+    if skip_halted:
+        matched_ids = pending_in_holding_pen(obj, _not_completed_or_halted)
+    else:
+        matched_ids = pending_in_holding_pen(obj, _non_completed)
+
+    if skip_blocked:
+        matched_ids = [_id for _id in matched_ids if
+                       not is_workflow_blocked_by_another_workflow(_id)]
+    obj.extra_data['holdingpen_matches'] = matched_ids
+    save_workflow(obj, None)
```
drjova: is it safe to save without eng?

tests/integration_async/test_workflows.py
```diff
@@ -0,0 +1,384 @@
...
+            raise AssertionError
+        time.sleep(1)
+
+
+def test_wf_not_stops_when_blocking_another_one_after_restarted_on_running(
+    app,
+    celery_app_with_context,
+    celery_session_worker
+):
+    def build_workflow(workflow_data, data_type='hep', **kwargs):
```
ammirate: This might be useful to have it as a fixture in `conftest.py` for workflow, don't you think?

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -244,7 +244,8 @@ def decorator(func):
         @wraps(func)
         def wrapper(*args, **kwargs):
             timeout_time = current_app.config[config_key]
-            return timeout(timeout_time)(func)(*args, **kwargs)
+            use_signals = current_app.config.get('USE_SIGNALS_ON_TIMEOUT', True)
+            return timeout(timeout_time, use_signals=use_signals)(func)(*args, **kwargs)
```
drjova: why we need this?

tests/integration_async/test_workflows.py
```diff
@@ -0,0 +1,384 @@
...
+    assert wf2.extra_data.get('restarted-by-wf') == [1]
+    assert wf3.extra_data.get('restarted-by-wf') == [2]
+
+
+def test_wf_not_stops_when_blocking_another_one_after_restarted_on_init(
+    app,
+    celery_app_with_context,
+    celery_session_worker
+):
+    def build_workflow(workflow_data, data_type='hep', **kwargs):
```
ammirate: Copy/paste supporting use of fixture :P

tests/integration_async/test_workflows.py
```diff
@@ -0,0 +1,384 @@
...
+
+    """
+    start = datetime.now()
+    end = start + timedelta(seconds=max_time)
+    while (True):
+        db.session.close()
+        if workflow_object_class.get(workflow_id).status == desired_status:
+            return
+        if datetime.now() > end:
+            raise AssertionError
```
drjova: AssertionError isn't it a bit arbitrary? Maybe a custom exception would be easier to understand

tests/integration_async/test_workflows.py
```diff
@@ -0,0 +1,384 @@
...
+                    'data': deepcopy(workflow_data),
+                    'extra_data': {},
+                }
+            },
+            **kwargs
+        )
+        return workflow_object
+
+    app.config['FEATURE_FLAG_ENABLE_UPDATE_TO_LEGACY'] = False
+    app.config['PRODUCTION_MODE'] = False
```
drjova: why we need to disable them?

inspirehep/modules/workflows/utils/__init__.py
pazembrz: Without this `async` tests are failing, as `timeout()` function here by default uses signals, and apparently signals can be used 'only in main thread'. So this is mainly a flag for fixing async workflows tests.

tests/integration_async/test_workflows.py
pazembrz: In usual (not async) case I would just use `assert` here, so using `AssertionError` looks more natural and intuitive here.

tests/integration_async/test_workflows.py
pazembrz: We want workflow to finish without hanging on `sending_to_legacy` etc, and as we are async here, and workflow itself is running in another process and it's not trivial to mock them from test function.

inspirehep/modules/workflows/tasks/matching.py
pazembrz: Not  anymore :P

tests/integration_async/test_workflows.py
drjova: ``assert`` in general hides the real error, I'd still go for a custom error to understand why failed

tests/integration_async/test_workflows.py
```diff
@@ -0,0 +1,384 @@
...
+
+def check_wf_state(workflow_id, desired_status, max_time=600):
+    """Waits for the workflow to go to desired status
+    Args:
+        workflow: workflow to check
+        desired_state: desired state
+        max_time: maximum time to wait in seconds, otherwise raise exception
+
+    Returns: None
+
```
drjova: cosmetics in general :P spaces / new lines

inspirehep/modules/workflows/tasks/matching.py
pazembrz: Yes, eng is not used in there.

tests/integration_async/test_workflows.py
pazembrz: Custom exception is too much here, but I've added a message which will explain more.

PR #3814, opened 2019-01-24
workflows: make robotupload priority configurable INSPIR-1899

michamos:
No tests for this as it's hard to test: requests in the workflow unit tests are tested with `request-mock`, which doesn't provide a simple way to match on POST request arguments. So it would require either implementing custom matching logic or using vcr instead.


drjova:
@michamos some unittests are failing

PR #3827, opened 2019-02-20
workflows: add literature sub wf endpoint

harunurhan:
That create a workflow for a literature submission.

### DIFF
--- a/inspirehep/modules/workflows/views.py
+++ b/inspirehep/modules/workflows/views.py
@@ -745,6 +745,30 @@ def start_author_workflow():
     return jsonify({'workflow_object_id': workflow_object_id})
 
 
+@workflow_blueprint.route('/literature', methods=['POST'])
+def start_workflow_for_literature_submission():
+    json = request.get_json()
+    workflow_object = workflow_object_class.create(
+        data=json['data'],
+        id_user=current_user.get_id(),
+        data_type="hep"
+    )
+    workflow_object.extra_data['formdata'] = json['form_data']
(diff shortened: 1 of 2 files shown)


inspirehep/modules/workflows/views.py
```diff
@@ -745,6 +745,31 @@ def start_author_workflow():
     return jsonify({'workflow_object_id': workflow_object_id})
 
 
+@workflow_blueprint.route('/literature', methods=['POST'])
+def start_workflow_for_literature_submission():
+    json = request.get_json()
+    workflow_object = workflow_object_class.create(
+        data={},
```
ammirate: `data=json['data']`

inspirehep/modules/workflows/views.py
```diff
@@ -745,6 +745,31 @@ def start_author_workflow():
...
+        id_user=current_user.get_id(),
+        data_type="hep"
+    )
+    workflow_object.extra_data['formdata'] = json['form_data']
+    workflow_object.data = json['data']
+    workflow_object.extra_data['source_data'] = {
```
ammirate: Maybe we can create a util function like https://github.com/inspirehep/inspire-next/blob/d67c86279968cd8843d657071c38e0838d0780aa/tests/integration_async/test_workflows.py#L36-L47 since it will be used also for the author endpoint

harunurhan: I agree that we should do it, but I am not feeling with comfortable changing all other parts that creates workflow.

PR #3828, opened 2019-02-20
workflow: avoid duplicate records when workflow fails

ammirate:

### DIFF
--- a/tests/integration/workflows/test_workflows_tasks_upload.py
+++ b/tests/integration/workflows/test_workflows_tasks_upload.py
@@ -27,7 +27,10 @@ from flask import current_app
 
 from invenio_workflows import workflow_object_class
 
+from inspirehep.modules.records.api import InspireRecord, RecordMetadata
+
 from factories.db.invenio_records import TestRecordMetadata
+
 
 # FIXME: otherwise this task is not found by Celery.
 from inspirehep.modules.orcid.tasks import orcid_push  # noqa: F401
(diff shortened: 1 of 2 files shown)


ammirate:
Related User Story: https://its.cern.ch/jira/browse/INSPIR-1994

tests/integration/workflows/test_workflows_tasks_upload.py
```diff
@@ -173,3 +178,33 @@ def test_is_stale_data_returns_false_if_is_update_is_falsy(workflow_app):
...
+    def raise_func(*args, **kwargs):
+        raise Exception()
+
+    record_count = RecordMetadata.query.count()
+    assert record_count == 0
+
+    with patch.object(
+        InspireRecord,
+        'download_documents_and_figures',
+        raise_func
```
michamos: Minor: I think you can raise exceptions with `side_effect=Exception` instead of having to define a function to do it.

tests/integration/workflows/test_workflows_tasks_upload.py
```diff
@@ -173,3 +178,33 @@ def test_is_stale_data_returns_false_if_is_update_is_falsy(workflow_app):
...
+    ):
+        # pytest.raises catches the exception and makes the test passing immediately
```
michamos: not sure I understand this comment: do you mean
```python
with pytest.raises(Exception):
    store_record(obj, eng)
record_count = RecordMetadata.query.count()
assert  record_count == 0
```
doesn't work? (doesn't really matter though, this way of doing it is fine, as you're not testing that it raises, but that the db state is correct).

ammirate: Exactly. I wrote this test (using `pytest.raises`) before writing the bugfix, expecting it to fail but it was passing, since the exception was caught properly. This way it failed properly.

tests/integration/workflows/test_workflows_tasks_upload.py
```diff
@@ -22,13 +22,18 @@
 
 from __future__ import absolute_import, division, print_function
 
+import pytest
```
michamos: unused import

PR #3831, opened 2019-02-25
Fix labs

drjova:


inspirehep/config.py
```diff
@@ -162,7 +162,7 @@
 
 SECURITY_CONFIRM_SALT = "CHANGE_ME"
 SECURITY_EMAIL_SENDER = "admin@inspirehep.net"
-SECURITY_EMAIL_SUBJECT_REGISTER = "Welcome to INSPIRE Labs!"
+SECURITY_EMAIL_SUBJECT_REGISTER = "Welcome to INSPIRE beta!"
```
michamos: :man_facepalming:

StellaCh: Change to "Welcome to INSPIRE!" , no need for 'beta' as this is the same message for the submission forms

PR #3839, opened 2019-03-06
Distributed continuous migration INSPIR-2081

michamos:
This should allow the continuous migration to run much faster by using the same trick as the migration from files (insert copies of legacy records into the mirror synchronously, then spawn tasks to do the migration proper from the mirror in chunks).

It requires extensive testing on QA to make sure that all the Celery stuff is working correctly, as it's spawning tasks from a celery task that then waits for those tasks to finish. I didn't manage to make it work using `integration_async`.

### DIFF
--- a/inspirehep/modules/migrator/tasks.py
+++ b/inspirehep/modules/migrator/tasks.py
@@ -33,7 +33,7 @@ from contextlib import closing
 import click
 import requests
 
-from celery import group, shared_task
+from celery import shared_task
 from elasticsearch.helpers import bulk as es_bulk
 from flask import current_app
 from flask_sqlalchemy import models_committed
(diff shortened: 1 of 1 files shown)


inspirehep/modules/migrator/tasks.py
```diff
@@ -224,27 +215,28 @@ def continuous_migration(skip_files=None):
...
+                ]
+                insert_into_mirror(decompressed_records)
+            migrate_from_mirror(wait_for_results=True, skip_files=skip_files)
+    except DistributedLockError:
         LOGGER.info("Continuous_migration already executed. Skipping.")
 
 
-@shared_task(ignore_result=False, queue='migrator')
+@shared_task(bind=True, queue='migrator')
 @disable_orcid_push
```
michamos: this is troublesome, as we don't want to disable the orcid push in the continuous migration. I will changed it.

michamos:
This doesn't work correctly as the bulk indexer is not handling lots of cases (deleted records, records that need reindexing because their citations have changed, etc.). That would take some work to fix and the indexer is moving to the new instance, so closing for now.

PR #3840, opened 2019-03-07
version: read version from git

drjova:

### DIFF
--- a/inspirehep/version.py
+++ b/inspirehep/version.py
@@ -22,5 +22,23 @@
 
 from __future__ import absolute_import, division, print_function
 
+import logging
+import pkg_resources
+from subprocess import check_output
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_git_tag_version():
+    try:
+        return check_output('git log -1 --pretty=format:%H'.split())
+    except Exception:
+        git_repo_path = pkg_resources.get_distribution('inspirehep').location
+        logger.error(
+            'Error reading ``SHA`` from path "{}".'.format(git_repo_path)
+        )
+        return -1
+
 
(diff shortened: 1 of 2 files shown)


inspirehep/version.py
```diff
@@ -22,5 +22,27 @@
...
+
+
+
+def _get_git_tag_version():
+    try:
+        git_repo_path = pkg_resources.get_distribution('inspirehep').location
+        repo = Repo(git_repo_path)
+        tags = sorted(repo.tags, key=lambda t: t.commit.committed_datetime)
+        latest_tag = tags[-1]
+        logger.info('Running in git tag: {}'.format(latest_tag))
```
harunurhan: maybe put logging outside in `__init__` to be more obvious. Since logging is the whole reason why we're doing this s**t

inspirehep/version.py
```diff
@@ -22,5 +22,27 @@
...
+        return latest_tag
```
michamos: what if the current commit is not tagged? isn't that misleading?

harunurhan: I think it's getting latest one tag of all tags, not from the latest commit.

drjova: I'll change it to the latest ``sha``

michamos: not the current `sha`?

drjova: yep, the head ``sha ``

inspirehep/version.py
drjova: good point 👍

drjova:
@michamos @harunurhan updated

drjova:
updated

PR #3841, opened 2019-03-11
v0.12.23

drjova:


inspirehep/modules/workflows/views.py
```diff
@@ -745,6 +763,30 @@ def start_author_workflow():
     return jsonify({'workflow_object_id': workflow_object_id})
 
 
+@workflow_blueprint.route('/literature', methods=['POST'])
```
monaawi: we should add @require_api_auth here as well

drjova: I'll create US for this

PR #3844, opened 2019-03-12
migration: do not index deleted records

monaawi:
INSPIR-2061


monaawi:
We should also reindex so that we fix the current records

PR #3855, opened 2019-04-30
cli: decrease query time for simpleindex command

ammirate:
This commit removes from the query for reindexing the check for deleted record. This is possible because the `batch_reindex` task already filters out deleted record when the metadata is loaded to be enhanced before reindexing.

* INSPIR-2353

### DIFF
--- a/inspirehep/modules/records/cli.py
+++ b/inspirehep/modules/records/cli.py
@@ -58,11 +58,7 @@ from inspirehep.modules.search.api import LiteratureSearch
 
 
 from sqlalchemy import (
-    String,
     cast,
-    type_coerce,
-    or_,
-    not_
 )
 
 from sqlalchemy.dialects.postgresql import JSONB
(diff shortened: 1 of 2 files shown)


ammirate:
Estimated time for querying and sending indexing tasks is reduced from 46 min ca. to 8 min ca.

inspirehep/modules/records/cli.py
```diff
@@ -133,12 +130,6 @@ def get_query_records_to_index(pid_types):
             PersistentIdentifier.pid_type.in_(pid_types),
             PersistentIdentifier.object_type == 'rec',
             PersistentIdentifier.status == PIDStatus.REGISTERED,
-            or_(
-                not_(
-                    type_coerce(RecordMetadata.json, JSONB).has_key('deleted')
-                ),
-                RecordMetadata.json["deleted"] == cast(False, JSONB)
-            )
```
michamos: is the `join` still needed now you don't need to access the record any more?

ammirate: yep

PR #3859, opened 2019-05-09
workflows: Allow workflows to create records via HTTP with inspirehep

pazembrz:
* add: Feature flag: ENABLE_INSPIREHEP_REMOTE_RECORD_MANAGEMENT - when set to true will save records through inspirehep records API

* INSPIR-2311

### DIFF
--- a/inspirehep/modules/workflows/tasks/upload.py
+++ b/inspirehep/modules/workflows/tasks/upload.py
@@ -24,7 +24,9 @@
 
 from __future__ import absolute_import, division, print_function
 
+import requests
 from flask import current_app
+from invenio_workflows.errors import WorkflowsError
 
 from inspire_schemas.readers import LiteratureReader
 from invenio_db import db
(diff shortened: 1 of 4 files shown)


inspirehep/modules/workflows/tasks/upload.py
```diff
@@ -43,34 +45,106 @@ def store_record(obj, eng):
...
-            record.clear()
-            record.update(obj.data, files_src_records=[obj])
+@with_debug_logging
+def store_record_inspirehep_api(obj, eng, is_update, is_authors):
+    """Saves record through inspirehep api by posting/pushing record to proper endpoint
+     in inspirehep"""
+    headers = {
+        "content-type": "application/json",
+        "Authorization": "Bearer {token}".format(
+            token=current_app.config['AUTHENTICATION_TOKEN']
```
michamos: make the config key name less generic? `INSPIREHEP_AUTHENTICATION_TOKEN`?

inspirehep/modules/workflows/tasks/upload.py
```diff
@@ -43,34 +45,106 @@ def store_record(obj, eng):
...
-                )
-                return
 
-            record = InspireRecord.get_record(obj.extra_data['head_uuid'])
-            obj.data['control_number'] = record['control_number']
-            record.clear()
-            record.update(obj.data, files_src_records=[obj])
+@with_debug_logging
+def store_record_inspirehep_api(obj, eng, is_update, is_authors):
+    """Saves record through inspirehep api by posting/pushing record to proper endpoint
```
michamos: `Save`

ammirate:
I would add at least a test for a full workflow run to check that things work properly all together, since in the current tests we are injecting "fake" data in the workflow to make things happen

inspirehep/modules/workflows/tasks/upload.py
pazembrz: We have already the same key name for Bearer token in inspirehep. We should keep it with the same name in both projects.

inspirehep/modules/workflows/tasks/upload.py
pazembrz: It's already at the end of `store_record_inspirehep_api` but I had to fix indentation.

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -1147,3 +1149,160 @@ def test_match_in_holdingpen_different_sources_continues(
...
+    assert obj.extra_data["holdingpen_matches"] == []
+    assert obj.extra_data["previously_rejected"] is False
+    assert not obj.extra_data.get("stopped-matched-holdingpen-wf")
+    assert obj.extra_data["is-update"]
+    assert obj.extra_data["exact-matched"]
+    assert obj.extra_data["matches"]["exact"] == [record.get("control_number")]
+    assert obj.extra_data["matches"]["approved"] == record.get("control_number")
+    assert obj.extra_data["approved"]
+    assert obj.status == ObjectStatus.COMPLETED
+
```
ammirate: Great, we should only check that `requests_mocker` has been called once with the proper args

inspirehep/modules/workflows/tasks/upload.py
michamos: I mean the first word should be `Save` instead of `Saves`

inspirehep/modules/workflows/tasks/upload.py
michamos: should we? in one case it's its own token, in the other case it's a foreign token. As long as it's clear what this authentication token is.

PR #3864, opened 2019-05-17
workflow: record update also should use inspirehep API

pazembrz:


inspirehep/modules/workflows/workflows/edit_article.py
```diff
@@ -43,15 +44,19 @@ def change_status_to_waiting(obj, eng):
...
-    record = get_db_record('lit', control_number)
-    record.update(obj.data)
-    record.commit()
-
-    user_id = obj.id_user
-    if user_id:
-        _set_transaction_user_id_for_last_record_update(control_number, user_id)
-
-    db.session.commit()
+    if current_app.config.get("ENABLE_INSPIREHEP_REMOTE_RECORD_MANAGEMENT"):
```
drjova: is this feature flag already in place? all of our feature flags are like ``FEATURE_FLAG_ENABLE_FOO``

tests/integration/workflows/test_edit_article.py
```diff
@@ -383,3 +385,94 @@ def test_edit_article_callback_does_not_have_redirect_url_if_no_ticket_id(
     assert response.status_code == 200
     data = json.loads(response.data)
     assert 'redirect_url' not in data
+
+
+def test_edit_article_workflow_sending_to_hep(workflow_app, mocked_external_services):
+    app_client = workflow_app.test_client()
```
drjova: do we an test case if hep replies with ``500``?

inspirehep/modules/workflows/workflows/edit_article.py
pazembrz: Yes, this flag is already in place on QA. I know that we are using `FEATURE_FLAG_ENABLE_` prefix but this flag felt too long, Still i can add this prefix to it's name if you wish.

tests/integration/workflows/test_edit_article.py
pazembrz: There are no tests for 500, only for 401, but all error codes are handled the same way.

tests/integration/workflows/test_workflows_tasks_upload.py
```diff
@@ -233,7 +232,9 @@ def test_store_record_inspirehep_api_literature_new(workflow_app):
                 headers={'content-type': 'application/json'},
                 status_code=201,
                 json={
-                    'control_number': expected_control_number,
+                    "metadata": {
```
drjova: why now is different? :/ we haven't changed anything

pazembrz: It's not different, previously it was wrong.

drjova: how it was passing the tests :/

pazembrz: Data I copied as a mocked response from inspirehep were wrong. It was my mistake. When I was copying it i missed one level of nested data.

PR #3868, opened 2019-06-14
loader: fix broken test with latest inspire-schemas

drjova:

### DIFF
--- a/tests/integration/records/test_records_json_ref_loader.py
+++ b/tests/integration/records/test_records_json_ref_loader.py
@@ -22,15 +22,10 @@
 
 from __future__ import absolute_import, division, print_function
 
-from inspire_schemas.utils import load_schema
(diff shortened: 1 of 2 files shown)


tests/integration/records/test_records_json_ref_loader.py
```diff
@@ -29,7 +29,7 @@
 def test_load_resolved_schema(app):
     unresolved_schema = load_schema('hep')
 
-    assert '$ref' in unresolved_schema['properties']['urls']['items']
+    assert '$ref' not in unresolved_schema['properties']['urls']['items']
```
michamos: I would remove the `assert`, as the behavior we really care about is that at the end schemas are resolved.

drjova: done

michamos:
unused import: https://travis-ci.org/inspirehep/inspire-next/jobs/546625162#L1312

PR #3869, opened 2019-06-17
edit_article: add ``head_uuid``

drjova:


inspirehep/modules/workflows/views.py
```diff
@@ -705,7 +705,7 @@ def start_edit_article_workflow(recid):
         if ticket_match:
             ticket_id = int(ticket_match.group('ticket_id'))
             workflow.extra_data['curation_ticket_id'] = ticket_id
-
+    workflow.extra_data['head_uuid'] = record.id
```
michamos: the other workflows do `str(head_uuid)`, I guess otherwise it's not serializable to JSON as it's a UUID.

PR #3877, opened 2019-07-15
workflows: rename id_ to uuid

monaawi:
* Receive uuid instead of _id in store record because Inspirehep no longer send id_, it now sends uuid 
* INSPIR-2572


monaawi:
Merge with the inspirehep fix

PR #3878, opened 2019-07-30
Use the new docker hub repos

benjamin-bergia:


michamos:
@benjamin-bergia is this still needed?

PR #3880, opened 2019-08-06
workflows: retry while managing record via hep

harunurhan:
* Retries when inspirehep returns bad gateway error possibly due to
network issues.

### DIFF
--- a/inspirehep/modules/workflows/tasks/upload.py
+++ b/inspirehep/modules/workflows/tasks/upload.py
@@ -28,12 +28,14 @@ import requests
 from flask import current_app
 from invenio_workflows.errors import WorkflowsError
 from simplejson import JSONDecodeError
+import backoff
 
 from inspire_schemas.readers import LiteratureReader
 from invenio_db import db
 
 from inspirehep.modules.records.api import InspireRecord
 from inspirehep.modules.workflows.models import WorkflowsRecordSources
+from inspirehep.modules.workflows.errors import BadGatewayError
 from inspirehep.modules.workflows.utils import (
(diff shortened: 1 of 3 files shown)


drjova:
@harunurhan you can rebase now

inspirehep/modules/workflows/tasks/upload.py
```diff
@@ -143,13 +146,15 @@ def send_record_to_hep(obj, endpoint, control_number=None):
...
+        raise BadGatewayError()
+
     try:
         error_msg = response.json()
     except JSONDecodeError:
         error_msg = response.text
     raise WorkflowsError(
-        "Response code from Inspirehep is {code}."
-        " This is wrong! Message from the server: {message}".format(
+        "Error from inspirehep [${code}]: {message}".format(
```
drjova: python :P

tests/integration/workflows/test_workflows_tasks_upload.py
```diff
@@ -359,6 +359,44 @@ def test_store_record_inspirehep_api_author_update(workflow_app):
...
+                        'uuid': expected_head_uuid
+                    }}
+                ],
+            )
+            store_record(workflow, eng)
+
+    assert workflow.data['control_number'] == expected_control_number
+    assert workflow.extra_data['head_uuid'] == expected_head_uuid
+
+
```
drjova: can we add a test with the update? ``200``

harunurhan: there is already one

PR #3884, opened 2019-08-16
workflows: is_pdf checks now more than first characters of PDF

pazembrz:
Some of the PDFs from arxiv have other data before '%PDF'. This change also considers this PDFs as correct ones.

INSPIR-2645


inspirehep/utils/url.py
```diff
@@ -50,8 +50,8 @@ def make_user_agent_string(component=""):
 def is_pdf_link(url):
     """Return ``True`` if ``url`` points to a PDF.
 
-    Returns ``True`` if the first non-whitespace characters of the response are
-    ``%PDF``.
+    Returns ``True`` if the first 512 haracters of the response contains
```
monaawi: characters

inspirehep/utils/url.py
```diff
@@ -65,9 +65,9 @@ def is_pdf_link(url):
     except requests.exceptions.RequestException:
         return False
 
-    significant_bytes = next(response.iter_content(10000), '').strip()
+    found = next(response.iter_content(10000), '').find('%PDF')
```
drjova: why rename it from ``significant_bytes`` and not do the test speratly? It's a bit confusing IMO :/

pazembrz: Because this are not significant bytes anymore? This variable name was confusing.

harunurhan: I see that this has been done like that before, but I do not get why we just don't check `Content-Type: application/pdf` response header :/

pazembrz: Looks like this is a requirement not to check it: https://github.com/inspirehep/inspire-next/blob/28630eab052ef798a874f9ca4a731c5ffeab9782/tests/unit/forms/test_forms_validators_simple_fields.py#L139

tsgit:
this is technically not correct. The test file 
http://arxiv.org/pdf/physics/0404071
is not a valid PDF file. It is a macbinary file that should be properly unpacked

```
$ file 0404071.pdf 
0404071.pdf: MacBinary II, Fri Apr 16 04:50:21 2004, modified Fri Apr 16 09:51:55 2004, creator 'CARO', type Portable Document Format, 523329 bytes "LS_proceedings9.pdf" PDF document, version 1.3
```

```
$ macunpack -v -i  ~/0404071.pdf
This is "MacBinary" input.
name="LS_proceedings9.pdf", type=PDF_, author=CARO, data=523329, rsrc=0
        Data: No compression, Rsrc: empty.
```
```
$ macunpack -v -3  ~/0404071.pdf
This is "MacBinary" input.
name="LS_proceedings9.pdf", type=PDF_, author=CARO, data=523329, rsrc=0
        Data: No compression, Rsrc: empty.
```
results in
```
-rw-r--r--. 1 thorsten thorsten 523329 Sep  4 15:06 LS_proceedings9.pdf.data
-rw-r--r--. 1 thorsten thorsten    128 Sep  4 15:06 LS_proceedings9.pdf.info
-rw-r--r--. 1 thorsten thorsten      0 Sep  4 15:06 LS_proceedings9.pdf.rsrc
```
with
```
$ file LS_proceedings9.pdf.data
LS_proceedings9.pdf.data: PDF document, version 1.3
```
If nothing else the leading and trailing stuff should be stripped, e.g. everything outside of `%PDF ... %%EOF`

Some pdf tools can handle such files. Nevertheless they are not compliant with the spec.

Also, I would inform arXiv and have them fix that file.
There are other issues with this PDF, `creator CARO` is creating bad PDF and arXiv has some tools to amend that.

tsgit:
@pazembrz

inspirehep/utils/url.py
tsgit: you absolutely don't want to rely on `Content-Type` for file format verification. Most sites (web servers) use simple mime type based on file extension to set `Content-Type` for static files. It doesn't guarantee that the file actually is what it claims. A proper check is to do introspection,
at least use file magic (there's pleny of libraries for that) or if you want to be more thorough use JHOVE https://jhove.openpreservation.org/  or similar

PR #3885, opened 2019-08-16
setup: deploy on QA webhook

drjova:

### DIFF
--- a/deploy-docker-image.sh
+++ b/deploy-docker-image.sh
@@ -19,9 +19,10 @@ buildPush() {
   DOCKERFILE="${2:-Dockerfile}"
   GIT_DESC="$(git describe --always || echo)"
 
+
   echo "Building docker image"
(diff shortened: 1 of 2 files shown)


deploy-docker-image.sh
```diff
@@ -45,12 +45,22 @@ logout() {
   retry docker logout
 }
 
+deployQA() {
+  curl -X POST \
+    -F token=${DEPLOY_QA_TOKEN} \
+    -F ref=master \
+    -F variables[IMAGE_NAME]=inspirehep/hep \
```
benjamin-bergia:
```suggestion
    -F variables[IMAGE_NAME]=inspirehep/next \
```

benjamin-bergia: there is also a call to make for `inspirehep/next-scrapyd` and `inspirehep/next-assets`, all using the same tag

drjova: true :/

PR #3891, opened 2019-10-01
cli: add workflows CLI with purge command

ammirate:
* INSPIR-2744

Signed-off-by: Antonio Cesarano <cesarano2607@gmail.comk>


inspirehep/modules/workflows/cli.py
```diff
@@ -0,0 +1,76 @@
...
+    """Workflows management command."""
+    pass
+
+
+@workflows.command()
+@click.option('--yes-i-know', is_flag=True)
+@with_appcontext
+def purge(yes_i_know):
+    """Removes every entry from DB and ES related to workflows"""
+    query = "TRUNCATE {tables} RESTART IDENTITY CASCADE"
```
michamos: `cascade` shouldn't be needed. Why have you added this?

inspirehep/modules/workflows/cli.py
```diff
@@ -0,0 +1,76 @@
...
+        click.confirm(
+            'This operation will irreversibly remove data from DB and ES, do you want to continue?',
+            abort=True
+        )
+    click.secho('Truncating tables from DB:\n* {}'.format('\n* '.join(TABLES)))
+    db.session.execute(query.format(tables=', '.join(TABLES)))
+    db.session.commit()
+
+    click.secho('Removing workflows indices from ES...')
+    es.indices.delete(index='holdingpen', ignore=[400, 404])
```
michamos: why change this from the script?

inspirehep/modules/workflows/cli.py
```diff
@@ -0,0 +1,76 @@
...
+
+    click.secho('Recreating indices...')
+    for index in ES_INDICES:
+        es.indices.create(index=index, ignore=400)
```
michamos: same

ammirate: To comply to the above statement, rather than using both `current_search_client` and `current_search`.

inspirehep/modules/workflows/cli.py
ammirate: Leftover :+1: It is not needed if you truncate the tables all together but if you do it one by one, than you need it since there's a foreign key between some tables.

setup.py
```diff
@@ -41,6 +41,7 @@
 setup_requires = []
 
 install_requires = [
+    'amqp===2.5.1',  # FIXME: required by kombu, Travis fails without it
```
pazembrz: It's not the correct way to fix this. `amqp` 2.5.2 is forced by `kombu` version 4.6.5 which actually required `amqp` in version 2.5.1, its better to skip `kombu` in version 4.6.5 as the only change it introduces is `amqp` requirement updated from 2.5.1 to 2.5.2 https://github.com/celery/kombu/commit/b51d1d678e198a80d7e5fd95f32674c7d8e04a75#diff-f7b7288f81db184bb127adf1389dafc4

inspirehep/modules/workflows/cli.py
```diff
@@ -0,0 +1,78 @@
...
```
pazembrz: Here you would prefer to use current_search.create(ignore_existing=True) instead of current_search_client.create(index) as for now invenio is creating indexes with postfix and under expected index name alias is created. If you uise current_search_engine.create it will create index without alias and it won't work with our code now.

PR #3902, opened 2019-11-20
cli: add restart wf by error message command

ammirate:
* INSPIR-2931

### DIFF
--- a/inspirehep/modules/workflows/cli.py
+++ b/inspirehep/modules/workflows/cli.py
@@ -29,6 +29,8 @@ import click
 from flask.cli import with_appcontext
 from invenio_db import db
 from invenio_search import current_search
+from invenio_workflows import workflow_object_class, ObjectStatus
+from invenio_workflows.models import WorkflowObjectModel
 
 
 TABLES = [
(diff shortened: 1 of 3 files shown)


inspirehep/modules/workflows/cli.py
```diff
@@ -75,3 +77,34 @@ def purge(yes_i_know):
...
+    errors = WorkflowObjectModel.query.filter_by(status=ObjectStatus.ERROR).all()
+    to_restart = [e.id for e in errors if error_message in e.extra_data['_error_msg']]
+
+    click.secho("Found {} workflows to restart from {}".format(
+        len(to_restart),
+        "first step" if from_beginning else "current step"
+    ))
+
+    for wf_id in to_restart:
+        obj = workflow_object_class.get(wf_id)
```
pazembrz: You could also change state from error to running or new if started from scratch as otherwise it will be in erorr state until status will be changed which may happen few minutes later.

ammirate: I set `INITIAL` for workflows restarted from the beginning and `RUNNING` for those restarted from current step. :+1:

PR #3908, opened 2019-12-09
setup: bump invenio-oauth2server to v1.0.4

drjova:


ammirate:
After deploy to prod, we can also delete our fork.

PR #3910, opened 2019-12-11
setup: pin coverage.py

jacquerie:
Collecting the coverage metrics on Coveralls has been broken since
September 2018 due to a change in the format of the .coverage file
that happened in coverage==5.0a2. So we can restore it by pinning
the previous and current stable version.

Last build that passed: https://coveralls.io/builds/18814566
Offending release of `coverage`: https://pypi.org/project/coverage/5.0a2/
This build passes: https://coveralls.io/builds/27541479

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -147,6 +147,7 @@ docs_require = [
 ]
 
 tests_require = [
+    'coverage~=4.0,>=4.5.4',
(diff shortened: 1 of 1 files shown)


drjova:
@jacquerie thank you

PR #3913, opened 2020-01-07
global: addition of poetry

drjova:


Dockerfile
```diff
@@ -64,7 +64,10 @@ RUN npm install -g \
...
- pip install --no-cache-dir --upgrade setuptools && \
- pip install --no-cache-dir --upgrade wheel && \
- pip install --no-cache-dir -e .[all,xrootd] -r requirements.txt
+ENV PATH="/root/.poetry/bin:${PATH}" \
+    POETRY_VIRTUALENVS_CREATE=false
+
+RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python \
+    && poetry --version
+
+RUN poetry install -E all
```
pazembrz: Maybe you should also add `-n` for no interaction.

docker-compose.deps.yml
```diff
@@ -27,7 +27,7 @@ services:
     extends:
       file: services.yml
       service: base
-    command: pip install --editable .[all] --find-links /pip-cache --pre --requirement requirements.txt --upgrade
+    command: poetry install -E all
```
pazembrz: same

Dockerfile
michamos: no `xrootd` on purpose?

docker-compose.deps.yml
```diff
@@ -27,7 +27,7 @@ services:
     extends:
```
michamos: so `docker-compose -f docker-compose.deps.yml run --rm pip` now runs `poetry` maybe also change the name of the service?

docker-compose.deps.yml
```diff
@@ -27,7 +27,7 @@ services:
...
       file: services.yml
       service: base
-    command: pip install --editable .[all] --find-links /pip-cache --pre --requirement requirements.txt --upgrade
+    command: bash -c "pip install poetry && poetry export --without-hashes -f requirements.txt > requirements.txt && pip install --pre --requirement requirements.txt"
```
pazembrz: You should also install here inspirehep itself

drjova:
@pazembrz updated

PR #3934, opened 2020-02-04
holdingpen: Show more than 10k in boxes total number

miguelgrc:
* INSPIR-3108


miguelgrc:
Not testing it since it requires to create 10k+ workflows, which takes several minutes so it's not worth it. This change only affects UI anyway.

PR #3935, opened 2020-02-05
workflow: add submission_pdf from formdata.url

pazembrz:


inspirehep/modules/workflows/views.py
```diff
@@ -780,7 +780,10 @@ def start_workflow_for_literature_submission():
         'extra_data': copy.deepcopy(workflow_object.extra_data),
         'data': copy.deepcopy(workflow_object.data),
     }
-
+    # Add submission pdf from formdata.url
+    form_url = workflow_object.extra_data['formdata'].get('url')
+    if form_url and 'arxiv.org' not in form_url:
+        workflow_object.extra_data['submission_pdf'] = form_url
```
drjova: do we have a test that the files is indeed attached to the workflow?

pazembrz: It's added in this PR.

pazembrz: There is also a test which checks if document from submission_pdf is added to documents: https://github.com/inspirehep/inspire-next/blob/02d98c243586894ce4390aee69d6bd2b49000f86/tests/unit/workflows/test_workflows_tasks_actions.py#L875

PR #3938, opened 2020-02-12
workflows: quick fix formdata_url

pazembrz:
submission_url should be filled before extra_data will be moved to sourcE_data

### DIFF
--- a/inspirehep/modules/workflows/views.py
+++ b/inspirehep/modules/workflows/views.py
@@ -776,14 +776,19 @@ def start_workflow_for_literature_submission():
     submission_data['acquisition_source']['submission_number'] = str(workflow_object.id)
     workflow_object.data = submission_data
     workflow_object.extra_data['formdata'] = json['form_data']
-    workflow_object.extra_data['source_data'] = {
-        'extra_data': copy.deepcopy(workflow_object.extra_data),
-        'data': copy.deepcopy(workflow_object.data),
-    }
+
     # Add submission pdf from formdata.url
     form_url = workflow_object.extra_data['formdata'].get('url')
     if form_url and 'arxiv.org' not in form_url:
         workflow_object.extra_data['submission_pdf'] = form_url
+
+    # Remember that source_data should be created at the end, with all fields already
+    # filled. As first step in WF will overwrite everything with data form source_data
+    workflow_object.extra_data['source_data'] = {
+        'extra_data': copy.deepcopy(workflow_object.extra_data),
+        'data': copy.deepcopy(workflow_object.data),
+    }
+
     workflow_object.save()
     db.session.commit()
 


inspirehep/modules/workflows/views.py
```diff
@@ -776,14 +776,19 @@ def start_workflow_for_literature_submission():
...
-        'data': copy.deepcopy(workflow_object.data),
-    }
+
     # Add submission pdf from formdata.url
     form_url = workflow_object.extra_data['formdata'].get('url')
     if form_url and 'arxiv.org' not in form_url:
         workflow_object.extra_data['submission_pdf'] = form_url
+
+    # Remember that source_data should be created at the end, with all fields already
+    # filled. As first step in Wf will overwrite exerythiong with data form source_data
```
drjova: I'd just remove this comment :)

pazembrz: The problem was exactly because no one noticed that previous fix was adding `submission_pdf` after setting source_data. I added it so this not happen again. Why you would like to remove this comment?

drjova: if you'd like to keep it's ok just please fix the typos :)

```suggestion
    # filled. As first step in Wf will overwrite everything with data form source_data
```

pazembrz: :+1:

inspirehep/modules/workflows/views.py
```diff
@@ -776,14 +776,19 @@ def start_workflow_for_literature_submission():
...
+
     # Add submission pdf from formdata.url
     form_url = workflow_object.extra_data['formdata'].get('url')
     if form_url and 'arxiv.org' not in form_url:
         workflow_object.extra_data['submission_pdf'] = form_url
+
+    # Remember that source_data should be created at the end, with all fields already
+    # filled. As first step in WF will overwrite everything with data form source_data
+    workflow_object.extra_data['source_data'] = {
+        'extra_data': copy.deepcopy(workflow_object.extra_data),
```
michamos: I hate to be this guy, but I have to ask, why do you need a `deepcopy`? In order to avoid the circular reference, it looks to me that a `copy` is enough.
```suggestion
        'extra_data': copy.copy(workflow_object.extra_data),
```

inspirehep/modules/workflows/views.py
```diff
@@ -776,14 +776,19 @@ def start_workflow_for_literature_submission():
...
+        'data': copy.deepcopy(workflow_object.data),
```
michamos: I don't think you even need the copy here, as you're storing `data` in `extra_data`.
```suggestion
        'data': workflow_object.data,
```

inspirehep/modules/workflows/views.py
pazembrz: I'm not author of this code.

inspirehep/modules/workflows/views.py
pazembrz: I'm not author of this code.

inspirehep/modules/workflows/views.py
michamos: Now you are :)

PR #3939, opened 2020-02-17
refextract: handle additional pubnotes

michamos:
* INSPIR-3190

No tests, as I don't want to touch workflow tests and the feature is already tested in inspire-schemas and inspire-dojson.


drjova:
@michamos merged dojson part

michamos:
@drjova thanks. Lots of integration tests are failing, I guess due to invenio updates (I ran `poetry update`). Should I revert the lockfile and just run `poetry update inspire-schemas`? or do you want to pin packages?

drjova:
the packages are pinned https://github.com/inspirehep/inspire-next/blob/master/pyproject.toml#L46-L71 that's strange, could you please just update schemas?

PR #3940, opened 2020-02-24
poetry: bump refextract to v0.2.22

michamos:


poetry.lock
```diff
@@ -2716,7 +2716,7 @@ description = "MessagePack (de)serializer."
 name = "msgpack"
 optional = false
 python-versions = "*"
-version = "0.6.2"
+version = "1.0.0"
```
michamos: maybe we should pin it ? see https://github.com/msgpack/msgpack-python#major-breaking-changes-in-msgpack-10

drjova: 👍

drjova:
Sometimes it happens, I just restarted the build

PR #3945, opened 2020-04-06
poetry: bump inspire-dojson

michamos:


poetry.lock
```diff
@@ -1385,13 +1388,13 @@ description = "INSPIRE-specific rules to transform from MARCXML to JSON and back
 name = "inspire-dojson"
 optional = false
 python-versions = "*"
-version = "63.0.2"
+version = "63.0.8"
```
michamos, line 1391: this should be `63.0.9` but can't convince poetry to get that one, although it's on PyPI :cry:

PR #3946, opened 2020-04-07
RT: submission approval ticket link to labs

miguelgrc:
* INSPIR-3370


pazembrz:
Maybe a test which would check url in context for a ticket would be appropriate? :)

PR #3948, opened 2020-04-16
views: move revision endpoints to hep

miguelgrc:
* INSPIR-3118


drjova:
LGTM, I'd leave this PR open until we'll make sure the new routes setup in k8s works properly

PR #3950, opened 2020-06-17
Dockerfile: upgrade pip when building image

michamos:


michamos:
I checked locally that this builds correctly, so I'll just merge it.

PR #1134, opened 2016-05-19
profiles: gravatar styling

jacenkow:
* Small cosmetics applied to gravatar image.

Signed-off-by: Grzegorz Jacenków <grzegorz.jacenkow@cern.ch>

PR #3954, opened 2020-07-24
workflows: disable wait_for_webcoll for articles

pazembrz:
ref #1134

### DIFF
--- a/inspirehep/modules/workflows/views.py
+++ b/inspirehep/modules/workflows/views.py
@@ -273,6 +273,11 @@ def _is_an_update(workflow_id):
     return bool(workflow_object.extra_data.get('is-update'))
 
 
+def _is_an_authors_workflow(workflow_id):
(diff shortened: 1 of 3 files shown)


inspirehep/modules/workflows/views.py
```diff
@@ -288,30 +293,32 @@ def _parse_robotupload_result(result, workflow_id):
...
-        return response
 
-    if not _is_an_update(workflow_id):
-        pending_entry = WorkflowsPendingRecord(
-            workflow_id=workflow_id,
-            record_id=recid,
-        )
-        db.session.add(pending_entry)
-        db.session.commit()
+    if _is_an_authors_workflow(workflow_id):
```
drjova, line 297: Shall we move `already_pending_ones` inside the `if` (one query less)?

pazembrz: :+1:

PR #3955, opened 2020-07-27
poetry: update inspire-json-merger

pazembrz:
Also use celery from official branch

ref: inspirehep/inspire-json-merger#81

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ babel = "==2.*,>=2.0.0,>=2.5.1"
 backoff = "==1.*,>=1.0.0,>=1.4.3"
 "backports.tempfile" = ">=1.0rc1"
 beard = "==0.*,>=0.0.0,>=0.2.0"
-celery = {git = "https://github.com/michamos/celery.git", branch = "fix-redis-connection-result-consumer"}
+celery = "==4.4.2"
 click = "<7.0,==6.*,>=6.7.0"
 click-spinner = "==0.*,>=0.0.0,>=0.1.8"
 coverage = {version = "==4.*,>=4.0.0,>=4.5.4", optional = true}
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -14,7 +14,7 @@ babel = "==2.*,>=2.0.0,>=2.5.1"
 backoff = "==1.*,>=1.0.0,>=1.4.3"
 "backports.tempfile" = ">=1.0rc1"
 beard = "==0.*,>=0.0.0,>=0.2.0"
-celery = {git = "https://github.com/michamos/celery.git", branch = "fix-redis-connection-result-consumer"}
+celery = "==4.4.6"
```
michamos, line 17: why not
```suggestion
celery = "^4.4.6"
```

pazembrz: On HEP we are pinning to specific version as we decided some time ago that all packages should be pinned manually to specific version. Don't we want to do the same thing on NEXT?

michamos: I was not aware of this policy. I don't really see the point (as packages are locked by the `poetry.lock` file anyway), but if that's the policy that's ok for me.

pyproject.toml
```diff
@@ -14,7 +14,7 @@ babel = "==2.*,>=2.0.0,>=2.5.1"
 backoff = "==1.*,>=1.0.0,>=1.4.3"
 "backports.tempfile" = ">=1.0rc1"
 beard = "==0.*,>=0.0.0,>=0.2.0"
-celery = {git = "https://github.com/michamos/celery.git", branch = "fix-redis-connection-result-consumer"}
+celery = "==4.4.2"
```
michamos, line 17: again, I don't understand this approach. If you had specified a less strict version, poetry would have computed for you which versions of packages to get to solve the sphinx version conflict issue that you mentioned IRL. In fact, that's the whole reason to use something like poetry. If you're pinning everything, you might as well use `requirements.txt` with pinned deps.

PR #3966, opened 2020-09-24
matcher: replace match_references with call to hep

monaawi:
* ref:  inspirehep/inspirehep#1291


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -402,6 +405,32 @@ def download_documents(obj, eng):
                 'Cannot download document from %s', url)
 
 
+def match_references_hep(references):
```
drjova, line 408: Could we add `backoff` for network issues?

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -402,6 +405,32 @@ def download_documents(obj, eng):
...
+        return response.json()['references']
+    else:
+        create_error(response)
+
+
+def match_references_based_on_flag(references):
+    if current_app.config.get("FEATURE_FLAG_ENABLE_MATCH_REFERENCES_HEP"):
+        return match_references_hep(references)
+    else:
+        return match_references(references)
```
drjova, line 431:
```suggestion
def match_references_based_on_flag(references):
    if current_app.config.get("FEATURE_FLAG_ENABLE_MATCH_REFERENCES_HEP"):
        return match_references_hep(references)
    return match_references(references)
```

PR #3967, opened 2020-09-29
matcher: replace match_references with call to hep

monaawi:
* ref:  inspirehep/inspirehep#1291

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -26,7 +26,10 @@ from __future__ import absolute_import, division, print_function
 
 import sys
 import time
+import backoff
 
+import json
+import requests
 from copy import deepcopy
 from functools import wraps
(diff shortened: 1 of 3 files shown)


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -402,6 +407,32 @@ def download_documents(obj, eng):
...
+        "{inspirehep_url}/api/matcher/linked_references/".format(
+            inspirehep_url=inspirehep_url,
+        ),
+        headers=headers,
+        json=json.dumps(data)
+    )
+    if response.status_code == 200:
+        return response.json()['references']
+    else:
+        create_error(response)
```
drjova, line 427:
```suggestion
    if response.status_code == 200:
        return response.json()['references']
    create_error(response)
```

drjova (dismissed):
LGTM! Just the comment (is not important though)

PR #3968, opened 2020-10-01
editor: add replace xrootd with http service

drjova:


inspirehep/modules/editor/api.py
```diff
@@ -224,8 +236,7 @@ def upload_files():
     full_url = fs.getpathurl(filename, allow_none=True)
     if not full_url:
         full_url = fs.getsyspath(filename)
-
-    return jsonify({'path': full_url})
+    return jsonify({'path': _replace_xrootd_with_http(full_url)})
```
michamos, line 239: I think I'm missing the context. Is the idea to get the file directly from EOS HTTP rather than xrootd because `hep` doesn't know about it, so it wouldn't work? what about auth & ACLs? what about the "real" editor on `hep` that we'll use once legacy is gone?

PR #3972, opened 2020-10-15
editor: add endpoint to get workflow and schema

harunurhan:
Ref #1367

### DIFF
--- a/inspirehep/modules/editor/views.py
+++ b/inspirehep/modules/editor/views.py
@@ -24,10 +24,13 @@
 
 from __future__ import absolute_import, division, print_function
 
-from flask import Blueprint, current_app, render_template, request
+from flask import Blueprint, current_app, render_template, request, jsonify
 from flask_login import login_required
+from invenio_workflows import workflow_object_class
+from inspire_schemas.api import load_schema
 
 from inspirehep.modules.records.utils import get_endpoint_from_record
+from inspirehep.modules.editor.permissions import editor_use_api_permission
 
 blueprint = Blueprint(
     'inspirehep_editor',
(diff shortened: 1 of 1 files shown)


inspirehep/modules/editor/views.py
```diff
@@ -54,3 +55,23 @@ def preview():
...
+@blueprint.route('/holdingpen/<int:workflow_id>', methods=['GET'])
+@login_required
+def get_workflow_and_schema(workflow_id):
+    workflow = workflow_object_class.get(workflow_id)
+    schema_url = workflow.data['$schema']
+    schema_name = (
+        six.moves.urllib.parse.urlsplit(schema_url)
+        .path.split("/")[-1]
+        .split(".")[0]
+    )
```
michamos, line 69: I don't think you need this, you can pass the URL directly to `load_schema` which will try to extract the schema from there, using https://github.com/inspirehep/inspire-schemas/blob/ab35f92f667cea0974cb54f717ef6913ba30cf65/inspire_schemas/utils.py#L595

harunurhan: well that's not that important here that much, but I actually took it from `inspirehep` I think we should replace it there too!

michamos: I agree :+1:

harunurhan: @michamos perhaps it would be good to make `load_schema` to accept absolute schema urls too

harunurhan: turns out we really need the name in that case and it's not for `load_schema`

inspirehep/modules/editor/views.py
```diff
@@ -54,3 +56,18 @@ def preview():
     endpoint = get_endpoint_from_record(record)
     template = current_app.config['RECORDS_UI_ENDPOINTS'][endpoint]['template']
     return render_template(template, record=record)
+
+
+@blueprint.route('/holdingpen/<int:workflow_id>', methods=['GET'])
+@login_required
```
drjova, line 62: Maybe just for superusers and curators? That means anyone logged can retrieve info about this workflow

drjova (approved):
LGTM, just the comment. It's up to you

drjova (approved):

PR #3973, opened 2020-10-15
author workflow: remove legacy dependencies

MJedr:
Ref: inspirehep/inspirehep#1346


drjova (approved):

drjova:
It's interesting that no tests broke after this change.

PR #3975, opened 2020-10-16
editor: add _workflow to the reource api

harunurhan:


drjova (approved):

inspirehep/modules/editor/api.py
```diff
@@ -248,6 +248,7 @@ def get_workflow_and_schema(workflow_id):
         "workflow": {
             "id": workflow.id,
             "_extra_data": workflow.extra_data,
+            "_workflow": workflow.workflow,
```
michamos, line 251: `workflow._workflow`, you're copying ES `hits.hits` :smile:

michamos (approved):

inspirehep/modules/editor/api.py
harunurhan: well there was no option tbh, I had to imitate the `/api/holdingpen/<id>`

PR #3981, opened 2020-10-26
Fix on author update

pazembrz:
* Fix sending author record in correct mode to legacy when updating.

### DIFF
--- a/inspirehep/modules/workflows/workflows/author.py
+++ b/inspirehep/modules/workflows/workflows/author.py
@@ -53,7 +53,7 @@ from inspirehep.modules.workflows.utils import do_not_repeat
 
 
 SEND_TO_LEGACY = [
-    send_robotupload(mode="insert", priority_config_key='LEGACY_ROBOTUPLOAD_PRIORITY_AUTHOR'),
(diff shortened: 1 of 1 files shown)


inspirehep/modules/workflows/workflows/author.py
```diff
@@ -53,7 +53,7 @@
 
 
 SEND_TO_LEGACY = [
-    send_robotupload(mode="insert", priority_config_key='LEGACY_ROBOTUPLOAD_PRIORITY_AUTHOR'),
```
michamos, line 56: why not simply hardcode `replace` here instead?

pazembrz: No. It's ugly and it's simplifying nothing.

michamos: It prevents bugs. There's no need to make it metadata dependent, as it can be inferred by where it's called in the workflow: we know that `replace` should work because it already has a control number. If there is no control number at that stage, it will be rejected and we'll notice there's a bug.

pazembrz: :/ ok... Ugly but ok.

michamos (approved):

PR #3985, opened 2020-11-05
utils: addition of get records from hep

drjova:


inspirehep/config.py
```diff
@@ -82,7 +85,7 @@
 
 # Database
 # ========
-SQLALCHEMY_DATABASE_URI = "postgresql+psycopg2://inspirehep:dbpass123@localhost:5432/inspirehep"
+SQLALCHEMY_DATABASE_URI = "postgresql+psycopg2://inspirehep:inspirehep@localhost:5432/inspirehep"
```
pazembrz, line 88: I think that you should also change password in services.yml for db service if you want to change it correctly.

inspirehep/modules/workflows/tasks/manual_merging.py
```diff
@@ -143,45 +145,24 @@ def _merge_roots(new_uuid, head_roots, update_roots):
...
     head_control_number = obj.extra_data['head_control_number']
     update_control_number = obj.extra_data['update_control_number']
-
-    head = get_db_record('lit', head_control_number)
-    update = get_db_record('lit', update_control_number)
-
-    # 1. Updates the head so that it contains the result of the merge.
-    head.clear()
+    head = {}
     head.update(obj.data)
```
pazembrz, line 164:
```suggestion
    head = dict(obj.data)
```

inspirehep/config.py
drjova: my bad!

drjova:
@pazembrz updated

pazembrz (dismissed):
Great Job! :+1:

PR #3987, opened 2020-11-11
workflows: article add additional step to add to hidden collection 

pazembrz:
* Adds hidden collection if the affiliations are matching "IN2P3",
  "CERN" or "Fermilab".

* inspirehep/inspirehep#1515


tests/integration/workflows/conftest.py
```diff
@@ -224,8 +227,21 @@ def mocked_external_services(workflow_app):
             status_code=200,
             text='Irrelevant part 1 of message \nIrrelevant part 2 of message \n# Ticket 1 updated.'
         )
-
-        yield
+        if 'INSPIREHEP_URL' in workflow_app.config:
```
drjova, line 230: why not `vcr`? :D

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -703,3 +703,29 @@ def increase_restart_count_or_error(obj, eng):
         save_workflow(obj, eng)
     else:
         raise WorkflowsError('Workflow restarted too many times')
+
+
+@with_debug_logging
+def find_all_interesting_affiliations(obj):
```
drjova, line 709:
```suggestion
def affiliations_for_hidden_collections(obj):
```

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -703,3 +703,29 @@ def increase_restart_count_or_error(obj, eng):
...
+    affiliations_mapping = current_app.config.get("AFFILIATIONS_TO_HIDDEN_COLLECTIONS_MAPPING", {})
+    affiliations = {affiliation for affiliations_list in get_value(obj.data, 'authors.raw_affiliations.value') for
+                    affiliation_raw in affiliations_list for affiliation in affiliation_raw.split()}
```
drjova, line 712: can we do something simpler? Like intercept?

drjova:
@pazembrz also we could change the commit message to something like:
```
workflows: article add additional step to add to hidden collection

* Adds hidden collection if the affiliations are matching [.....]
```

tests/integration/workflows/conftest.py
pazembrz: This is much simpler than creating a record to external service (which on my local setup I cannot even run together with next). Additionally it's not necessary to have proper response from HEP here, as we only want to check if we send correct stuff.

inspirehep/modules/workflows/tasks/actions.py
pazembrz: WDYM?

inspirehep/modules/workflows/tasks/actions.py
pazembrz: :+1:

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -703,3 +703,39 @@ def increase_restart_count_or_error(obj, eng):
...
+def affiliations_for_hidden_collections(obj):
+    affiliations_mapping = current_app.config.get("AFFILIATIONS_TO_HIDDEN_COLLECTIONS_MAPPING", {})
+    affiliations = flatten_list(get_value(obj.data, 'authors.raw_affiliations.value'))
+
+    affiliations_set = {}
+    for aff in affiliations:
+        affiliations_set.update(aff.split())
+    aff_keys = set(affiliations_mapping.keys())
+    spotted_affiliations = aff_keys.intersection(affiliations_set)
+    return [affiliations_mapping[affiliation] for affiliation in spotted_affiliations]
```
drjova, line 726: shouldn't be the check case insensitive?

pazembrz: I have no idea actually. There was nothing about this in the task actually. @michamos  ?

michamos: yes, better make the lookup case insensitive.

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -310,26 +310,35 @@
             is_marked('auto-approved'),
             mark('approved', True),
             [
-                IF_NOT(
+                IF_ELSE(
                     is_record_relevant,
```
michamos, line 314: this is only one case. There's also the case where the record gets manually rejected by a curator (where it should be added to the private collection instead and go through all the same workflow steps).

pazembrz: Yeah, you are right. Fixed now and I added additional test to check this.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -703,3 +703,39 @@ def increase_restart_count_or_error(obj, eng):
...
+
+
+@with_debug_logging
+def should_be_hidden(obj, eng):
+    """Checks if paper attached to this workflow is affiliated with one of affiliations which interests us."""
+    if affiliations_for_hidden_collections(obj):
+        return True
+    return False
```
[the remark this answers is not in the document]
pazembrz: :+1:

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -703,3 +703,39 @@ def increase_restart_count_or_error(obj, eng):
...
+        return [
+            element for innerList in input_list for element in flatten_list(innerList)
+        ]
+    return [input_list]
+
+
+@with_debug_logging
+def affiliations_for_hidden_collections(obj):
+    affiliations_mapping = current_app.config.get("AFFILIATIONS_TO_HIDDEN_COLLECTIONS_MAPPING", {})
+    affiliations = flatten_list(get_value(obj.data, 'authors.raw_affiliations.value'))
```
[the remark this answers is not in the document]
pazembrz: :+1:

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -703,3 +703,37 @@ def increase_restart_count_or_error(obj, eng):
...
+
+
+@with_debug_logging
+def affiliations_for_hidden_collections(obj):
+    affiliations_mapping = current_app.config.get("AFFILIATIONS_TO_HIDDEN_COLLECTIONS_MAPPING", {})
+    affiliations = flatten_list(get_value(obj.data, 'authors.raw_affiliations.value', []))
+
+    affiliations_set = set()
+    for aff in affiliations:
+        affiliations_set.update(aff.upper().split())
```
michamos, line 723: could you use a case-insensitive regex on word boundaries here instead? so things like `CERN,` work too.

pazembrz: Why? I'm not sure if I understood you correctly as things like `CERN` works. IMHO regex here is unnecessary. I [updated test](https://github.com/inspirehep/inspire-next/pull/3987/files#diff-90a38869ac7ba833b44a3a7b5dfdc754fd7ad18c91f029c882e074150d171493R1454) for this case to be sure.

michamos: I didn't write `CERN` but `CERN,` (with a trailing comma).

pazembrz: Ah. You are right. :+1:  Fixed now.

tests/unit/search/test_search_queries.py
```diff
@@ -72,6 +73,7 @@ def test_select_source_function_in_author_search():
     assert search_source is None
 
 
+@pytest.mark.xfail(reason="invenio-query-parser updated and not compatible anymore.")
```
drjova, line 76: we could also delete the whole file

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -1307,3 +1308,164 @@ def test_update_record_goes_through_api_version_of_store_record_connection_timeo
...
+    mocked_arxiv_download,
+    workflow_app,
+    mocked_external_services,
+):
+    """Test a full harvesting workflow."""
+    record = generate_record()
+    workflow_id = build_workflow(record).id
+    with patch.dict(workflow_app.config, {
+        'FEATURE_FLAG_ENABLE_REST_RECORD_MANAGEMENT': True,
+        'INSPIREHEP_URL': "http://web:8000"
```
drjova, line 1402: shouldn't this be part of the conftest config?

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -703,3 +704,37 @@ def increase_restart_count_or_error(obj, eng):
...
+def should_be_hidden(obj, eng):
+    """Checks if paper attached to this workflow is affiliated with one of affiliations which interests us."""
+    return bool(affiliations_for_hidden_collections(obj))
+
+
+@with_debug_logging
+def replace_collection_to_hidden(obj, eng):
+    """Replaces collection to hidden based on authors affiliations"""
+    obj.data["_collections"] = affiliations_for_hidden_collections(obj)
+    return obj
```
drjova, line 740: shouldn't we also save?

drjova (approved):
LGTM 🥇 Just few comments!

tests/unit/search/test_search_queries.py
pazembrz: We are still using some ES queries in workflows as I recall. We should clean this first.

tests/integration/workflows/test_arxiv_workflow.py
pazembrz: We don't want to set this for all tests we have in there...

inspirehep/modules/workflows/tasks/actions.py
pazembrz: No. It's not action who should save a workflow. We have separate steps on every workflow which saves state of it.

tests/integration/workflows/conftest.py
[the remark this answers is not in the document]
pazembrz: Not really. What with the tests which are testing lack of this key? And not all tests probably should have this key also.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -703,3 +704,37 @@ def increase_restart_count_or_error(obj, eng):
...
+        ]
+    return [input_list]
+
+
+@with_debug_logging
+def affiliations_for_hidden_collections(obj):
+    affiliations_mapping = current_app.config.get("AFFILIATIONS_TO_HIDDEN_COLLECTIONS_MAPPING", {})
+    affiliations = flatten_list(get_value(obj.data, 'authors.raw_affiliations.value', []))
+
+    query = "|".join(affiliations_mapping.keys())
```
michamos, line 722: I said word boundary for a specific reason, we don't want to match on things like `concern`
```suggestion
    query = "|".join(f"\b{key}\b" for key in affiliations_mapping.keys())
```

pazembrz: Argh... Sorry, this comment appeared on the moment when I merged... I will create a fix.

PR #3991, opened 2020-11-19
workflows: create ticket when there is a conflict on wf

pazembrz:
inspirehep/inspirehep#1513

### DIFF
--- a/inspirehep/modules/workflows/tasks/merging.py
+++ b/inspirehep/modules/workflows/tasks/merging.py
@@ -27,9 +27,11 @@ from datetime import datetime
 
 from copy import deepcopy
 from flask import current_app
+from idutils import is_arxiv_post_2007
 
 from inspire_json_merger.api import merge
 from inspire_schemas.readers import LiteratureReader
+from inspire_utils.record import get_value
 from invenio_pidstore.models import PersistentIdentifier
 
 from inspirehep.modules.records.api import InspireRecord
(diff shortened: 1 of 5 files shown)


inspirehep/modules/workflows/tasks/merging.py
```diff
@@ -95,3 +95,14 @@ def merge_articles(obj, eng):
         obj.extra_data['callback_url'] = \
             get_resolve_merge_conflicts_callback_url()
     obj.save()
+
+
+def conflicts_ticket_context(user, obj):
+    server_name = current_app.config['SERVER_NAME']
+    workflow_id = obj.id
+    subject = "Merge conflict on workflow {} needs to be resolved.".format(workflow_id)
```
michamos, line 103: could you reuse the logic from the core curation tickets which display arXiv ID, DOI, etc.?

pazembrz: I would prefer not to reuse whole context from there, but I can update this one with more data from the workflow to display arXiv, DOI , ID etc.

pazembrz: Is this solution better? @michamos

inspirehep/modules/workflows/tasks/merging.py
```diff
@@ -95,3 +97,33 @@ def merge_articles(obj, eng):
...
+        "doi:{0}".format(doi)
+        for doi in get_value(obj.data, 'dois.value') or []
+    ]
+    recid = obj.extra_data.get('recid')
+    subject = ' '.join(filter(
+        lambda x: x is not None,
+        arxiv_ids + dois + report_numbers + ['(#{0})'.format(recid)]
+    ))
+
+    subject = "Merge conflict on workflow {} needs to be resolved. {}".format(workflow_id, subject)
```
michamos, line 120: leftover

inspirehep/modules/workflows/templates/workflows/tickets/conflicts-ticket-template.html
```diff
@@ -0,0 +1,37 @@
...
+arXiv IDS: {{ arxiv_ids|safe }}
+{% endif %}
+
+{% if dois %}
+DOIs: {{ dois|safe }}
+{% endif %}
+
+{% if report_numbers %}
+Report numbers: {{ report_numbers|safe }}
+{% endif %}
```
michamos, line 37: I don't think we need all of this, it's enough to put the data in the subject.

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -298,10 +298,19 @@
             merge_articles,
             IF(
                 has_conflicts,
-                halt_record(
-                    action='merge_approval',
-                    message='Submission halted for merging conflicts.'
-                ),
+                [
+                    create_ticket(
```
michamos, line 302: what happens if the workflow is restarted? don't you need to do some gymnastics to avoid creating an extra ticket?

inspirehep/modules/workflows/tasks/merging.py
pazembrz: Not really leftover. It's merging subject generated above with this text. If you prefer I can leave only thing generated above.

inspirehep/modules/workflows/templates/workflows/tickets/conflicts-ticket-template.html
pazembrz: Ah, ok then :+1:

inspirehep/modules/workflows/workflows/article.py
pazembrz: If the workflow will be restarted then I assume we should create another ticket as conflict will appear again, and probably old one will be closed already. Chance for restarting when it's halted waiting for curation is quite small, and workflows do not allow easily to generate a ticket if previous one is closed and do nothing if previous one is still opened :/

michamos (approved):

PR #3993, opened 2020-11-26
workflows: in author wf send record to legacy as an update instead to legacy holdingpen

pazembrz:
* #1540

### DIFF
--- a/inspirehep/utils/url.py
+++ b/inspirehep/utils/url.py
@@ -101,3 +101,14 @@ def get_legacy_url_for_recid(recid):
     """
     pattern = current_app.config['LEGACY_RECORD_URL_PATTERN']
     return record_url_by_pattern(pattern, recid)
+
+
+def get_hep_url_for_recid(recid, endpoint):
+    server_name = current_app.config.get("SERVER_NAME")
+    protocol = current_app.config.get("PREFERRED_URL_SCHEME", 'http')
+    return "{protocol}://{server_name}/{endpoint}/{control_number}".format(
+        protocol=protocol,
+        server_name=server_name,
+        endpoint=endpoint,
+        control_number=recid
+    )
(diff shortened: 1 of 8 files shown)


inspirehep/utils/url.py
```diff
@@ -101,3 +101,8 @@ def get_legacy_url_for_recid(recid):
     """
     pattern = current_app.config['LEGACY_RECORD_URL_PATTERN']
     return record_url_by_pattern(pattern, recid)
+
+
+def get_hep_url_for_recid(recid, endpoint):
+    inspirehep_url = current_app.config.get("INSPIREHEP_PUBLIC_URL")
```
drjova, line 107: isn't it `INSPIREHEP_URL` enough?

pazembrz: `INSPIREHEP_URL` is actually internal url (no domain, but ip address and it points to /api endpoint)

drjova: oh I see, we could use `SERVER_NAME` then :)

pazembrz: Yeah, but this is only domain without protocol. I can use it if you prefer but I will still have to build proper url.

pazembrz: Additionally on tests it's `localhost:5000` so it will make tests less readable as it will point to `next` instance.

drjova:
```python
In [4]: app.config['PREFERRED_URL_SCHEME']
Out[4]: 'https'

In [5]: app.config['SERVER_NAME']
Out[5]: 'inspirehep.net'
```
We can combine them together instead of introducing yet another config :)

pazembrz: Yes I know, this is what we are already doing on hep. Still on next as I said SERVER_NAME in tests is "localhost:5000" which means it's pointing to next instance not to hep instance. And logically this address should not point to NEXT but to HEP

drjova: hmm what is the problem on tests if it's `localhost:5000`? You could even patch it or we just check if the url is correct with whatever is in `SERVER_NAME`

pazembrz: Yeah, I also know this but this just feels wrong, as logically "SERVER_NAME" is server name for NEXT not fo HEP and the URL we want to generate in there is for HEP.

pazembrz: Ok, I changed it but we have to remember that if we change SERVER_NAME it will change in curator tickets also.

drjova (approved):

drjova (approved):

PR #1553, opened 2016-09-09
WIP dojson: fix a bug in conference series

tomaszgy:
* Stop dojson rule from producing conference series number
without a name.

* Bug found in #1547.

Signed-off-by: Tomasz Gargas <tomaszgy@gmail.com>

PR #3996, opened 2020-12-01
workflows:fix refextract to ignore references without authors

pazembrz:
* #1553


inspirehep/utils/references.py
```diff
@@ -95,6 +95,9 @@ def map_refextract_to_schema(extracted_references, source=None):
     result = []
 
     for reference in extracted_references:
+        # Skip references without author
+        if not reference.get('author'):
+            continue
```
drjova, line 100: I'm sorry if I missed the conversation, but is it really what we want to achieve? Aren't we loosing references?

michamos: no, this is absolutely not what we want. We don't care about authors in references. If there are not authors in references, we should just skip adding authors (can probably be fixed in `ReferenceBuilder.add_refextract_authors_str`).

PR #4004, opened 2021-01-13
validation: endpoint changed to POST.

oguzdemirbasci:
Ref: https://github.com/inspirehep/inspirehep/issues/1474

### DIFF
--- a/inspirehep/modules/editor/api.py
+++ b/inspirehep/modules/editor/api.py
@@ -254,21 +254,23 @@ def get_workflow_and_schema(workflow_id):
     })
 
 
-@blueprint_api.route('/validate_workflow/<int:workflow_id>', methods=['GET'])
-@editor_use_api_permission.require(http_exception=403)
-def validate_workflow(workflow_id):
-    try:
-        workflow = workflow_object_class.get(workflow_id)
-    except WorkflowsMissingObject:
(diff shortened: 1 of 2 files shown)


inspirehep/modules/editor/api.py
```diff
@@ -254,21 +254,23 @@ def get_workflow_and_schema(workflow_id):
...
+
+        workflow_db.extra_data['validation_errors'] = errors
+        workflow_db.save()
         db.session.commit()
 
-        return jsonify(workflow.extra_data), 400
+        return jsonify(errors), 400
 
-    return jsonify(workflow.extra_data)
+    return jsonify({'result': 'success'})
```
drjova, line 276: Why do we need success? isn't the status code enough? Is it expected by record-editor?

oguzdemirbasci: Client in record-editor throws json conversion error when only status code returned.

drjova: Is the editor expects `result`?

oguzdemirbasci: No, it should be a valid JSON format. So I just write it like this. It throws an error when `{'success'}` returned.

drjova: `"Success"` is a valid json

drjova (approved):

PR #4008, opened 2021-01-21
rt: close ticket with proper owner

pazembrz:


inspirehep/utils/tickets.py
```diff
@@ -214,34 +221,75 @@ def get_queues():
...
+
+    Args:
+        email(str): Email of user to look for.
+
+    Returns:
+        dict: dictionary of the user {"id":"user/<id>", "Name":"<name>", "EmailAddress": "<e-mail>"}.
+
+    """
+    def _parse_users_with_emails(lines):
+        data = zip(lines[0::3], lines[1::3], lines[2::3])
```
michamos, line 246: nice API :grimacing:

inspirehep/utils/tickets.py
```diff
@@ -214,34 +221,75 @@ def get_queues():
...
+
+    def _get_rt_users():
+        _lines = query_rt('user', fields="Name,EmailAddress")
+        _users_map = _parse_users_with_emails(_lines)
+        if not _users_map:
+            raise EmptyResponseFromRT
+        return _users_map
+
+    users_map = _get_from_cache("rt_users_with_emails", _get_rt_users, force_update)
+    result = filter(lambda user: user['EmailAddress'] == email, users_map)
```
michamos, line 261:
```suggestion
    result = [user for user in users_map if user['EmailAddress'] == email]
```
is more readable

inspirehep/utils/tickets.py
```diff
@@ -214,34 +221,75 @@ def get_queues():
...
+        data = zip(lines[0::3], lines[1::3], lines[2::3])
+        users = []
+        for user_data in data:
+            if user_data and all(user_data):
+                users.append(dict([map(unicode.strip, element.split(':', 1)) for element in user_data]))
+        return users
+
+    def _get_rt_users():
+        _lines = query_rt('user', fields="Name,EmailAddress")
+        _users_map = _parse_users_with_emails(_lines)
```
michamos, line 255: misnomer, this is not a mapping, why not simply `users_data` (what's with the `_` BTW)?

inspirehep/utils/tickets.py
```diff
@@ -214,34 +221,75 @@ def get_queues():
...
+    def _parse_users_with_emails(lines):
+        data = zip(lines[0::3], lines[1::3], lines[2::3])
+        users = []
+        for user_data in data:
+            if user_data and all(user_data):
+                users.append(dict([map(unicode.strip, element.split(':', 1)) for element in user_data]))
+        return users
+
+    def _get_rt_users():
+        _lines = query_rt('user', fields="Name,EmailAddress")
```
michamos, line 254: just to make sure: does this list contain only priviledged users, or also all users who ever sent an email to RT? the latter is huge.

inspirehep/utils/tickets.py
pazembrz: You cannot do much more with this beautiful response from RT :P

inspirehep/utils/tickets.py
pazembrz: I prefer filter IMHO but ok :+1:

inspirehep/utils/tickets.py
pazembrz: :+1:  
I added `_`  as my IDE is complaining that those are shadowing vars from outer scope. It's not breaking anything but I do not like to add more "warnings" to this beautiful code on next ;)

inspirehep/utils/tickets.py
pazembrz: When I ran this query it only returned a dozen or so results. So as I assume it returns only privileged users, but it's difficult to say as /search/user looks to be undocumented :/ https://rt-wiki.bestpractical.com/wiki/REST

PR #4011, opened 2021-01-28
Do not stop WF before all tickets are closed

pazembrz:


inspirehep/modules/workflows/workflows/article.py
```diff
@@ -349,12 +349,10 @@
                         [
                             reject_record('Article automatically rejected'),
                             mark('approved', False),
-                            save_workflow,
-                            stop_processing,
                         ],
-                    )
-                )
-            ]
+                    ),
```
drjova, line 353: we don't have to `save_workflow`?

pazembrz: we can but we don't have to as wf is not stopped. It's saved literally 3 lines below and this executes for all paths.

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -349,12 +349,10 @@
                         [
                             reject_record('Article automatically rejected'),
```
michamos, line 350: this is incorrect, you reach this also if it's manually rejected and not for a hidden collection.

pazembrz: Do we need an exact reason here or can I just put "Article rejected"?

michamos: no, that's fine

michamos (approved):

PR #4016, opened 2021-02-16
workflow: Normalize journal_title in references

pazembrz:
* inspirehep/inspirehep#1673

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -555,7 +555,7 @@ def normalize_journal_titles(obj, eng):
     """Normalize the journal titles
 
     Normalize the journal titles stored in the `journal_title` field of each object
-    contained in `publication_info`.
+    contained in `publication_info` and for each `publication_info.journal_title` in references.
 
     Note:
         The DB is queried in order to get the `$ref` of each journal and add it in
@@ -571,22 +571,34 @@ def normalize_journal_titles(obj, eng):
     Returns:
        None
     """
-    publications = obj.data.get('publication_info')
+    publications = obj.data.get('publication_info', [])
 
-    if not publications:
-        return None
+    for publication in publications:
+        normalize_journal_title_entry(publication)
 
-    for index, publication in enumerate(publications):
-        if 'journal_title' in publication:
-            normalized_title = normalize_journal_title(publication['journal_title'])
-            obj.data['publication_info'][index]['journal_title'] = normalized_title
-            ref_query = RecordMetadata.query.filter(
-                RecordMetadata.json['_collections'].op('?')('Journals')).filter(
-                cast(RecordMetadata.json['short_title'], String) == type_coerce(normalized_title, JSON))
-            result = db.session.execute(ref_query).fetchone()
+    references = obj.data.get("references", [])
+    for reference in references:
+        publication_info = get_value(reference, 'reference.publication_info')
+        if not publication_info:
+            continue
+        normalize_journal_title_entry(publication_info)
 
-            if result:
-                obj.data['publication_info'][index]['journal_record'] = result.records_metadata_json['self']
+
+def normalize_journal_title_entry(publication_info):
+    if 'journal_title' not in publication_info:
+        return
+
+    normalized_title = normalize_journal_title(publication_info['journal_title'])
+    publication_info['journal_title'] = normalized_title
+
+    ref_query = RecordMetadata.query.filter(
+        RecordMetadata.json['_collections'].op('?')('Journals')).filter(
+        cast(RecordMetadata.json['short_title'], String) == type_coerce(normalized_title, JSON))
+    result = db.session.execute(ref_query).fetchone()
+    journal_record = result.records_metadata_json['self'] if result else None
+
+    if journal_record:
+        publication_info['journal_record'] = journal_record
 
 
 @with_debug_logging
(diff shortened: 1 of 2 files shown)


oguzdemirbasci (approved):

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -571,22 +571,40 @@ def normalize_journal_titles(obj, eng):
...
+
+def normalize_journal_title_entry(publication_info):
+    if 'journal_title' not in publication_info:
+        return None, None
+
+    normalized_title = normalize_journal_title(publication_info['journal_title'])
+    ref_query = RecordMetadata.query.filter(
+        RecordMetadata.json['_collections'].op('?')('Journals')).filter(
+        cast(RecordMetadata.json['short_title'], String) == type_coerce(normalized_title, JSON))
+    result = db.session.execute(ref_query).fetchone()
```
drjova, line 603: could we do an http request on hep for that?

pazembrz: This is different from task requirements and refinement. It will require new endpoint on HEP and task was estimated only for 3 points. If you want to do this kind of changes it should go as new task.

drjova: I'm sorry the goal is not to use the db directly on next as much as possible, and this adds an extra dependency. Isn't it just a search request? ```q=_collections:Journals and short_title: bla```

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -571,22 +571,40 @@ def normalize_journal_titles(obj, eng):
...
+            obj.data['publication_info'][index]['journal_record'] = journal_record
+
+    references = obj.data.get("references", [])
+    for index, reference in enumerate(references):
+        publication_info = get_value(reference, 'reference.publication_info')
+        if not publication_info:
+            continue
+        normalized_title, journal_record = normalize_journal_title_entry(publication_info)
+        if normalized_title:
+            obj.data['references'][index]['reference']['publication_info']['journal_title'] = normalized_title
```
michamos, line 590:
```suggestion
            reference['reference']['publication_info']['journal_title'] = normalized_title
```
and same thing on L592, so you don't need to `enumerate`.

inspirehep/modules/workflows/tasks/actions.py
pazembrz: I know that goal is to not create new db queries on next but this code is not new code. It's moved from another function only.

inspirehep/modules/workflows/tasks/actions.py
pazembrz: :+1:

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -571,22 +571,34 @@ def normalize_journal_titles(obj, eng):
...
+    for reference in references:
+        publication_info = get_value(reference, 'reference.publication_info')
+        if not publication_info:
+            continue
+        normalize_journal_title_entry(publication_info)
+
+
+def normalize_journal_title_entry(publication_info):
+    if 'journal_title' not in publication_info:
+        return None, None
```
michamos, line 589: why are you returning this? the main branch doesn't explicitly return anything (it mutates the input)
```suggestion
        return
```

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -571,22 +571,34 @@ def normalize_journal_titles(obj, eng):
...
+
+    for publication in publications:
+        normalize_journal_title_entry(publication)
+
+    references = obj.data.get("references", [])
+    for reference in references:
+        publication_info = get_value(reference, 'reference.publication_info')
+        if not publication_info:
+            continue
+        normalize_journal_title_entry(publication_info)
```
michamos, line 584: Not sure it's worth it, but it would be better to build a map `journal_title`->`(normalized_title, normalized_record)` to avoid having to lookup the same journal title multiple times. Even better would be to allow normalizing all titles with one query only, which is much more efficient.

inspirehep/modules/workflows/tasks/actions.py
michamos: The reason for this is that indexing is async, so having the contents of the DB depend on stuff in ElasticSearch is inconsistent in principle. In practice it probably doesn't matter here, so I would be ok with doing the query on ES only and getting the record ID from there.

pazembrz: So maybe we could create a task for this? We agreed not to change requirements when task is during review... :/

inspirehep/modules/workflows/tasks/actions.py
pazembrz: You are right. Leftover from time when this was returning modified title and record.

inspirehep/modules/workflows/tasks/actions.py
pazembrz: Idea is nice but I'm not sure how much we will benefit form it here, as we want in nearest future move workflows from NEXT so this will be rewritten as it will probably won't have access to DB at all.

michamos: Even on ES it's inefficient to do multiple queries where one would suffice (`normalize_journal_title` is doing an ES query under the hood).

pazembrz: Maybe we could create a new task to do this on ES in one query and not on db like now? This (maybe) will be useful for new implementation of workflows also.

michamos: ok, can you create a task then?

michamos (approved):

inspirehep/modules/workflows/tasks/actions.py
pazembrz: https://github.com/inspirehep/inspirehep/issues/1723

Issue #1782, opened 2016-12-01
workflows: ids of the new workflows start from 1


david-caro:
It might be that the migration of the old records did not work properly.

kaplun:
http://stackoverflow.com/questions/787722/postgresql-autoincrement

jmartinm:
And http://stackoverflow.com/a/244265/890185

jmartinm:
I can confirm that running:

```
SELECT setval('workflows_object_id_seq', (SELECT MAX(id) FROM workflows_object));
```

fixes the problem, will add to the deployment instructions.

PR #4024, opened 2021-03-30
workflow: fix and improve grobid authors extraction

pazembrz:
* Ignore authors without surname
* Replace authors in metadata if there are no authors originally and there are authors extracted from grobid

#1782

### DIFF
--- a/inspirehep/modules/workflows/utils/grobid_authors_parser.py
+++ b/inspirehep/modules/workflows/utils/grobid_authors_parser.py
@@ -32,7 +32,7 @@ class GrobidAuthors(object):
             xml_text = xml_text.decode('utf-8')
         self._xml = Selector(text=xml_text, type="xml")
         self._xml.remove_namespaces()
-        self._parsed_authors = self._xml.xpath("//author")
+        self._parsed_authors = self._xml.xpath("//author[persName/surname[string-length(normalize-space()) > 0]]")
         self._builder = None
 
     def __getitem__(self, item):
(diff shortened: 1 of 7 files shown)


inspirehep/modules/workflows/utils/grobid_authors_parser.py
```diff
@@ -66,9 +66,9 @@ def __init__(self, author_selector):
     @staticmethod
     def _extract(source, path, type=None, text=False):
         if type:
-            path += "[@type='{type}']".format(type=type)
+            path += "[@type='{type}' and string-length(normalize-space()) > 0]".format(type=type)
         if text:
-            path += "/text()"
+            path += "[string-length(normalize-space()) > 0]/text()"
```
michamos, line 71: won't this fail now if you provide both `type` and `text`? I'm not sure you can do `node[selector1][selector2]`.

pazembrz: Actually It's working: 
```python 
> /home/pazembrz/dev/inspire-next/inspirehep/modules/workflows/utils/grobid_authors_parser.py(72)_extract()
     71             path += "[string-length(normalize-space()) > 0]/text()"
---> 72             return source.xpath(path)
     73         return source.xpath(path)

ipdb> path
"orgName[@type='institution' and string-length(normalize-space()) > 0][string-length(normalize-space()) > 0]/text()"
ipdb> source.xpath(path)
[<Selector xpath="orgName[@type='institution' and string-length(normalize-space()) > 0][string-length(normalize-space()) > 0]/text()" data=u'Bose National Centre for Basic Sciences'>, <Selector xpath="orgName[@type='institution' and string-length(normalize-space()) > 0][string-length(normalize-space()) > 0]/text()" data=u'JD Block'>]
```
https://www.w3.org/TR/1999/REC-xpath-19991116/#location-paths It's recognized as concatenation of those two selectors. But I have to agree that this is quite ugly as in both selectors one part is identical. I will change it a little.

michamos (approved):

PR #4029, opened 2021-04-13
workflow: hep send record first to legacy

drjova:


inspirehep/modules/workflows/workflows/article.py
```diff
@@ -578,8 +578,8 @@ class Article(object):
                 is_record_accepted,
                 (
                     POSTENHANCE_RECORD +
-                    STORE_RECORD +
                     SEND_TO_LEGACY +
+                    STORE_RECORD +
```
michamos, line 582: Why? I remember there being lots of back and forth about this, there were some issues I don't quite remember right know in doing things in a different order.

drjova: Because of that https://inspirehep.zulipchat.com/#narrow/stream/122199-ops/topic/bugninja/near/234261081

michamos: I think you'd need to add explicit validation before then. Currently this happens in `STORE_RECORD`, but if you send it to legacy first you have the risk of the record being invalid, but already being created on legacy. But then we need to make sure the version of `inspire-schemas` in `next` is 100% in sync with the one in `hep`, or have a validation API on `hep`.

michamos: Also something I'm thinking off right now: this will prevent harvest on QA, as the legacy test instance is not processing bibsched.

PR #4032, opened 2021-04-20
workflows: put classifier keywords into core records

pazembrz:
inspirehep/inspirehep#1677

### DIFF
--- a/inspirehep/modules/workflows/tasks/classifier.py
+++ b/inspirehep/modules/workflows/tasks/classifier.py
@@ -76,33 +76,48 @@ def classify_paper(taxonomy=None, rebuild_cache=False, no_cache=False,
             extract_acronyms=extract_acronyms
         )
 
-        fulltext_used = True
+        try:
+            data = get_value(obj.data, 'titles.title', [])
+            data.extend(get_value(obj.data, 'titles.subtitle', []))
+            data.extend(get_value(obj.data, 'abstracts.value', []))
+            data.extend(get_value(obj.data, 'keywords.value', []))
+            if not data:
+                obj.log.error("No classification done due to missing data.")
+                data_result = None
+            else:
+                data_result = get_keywords_from_text(data, **params)
+                if data_result:
+                    data_result['complete_output'] = clean_instances_from_data(
+                        data_result.get("complete_output", {})
+                    )
+        except ClassifierException as e:
+            obj.log.exception(e)
+            data_result = None
+
         with get_document_in_workflow(obj) as tmp_document:
             try:
                 if tmp_document:
                     result = get_keywords_from_local_file(tmp_document, **params)
+                    fulltext_used = True
+                    if result:
+                        result['complete_output'] = clean_instances_from_data(
+                            result.get("complete_output", {})
+                        )
                 else:
-                    data = get_value(obj.data, 'titles.title', [])
-                    data.extend(get_value(obj.data, 'titles.subtitle', []))
-                    data.extend(get_value(obj.data, 'abstracts.value', []))
-                    data.extend(get_value(obj.data, 'keywords.value', []))
-                    if not data:
-                        obj.log.error("No classification done due to missing data.")
-                        return
-                    result = get_keywords_from_text(data, **params)
+                    result = data_result
                     fulltext_used = False
             except ClassifierException as e:
                 obj.log.exception(e)
-                return
-
-        result['complete_output'] = clean_instances_from_data(
-            result.get("complete_output", {})
-        )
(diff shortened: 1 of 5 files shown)


inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -330,21 +332,28 @@ def filter_keywords(obj, eng):
...
+    prediction_keywords = get_value(obj.extra_data, 'keywords_prediction.keywords', [])
     keywords = obj.data.get('keywords', [])
-    for keyword in prediction.get('keywords', []):
+    for keyword in prediction_keywords:
         # TODO: differentiate between curated and guessed keywords
         keywords.append(
             {
                 'value': keyword['label'],
                 'source': 'curator' if keyword.get('curated') else 'magpie',
             }
```
michamos, line 343: this is not actually ending up in the record, so either this task doesn't get called or there's some removal at a later stage.
In any case, we don't want these predicted keywords to end up in the record, but only those in this issue, and you should make sure they are really part of the record at the end.

inspirehep/modules/workflows/tasks/classifier.py
```diff
@@ -76,33 +76,43 @@ def _classify_paper(obj, eng):
...
-        )
-        result["fulltext_used"] = fulltext_used
-
-        # Check if it is not empty output before adding
-        if any(result.get("complete_output", {}).values()):
-            obj.extra_data['classifier_results'] = result
+        if result:
+            result['complete_output'] = clean_instances_from_data(
+                result.get("complete_output", {})
+            )
```
michamos, line 107: This will only affect `data_result` in case `result == data_result` and L99 is reached. It should be called exactly once for `data_result` in any case.

pazembrz: :+1:

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -330,21 +332,28 @@ def filter_keywords(obj, eng):
...
-
-    obj.data['keywords'] = keywords
+    for keyword in prediction_keywords:
+        if keyword.get('curated'):
+            keywords.append(
+                {
+                    'value': keyword['label'],
+                    'source': 'curator',
+                }
+            )
```
michamos, line 344: no, we don't need this thing (related to magpie too), you can remove it.

inspirehep/modules/workflows/tasks/classifier.py
```diff
@@ -76,33 +76,48 @@ def _classify_paper(obj, eng):
...
-                    data.extend(get_value(obj.data, 'keywords.value', []))
-                    if not data:
-                        obj.log.error("No classification done due to missing data.")
-                        return
-                    result = get_keywords_from_text(data, **params)
+                    result = data_result
                     fulltext_used = False
             except ClassifierException as e:
                 obj.log.exception(e)
                 return
```
michamos, line 111: You probably shouldn't `return` here, as you still want to get the keywords from metadata into the record in this case.

michamos (approved):
small issue, but looks ok for the rest :+1:

inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -330,21 +332,19 @@ def filter_keywords(obj, eng):
...
-
-    obj.data['keywords'] = keywords
+    if obj.data.get('core'):
+        extracted_keywords = obj.extra_data.get('extracted_keywords', [])
+        for keyword in extracted_keywords:
+            if keyword:
+                keywords.append(
+                    {
+                        "value": keyword,
+                        "schema": "INSPIRE"
```
michamos, line 343: sorry, small addition: it would be nice to add `"source": "classifier"` here so we know where they come from

inspirehep/modules/workflows/tasks/classifier.py
pazembrz: :+1:

inspirehep/modules/workflows/tasks/submission.py
pazembrz: :+1:

inspirehep/modules/workflows/tasks/submission.py
pazembrz: :+1:

inspirehep/modules/workflows/tasks/submission.py
pazembrz: Discussed IRL.

michamos (approved):

PR #4041, opened 2021-05-12
collaboration linking: don't link when match is ambiguous

MJedr:
ref: inspirehep/inspirehep#1862


drjova (approved):

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -907,7 +907,8 @@ def normalize_collaborations(obj, eng):
         else:
             collaboration_normalized_name = response[0].collaboration.value
         LOGGER.info(u"(%s) collaboration normalization: normalized: %s ==> %s", obj.id, collaboration['value'], collaboration_normalized_name)
-
+        if len(response.hits) > 1:
+            continue
```
michamos, line 911: @MJedr it would have been good to add some logging here. The idea is that a curator will look at the logs and manually fix ambiguous cases. Sorry if it wasn't clear in the story.

MJedr: Ok, I'll add

PR #4043, opened 2021-05-18
workflows: affiliation disambiguation

MJedr:
ref:inspirehep/inspirehep#1866

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -72,6 +72,7 @@ from inspirehep.modules.workflows.tasks.refextract import (
     extract_references_from_text,
 )
 from inspirehep.modules.refextract.matcher import match_references
+from inspirehep.modules.search import InstitutionsSearch
 from inspirehep.modules.workflows.tasks.upload import create_error
 from inspirehep.modules.workflows.errors import BadGatewayError, CannotFindProperSubgroup
 from inspirehep.modules.workflows.utils import (
@@ -170,6 +171,11 @@ def add_core(obj, eng):
     """Mark a record as CORE if it was approved as CORE."""
     if 'core' in obj.extra_data:
         obj.data['core'] = obj.extra_data['core']
+
+
+def is_core(obj, eng):
+    """Check if workflow is CORE."""
+    return obj.data.get("core")
 
 
 def halt_record(action=None, message=None):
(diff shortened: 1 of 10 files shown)


pazembrz (approved):

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -920,3 +921,81 @@ def normalize_collaborations(obj, eng):
...
+            "nested",
+            path="authors",
+            query=(
+                Q("match", authors__raw_affiliations__value=raw_aff)
+                & Q("exists", field="authors.affiliations.value")
+            ),
+            inner_hits={},
+        )
+        & Q("term", _collections="Literature")
+        & Q("term", curated=True)
```
michamos, line 938: these should be part of the [filter context](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-filter-context.html#filter-context), they're not relevant for scoring.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -920,3 +921,81 @@ def normalize_collaborations(obj, eng):
...
+    )
+    result = (
+        Search(index="records-hep", using=current_search_client)
+        .query(query)
+        .execute()
```
michamos, line 943: you only care about the first result, so you can set `size=1`.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -920,3 +921,81 @@ def normalize_collaborations(obj, eng):
...
+        .hits
+    )
+    if result:
+        matched_author = result[0].meta.inner_hits.authors.hits[0].to_dict()
+        return matched_author["affiliations"][0]
+
+
+def _match_institution(matched_affiliation):
+    query = Q("match", legacy_ICN=matched_affiliation["value"])
+    result = InstitutionsSearch().query(query).execute()
```
michamos, line 953: also `size=1`

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -920,3 +921,81 @@ def normalize_collaborations(obj, eng):
...
+    )
+    result = (
+        Search(index="records-hep", using=current_search_client)
+        .query(query)
+        .execute()
+        .hits
+    )
+    if result:
+        matched_author = result[0].meta.inner_hits.authors.hits[0].to_dict()
+        return matched_author["affiliations"][0]
```
michamos, line 948: you want to assign *all* affiliations here, not only the first.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -920,3 +921,81 @@ def normalize_collaborations(obj, eng):
...
+):
+    if "record" not in matched_author_affiliation:
+        matched_author_affiliation = _assign_institution_ref_to_affiliation(
+            matched_author_affiliation
+        )
+    author_affiliations.append(matched_author_affiliation)
+    return matched_author_affiliation
+
+
+def normalize_affiliations(obj, eng):
```
michamos, line 976: does this ever get called? As far as I can tell, this creates the workflow step but doesn't use it anywhere.

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -199,6 +200,7 @@
     guess_keywords,
     guess_coreness,
     normalize_collaborations,
+    normalize_affiliations
```
michamos, line 203: This should only happen for core records, so it can't happen here, you need to do it in post-enhance (so you know if a record is core) and check whether the record is core.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -938,3 +939,83 @@ def normalize_collaborations(obj, eng):
...
+    for author in obj.data.get("authors", []):
+        author_affiliations = author.get("affiliations", [])
+        if author_affiliations:
+            continue
+        raw_affs = get_value(author, "raw_affiliations.value", [])
+        for raw_aff in raw_affs:
+            if raw_aff in matched_affiliations:
+                author_affiliations.append(matched_affiliations[raw_aff])
+                continue
+            matched_author_affiliation = _match_lit_author_affiliation(raw_aff)
```
michamos, line 1007:
```suggestion
            matched_author_affiliations = _match_lit_author_affiliations(raw_aff)
```
as it's a list now.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -938,3 +939,83 @@ def normalize_collaborations(obj, eng):
...
+    result = InstitutionsSearch().query(query).params(size=1).execute()
+    if result:
+        return result.hits[0]["self"]
+
+
+def _assign_institution_ref_to_affiliation(matched_affiliation):
+    institution_reference = _match_institution(matched_affiliation)
+    if institution_reference:
+        matched_affiliation["record"] = institution_reference
+    return matched_affiliation
```
michamos, line 979: you're mutating the input `dict`, so no need to `return` it.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -938,3 +939,83 @@ def normalize_collaborations(obj, eng):
...
+        ),
+        inner_hits={},
+    )
+    query_filters = Q("term", _collections="Literature") & Q("term", curated=True)
+    result = (
+        Search(index="records-hep", using=current_search_client)
+        .query(query)
+        .filter(query_filters)
+        .params(size=1)
+        .execute()
```
michamos, line 960: you can also disable the `source`, as you only care about the inner hits.

inspirehep/modules/workflows/workflows/article.py
MJedr: Ah ok, thank you!

michamos (approved):
Looks good, let's put it to the test :rocket:

PR #4045, opened 2021-05-25
collaboration linking: add logging if ambiguous match

MJedr:


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -908,6 +908,12 @@ def normalize_collaborations(obj, eng):
             collaboration_normalized_name = response[0].collaboration.value
         LOGGER.info(u"(%s) collaboration normalization: normalized: %s ==> %s", obj.id, collaboration['value'], collaboration_normalized_name)
         if len(response.hits) > 1:
+            matched_collaboration_names = [matched_collaboration.collaboration.value for matched_collaboration in response]
+            LOGGER.info(
+                "(wf: {0}) ambiguous match for collaboration {1}. Matched collaborations: {2}".format(
```
michamos, line 913: the format string needs to be unicode, not bytes.

tests/integration/workflows/fixtures/experiment_with_ambiguous_collaboration_1.json
```diff
@@ -41,10 +40,10 @@
...
       }
-    ]
-  },
+    ],
   "id": "1109131",
   "links": {
     "json": "http://localhost:5000/api/experiments/1109131?format=json"
-  }
+  },
+  "normalized_name_variants": ["SHIP"]
```
michamos, line 48: why is this changing?

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -902,12 +902,30 @@ def normalize_collaborations(obj, eng):
...
+            elif len(response.hits) > 1:
+                matched_collaboration_names = [
+                    matched_collaboration.collaboration.value for matched_collaboration in response
+                ]
+                LOGGER.info(
+                    u"(wf: {0}) ambiguous match for collaboration {1}. Matched collaborations: {2}".format(
+                        obj.id, collaboration['value'], ', '.join(matched_collaboration_names)
+                    )
+                )
+                continue
```
michamos, line 914: this is the branch where you normalize based on a subgroup name, not based on the name of the full collaboration. The flow is a bit complex, but also there it should be unique.

MJedr: So it's the wrong place or logging should list subgroup names instead of collaborations?

michamos: you should make it clear that it's about collaboration subgroups (`ambiguous match for collaboration subgroup`), and you should also add a check for several results to the non-subgroup matching.

michamos (approved):

PR #4046, opened 2021-05-25
workflows: Add core_selection workflow

pazembrz:
inspirehep/inspirehep#1690

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -53,7 +53,7 @@ from sqlalchemy import (
 from werkzeug import secure_filename
 
 from invenio_db import db
-from invenio_workflows import ObjectStatus
+from invenio_workflows import ObjectStatus, workflow_object_class, start
 from invenio_workflows.errors import WorkflowsError
 from invenio_records.models import RecordMetadata
 from inspire_json_merger.api import merge
@@ -74,7 +74,7 @@ from inspirehep.modules.workflows.tasks.refextract import (
 from inspirehep.modules.refextract.matcher import match_references
 from inspirehep.modules.search import InstitutionsSearch
 from inspirehep.modules.workflows.tasks.upload import create_error
-from inspirehep.modules.workflows.errors import BadGatewayError, CannotFindProperSubgroup
+from inspirehep.modules.workflows.errors import BadGatewayError, CannotFindProperSubgroup, MissingRecordControlNumber
 from inspirehep.modules.workflows.utils import (
     copy_file_to_workflow,
     download_file_to_workflow,
@@ -1028,3 +1028,43 @@ def link_institutions_with_affiliations(obj, eng):
             _assign_institution_reference_to_affiliations(author_affiliations, affiliations)
 
     return obj
+
+
+def core_selection_wf_already_created(control_number):
+    query_wf_with_cn = Q("match", metadata__control_number=control_number)
+    query_core_selection_wf = Q("match", _workflow__workflow_name="CORE_SELECTION")
+    query = query_wf_with_cn & query_core_selection_wf
+
+    return Search(index="holdingpen-hep", using=current_search_client).query(query).source(False).execute()
+
+
+def create_core_selection_wf(obj, eng):
+    record_control_number = obj.data.get('control_number')
+    if not record_control_number:
+        raise MissingRecordControlNumber
+    if is_core(obj, eng) or not _is_auto_approved(obj) or core_selection_wf_already_created(record_control_number):
+        LOGGER.info("No core selection needed for %s workflow with record %s", obj.id, record_control_number)
+        return obj
+
+    extra_data_to_clean = ['_task_history', '_error_msg', 'source_data', 'restart-count']
+    extra_data = dict(obj.extra_data.copy())
+    for key in extra_data_to_clean:
+        if key in extra_data:
+            del extra_data[key]
+    data = obj.data.copy()
+
+    workflow_object = workflow_object_class.create(
+        data={},
+        id_user=None,
+        data_type="hep"
+    )
+
+    workflow_object.data = data
+
+    extra_data['source_data'] = {}
+    workflow_object.extra_data = extra_data
+
+    workflow_object.save()
+    db.session.commit()
+
+    start.delay("core_selection", object_id=workflow_object.id)
--- a/inspirehep/modules/workflows/workflows/core_selection.py
+++ b/inspirehep/modules/workflows/workflows/core_selection.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+from __future__ import absolute_import, division, print_function
+
+from inspirehep.modules.literaturesuggest.tasks import curation_ticket_context
+from inspirehep.modules.records.errors import MissingInspireRecordError
+from inspirehep.modules.workflows.tasks.actions import halt_record, normalize_affiliations, \
+    link_institutions_with_affiliations
+from inspirehep.modules.workflows.tasks.submission import prepare_keywords, create_ticket
+from inspirehep.modules.workflows.tasks.upload import store_record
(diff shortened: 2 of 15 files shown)


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -938,3 +938,43 @@ def normalize_collaborations(obj, eng):
...
+    query = query_wf_with_cn & query_core_selection_wf
+
+    return Search(index="holdingpen-hep", using=current_search_client).query(query).source(False).execute()
+
+
+def create_core_selection_wf(obj, eng):
+    record_control_number = obj.data.get('control_number')
+    if not record_control_number:
+        raise MissingRecordControlNumber
+    if obj.data.get('core') or core_selection_wf_already_created(record_control_number):
```
michamos, line 955: won't this run for all records that have been manually accepted as non-core? It should only run for records that have been *automatically* harvested as non-core: https://github.com/inspirehep/inspire-next/blob/0662a4a965db24b8bd0e3b0c4ecb3523264e8d40/inspirehep/modules/workflows/tasks/actions.py#L228-L229

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -938,3 +938,43 @@ def normalize_collaborations(obj, eng):
...
+        LOGGER.info("No core selection needed for %s workflow with record %s",)
+        return obj
+
+    extra_data_to_clean = ['_task_history', '_error_msg', 'source_data', 'restart-count']
+    extra_data = dict(obj.extra_data)
```
michamos, line 960: this only makes a copy if `extra_data` is not a `dict` itself. Otherwise, you need to do `obj.extra_data.copy()`

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -938,3 +938,43 @@ def normalize_collaborations(obj, eng):
...
+    for key in extra_data_to_clean:
+        if key in extra_data:
+            del extra_data[key]
+    data = dict(obj.data)
```
michamos, line 964: same

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -938,3 +938,43 @@ def normalize_collaborations(obj, eng):
...
+
+    workflow_object = workflow_object_class.create(
+        data={},
+        id_user=None,
+        data_type="hep"
+    )
+
+    workflow_object.data = data
+
+    extra_data['source_data'] = {}
```
michamos, line 974: what happens if there's an error and the workflow gets restarted?

inspirehep/modules/workflows/workflows/core_selection.py
```diff
@@ -0,0 +1,53 @@
...
+    data_type = "hep"
+
+    workflow = (
+        halt_record(
+            action='auto_non_core_record',
+            message='Submission halted Waiting for curator to decide if record is CORE.'
+        ),
+        load_record_from_hep,
+        set_core,
+        prepare_keywords,
```
michamos, line 43: we'll need to do also the affiliation normalization from #4043.

inspirehep/modules/workflows/tasks/actions.py
pazembrz: :+1:

inspirehep/modules/workflows/tasks/actions.py
pazembrz: Record is loaded from HEP based on control_number in wf.data. Nothing is loaded from source_data as there is no step to do this as we always want the latest record in this WF. Still source_data has to be initialized as it used in do_not_repeat action to store persistent data.

inspirehep/modules/workflows/workflows/core_selection.py
pazembrz: It's not merged yet so I can't do anything about this right now.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -938,3 +938,43 @@ def normalize_collaborations(obj, eng):
...
+    query = query_wf_with_cn & query_core_selection_wf
+
+    return Search(index="holdingpen-hep", using=current_search_client).query(query).source(False).execute()
+
+
+def create_core_selection_wf(obj, eng):
+    record_control_number = obj.data.get('control_number')
+    if not record_control_number:
+        raise MissingRecordControlNumber
+    if obj.data.get('core') or core_selection_wf_already_created(record_control_number) or not _is_auto_approved():
```
michamos, line 955: small perf optimisation to avoid doing an ES query in some cases:
```suggestion
    if obj.data.get('core') or not _is_auto_approved() or core_selection_wf_already_created(record_control_number):
```

inspirehep/modules/workflows/workflows/core_selection.py
michamos: It is now :)

tests/integration/workflows/test_workflow_core_selection.py
```diff
@@ -0,0 +1,236 @@
...
+        with requests_mock.Mocker() as mock:
+            mock.register_uri('GET', mocked_url, json=load_json_record('hep_record_no_core.json'))
+            # It's update so it should be PUT
+            mock.register_uri('PUT', "http://web:8000/literature/{control_number}".format(control_number=pid_value), json={"metadata": {"control_number": pid_value}})
+
+            start("article", object_id=workflow_object.id)
+
+            assert WorkflowObjectModel.query.filter(WorkflowObjectModel.workflow.has(name="core_selection")).count() == 0
+
+            workflow_object.callback_pos = [34, 1, 13]
```
michamos, line 84: seems very brittle, isn't there a better way?

pazembrz: There is always a better way but making a WF in tests to go through whole process and finish is not easy and creating test like that will take time :/ Still it's not required for WF to really go through as we do not care for it really in this test We just want to test if it's creating another workflow at the end if record is not core.

michamos: I'm worried about maintainability of this when the workflow changes. But maybe it's too late for maintainability of workflow tests anyway :)

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -1024,3 +1024,43 @@ def normalize_affiliations(obj, eng):
...
+
+    workflow_object.data = data
+
+    extra_data['source_data'] = {}
+    workflow_object.extra_data = extra_data
+
+    workflow_object.save()
+    db.session.commit()
+
+    start.delay("core_selection", object_id=workflow_object.id)
```
drjova, line 1066: how the workflow would be part of `auto_non_core_record`?

drjova (approved):

inspirehep/modules/workflows/workflows/core_selection.py
```diff
@@ -0,0 +1,54 @@
...
+        normalize_affiliations,
+        do_not_repeat('create_ticket_curator_core_publisher')(
+            create_ticket(
+                template='literaturesuggest/tickets/curation_core.html',
+                queue='HEP_publishing',
+                context_factory=curation_ticket_context,
+                ticket_id_key='curation_ticket_id',
+            ),
+        ),
+        store_record,
```
drjova, line 53: we're missing the send to legacy step for the moment

drjova: @pazembrz 👆

inspirehep/modules/workflows/workflows/core_selection.py
michamos: a new one to add: https://github.com/inspirehep/inspire-next/blob/b2582d2cc0c3e9d24b3783f4fcc529af3d8b83b7/inspirehep/modules/workflows/workflows/article.py#L75

inspirehep/modules/workflows/workflows/core_selection.py
pazembrz: I wasn't aware that we should send to legacy from this workflow. It wasn't mentioned anywhere. I will add it.

michamos (dismissed):

drjova (approved):

PR #4049, opened 2021-06-01
affiliation linking: assign institution ref when not present

MJedr:
ref: inspirehep/inspirehp#1866

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -971,31 +971,26 @@ def _match_lit_author_affiliation(raw_aff):
         return matched_author["affiliations"]
 
 
-def _match_institution(matched_affiliation):
+def _assign_institution(matched_affiliation):
     query = Q("match", legacy_ICN=matched_affiliation["value"])
     result = InstitutionsSearch().query(query).params(size=1).execute()
     if result:
-        return result.hits[0]["self"]
+        matched_affiliation["record"] = result.hits[0]["self"]
+        return matched_affiliation
 
 
-def _assign_institution_ref_to_affiliation(matched_affiliation):
-    institution_reference = _match_institution(matched_affiliation)
-    if institution_reference:
-        matched_affiliation["record"] = institution_reference
-
-
-def _assign_matched_affiliations_to_author(
-    author_affiliations, matched_author_affiliations
-):
-    matched_affiliations_to_assign = []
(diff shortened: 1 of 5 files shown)


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -984,25 +984,20 @@ def _assign_institution_ref_to_affiliation(matched_affiliation):
...
-    author_affiliations.extend(matched_affiliations_to_assign)
-    return matched_affiliations_to_assign
 
 
 def normalize_affiliations(obj, eng):
     matched_affiliations = {}
     for author in obj.data.get("authors", []):
         author_affiliations = author.get("affiliations", [])
         if author_affiliations:
+            _assign_institution_reference_to_incomplete_affiliations(author_affiliations)
```
michamos, line 1000: this is doing one search per author, instead of one search per unique affiliation. Also you're doing it only for core records now. I think it would be cleaner to take it out into a different workflow task. If you run it after normalization, you don't need to call it here at all (on L1009 currently).

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -1011,16 +1005,26 @@ def normalize_affiliations(obj, eng):
                 continue
             matched_author_affiliations = _match_lit_author_affiliation(raw_aff)
             if matched_author_affiliations:
-                assigned_affiliations = _assign_matched_affiliations_to_author(
-                    author_affiliations, matched_author_affiliations
-                )
-                matched_affiliations[raw_aff] = assigned_affiliations
+                _assign_institution_reference_to_incomplete_affiliations(matched_author_affiliations)
```
michamos, line 1008: You don't need to do this here as you'll be doing it in the next workflow step anyway

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -1011,16 +1005,26 @@ def normalize_affiliations(obj, eng):
...
+                    author["full_name"],
+                    " ".join(raw_affs),
+                    author_affiliations
+                )
             )
     return obj
+
+
+def link_institutions_with_affiliations(obj, eng):
+    for author in obj.data.get("authors", []):
```
michamos, line 1025: There will very often be affiliations in common between the different authors. Like for affiliation normalisation, you should do it only once per affiliation instead of doing it again for every author.

michamos (approved):

PR #4052, opened 2021-06-02
workflows: add feature flag to disable send_to_legacy

MJedr:
ref:inspirehep/inspirehep#1905


inspirehep/modules/workflows/tasks/submission.py
```diff
@@ -291,6 +291,12 @@ def send_to_legacy(priority_config_key=None):
             default priority is used.
     """
     def _send_to_legacy(obj, eng):
+        if current_app.config.get('FEATURE_FLAG_DISABLE_SEND_TO_LEGACY', False):
```
drjova, line 294: I would do `ENABLE` as all of our features flags

drjova (approved):

PR #4054, opened 2021-06-03
setup: bump refextract

pazembrz:


pyproject.toml
```diff
@@ -108,6 +107,7 @@ urllib3 = "<1.25,==1.*,>=1.0.0,>=1.24.1"
 vine = "==1.*,>=1.0.0"
 workflow = "==2.*,>=2.0.0,>=2.1.3"
 
+refextract = {git = "https://github.com/inspirehep/refextract.git", branch = "maintenance/0.2.18" }
```
drjova, line 110: I would put specific `SHA`

pazembrz: Lockfile is doing this for us isn't it?

drjova: Not sure with the branch

drjova (approved):

PR #4057, opened 2021-06-08
workflows: addition of `If-Match` header

drjova:


inspirehep/modules/workflows/tasks/upload.py
```diff
@@ -107,8 +107,14 @@ def store_record_inspirehep_api(obj, eng, is_update, is_authors):
 def send_record_to_hep(obj, pid_type, control_number=None):
     try:
         if control_number:
+            head_version_id = obj.extra_data.get('head_version_id')
```
pazembrz, line 110: This is set only in Article WF and only if we are doing Merge if I'm correct. What with Article WF and core-selection?

pazembrz: edit_article is also missing this key

drjova: Are you sure, all the updates should have the `head_version_id`

drjova: we alway `merge_articles` hence it always have `head_version_id` I'm not sure if I'm missing something here

pazembrz: For article workflow you are right but for other workflows: 
On edit_article we are not running merge_articles https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/workflows/edit_article.py
On Core-selection neither https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/workflows/core_selection.py We might have `head_version_id` here but it's a copy from workflow which creates core-selection and it might be outdated)
on Authors WF we are also not running merge https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/workflows/author.py (but here I'm not sure if we use this WF for authors updates?)

drjova:
@pazembrz updated, for authors we don't care because we update directly from the form we don't have update workflow step it goes directly to legacy.

pazembrz:
Can you also update core-selection workflow with this? Or something similar?

drjova:
It has it already, as it runs after the merge_records on `article`

pazembrz:
> It has it already, as it runs after the merge_records on `article`

But if there will be any update to the record then this version will be outdated. This is why core-selection is pulling latest record from HEP, but it's not storing version as it was not required before.

inspirehep/modules/workflows/workflows/core_selection.py
```diff
@@ -41,6 +41,7 @@ class CoreSelection(object):
             message='Submission halted Waiting for curator to decide if record is CORE.'
         ),
         load_record_from_hep,
+        store_head_version,
```
pazembrz, line 44: Now the question is do we want this to be loaded like that? load_record_from_hep is loading record through REST API, maybe it will be just better to update `load_record_from_hep` to load it with the version?

drjova: Since we have it like that in many places, I would just leave it the same. We should cleanup many parts in workflows

pazembrz: Can we create task to stop reading records from DB in workflows and read them only form HEP?

drjova: I will create a taks

pazembrz (approved):

PR #4058, opened 2021-06-09
workflows: check if france in fulltext

MJedr:
ref:inspirehep/inspirehep#1854


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -1068,3 +1069,29 @@ def create_core_selection_wf(obj, eng):
...
+def get_fulltext(obj):
+    with get_document_in_workflow(obj) as tmp_document:
+        if not tmp_document:
+            return
+        with open(tmp_document, 'rb') as document_file:
+            document = document_file.read()
+        data = {'input': document}
+        grobid_url = current_app.config["GROBID_URL"]
+        api_path = "api/processFulltextDocument"
+        response = requests.post(urljoin(grobid_url, api_path), files=data)
```
drjova, line 1083: I would create a separate function

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -1068,3 +1069,29 @@ def create_core_selection_wf(obj, eng):
...
+
+    xml_data = response.text
+    xml = minidom.parseString(xml_data.encode("utf-8"))
```
drjova, line 1086: We should use `parsel` instead

drjova (dismissed):
Just the comment

drjova (approved):

Issue #1940, opened 2017-03-02
schema: HEP 210__a  title variations: RPP


annetteholtkamp:
The 210 field mostly contains synonyms, expansions of acronyms etc - which can probably be ignored in the future. But it also contains the acronym RPP which is used for the citesummary option to exclude the RPP. These records we need to tag somehow.

kaplun:
Issue moved to [inspirehep/inspire-schemas #121](https://github.com/inspirehep/inspire-schemas/issues/121) via [**ZenHub**](https://www.zenhub.com/)

PR #4062, opened 2021-06-11
workflows: fix #1944 and #1940

pazembrz:


inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -227,14 +227,16 @@ def set_wf_not_completed_ids_to_wf(obj, skip_blocked=True, skip_halted=False):
         skip_halted: boolean, if True, then it skips HALTED workflows when
             looking for matched workflows
     """
+    def _ignore_core_selection(base_record, match_result):
+        return get_value(match_result, '_source._workflow.workflow_name') != "CORE_SELECTION"
```
michamos, line 231: I would restrict this to `Article` actually. It's probably a bug that we're matching `edit_article` here.

pazembrz: :+1 I have to fix few tests now as this change is breaking them ;)

michamos (dismissed):

tests/integration/workflows/helpers/workflow_utils.py
```diff
@@ -22,22 +22,31 @@
...
+    if 'source_data' not in extra_data:
+        extra_data = {
             'source_data': {
                 'data': deepcopy(workflow_data),
-                'extra_data': extra_data or {},
+                'extra_data': extra_data,
             }
-        },
+        }
+    wf = Workflow(name='article', extra_data=extra_data, uuid=uuid.uuid4())
```
michamos, line 40: why do you need this? can't you pass the `uuid` and `name` to `workflow_object_class.create`?

.github/workflows/backend-tests-on-docker.yml
```diff
@@ -106,7 +106,8 @@ jobs:
           --ignore tests/integration/workflows/test_workflows_tasks_actions.py \
           --ignore tests/integration/workflows/test_workflows_cli.py \
           --ignore tests/integration/workflows/test_workflow_utils.py \
-          --ignore tests/integration/workflows/test_workflow_core_selection.py
+          --ignore tests/integration/workflows/test_workflow_core_selection.py \
+          --ignore tests/integration/workflows/test_matching.py
```
michamos, line 110: haven't you made changes to the matching logic? do we really want to stop testing it?

tests/integration/workflows/helpers/workflow_utils.py
pazembrz: Unfortunately No. `workflow_object_class` is actually [invenio_workflows_files.api.WorkflowObject](https://github.com/inveniosoftware-contrib/invenio-workflows-files/blob/1e2b07ea4e46f89e3ca30211919e8cf7680c030c/invenio_workflows_files/api.py#L53) Which is mostly actually [invenio_workflows.api.WorkflowObject](https://github.com/inveniosoftware-contrib/invenio-workflows/blob/892a9c3cc2134566bcd575c43e55865d6381a4a0/invenio_workflows/api.py#L47) and in `create` we are not even able to properly create `Workflow` object which is holding data about workflow type, workflow class etc. Which is required now to properly filter-out workflows. Proper way of creating WF with `Workflow` Object properly assigned to is it only through [`save`](https://github.com/inveniosoftware-contrib/invenio-workflows/blob/892a9c3cc2134566bcd575c43e55865d6381a4a0/invenio_workflows/api.py#L98-L136) method which is doing this if if_workflow of already created `Workflow` object is passed. It's messy. It took me yesterday some time to understand how this works in the backend :/

.github/workflows/backend-tests-on-docker.yml
pazembrz: I just moved this test from outside workflows suite into workflow suite, and I disabled it form |Workflows-II| so it will run only on "Wokflows-I"

tests/integration/workflows/helpers/workflow_utils.py
pazembrz: I can explain it on zoom if you would like to ;)

michamos: It's ok, I can wait for the presentation :wink:

michamos (approved):

PR #4065, opened 2021-06-16
next: refactor to use `revision_id` instead of version_id

drjova:


inspirehep/modules/workflows/tasks/upload.py
```diff
@@ -107,11 +107,11 @@ def store_record_inspirehep_api(obj, eng, is_update, is_authors):
 def send_record_to_hep(obj, pid_type, control_number=None):
     try:
         if control_number:
-            head_version_id = obj.extra_data.get('head_version_id')
+            revision_id = obj.extra_data.get('revision_id')
```
michamos, line 110: this seems like a bad idea, it will make all existing workflows invalid and they will need to be restarted from the beginning. Also, why is this a `get`? if the variable is not there, it should fail, as you need it to perform the update.

drjova: what do you mean it's bad idea? There is not other way around that, invenio-records-rest and editor expect a `revision_id` I wouldn't change in these parts.

michamos: I mean computing the `revision_id` from the `version_id`, so do `head_version_id - 1` when doing the API call to `hep`.

drjova: I see, but this is super hackish at the end we are using different versions in different modules

michamos: yeah, I know, but is it worth breaking all existing workflows to avoid the hack? it's up to you!

PR #4066, opened 2021-06-16
workflows: calculate the version on the fly

drjova:


drjova:
@michamos updated, could you please have another look?

inspirehep/modules/workflows/tasks/upload.py
```diff
@@ -107,11 +107,11 @@ def store_record_inspirehep_api(obj, eng, is_update, is_authors):
 def send_record_to_hep(obj, pid_type, control_number=None):
     try:
         if control_number:
-            head_version_id = obj.extra_data.get('head_version_id')
+            head_version_id = obj.extra_data['head_version_id']
```
MJedr, line 110: Is that always present?

MJedr (approved):

inspirehep/modules/workflows/tasks/upload.py
drjova: It should be!

PR #4070, opened 2021-06-21
workflows: improve affiliation linking for ambiguous matches

MJedr:
ref: #inspirehep/inspirehep#1930

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -941,6 +941,43 @@ def normalize_collaborations(obj, eng):
     return obj
 
 
+def _clean_up_affiliation_data(affiliations):
+    cleaned_affiliations = []
+    for aff in affiliations:
+        cleaned_affiliations.append({key: val for key, val in aff.items() if key in ['value', 'record']})
+    return cleaned_affiliations
+
+
+def _raw_aff_highlight_len(highlighted_raw_aff):
+    matches = re.findall(r"<em>(.*?)</em>", highlighted_raw_aff)
+    return sum(len(match) for match in matches)
+
+
+def _extract_matched_aff_from_highlight(highlighted_raw_affs, author_raw_affs, author_affs):
+    raw_aff_highlight_lenghts = [_raw_aff_highlight_len(raw_aff) for raw_aff in highlighted_raw_affs]
+    longest_highlight_idx = raw_aff_highlight_lenghts.index(max(raw_aff_highlight_lenghts))
+    extracted_raw_aff = re.sub('<em>|</em>', "", highlighted_raw_affs[longest_highlight_idx])
+    for raw_aff, aff in zip(author_raw_affs, author_affs):
+        if raw_aff['value'] == extracted_raw_aff:
+            return [aff]
+
+
+def _find_unambiguous_affiliation(result):
+    for matched_author in result:
+        matched_author_data = matched_author.meta.inner_hits.authors.hits[0].to_dict()
+        matched_author_raw_affs = matched_author_data['raw_affiliations']
+        matched_author_affs = matched_author_data['affiliations']
+        if len(matched_author_raw_affs) == 1:
+            matched_aff = matched_author_affs
+        elif len(matched_author_raw_affs) == len(matched_author_affs):
+            matched_aff = _extract_matched_aff_from_highlight(
+                matched_author.meta.highlight['authors.raw_affiliations.value'],
+                matched_author_raw_affs,
+                matched_author_affs
+            )
+        return _clean_up_affiliation_data(matched_aff)
+
+
 def _match_lit_author_affiliation(raw_aff):
     query = Q(
         "nested",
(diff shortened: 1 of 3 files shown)


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -941,6 +941,38 @@ def normalize_collaborations(obj, eng):
...
 
+def _get_correct_affiliation_data(affiliations):
+    cleaned_affiliations = []
+    for aff in affiliations:
+        cleaned_affiliations.append({key: val for key, val in aff.items() if key in ['value', 'record']})
+    return cleaned_affiliations
+
+
+def _extract_matched_aff_from_highlight(highlighted_raw_affs, author_raw_affs, author_affs):
+    raw_aff_highlight_lenghts = [len(re.findall('<em>.*?<\/em>', raw_aff)) for raw_aff in highlighted_raw_affs]
```
michamos, line 952: doesn't `len(re.findall(...))` compute the number of pieces? I think you want to find the total matching length instead (without delimiters).

So if you have 
```
highlighted_raw_affs = [ "<em>a</em> string <em>with</em> many <em>short</em> matches", "a string <em>with a much longer single match</em>"]
```
you want to pick the second, not the first, which I believe your code is currently doing (you can add a unit test for that).

MJedr: Currently (I think that it's because `raw_affiliations.value` is a text field) highlighter is splitting it into tokens and then each matched token is highlighted. I couldn't find any solution to change it. 

So the example output for `"authors.raw_affiliations.value": "Institute of Physics Belgrade, Belgrade, Serbia"` is: `"<em>Institute</em> <em>of</em> <em>Physics</em> <em>Belgrade</em>, <em>Belgrade</em>, <em>Serbia</em>"`

michamos: ok, so you're counting the number of matched tokens. I think total matched length would make more sense though. Do you agree?

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -941,6 +941,39 @@ def normalize_collaborations(obj, eng):
...
+def _get_correct_affiliation_data(affiliations):
+    cleaned_affiliations = []
+    for aff in affiliations:
+        cleaned_affiliations.append({key: val for key, val in aff.items() if key in ['value', 'record']})
+    return cleaned_affiliations
+
+
+def _extract_matched_aff_from_highlight(highlighted_raw_affs, author_raw_affs, author_affs):
+    highlight_matches = [re.finditer('(?=<em>).*?(?<=<\/em>)', raw_aff) for raw_aff in highlighted_raw_affs]
+    raw_aff_highlight_lenghts = [sum([match.span()[1] - match.span()[0] for match in aff_matches]) for aff_matches in highlight_matches]
```
michamos, line 953: Instead of using lookaround and computing position differences, you can also improve the previous approach to use a capturing group.
If you define
```python
def _raw_aff_highlight_len(highlighted_raw_aff):
    matches = re.finditer(r"<em>(.*?)</em>", highlighted_raw_aff)
    return sum(len(match) for match in matches)
```
you can do
```suggestion
    raw_aff_highlight_lenghts = [_raw_aff_highlight_len(raw_aff) for raw_aff in highlighted_raw_affs]
```

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -941,6 +941,39 @@ def normalize_collaborations(obj, eng):
...
+    longest_highlight_idx = raw_aff_highlight_lenghts.index(max(raw_aff_highlight_lenghts))
+    extracted_raw_aff = re.sub('<em>|</em>', "", highlighted_raw_affs[longest_highlight_idx])
+    for ix, raw_aff in enumerate(author_raw_affs):
+        if raw_aff['value'] == extracted_raw_aff:
+            return [author_affs[ix]]
```
michamos, line 958: if you want to iterate on two lists together, you should use `zip`, not `enumerate`:
```suggestion
    for raw_aff, aff in zip(author_raw_affs, author_affs):
        if raw_aff['value'] == extracted_raw_aff:
            return aff
```

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -941,6 +941,39 @@ def normalize_collaborations(obj, eng):
...
+        matched_author_affs = matched_author_data['affiliations']
+        if len(matched_author_raw_affs) == 1:
+            return _get_correct_affiliation_data(matched_author_affs)
+        elif len(matched_author_raw_affs) == len(matched_author_affs):
+            matched_aff = _extract_matched_aff_from_highlight(
+                matched_author.meta.highlight['authors.raw_affiliations.value'],
+                matched_author_raw_affs,
+                matched_author_affs
+            )
+            return _get_correct_affiliation_data(matched_aff)
```
michamos, line 974: you're calling this function in every branch, better take it out of this function. Also this should be called `_clean_up_affiliation_data` to make its purpose more clear.

michamos (approved):

PR #4071, opened 2021-06-22
workflows: fix logging when no collaborations

oguzdemirbasci:
Ref: https://github.com/inspirehep/inspirehep/issues/1969

### DIFF
--- a/tests/integration/workflows/test_workflows_tasks_actions.py
+++ b/tests/integration/workflows/test_workflows_tasks_actions.py
@@ -904,7 +904,7 @@ def test_normalize_collaborations(workflow_app, insert_experiments_into_db):
     assert obj.data['accelerator_experiments'] == expected_accelerator_experiments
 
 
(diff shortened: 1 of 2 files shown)


tests/integration/workflows/test_workflows_tasks_actions.py
```diff
@@ -904,7 +904,7 @@ def test_normalize_collaborations(workflow_app, insert_experiments_into_db):
     assert obj.data['accelerator_experiments'] == expected_accelerator_experiments
 
 
-def test_normalize_collaborations_one_collaboration(workflow_app, insert_experiments_into_db):
+def test_normalize_collaborations_no_collaboration(workflow_app, insert_experiments_into_db):
```
michamos, line 907: why not keep both tests?

oguzdemirbasci: I added this test in the previous pr. But it doesn't make sense now. We need to test multiple collaborations and 0 collaboration. No need for testing a single collaboration.

michamos (approved):

PR #4072, opened 2021-06-23
workflows: load records from HEP only

pazembrz:


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -1139,3 +1140,15 @@ def check_if_france_in_raw_affiliations(obj, eng):
     for aff in chain.from_iterable(raw_affs):
         if "france" in aff.lower():
             return True
+
+
+def load_record_from_hep(obj, wf):
```
MJedr, line 1145: do we use it only in core selection workflow?

MJedr (approved):

inspirehep/modules/workflows/tasks/actions.py
pazembrz: Yes, I think so.

PR #4082, opened 2021-07-01
inspire-next: record url not displayed in holdingpen

hanadikaleel:
#2005

### DIFF
--- a/inspirehep/modules/workflows/tasks/upload.py
+++ b/inspirehep/modules/workflows/tasks/upload.py
@@ -30,6 +30,8 @@ from invenio_workflows.errors import WorkflowsError
 from simplejson import JSONDecodeError
 import backoff
 
+from os.path import join
+from inspire_utils.urls import ensure_scheme
 from inspire_schemas.readers import LiteratureReader
 from invenio_db import db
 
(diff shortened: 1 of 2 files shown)


drjova:
This https://github.com/inspirehep/inspirehep-search-js/blob/master/src/inspirehep-holdingpen-js/templates/inspirehep-holdingpen-js/details/details.html#L17-L19 should change to ``vm.record._extra_data.recid``

inspirehep/modules/workflows/tasks/upload.py
```diff
@@ -76,6 +80,7 @@ def store_record(obj, eng):
 
                 obj.data['control_number'] = record['control_number']
                 obj.extra_data['recid'] = record['control_number']
+                obj.extra_data['recid'] = join(base_url, 'record', obj.extra_data['recid'])
```
drjova, line 83:
```suggestion
                base_url = ensure_scheme(current_app.config["SERVER_NAME"])
                obj.extra_data['url'] = join(base_url, 'record', obj.extra_data['recid'])
```

tests/integration/workflows/test_workflows_tasks_upload.py
```diff
@@ -245,6 +247,8 @@ def test_store_record_inspirehep_api_literature_new(workflow_app):
     assert workflow.data['control_number'] == expected_control_number
     assert workflow.extra_data['recid'] == expected_control_number
     assert workflow.extra_data['head_uuid'] == expected_head_uuid
+    base_url = ensure_scheme(current_app.config["SERVER_NAME"])
+    assert workflow.extra_data['url'] == join(base_url, 'record', workflow.extra_data['recid'])
```
michamos, line 251: it's not very useful to have a test that basically contains all the same logic as the code. If for some reason it doesn't work correctly, this test will never tell you. You can read more about this in the great Google testing blog: https://testing.googleblog.com/2014/07/testing-on-toilet-dont-put-logic-in.html.
```suggestion
    assert workflow.extra_data['url'] == "http://localhost:5000/record/111"
```

drjova (approved):

PR #4092, opened 2021-07-13
holdingpen: enable batch actions for core selection

MJedr:
ref: inspirehep/inspirehep#1976

### DIFF
--- a/inspirehep/modules/workflows/workflows/core_selection.py
+++ b/inspirehep/modules/workflows/workflows/core_selection.py
@@ -27,7 +27,7 @@ class CoreSelection(object):
 
     workflow = (
         halt_record(
-            action='auto_non_core_record',
+            action='hep_approval',
             message='Submission halted Waiting for curator to decide if record is CORE.'
         ),
         load_record_from_hep,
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -167,6 +167,7 @@
             'hep_approval = inspirehep.modules.workflows.actions.hep_approval:HEPApproval',
             'merge_approval = inspirehep.modules.workflows.actions.merge_approval:MergeApproval',
             'match_approval = inspirehep.modules.workflows.actions.match_approval:MatchApproval',
+            'auto_non_core_record = inspirehep.modules.workflows.workflows:CoreSelection'
```
michamos, line 170: are you sure this works? based on the other examples, it looks like the entry point should be a class with a `resolve` static method that deals with the payload, for example https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/actions/hep_approval.py#L33. I don't see anything like that in the class you reference (which is a workflow).

MJedr: Now for sure it's not working because this action is not registered at all, so I think that's the issue there https://sentry.inspirehep.net/inspire-prod/next/issues/36047/?query=is%3Aunresolved

michamos: I know, but the thing you're registering doesn't have the right shape, it's a workflow class, not an action class (which has a `resolve` static method). Look at the example I linked.

inspirehep/modules/workflows/workflows/core_selection.py
```diff
@@ -27,7 +27,7 @@ class CoreSelection(object):
 
     workflow = (
         halt_record(
-            action='auto_non_core_record',
+            action='hep_approval',
```
drjova, line 30: oh wait this will not work, as we want to aggregate the workflows differently for `auto_non_core_records`

PR #4093, opened 2021-07-21
WIP: Fuzzy matcher should handle abstracts with MathML

hanadikaleel:
#1984


inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -87,9 +88,12 @@ def fuzzy_match(obj, eng):
     """
     if not current_app.config.get('FEATURE_FLAG_ENABLE_FUZZY_MATCHER'):
         return False
-
+    math_ml_latex_regex = '<mml[^>]*>([^<]+)<\/mml> | <latex[^>]*>([^<]+)<\/latex>'
```
michamos, line 91: Both sides of these regexes are wrong.
* LaTeX is not an XML language, so isn't delimited by XML-like tags. Instead, it uses `$...$` or `\(...\)` or `\[...\]` as delimiters. You should use some variation on https://github.com/inspirehep/inspirehep/blob/cb15e67376042277a0550a2ae0f8b7147a8d65ae/backend/inspirehep/records/marshmallow/literature/utils.py#L20 (also adding the last pattern I pointed out).
* where is your examples coming from? `mml` is an XML namespace prefix, not the actual tag. In INSPIRE (and [HTML5](https://github.com/inspirehep/inspirehep/blob/cb15e67376042277a0550a2ae0f8b7147a8d65ae/backend/inspirehep/records/marshmallow/literature/utils.py#L20)), the namespace shouldn't be included. You should match `<math>` tags instead (and use a more realistic example in the tests, like the abstract of https://inspirehep.net/api/literature/1870659). Also your regex should be non-greedy (currently you're matching everything in `<mml>1</mml>middle text outside<mml>2</mml>`, but the middle text outside of the tags shouldn't be). It's also not clear to me if you want to strip all markup, or strip also the content of the tags (with my previous example, do you expect `1middle text outside2` or `middle text outside`).

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -87,9 +88,12 @@ def fuzzy_match(obj, eng):
...
     fuzzy_match_config = current_app.config['FUZZY_MATCH']
-    matches = dedupe_list(match(obj.data, fuzzy_match_config))
+    obj_data_without_math_ml_latex = deepcopy(obj.data)
```
michamos, line 93: this is wasteful if it's a big record! you're only touching abstracts, so it makes more sense to make a shallow copy, and replace the `abstracts` of that copy only.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -24,8 +24,9 @@
 
 from __future__ import absolute_import, division, print_function
 
+import re
 from flask import current_app
-
+from copy import deepcopy
```
MJedr, line 29: It's not needed anymore

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -87,9 +89,12 @@ def fuzzy_match(obj, eng):
     """
     if not current_app.config.get('FEATURE_FLAG_ENABLE_FUZZY_MATCHER'):
         return False
-
+    math_ml_latex_regex = '<math[^>]*>|<\/math> | (?<!\\)\$.*?(?<!\\)\$|(?<!\\)\\\(.*?(?<!\\)\\\)'
```
michamos, line 92: this is wrong too, you don't want to strip only math tags, take a look at the example I gave you previously (and use that one as a test case)! Also, you didn't include `\[...\]`in the LaTeX pattern.

tests/integration/workflows/test_matching.py
```diff
@@ -167,3 +178,30 @@ def test_stop_matched_holdingpen_wfs(app, simple_record):
...
+
+    obj = workflow_object_class.create(
+        data=record,
+        id_user=1,
+        data_type='hep'
+    )
+
+    eng = None
+
+    assert not fuzzy_match(obj, eng)
```
michamos, line 207: could you explain what you're testing here?

tests/integration/workflows/fixtures/literature_1863053.json
```diff
@@ -452,6 +452,10 @@
     {
       "value": "We apply object detection techniques based on Convolutional Neural Networks to jet reconstruction and identification at the CERN Large Hadron Collider. In particular, we focus on CaloJet reconstruction, representing each event as an image composed of calorimeter cells and using a Single Shot Detection network, called Jet-SSD. The model performs simultaneous localization and classification and additional regression tasks to measure jet features. We investigate Ternary Weight Networks with weights constrained to {-1, 0, 1} times a layer- and channel-dependent scaling factors. We show that the quantized version of the network closely matches the performance of its full-precision equivalent.",
       "source": "arXiv"
+    },
+    {
+      "value": "<mml:msub>\n<mml:mi>Z</mml:mi>\n<mml:mn>2</mml:mn>\n</mml:msub>\n</mml:math>",
```
michamos, line 457: there are not realistic values, they have a `mml` prefix which we don't have in the records. Please use one of the real abstracts we have that contain MathML.

tests/integration/workflows/test_matching.py
```diff
@@ -53,6 +55,7 @@ def simple_record(app):
             'titles': [{'title': 'Superconductivity'}],
             'acquisition_source': {'source': 'arXiv'},
             'dois': [{'value': '10.3847/2041-8213/aa9110'}],
+            'abstracts': [{'value': '<mml:msub>\n<mml:mi>Z</mml:mi>\n<mml:mn>2</mml:mn>\n</mml:msub>\n</mml:math>'}],
```
michamos, line 58: same

PR #4095, opened 2021-07-26
records: document url should be escaped

MJedr:
ref:inspirehep/inspirehep#2059

### DIFF
--- a/inspirehep/modules/records/api.py
+++ b/inspirehep/modules/records/api.py
@@ -26,6 +26,7 @@ from __future__ import absolute_import, division, print_function
 
 from copy import deepcopy
 from datetime import datetime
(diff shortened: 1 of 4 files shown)


inspirehep/modules/search/bundles.py
```diff
@@ -34,6 +34,6 @@
     npm={
         'invenio-search-js': '~1.4.0',
         'angular-loading-bar': '~0.9.0',
-        'inspirehep-search-js': '~2.0.0'
+        'inspirehep-search-js': '==2.0.10'
```
drjova, line 37: I would revert this change

drjova (dismissed):

drjova (approved):

PR #4096, opened 2021-07-29
setup: improve local setup script

MJedr:
ref: inspirehep/inspirehep#2050

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,4 +1,8 @@
-[tool]
+# The next version of INSPIRE.
+# If you are updating this file (eg. adding entry points) don't forget to update
+# setup.py which is used for setting up docker environment. The guide how to do it might 
+# be found here: https://python-poetry.org/docs/pyproject/
+
 [tool.poetry]
 name = "Inspirehep"
 version = "0.1.0"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -129,3 +128,106 @@ web-node = ["gunicorn", "honcho"]
...
+
+[tool.poetry.plugins."invenio_workflows_ui.actions"]
+author_approval = "inspirehep.modules.workflows.actions.author_approval:AuthorApproval"
+hep_approval = "inspirehep.modules.workflows.actions.hep_approval:HEPApproval"
+merge_approval = "inspirehep.modules.workflows.actions.merge_approval:MergeApproval"
+match_approval = "inspirehep.modules.workflows.actions.match_approval:MatchApproval"
+
+[build-system]
+requires = ["setuptools >= 40.6.0", "wheel"]
+build-backend = "setuptools.build_meta"
```
michamos, line 233: not sure this will work. If not, you can explicitly do `pip install setuptools` in the Dockerfile.

michamos:
One thing I'm worried about with the current approach is the two files getting out of sync, causing tests to pass locally (installing with `poetry`) but not on the CI (installing with `pip`). I don't see an easy solution there, maybe put some comment in both files with instructions to update the other file?

MJedr:
Maybe we can use something which will automatically convert `pyproject.toml` to `setup.py`? Like that https://pypi.org/project/poetry2setup/

michamos (dismissed):

drjova (approved):

PR #4097, opened 2021-08-03
workflows: add non_core_selection workflow

MJedr:
ref:inspirehep/inspirehep#2057


.github/workflows/backend-tests-on-docker.yml
```diff
@@ -107,6 +107,7 @@ jobs:
           --ignore tests/integration/workflows/test_workflows_cli.py \
           --ignore tests/integration/workflows/test_workflow_utils.py \
           --ignore tests/integration/workflows/test_workflow_core_selection.py \
+          --ignore tests/integration/workflows/test_workflow_non_core_selection.py \
```
drjova, line 110: why ignoring it?

tests/integration/workflows/test_workflow_non_core_selection.py
```diff
@@ -0,0 +1,196 @@
...
+            "method": "hepcrawl",
+            "submission_number": "978",
+            "source": "Elsevier",
+        },
+    }
+    workflow_id = build_workflow(record).id
+    start("article", object_id=workflow_id)
+    wf = workflow_object_class.get(workflow_id)
+    wf.continue_workflow()
+    wf.extra_data["approved"]
```
drjova, line 111: left over?

tests/integration/workflows/test_workflow_non_core_selection.py
```diff
@@ -0,0 +1,196 @@
...
+            "datetime": "2020-11-12T04:49:13.369515",
+            "method": "hepcrawl",
+            "submission_number": "978",
+            "source": "Elsevier",
+        },
+    }
+    workflow_id = build_workflow(record).id
+    start("article", object_id=workflow_id)
+    wf = workflow_object_class.get(workflow_id)
+    wf.continue_workflow()
```
drjova, line 110: do we need `conitnue` cause it's already started?

.github/workflows/backend-tests-on-docker.yml
MJedr: because we split tests into 2 workflows, so no sense in running it in both

tests/integration/workflows/test_workflow_non_core_selection.py
MJedr: no, otherwise it was halted

drjova (dismissed):

drjova (approved):

PR #4103, opened 2021-08-18
Fuzzy matcher should handle abstracts with MathML

hanadikaleel:


tests/integration/workflows/test_matching.py
```diff
@@ -167,3 +170,82 @@ def test_stop_matched_holdingpen_wfs(app, simple_record):
...
+    )
+    eng = object()
+
+    assert fuzzy_match(obj, eng)
+    assert 'matches' in obj.extra_data
+
+    matches = get_value(obj.extra_data, 'matches.fuzzy')
+
+    # assert the result with correct abstract comes as a first result
+    assert matches[0]['control_number'] == record_1.json['control_number']
```
michamos, line 251: looks great :+1: 
could you also assert that you're not mutating the abstract? that is, before running the fuzzy matcher on L245 you do
```
original_abstract = obj.data["abstracts"][0]["value"]
```
then at the end you add
```
assert obj.data["abstracts"][0]["value"] == original_abstract
```
The reason I'm asking is that I have a strong suspicion that this will fail, because on L94 of the code you do a shallow copy of `obj.data`, but then you reassign `abstract["value"]` which is still shared with the original `obj.data`.

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -87,9 +89,14 @@ def fuzzy_match(obj, eng):
     """
     if not current_app.config.get('FEATURE_FLAG_ENABLE_FUZZY_MATCHER'):
         return False
-
+    math_ml_latex_regex = r"(<math(.*?)<\/math>|(?<!\\)\$.*?(?<!\\)\$|(?<!\\)\\(.*?(?<!\\)\\)|(?<!\\)\\[.*?(?<!\\)\\])"
     fuzzy_match_config = current_app.config['FUZZY_MATCH']
-    matches = dedupe_list(match(obj.data, fuzzy_match_config))
+    obj_data_without_math_ml_latex = copy(obj.data)
+    if "abstract" in obj.data:
+        abstract_dup = deepcopy(obj.data["abstracts"])
```
michamos, line 96: you're never assigning this back to the `obj.data["abstracts"]`. How come the test passes then?

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -87,9 +89,14 @@ def fuzzy_match(obj, eng):
     """
     if not current_app.config.get('FEATURE_FLAG_ENABLE_FUZZY_MATCHER'):
         return False
-
+    math_ml_latex_regex = r"(<math(.*?)<\/math>|(?<!\\)\$.*?(?<!\\)\$|(?<!\\)\\(.*?(?<!\\)\\)|(?<!\\)\\[.*?(?<!\\)\\])"
     fuzzy_match_config = current_app.config['FUZZY_MATCH']
-    matches = dedupe_list(match(obj.data, fuzzy_match_config))
+    obj_data_without_math_ml_latex = copy(obj.data)
+    if "abstract" in obj.data:
```
michamos, line 95:
```suggestion
    if "abstracts" in obj.data:
```

michamos (approved):

PR #4114, opened 2021-09-27
Remove RT Ticket

hanadikaleel:


inspirehep/modules/workflows/workflows/author.py
```diff
@@ -88,17 +87,6 @@
 ]
 
 
-SEND_UPDATE_NOTIFICATION = [
-    do_not_repeat('create_ticket_author_submission_curator_update')(
-        create_ticket(
-            template="authors/tickets/curator_update.html",
```
MJedr, line 94: Is the template still in use? If not, maybe you can delete it?

hanadikaleel: It's not used- removed it

hanadikaleel: It's not used- removed it

PR #4131, opened 2021-11-03
workflows: automatically assign categories to journals

MJedr:
ref: inspirehep/inspirehep#2273

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -580,7 +580,7 @@ def normalize_journal_titles(obj, eng):
 
     Note:
         The DB is queried in order to get the `$ref` of each journal and add it in
-        `journal_record`.
+        `journal_record` as well as inspire categories.
 
     TODO:
         Refactor: it must be checked that `normalize_journal_title` is appropriate.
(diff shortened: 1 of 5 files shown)


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -616,11 +616,26 @@ def normalize_journal_title_entry(publication_info):
...
+
+    if not journal_data:
+        return
+
+    journal_record = journal_data['self']
 
     if journal_record:
         publication_info['journal_record'] = journal_record
 
+    if not check_mark(obj, 'is-update') and journal_data.get('inspire_categories'):
```
michamos, line 629: do we need the first check? we can always write the categories, even if we don't need them later.

michamos (approved):
minor (optional) comment, but looks ok

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -616,11 +616,26 @@ def normalize_journal_title_entry(publication_info):
...
+
+    journal_record = journal_data['self']
 
     if journal_record:
         publication_info['journal_record'] = journal_record
 
+    if journal_data.get('inspire_categories'):
+        current_categories = obj.extra_data.get('journal_inspire_categories', [])
+        current_categories.extend(journal_data['inspire_categories'])
+        obj.extra_data['journal_inspire_categories'] = current_categories
```
michamos, line 632: didn't realize first, but this is a one-liner using `setdefault`
```suggestion
        obj.extra_data.setdefault('journal_inspire_categories', []).extend(journal_data['inspire_categories'])
```

michamos (approved):

PR #4147, opened 2022-02-04
add inspire-classifier to workflows

MJedr:
ref: inspirehep/inspirehep#2299


tests/unit/workflows/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+from __future__ import absolute_import, division, print_function
+
+import pytest
+
+
+IS_VCR_ENABLED = True
+IS_VCR_EPISODE_OR_ERROR = True  # False to record new cassettes.
+
+
+@pytest.fixture
```
drjova, line 32: I believe if the VCR configuration is the same you can add it to the `tests/conftest.py` and share it between suites

tests/unit/utils/test_utils_url.py
```diff
@@ -45,7 +45,6 @@ def test_is_pdf_link_handles_empty_requests():
         assert not is_pdf_link('http://example.org/empty-pdf')
 
 
-@pytest.mark.vcr()
```
drjova, line 48: this might fail, since it will try to reach arxiv server https://github.com/inspirehep/inspire-next/blob/a1e6356e2ed30a764825d6e5fa875a16efa2d3e0/inspirehep/utils/url.py#L64

drjova (approved):

PR #4149, opened 2022-02-17
don't stop processing workflows from other source when matched

MJedr:
ref: inspirehep/inspirehep#1730

### DIFF
--- a/inspirehep/modules/workflows/workflows/article.py
+++ b/inspirehep/modules/workflows/workflows/article.py
@@ -482,21 +482,10 @@ PROCESS_HOLDINGPEN_MATCH_HARVEST = [
     ),
     IF_ELSE(
         is_marked('already-in-holding-pen'),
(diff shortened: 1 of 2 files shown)


inspirehep/modules/workflows/workflows/article.py
```diff
@@ -511,9 +500,9 @@
 
         # stop the matched wf and continue this one
         [
-            stop_matched_holdingpen_wfs,
-            mark('stopped-matched-holdingpen-wf', True),
-            save_workflow
+            mark('stopped-matched-holdingpen-wf', False),
+            save_workflow,
+            stop_processing
```
michamos, line 505: I'm confused, why did you change this one?

MJedr: cause we talked to finish submission workflows, at least it's what I understood.

PR #4153, opened 2022-03-09
bump inspire-matcher and test validator

nooraangelva:
Ref: #inspirehep/inspirehep/issues/2377


tests/integration/workflows/test_matching.py
```diff
@@ -252,3 +252,90 @@ def test_fuzzy_match_without_math_ml_and_latex(workflow_app):
...
+    eng = object()
+
+    original_abstract = obj.data["abstracts"][0]["value"]
+    assert fuzzy_match(obj, eng)
+    assert "matches" in obj.extra_data
+    matches = get_value(obj.extra_data, "matches.fuzzy")
+
+    # assert the result with correct abstract comes as a first result
+    assert matches[0]["control_number"] == record_1.json["control_number"]
+    assert obj.data["abstracts"][0]["value"] == original_abstract
```
MJedr, line 341: I believe that in that test you don't want to get matches for same arxiv paper. Could you assert that the the incorrect match is not in results?

tests/integration/workflows/test_matching.py
```diff
@@ -252,3 +252,91 @@ def test_fuzzy_match_without_math_ml_and_latex(workflow_app):
...
+
+    original_abstract = obj.data["abstracts"][0]["value"]
+    assert fuzzy_match(obj, eng)
+    assert "matches" in obj.extra_data
+    matches = get_value(obj.extra_data, "matches.fuzzy")
+
+    # assert the result with correct abstract and not with the incorrect one
+    assert matches[0]["control_number"] == record_1.json["control_number"]
+    assert obj.data["abstracts"][0]["value"] == original_abstract
+    assert not any(record_2.json["control_number"] in sublist for sublist in matches)
```
MJedr, line 342: you can simply assert that the length of matches is one and the first match is record_1 (what you already have in your code ;))

nooraangelva: Keep it simple. As they say xD

MJedr (approved):

PR #4154, opened 2022-04-11
remove from object doc or fig if empty

nooraangelva:
Ref: https://github.com/inspirehep/inspirehep/issues/2455

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -410,7 +410,11 @@ def populate_submission_document(obj, eng):
             url=submission_pdf,
             original_url=submission_pdf,
         )
+
         obj.data = lb.record
+
+    if 'documents' in obj.data and len(obj.data['documents']) == 0:
+        del obj.data['documents']
 
 
 @with_debug_logging
--- a/inspirehep/modules/workflows/tasks/arxiv.py
+++ b/inspirehep/modules/workflows/tasks/arxiv.py
@@ -99,6 +99,7 @@ def populate_arxiv_document(obj, eng):
         original_url=url,
         url=url,
     )
+
     obj.data = lb.record
 
 
@@ -206,6 +207,10 @@ def arxiv_plot_extract(obj, eng):
             )
 
         obj.data = lb.record
+
+    if len(obj.data['figures']) == 0:
+        del obj.data['figures']
+    else:
         obj.log.info('Added {0} plots.'.format(len(plots)))
 
 
--- a/tests/unit/workflows/test_workflows_tasks_arxiv.py
+++ b/tests/unit/workflows/test_workflows_tasks_arxiv.py
@@ -181,6 +181,30 @@ def test_populate_arxiv_document_logs_on_pdf_not_existing():
         assert expected == result
 
 
+def test_populate_arxiv_document_figures_not_existing():
+    schema = load_schema('hep')
+    subschema = schema['properties']['arxiv_eprints']
+    data = {
+        'arxiv_eprints': [
+            {
+                'categories': [
+                    'physics.ins-det',
+                ],
+                'value': '1612.00626',
+            },
+        ],
+    }  # synthetic data
+    extra_data = {}
+    files = MockFiles({})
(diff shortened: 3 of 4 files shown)


tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -181,6 +181,98 @@ def test_populate_arxiv_document_logs_on_pdf_not_existing():
...
+            (response500,),
+        )
+        schema = load_schema('hep')
+        subschema = schema['properties']['arxiv_eprints']
+
+        data = {
+            'arxiv_eprints': [
+                {
+                    'categories': [
+                        'cs.CV',
```
MJedr, line 207: I think that in order to test the code that you wrote you need to add `documents` key, which should be emptied [here](https://github.com/inspirehep/inspire-next/blob/87326f7b0417a14dced1bc59f43e17dc29ca2ff3/inspirehep/modules/workflows/tasks/actions.py#L401-L403) cause we are deleting documents what could result in something like `documents: []` and then we assign workflow data to the builder. Same for other tests

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -410,7 +410,12 @@ def populate_submission_document(obj, eng):
             url=submission_pdf,
             original_url=submission_pdf,
         )
+
         obj.data = lb.record
+        if not obj.data .get('documents'):
+            del obj.data['documents']
+        else:
+            obj.data = lb.record
```
MJedr, line 418:
```suggestion
        if not obj.data.get('documents'):
            del obj.data['documents']
```

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -99,7 +99,12 @@ def populate_arxiv_document(obj, eng):
         original_url=url,
         url=url,
     )
+
     obj.data = lb.record
+    if not obj.data .get('documents'):
+        del obj.data['documents']
+    else:
+        obj.data = lb.record
```
MJedr, line 107:
```suggestion
    if not obj.data.get('documents'):
        del obj.data['documents']
```

MJedr (approved):

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -410,8 +410,12 @@ def populate_submission_document(obj, eng):
             url=submission_pdf,
             original_url=submission_pdf,
         )
+
         obj.data = lb.record
 
+    if not obj.data.get('documents') and 'documents' in obj.data:
+        del obj.data['documents']
```
MJedr, line 417: I think that this check is very unclear, maybe if you really need to check explicitly if it's an empty list you can check:

```
if 'documents' in obj.data and len(obj.data['documents']) == 0:
```

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -99,7 +99,10 @@ def populate_arxiv_document(obj, eng):
         original_url=url,
         url=url,
     )
+
     obj.data = lb.record
+    if not obj.data.get('documents') and 'documents':
```
MJedr, line 104: same (and the check is wrong here btw, you forgot `'documents' in obj.data` I think)

tests/unit/workflows/test_workflows_tasks_arxiv.py
```diff
@@ -181,6 +181,72 @@ def test_populate_arxiv_document_logs_on_pdf_not_existing():
...
+        }  # literature/1458302
+        extra_data = {}
+        files = MockFiles({})
+        assert validate(data['arxiv_eprints'], subschema) is None
+
+        obj = MockObj(data, extra_data, files=files)
+        eng = MockEng()
+
+        assert populate_arxiv_document(obj, eng) is None
+        assert not obj.data.get('documents')
```
MJedr, line 223: just last thing - can you assert `'documents' not in obj.data` (and same for figures? Cause here we don't know if the key is not there or maybe if the list is empty and we actually don't want this key to be present in wf

MJedr (dismissed):

MJedr (approved):

PR #4156, opened 2022-04-20
refextract: don't query deleted jounrals for journal dict

MJedr:
ref:#inspirehep/inspirehep#2399

### DIFF
--- a/inspirehep/modules/refextract/tasks.py
+++ b/inspirehep/modules/refextract/tasks.py
@@ -55,7 +55,7 @@ def create_journal_kb_file():
         FROM
             records_metadata AS r
         WHERE
-            (r.json -> '_collections')::jsonb ? 'Journals'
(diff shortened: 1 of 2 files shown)


inspirehep/modules/refextract/tasks.py
```diff
@@ -55,7 +55,7 @@ def create_journal_kb_file():
         FROM
             records_metadata AS r
         WHERE
-            (r.json -> '_collections')::jsonb ? 'Journals'
+            (r.json -> '_collections')::jsonb ? 'Journals' AND ((r.json -> 'deleted') IS NULL OR r.json['deleted'] != 'true'::jsonb)
```
michamos, line 58: isn't `r.json['deleted']` a SQLAlchemy thing? or is it new PostgreSQL syntax? also, isn't `NULL != 'true'::jsonb`, making the first clause redundant?

MJedr: Actually it doesn't work correctly without adding the clause with `NULL`. `r.json['deleted']` also works but I would change it to be consistent with the rest of the query

michamos (approved):

PR #4163, opened 2022-04-28
improve author.xml handling

nooraangelva:
Ref: https://github.com/cern-sis/issues-inspire/issues/19


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -250,31 +248,72 @@ def _author_list(obj, eng):
...
+                    name = normalize_name(author.xpath(".//familyName/text()").get() + ", " + author.xpath(".//givenName/text()").get())
+
+                    # building the info to a correct format with litratureBuilder()
+                    builder.add_author(
+                        builder.make_author(
+                            full_name=name,
+                            affiliations=affiliations,
+                            ids=ids,
                         )
+                    )
```
michamos, line 307: could you maybe extract this part into a function that takes a string containing the XML and returns the JSON translation? that way it's easier to test and we can more easily run it manually and reuse it.

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -250,31 +248,72 @@ def _author_list(obj, eng):
...
+
+                if isinstance(xml_content, str):
+                    xml_content = xml_content.decode('utf-8')
+
+                # Probably the %auto-ignore comment exists, so we skip the
+                # first line. See: inspirehep/inspire-next/issues/2195
+                if "%auto-ignore" in xml_content:
+                    content = Selector(text=xml_content.split('\n', 1)[1], type="xml")
+                else:
+                    content = Selector(text=xml_content, type="xml")
```
michamos, line 270:
```suggestion
                if "%auto-ignore" in xml_content:
                    xml_content.split('\n', 1)[1]
                content = Selector(text=xml_content, type="xml")
```

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -250,31 +248,72 @@ def _author_list(obj, eng):
...
+                        index = 1
+                    else:
+                        index = 0
+
+                    # Getting all the author ids
+                    for id in author.xpath(author_id_paths[index]):
+                        if id.xpath("./@source").get() == u'CCID':
+                            ids.append(['CERN', id.xpath("./text()").get()])
+                        elif id.xpath("./@source").get() == u'INSPIRE':
+                            ids.append([id.xpath("./@source").get() + ' ID', id.xpath("./text()").get()])
```
michamos, line 290: you shoudln't use `+` for string interpolation (anywhere), better use f-strings on Python 3 or `format` here.

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -250,31 +248,72 @@ def _author_list(obj, eng):
...
+                        else:
+                            ids.append([id.xpath("./@source").get(), id.xpath("./text()").get()])
+
+                    # Getting all the names for affiliated organizations using the organization ids from author
+                    for affiliation in author.xpath("./authorAffiliations/authorAffiliation/@organizationid").getall():
+                        affiliations.append(content.xpath("//organizations/Organization[@id='" + affiliation + "']/orgName[@source='spiresICN' or @source='INSPIRE']/text()").get())
```
michamos, line 296: you should also get the ROR, see example linked in issue.

inspirehep/modules/workflows/tasks/arxiv.py
michamos, line 290: can you also handle various placeholders for empty data, like in https://github.com/inspirehep/inspire-next/blob/a442fd80476e48fe37eb9a2c6e0854c487eb2caf/inspirehep/modules/workflows/utils/stylesheets/authorlist2marcxml.xsl#L71 ?

That is, if the tag contains `undefined`, `UNDEFINED`, `INSPIRE-`, `inspire-` or ``, skip it (instead of generating a validation error down the line).

inspirehep/modules/workflows/tasks/arxiv.py
michamos: and also ignore `undefined`/`UNDEFINED` or empty value.

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -215,68 +212,102 @@ def arxiv_plot_extract(obj, eng):
...
+
+    content.remove_namespaces()
+    # Going through all the authors in the file
+    for author in content.xpath("//Person"):
+
+        ids = []
+        affiliations = []
+
+        # Getting all the author ids
+        for source, id in itertools.izip(author.xpath('./authorIDs/authorID[@source!="undefined" and @source!="UNDEFINED" and @source!="INSPIRE-" and @source!="inspire-" and @source!="" and text()!="" and text()!="undefined" and text()!="UNDEFINED"]/@source | ./authorids/authorid[@source!="undefined" and @source!="UNDEFINED" and @source!="INSPIRE-" and @source!="inspire-" and @source!="" and text()!="" and text()!="undefined" and text()!="UNDEFINED"]/@source').getall(), author.xpath('./authorIDs/authorID[@source!="undefined" and @source!="UNDEFINED" and @source!="INSPIRE-" and @source!="inspire-" and @source!="" and text()!="" and text()!="undefined" and text()!="UNDEFINED"]/text() | ./authorids/authorid[@source!="undefined" and @source!="UNDEFINED" and @source!="INSPIRE-" and @source!="inspire-" and @source!="" and text()!="" and text()!="undefined" and text()!="UNDEFINED"]/text()').getall()):
```
michamos, line 291: maybe it would be easier to do all the string filtering in python rather than xpath?

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -215,68 +212,102 @@ def arxiv_plot_extract(obj, eng):
...
+            elif source == u'INSPIRE':
+                ids.append(['{} ID'.format(source), id])
             else:
-                set_mark(obj, "authors_xml", False)
+                ids.append([source, id])
 
-    return _author_list
+        # Getting all the names for affiliated organizations using the organization ids from author
+        for affiliation in author.xpath("./authorAffiliations/authorAffiliation/@organizationid").getall():
+            affiliations.append(content.xpath('//organizations/Organization[@id="' + affiliation + '"]/orgName[@source="spiresICN" or @source="INSPIRE" or @source="ROR" and text()!="undefined" and text()!="UNDEFINED" and text()!="" ]/text()').get())
```
michamos, line 301: * are you putting the ROR in the right place? in the end, it should end up in [`affiiliations_identifiers`](https://github.com/inspirehep/inspire-schemas/blob/270dbb0d25a8d652b669a3456d6ef02b29e0f824/inspire_schemas/records/hep.yml#L294-L320) in the record, not `affiliations` (see issue for test case).
* you shouldn't do string formatting with `+` (see previous comment).

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -215,68 +212,115 @@ def arxiv_plot_extract(obj, eng):
...
+    # Going through all the authors in the file
+    for author in content.xpath("//Person"):
+
+        ids = []
+        affiliations = []
+        affiliations_identifiers = []
+
+        # Getting all the author ids
+        for source, id in itertools.izip(author.xpath('./authorIDs/authorID[@source!="" and text()!=""]/@source | ./authorids/authorid[@source!="" and text()!=""]/@source').getall(), author.xpath('./authorIDs/authorID[@source!="" and text()!=""]/text() | ./authorids/authorid[@source!="" and text()!=""]/text()').getall()):
+            if not re.match("undefined|inspire-", source, flags=re.IGNORECASE) or re.match("undefined", id, flags=re.IGNORECASE):
```
michamos, line 293: `inspire-` can be present in the ID, not the source.

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -215,68 +212,115 @@ def arxiv_plot_extract(obj, eng):
...
+        # Getting all the names for affiliated organizations using the organization ids from author
+        for affiliation in author.xpath("./authorAffiliations/authorAffiliation/@organizationid").getall():
+            affiliations.append(content.xpath('//organizations/Organization[@id="{}"]/orgName[@source="spiresICN" or @source="INSPIRE" and text()!="undefined" and text()!="UNDEFINED" and text()!="" ]/text()'.format(affiliation)).get())
+
+            # Getting all the affiliations_identifiers for affiliated organizations using the organization ids from author
+
+            for value, source in itertools.izip(content.xpath('//organizations/Organization[@id="{}"]/orgName[@source="ROR" or @source="GRID" and text()!=""]/text()'.format(affiliation)).getall(), content.xpath('//organizations/Organization[@id="{}"]/orgName[@source="ROR" or @source="GRID" and text()!=""]/@source'.format(affiliation)).getall()):
+                if not re.match("undefined", source, flags=re.IGNORECASE) or re.match("undefined", id, flags=re.IGNORECASE):
+
+                    if source == 'ROR':
```
michamos, line 310: you select both GRID and ROR in L307, but use only the ROR here.

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -215,68 +212,115 @@ def arxiv_plot_extract(obj, eng):
...
+                        value = 'https://ror.org/{}'.format(value)
```
michamos, line 311: If the value already starts with `https://ror.org`, you should take it literally instead of prepending this to make it into a URL (ROR had the "great" idea to make their IDs into URLs, causing much confusion).

inspirehep/modules/workflows/tasks/arxiv.py
nooraangelva: I add them both, the ROR needs the ```https://ror.org``` to pass the schema check.
The ```affiliations_identifiers.append([source, value])``` is not inside the ```if```

michamos: oh, right, I missed that :+1:

inspirehep/modules/workflows/tasks/arxiv.py
nooraangelva: I'm only checking it from the source.

michamos: exactly, which is not where it can appear in the XML file, it might appear in the ID because it's a prefix of the correct ID, but should not be taken in that case.

nooraangelva: Ah now I understand. Thank you for explaining.

inspirehep/modules/workflows/tasks/arxiv.py
nooraangelva: I'm doing now most of the filtering in python. I'm still checking the empty ones in xpath, otherwise, the order in the list might not be correct and the number of items in source and id might differ.

michamos: that's a good point, I hadn't thought of that. But wouldn't it be the other way around? what if you have something like
`<authorID source="INSPIRE"></authorID>`. Wouldn't that mess up the alignement?

nooraangelva: Yes. That's why I'm checking the source and the value in both of the XPath() searches  ```/authorID[@source!="" and text()!=""]/@source```
```/authorID[@source!="" and text()!=""]/text()```

Comment (michamos) [reply inspirehep/modules/workflows/tasks/arxiv.py]: :facepalm: of course, you're right, I didn't look carefully enough

File: inspirehep/modules/workflows/tasks/arxiv.py [hunk 10]
```diff
@@ -215,68 +212,116 @@ def arxiv_plot_extract(obj, eng):
...
+
+            for value, source in itertools.izip(content.xpath('//organizations/Organization[@id="{}"]/orgName[@source="ROR" or @source="GRID" and text()!=""]/text()'.format(affiliation)).getall(), content.xpath('//organizations/Organization[@id="{}"]/orgName[@source="ROR" or @source="GRID" and text()!=""]/@source'.format(affiliation)).getall()):
+                if not re.match("undefined", source, flags=re.IGNORECASE) or not re.match("undefined", id, flags=re.IGNORECASE):
+
+                    if source == 'ROR' and not re.match("https://ror.org/*", value):
+                        value = 'https://ror.org/{}'.format(value)
+
+                    affiliations_identifiers.append([source, value])
+
+        name = normalize_name(author.xpath('.//familyName/text()').get() + ', ' + author.xpath('.//givenName/text()').get())
```
michamos, line 315:
```suggestion
        name = normalize_name("{}, {}".format(author.xpath('.//familyName/text()').get(), author.xpath('.//givenName/text()').get()))
```

michamos (approved):
just a minor comment remaining, looks good overall, well done :+1:

inspirehep/modules/workflows/tasks/arxiv.py
michamos: sorry, python 2, so `u"{}, {}"`

nooraangelva: This is why one should never copy their homework from others ;)

PR #4166, opened 2022-05-02
holdingpen: link workflow ids

karolina-siemieniuk-morawska:
* changed ids from plain strings to urls in error messages

* ref: cern-sis/issues-inspire#20

### DIFF
--- a/inspirehep/modules/workflows/tasks/matching.py
+++ b/inspirehep/modules/workflows/tasks/matching.py
@@ -36,6 +36,7 @@ import re
 
 from inspire_matcher.api import match
 from inspire_utils.dedupers import dedupe_list
+from inspirehep.utils.url import get_hep_url_for_recid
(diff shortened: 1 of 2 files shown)


inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -310,12 +310,14 @@ def raise_if_match_workflow(obj, eng):
         skip_halted=True
     )
     if bool(matched_ids):
+        urls = ['https://inspirehep.net/holdingpen/{}'.format(id) for id in matched_ids]
```
drjova, line 313: The link is hardcoded and it will not work for QA and beta. Instead you should do this:
https://github.com/inspirehep/inspire-next/blob/master/inspirehep/utils/url.py#L106

```suggestion
        urls = [ get_hep_url_for_recid(id, 'holdingpen') for id in matched_ids]
```

drjova (approved):

PR #4169, opened 2022-05-04
[WIP] use hep api for root table

MJedr:


drjova:
I think we could close this

PR #4170, opened 2022-05-05
fix for the refextract response

MJedr:


inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -154,7 +154,7 @@ def extract_references_from_pdf_url(url, custom_kbs_file, source=None):
         headers=refextract_request_headers,
         data=json.dumps({"url": url, "journal_kb_data": custom_kbs_file})
     )
-    if response.status_code != 200 and response.json.get("messsage"):
+    if response.status_code != 200 and response.json().get("messsage"):
         LOGGER.info("Couldn't extract references from url!")
```
michamos, line 158: I think we should fail the workflow in this case, rather than silently not extract refs.

drjova (approved):

PR #4171, opened 2022-05-06
refextract: pass url with inspire host

MJedr:


inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -272,7 +272,7 @@ def get_document_url_for_reference_extraction(obj):
 
     url = documents[0]['url']
     obj.log.info('Using document with url "%s"', url)
-    return url
+    return current_app.config['INSPIREHEP_URL'] + url
```
drjova, line 275: doesn't need a `/`?

MJedr: I don't think so, urls are starting with `/`

MJedr: [like `"url": "/api/files/be4ca558-1948-478c-a7f1-05af95b94282/2205.02190.pdf"`](https://github.com/inspirehep/inspire-next/blob/c42106fa7d47d89023748f62d0e900a839cc6976/inspirehep/modules/workflows/tasks/actions.py#L440)

drjova:
```
Instance: /usr/var/inspirehep-instance
In [1]: app.config['INSPIREHEP_URL']
Out[1]: 'http://10.254.122.111:80/api'
```

drjova: https://github.com/cern-sis/kubernetes/blob/master/next/base/config/inspirehep.cfg#L32

drjova: Maybe we should use something like https://github.com/inspirehep/inspire-next/blob/master/inspirehep/utils/url.py#L106-L114

inspirehep/modules/workflows/utils/__init__.py
```diff
@@ -610,3 +610,13 @@ def store_head_version(obj, eng):
     obj.extra_data['head_uuid'] = str(head_uuid)
     obj.extra_data['head_version_id'] = head_version
     obj.save()
+
+
+def _get_hep_url_for_document(file_url):
+    server_name = current_app.config.get("SERVER_NAME")
+    protocol = current_app.config.get("PREFERRED_URL_SCHEME", 'http')
+    return "{protocol}://{server_name}/{file_url}".format(
```
drjova, line 618: the `file_url` already contains the `/` no?

tests/unit/workflows/test_workflows_utils.py
```diff
@@ -29,6 +29,7 @@
 import requests
 import requests_mock
 import time
+from flask import current_app
```
drjova, line 32: this cannot work, in unit tests we don't have a flask app

tests/unit/workflows/test_workflows_utils.py
```diff
@@ -524,3 +526,23 @@ def test_grobid_empty_author():
...
+            {
+                'key': 'document.pdf',
+            },
+        ],
+    }
+    obj = MockObj(data, {})
+    url = get_document_url_for_reference_extraction(obj)
+    assert current_app.config['SERVER_NAME'] in url
+    assert file_url in url
+    assert 'http' in url
```
drjova, line 548: We have to mock the `current_app`, and we have to update the assert in order to catch errors like the previous I mentioned 😃 

```suggestion
    obj = MockObj(data, {})
    url = get_document_url_for_reference_extraction(obj)
    expected_url = 'http://{0}{1}'.format(server_name, file_url)
    assert expected_url == url
```

tests/unit/workflows/test_workflows_utils.py
MJedr: I thought so too, but actually they are..

PR #4180, opened 2022-05-24
refextract: handle extract journal info exception

MJedr:
ref: cern-sis/issues-inspire#51

### DIFF
--- a/inspirehep/modules/workflows/tasks/refextract.py
+++ b/inspirehep/modules/workflows/tasks/refextract.py
@@ -28,6 +28,7 @@ import backoff
 from itertools import chain
 import json
 from flask import current_app
+from requests.exceptions import RequestException
 import requests
 
 from invenio_workflows.errors import WorkflowsError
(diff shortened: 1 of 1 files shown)


inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -91,12 +92,22 @@ def extract_journal_info(obj, eng):
...
-            "{}/extract_journal_info".format(current_app.config["REFEXTRACT_SERVICE_URL"]),
-            headers=refextract_request_headers,
-            data=json.dumps({"publication_infos": publication_infos, "journal_kb_data": kbs_journal_dict})
-        )
-        extracted_publication_info = response.json()['extracted_publication_infos']
+        try:
+            response = requests.post(
+                "{}/extract_journal_info".format(current_app.config["REFEXTRACT_SERVICE_URL"]),
+                headers=refextract_request_headers,
+                data=json.dumps({"publication_infos": publication_infos, "journal_kb_data": kbs_journal_dict})
```
drjova, line 99: don't think you need `json.dumps` btw

drjova: maybe we can retry before we fail the workflow? especially if it's timeout or bad gateway?

drjova (approved):

PR #4185, opened 2022-06-01
Fix empty orgName in authors.xml

nooraangelva:
Ref: https://github.com/cern-sis/issues-inspire/issues/50


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -301,8 +301,9 @@ def extract_authors_from_xml(xml_content):
         # Getting all the names for affiliated organizations using the organization ids from author
         for affiliation in author.xpath("./authorAffiliations/authorAffiliation/@organizationid").getall():
             orgName = content.xpath('//organizations/Organization[@id="{}"]/orgName[@source="spiresICN" or @source="INSPIRE" and text()!="" ]/text()'.format(affiliation)).get()
-            if not re.match(undefined_or_none_value_regex, orgName):
-                affiliations.append(orgName)
+            if orgName:
+                if not re.match(undefined_or_none_value_regex, orgName):
+                    affiliations.append(orgName)
```
MJedr, line 306: can we have that in one if cause the number of indentations in this function is going crazy? 😅

drjova, line 306:
```suggestion
            if orgName and not re.match(undefined_or_none_value_regex, orgName):
                    affiliations.append(orgName)
```

MJedr (approved):

PR #4186, opened 2022-06-13
inspireid value is empty

nooraangelva:
Ref: https://github.com/cern-sis/issues-inspire/issues/61

### DIFF
--- a/inspirehep/modules/workflows/tasks/arxiv.py
+++ b/inspirehep/modules/workflows/tasks/arxiv.py
@@ -281,16 +281,20 @@ def extract_authors_from_xml(xml_content):
     content = Selector(text=xml_content, type="xml")
     content.remove_namespaces()
     undefined_or_none_value_regex = re.compile("undefined|none", re.IGNORECASE)
-    # Going through all the authors in the file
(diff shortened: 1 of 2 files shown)


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -290,7 +290,7 @@ def extract_authors_from_xml(xml_content):
 
         # Getting all the author ids
         for source, id in itertools.izip(author.xpath('./authorIDs/authorID[@source!="" and text()!=""]/@source | ./authorids/authorid[@source!="" and text()!=""]/@source').getall(), author.xpath('./authorIDs/authorID[@source!="" and text()!=""]/text() | ./authorids/authorid[@source!="" and text()!=""]/text()').getall()):
-            if not re.match("undefined", source, flags=re.IGNORECASE) or not re.match("undefined|inspire-", id, flags=re.IGNORECASE):
+            if not re.match("undefined", source, flags=re.IGNORECASE) and not re.match("undefined|inspire-$", id, flags=re.IGNORECASE):
```
MJedr, line 293: I would check also if there are no whitespaces after (maybe it would be parsed as something like `INSPIRE- `) and your regex will not match it

MJedr (approved):

PR #4196, opened 2022-08-01
Remove "\n" from text

nooraangelva:
ref: https://github.com/cern-sis/issues-inspire/issues/94


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -284,6 +284,7 @@ def extract_authors_from_xml(xml_content):
     undefined_or_empty_inspireid_value_regex = re.compile("undefined|inspire-\s*$", re.IGNORECASE)
     undefined_value_regex = re.compile("undefined", re.IGNORECASE)
     ror_path_value_regex = re.compile("https://ror.org/*")
+    remove_new_line_regex = re.compile(u"\n\s+")
```
MJedr, line 287: and if the spaces are before new line char?

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -304,12 +307,14 @@ def extract_authors_from_xml(xml_content):
 
         # Gets all the names for affiliated organizations using the organization ids from author
         for affiliation in author.xpath("./authorAffiliations/authorAffiliation/@organizationid").getall():
-            orgName = content.xpath(u'//organizations/Organization[@id="{}"]/orgName[@source="spiresICN" or @source="INSPIRE" and text()!="" ]/text()'.format(affiliation)).get()
+            orgName = re.sub(remove_new_line_regex, '', str(content.xpath(u'//organizations/Organization[@id="{}"]/orgName[@source="spiresICN" or @source="INSPIRE" and text()!="" ]/text()'.format(affiliation)).get()))
```
MJedr, line 310: maybe it would be clearer to extract the name first and then do `re.sub`

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -284,6 +284,7 @@ def extract_authors_from_xml(xml_content):
     undefined_or_empty_inspireid_value_regex = re.compile("undefined|inspire-\s*$", re.IGNORECASE)
     undefined_value_regex = re.compile("undefined", re.IGNORECASE)
     ror_path_value_regex = re.compile("https://ror.org/*")
+    remove_new_line_regex = re.compile(u"\s*\n\s+")
```
MJedr, line 287: are we sure that there's always `\s` char after new line?

nooraangelva: Hmm, at least in the problem record yes, but good point. I could change it also to ```\s*\n\s*```

MJedr: I think it would be safer

MJedr (approved):

PR #4199, opened 2022-08-01
fix remove none valued rors

nooraangelva:
Ref: https://github.com/cern-sis/issues-inspire/issues/95


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -310,7 +310,7 @@ def extract_authors_from_xml(xml_content):
 
             # Gets all the affiliations_identifiers for affiliated organizations using the organization ids from author
             for value, source in itertools.izip(content.xpath(u'//organizations/Organization[@id="{}"]/orgName[@source="ROR" or @source="GRID" and text()!=""]/text()'.format(affiliation)).getall(), content.xpath(u'//organizations/Organization[@id="{}"]/orgName[@source="ROR" or @source="GRID" and text()!=""]/@source'.format(affiliation)).getall()):
-                if not re.match(undefined_or_none_value_regex, source) or not re.match(undefined_or_none_value_regex, id):
+                if not re.match(undefined_or_none_value_regex, source) and not re.match(undefined_or_none_value_regex, value):
```
MJedr, line 313: maybe instead you can do 
```
if re.match(undefined_or_none_value_regex, source) and not re.match(undefined_or_none_value_regex, value):
    continue
```
so that we can remove one indentation level?

MJedr (approved):

PR #4200, opened 2022-08-02
core selection: keep only inspire_categories derived from non-core arxiv categories

MJedr:
ref: cern-sis/issues-inspire#63


nooraangelva (approved):
LGTM 👍 
Thanks for the very clear naming of the variables.

inspirehep/modules/workflows/actions/core_selection_approval.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+
+        obj.extra_data["core"] = value == "accept_core"
+        obj.remove_action()
+
+        obj.extra_data["reason"] = reason
+        obj.extra_data["user_action"] = value
+
+        obj.status = ObjectStatus.WAITING
+        obj.save()
```
drjova, line 61: don't we need `db.session.commit()`?

inspirehep/modules/workflows/workflows/core_selection.py
```diff
@@ -6,43 +6,50 @@
...
+                do_not_repeat("create_ticket_curator_core_publisher")(
+                    create_ticket(
+                        template="literaturesuggest/tickets/curation_core.html",
+                        queue="HEP_curation",
+                        context_factory=curation_ticket_context,
+                        ticket_id_key="curation_ticket_id",
+                    ),
+                ),
+                store_record,
+                SEND_TO_LEGACY,
```
drjova, line 51: we can just drop the SEND_TO_LEGACY :)

inspirehep/modules/workflows/actions/core_selection_approval.py
MJedr: Not necessarily cause `obj.save` is adding to the session and `obj.continue_workflow` is committing at the beginning of the function.

drjova:
Sorry if I missed any discussion, that means we have to remove also the Core/Accept buttons from the UI, is it what we really want?

MJedr:
definitely not, it is what we do for all the other actions which require user decision, eg. [HEPApproval](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/actions/hep_approval.py). The point is that until now we were stopping workflow right after the `Approve` click and we didn't need to pass anything in the payload from the decision. So now, since we need to continue the workflow to remove those categories so I had to refactor it in the same way as for other actions.

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -1149,3 +1149,27 @@ def load_record_from_hep(obj, wf):
...
 
     return obj
+
+
+def remove_inspire_categories_derived_from_core_arxiv_categories(obj, eng):
+    if not obj.data.get("arxiv_eprints"):
+        return
+    inspire_categories_without_arxiv_sourced = [
+        category
+        for category in obj.data.get("inspire_categories")
```
drjova, line 1159:
```suggestion
        for category in obj.data.get("inspire_categories", [])
```

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -1149,3 +1149,27 @@ def load_record_from_hep(obj, wf):
...
+        if category.get("source") != "arxiv"
+    ]
+    non_core_arxiv_categories = [
+        arxiv_category
+        for arxiv_category in obj.data["arxiv_eprints"][0]["categories"]
```
drjova, line 1164: are we sure that we get always `arxiv_eprints[0].categories`?

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -1149,3 +1149,27 @@ def load_record_from_hep(obj, wf):
...
     return obj
+
+
+def remove_inspire_categories_derived_from_core_arxiv_categories(obj, eng):
+    if not obj.data.get("arxiv_eprints"):
+        return
+    inspire_categories_without_arxiv_sourced = [
+        category
+        for category in obj.data.get("inspire_categories")
+        if category.get("source") != "arxiv"
```
drjova, line 1160: From the requirements, shouldn't be `=='arxiv'`?

> The easiest approach is probably to call classify_field again in the workflow and substitute all inspire_categories with source=arXiv with the newly derived ones.

MJedr: No, it's correct, later we process  arxiv categories and we reclassify them leaving non-arxiv sourced categories not touched.

inspirehep/modules/workflows/tasks/actions.py
MJedr: I thought that `inspire_categories` are always present when `arxiv_eprints` are in the record, but you are right, it might happen that there's an `arxiv_eprint` without a category 👍

inspirehep/modules/workflows/tasks/actions.py
MJedr: yes, you are right, `categories` are not required in `arxiv_eprints` 👍

inspirehep/modules/workflows/tasks/actions.py
MJedr: No, it's correct, later we process arxiv categories and we reclassify them leaving non-arxiv sourced categories not touched.

drjova (approved):

PR #4205, opened 2022-08-11
abstracts in holdingpen not interpreted as HTML now as string

nooraangelva:


MJedr:
firstly you need a release in `inspirehep-search-js`, but it is not working at the moment

PR #4209, opened 2022-08-22
don't stop wfs with different source when hp match

MJedr:
ref: cern-sis/issues-inspire#101

### DIFF
--- a/inspirehep/modules/workflows/tasks/matching.py
+++ b/inspirehep/modules/workflows/tasks/matching.py
@@ -31,7 +31,7 @@ from inspire_utils.record import get_value
 from invenio_db import db
 from invenio_workflows import workflow_object_class, WorkflowEngine, \
     ObjectStatus
-from invenio_workflows.errors import WorkflowsError
+from invenio_workflows.errors import WorkflowsError, HaltProcessing
 import re
 
 from inspire_matcher.api import match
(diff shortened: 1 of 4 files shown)


inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -446,14 +446,50 @@ def _get_wfs_same_source(obj, eng):
...
+            if _has_same_source(current_source, wf):
                 return True
         return False
 
     return _get_wfs_same_source
 
 
+def _has_same_source(current_wf_source, holdingpen_match_wf):
+    wf_source = get_value(holdingpen_match_wf.data, 'acquisition_source.source').lower()
+    if wf_source == current_wf_source:
```
drjova, line 458:
```suggestion
    if wf_source == current_wf_source.lower():
```

tests/integration/workflows/test_arxiv_workflow.py
```diff
@@ -1226,16 +1226,16 @@ def test_match_in_holdingpen_different_sources_continues(
...
+    start('article', object_id=workflow_2_id)
+    obj_2 = workflow_object_class.get(workflow_2_id)
 
-    assert obj.extra_data['already-in-holding-pen'] is True
-    assert obj.extra_data['holdingpen_matches'] == [wf_to_match]
-    assert obj.extra_data['previously_rejected'] is False
-    assert obj.extra_data.get('stopped-matched-holdingpen-wf')
+    # the first wf from another source is still halted in HP
     assert obj.status == ObjectStatus.HALTED
+    assert obj_2.status == ObjectStatus.HALTED
```
drjova, line 1235: shall we check also that the engine is HALTED?

inspirehep/modules/workflows/tasks/matching.py
```diff
@@ -446,14 +446,50 @@ def _get_wfs_same_source(obj, eng):
...
+        holdingpen_wf = workflow_object_class.get(holdingpen_wf_id)
+        holdingpen_wf_eng = WorkflowEngine.from_uuid(holdingpen_wf.id_workflow)
+
+        # halt workflow
+        if same_source_only and not _has_same_source(current_wf_source, holdingpen_wf):
+            _halt_record_if_match_from_different_source(holdingpen_wf_eng, holdingpen_wf)
+            continue
+
+        # stop this holdingpen workflow by replacing its steps with a stop step
+        _replace_matched_wf_steps_with_stop_steps(stopping_steps, holdingpen_wf_eng, holdingpen_wf)
```
drjova, line 490: is this just used here? Shall we just add the two lines here instead of a separate function?

drjova (approved):
LGTM

inspirehep/modules/workflows/workflows/article.py
```diff
@@ -491,18 +487,13 @@
 
 
 PROCESS_HOLDINGPEN_MATCH_SUBMISSION = [
+    handle_matched_holdingpen_wfs,
     IF_ELSE(
         has_same_source('holdingpen_matches'),
-        # form should detect this double submission
-        ERROR_WITH_UNEXPECTED_WORKFLOW_PATH,
```
michamos, line 497: could we keep this part? it should be very rare, but different submissions are not interchangeable in principle.

michamos (approved):
looks good, small comment only :)

PR #4213, opened 2022-08-30
workflows: move rejected wfs to hidden collections based on report number

MJedr:
ref: cern-sis/issues-inspire#108

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -808,16 +808,35 @@ def affiliations_for_hidden_collections(obj):
     return [affiliations_mapping[affiliation] for affiliation in affiliations_set]
 
 
+def reports_for_hidden_collections(obj):
+    hidden_collection_mapping = current_app.config.get(
+        "AFFILIATIONS_TO_HIDDEN_COLLECTIONS_MAPPING", {}
+    )
+    hidden_collections = set()
+    for report_number in obj.data.get("report_numbers", []):
+        hidden_collections.update(
+            [
+                match.upper()
+                for match in re.findall("cern|fermilab", report_number["value"], re.IGNORECASE)
+            ]
+        )
+    return {hidden_collection_mapping[collection] for collection in hidden_collections}
+
+
 @with_debug_logging
 def should_be_hidden(obj, eng):
     """Checks if paper attached to this workflow is affiliated with one of affiliations which interests us."""
-    return bool(affiliations_for_hidden_collections(obj))
+    return bool(affiliations_for_hidden_collections(obj)) or bool(
+        reports_for_hidden_collections(obj)
+    )
 
 
 @with_debug_logging
 def replace_collection_to_hidden(obj, eng):
     """Replaces collection to hidden based on authors affiliations"""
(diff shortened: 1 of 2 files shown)


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -808,16 +808,26 @@ def affiliations_for_hidden_collections(obj):
     return [affiliations_mapping[affiliation] for affiliation in affiliations_set]
 
 
+def reports_for_hidden_collections(obj):
+    hidden_collection_mapping = current_app.config.get("AFFILIATIONS_TO_HIDDEN_COLLECTIONS_MAPPING", {})
+    hidden_collections = set()
+    for report in obj.data.get('report_number', []):
```
drjova, line 814: shouldn't be ` report_numbers`?

drjova, line 814:
```suggestion
    for report_number in obj.data.get('report_numbers', []):
```

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -808,16 +808,26 @@ def affiliations_for_hidden_collections(obj):
...
+        hidden_collections.update([match.upper() for match in re.findall('cern|fermilab', report['value'], re.IGNORECASE)])
```
drjova, line 815: it would be more readable if we split this into more lines

drjova: also we could directly update the set with the `hidden_collection_mapping` value, instead of creating a new set

MJedr: But `hidden_collection_mapping` is a different thing, there are also other values. I would prefer to store matched collections separately

inspirehep/modules/workflows/tasks/actions.py
MJedr: yes, you are right

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -808,16 +808,35 @@ def affiliations_for_hidden_collections(obj):
     return [affiliations_mapping[affiliation] for affiliation in affiliations_set]
 
 
+def reports_for_hidden_collections(obj):
+    hidden_collection_mapping = current_app.config.get(
+        "AFFILIATIONS_TO_HIDDEN_COLLECTIONS_MAPPING", {}
+    )
+    hidden_collections = set()
+    for report_number in obj.data.get("report_number", []):
```
drjova, line 816:
```suggestion
    for report_number in obj.data.get("report_numbers", []):
```

:)

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -808,16 +808,35 @@ def affiliations_for_hidden_collections(obj):
...
+        hidden_collections.update(
+            [
+                match.upper()
```
drjova, line 819: if I'm not mistaken, if there is no match it will fail, `None` doesn't have `upper`

MJedr: No, I am using `re.findall` there which is returning empty list if there's no match

drjova (approved):
LGTM

PR #4215, opened 2022-08-31
classifier: update scoring for non-core

MJedr:
ref: cern-sis/issues-inspire#123

### DIFF
--- a/inspirehep/modules/workflows/tasks/classifier.py
+++ b/inspirehep/modules/workflows/tasks/classifier.py
@@ -87,6 +87,8 @@ def guess_coreness(obj, eng):
     relevance_score = max_score
     if decision == "CORE":
         relevance_score += 1
+    elif decision == "Non-CORE":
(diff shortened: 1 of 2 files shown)


inspirehep/modules/workflows/tasks/classifier.py
```diff
@@ -87,6 +87,8 @@ def guess_coreness(obj, eng):
     relevance_score = max_score
     if decision == "CORE":
         relevance_score += 1
+    elif decision == "Non-CORE":
+        relevance_score = 0.5 * scores["Non-CORE"] + scores["CORE"]
```
drjova, line 91: are these keys always present?

drjova (approved):
LGTM just the comment

inspirehep/modules/workflows/tasks/classifier.py
MJedr: yes, they are, they come from classifier

PR #4223, opened 2022-09-23
bug fix: unicode error with organisations in arxiv 

nooraangelva:
Bug fix for this error: https://inspirehep.net/holdingpen/4639376


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -307,7 +307,8 @@ def extract_authors_from_xml(xml_content):
 
         # Gets all the names for affiliated organizations using the organization ids from author
         for affiliation in author.xpath("./authorAffiliations/authorAffiliation/@organizationid").getall():
-            orgName = str(content.xpath(u'//organizations/Organization[@id="{}"]/orgName[@source="spiresICN" or @source="INSPIRE" and text()!="" ]/text()'.format(affiliation)).get())
+            orgName = content.xpath(u'string(//organizations/Organization[@id="{}"]/orgName[@source="spiresICN" or @source="INSPIRE" and text()!="" ]/text())'.format(affiliation)).get()
```
drjova, line 310: We don't need the `u` here :)

nooraangelva: ```UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 1: ordinal not in range(128)``` It seems that I still do need it 😅

drjova (approved):

PR #4228, opened 2022-10-20
update logging settings

MJedr:


inspirehep/config.py
```diff
@@ -160,6 +161,11 @@
     }
 })
 
+logger = logging.getLogger()
+handler = logging.StreamHandler()
+logger.addHandler(handler)
+logger.setLevel(logging.INFO)
```
drjova, line 167: unfortunately this will not work for all loggers but just the defined `logger` in this file.

MJedr: yes, but the code you sent me 
```
handler = logging.StreamHandler()
logging.addHandler(handler)
logging.setLevel(logging.INFO)
```

won't work either

MJedr: I'll add it to the app setup, maybe that'll work @drjova

PR #4238, opened 2022-11-21
add inspire-utils logger

MJedr:


inspirehep/config.py
```diff
@@ -1726,6 +1726,11 @@
                 "handlers": ["console"],
                 "propagate": True
             },
+            "inspire_utils": {
```
drjova, line 1729: I'm surprised that `""` doesn't work

drjova (approved):

PR #4241, opened 2022-11-30
workflows improvements

MJedr:
cern-sis/issues-inspire#201
cern-sis/issues-inspire#203


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -119,6 +119,12 @@
     'Instrumentation',
 ]
 
+GROBID_EXCEPTIONS = (
+    BadGatewayError, 
+    requests.exceptions.ConnectionError, 
+    requests.exceptions.HTTPError
+)
```
drjova, line 126: shall we just [RequestException](https://requests.readthedocs.io/en/latest/api/#requests.RequestException)?

inspirehep/modules/workflows/tasks/upload.py
```diff
@@ -97,7 +99,10 @@ def store_record(obj, eng):
             record.commit()
             obj.save()
     else:
-        store_record_inspirehep_api(obj, eng, is_update, is_authors)
+        try:
+            store_record_inspirehep_api(obj, eng, is_update, is_authors)
+        except WorkflowsError:
+            restart_workflow(obj)
```
drjova, line 105: hmm infinite loop? we should increase the restart counter first

inspirehep/modules/workflows/tasks/upload.py
```diff
@@ -149,22 +157,6 @@ def send_record_to_hep(obj, pid_type, control_number=None):
         obj.save()
 
 
-def create_error(response):
```
drjova, line 152: I'm surprised that this is not used anywhere else

drjova: oh ok you just moved it.

inspirehep/modules/workflows/tasks/upload.py
MJedr: afair it does it automatically, will check

MJedr: if I understand correctly the `restart_workflow` is restarting from step `0` by default so it goes through all the steps including `increase_restart_count` [in pre-processing](https://github.com/inspirehep/inspire-next/blob/a83bae4dc8be2f091d25b6b16c3576b07e8744ba/inspirehep/modules/workflows/workflows/article.py#L587), so imo it's not needed to increase restart count here.

tests/integration/workflows/test_workflows_tasks_upload.py
```diff
@@ -213,324 +202,539 @@ def test_regression_store_record_does_not_commit_when_error(workflow_app):
...
+            store_record(obj, eng)  # not throwing exception
+            wf = workflow_object_class.get(wf_id)
+            assert (
+                wf.extra_data["source_data"]["persistent_data"]["marks"][
+                    "restart-count"
+                ]
+                == 1
+            )
+
+
```
MJedr, line 407: @drjova here is the check for the restart count that is increased during the restart. If we add another `increase_restart_count` it the `store_record` we'll end up with a doubled number of restarts.

drjova (approved):

PR #4260, opened 2023-04-25
delete empty documents key

MJedr:


drjova (approved):

inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -108,6 +108,8 @@ def populate_arxiv_document(obj, eng):
     )
 
     obj.data = lb.record
+    if 'documents' in obj.data and len(obj.data['documents']) == 0:
+        del obj.data['documents']
```
drjova, line 112: can we actually just strip all empty? It happens also with figures

michamos:
I don't think so, it has `_extra_data.is-update=False` so it's not an update. In any case, `merger_root` is the future root corresponding to the current workflow stored at the end in the root table, not the root used when calling the merger for the current workflow.

PR #4263, opened 2023-04-27
improve logging in documents download

MJedr:


inspirehep/modules/workflows/tasks/arxiv.py
```diff
@@ -215,11 +220,8 @@ def arxiv_plot_extract(obj, eng):
             )
 
         obj.data = lb.record
-
-    if 'figures' in obj.data and len(obj.data['figures']) == 0:
-        del obj.data['figures']
-    else:
-        obj.log.info('Added {0} plots.'.format(len(plots)))
+    
+    LOGGER.info('Added {0} plots.'.format(len(plots)))
```
drjova, line 224: should we add also save here?

MJedr: Yes, let's save to be on the safe side 👍

drjova (approved):

PR #4293, opened 2023-06-28
refextract: do not fail on empty `pubinfo_freetext`

drjova:


michamos:
wouldn't it be better to check if you have a `pubinfo_freetext` before doing the request, and only do the request if needed? now you're doing a whole lot of useless calls, and ignoring exceptions that are unrelated to `pubinfo_freetext` not being present.

PR #4298, opened 2023-07-14
workflows: enable matcher on hep

drjova:
* ref: cern-sis/issues-inspire#347


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -475,7 +474,7 @@ def match_references_hep(references):
 
 
 def match_references_based_on_flag(references):
```
MJedr, line 476: I think we're using this function only once, maybe we can use `match_references_hep` directly?

MJedr (approved):

PR #4303, opened 2023-07-19
workflows: fix reference matcher url and payload

drjova:


tests/integration/workflows/conftest.py
```diff
@@ -319,10 +319,11 @@ def mocked_external_services(workflow_app):
             },
             headers=_get_headers_for_hep_root_table_request(),
             status_code=200,
-        )
+        )test_wf_replaces_old_workflow_which_is_in_halted_state
```
MJedr, line 322: new line?

MJedr (approved):

tests/integration/workflows/conftest.py
drjova: 🤦

PR #4315, opened 2023-09-22
fuzzy matching: change request payload

MJedr:


drjova:
hmm shouldn't this be here https://github.com/inspirehep/inspirehep/blob/a8d0b8cce1ed78a0d744a7dac7738a5140548c08/backend/inspirehep/matcher/api.py#L292?

MJedr:
I think filtering here would be better cause we don't need to pass for example references in request to hep (they are not used by matcher)

drjova (approved):

PR #4316, opened 2023-09-26
celery: add liveness probe

MJedr:


inspirehep/celery.py
```diff
@@ -27,31 +27,72 @@
...
 
+from celery import bootsteps
+from celery.signals import worker_ready, worker_shutdown
 from flask_celeryext import AppContextTask, create_celery_app
 from psycopg2 import OperationalError as Psycopg2OperationalError
 from sqlalchemy.exc import InvalidRequestError, OperationalError
 
 from inspirehep.factory import create_app
 
+HEARTBEAT_FILE = Path("/tmp/worker_heartbeat")
```
drjova, line 40: pathlib in 2.7?

MJedr: you are right, thanks for pointing out

drjova (approved):
let's try it

PR #4318, opened 2023-10-11
setup: bump hepcrawl

drjova:


poetry.lock
```diff
@@ -4132,1520 +4150,300 @@ xrootd = ["xrootdpyfs", "invenio-xrootd"]
...
+certifi = []
+cffi = []
+chardet = []
+click = []
+click-spinner = []
+colorama = []
+configobj = []
+configparser = []
+constantly = []
+contextlib2 = []
```
MJedr, line 4191: it seems to be wrong

drjova: good catch but I cannot understand 😕

PR #4326, opened 2023-11-09
normailize journal titles: use hep request

MJedr:


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -611,6 +610,25 @@ def preserve_root(obj, eng):
     obj.save()
 
 
+def _get_all_journal_titles_to_normalize(obj_data):
```
drjova, line 613: could we please add a backoff?

PR #4346, opened 2024-07-24
headers: add content-type json

PascalEgn:
The python 3.11 upgrade of inspirehep will require the requests to be passed with the content type json header.


drjova (approved):

drjova:
@PascalEgn at the end we don't need to add them in fuzzy matcher/matcher?

PascalEgn:
> @PascalEgn at the end we don't need to add them in fuzzy matcher/matcher?

https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/tasks/matching.py#L78
https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/tasks/matching.py#L125

PR #4350, opened 2024-08-27
workflows: remove prepare_keywords step

PascalEgn:
kept the actual function in case we need it again at some later stage

ref: https://github.com/cern-sis/issues-inspire/issues/554


inspirehep/modules/workflows/workflows/core_selection.py
```diff
@@ -13,8 +13,7 @@
     add_core, halt_record, is_core, link_institutions_with_affiliations,
     load_record_from_hep, normalize_author_affiliations,
     remove_inspire_categories_derived_from_core_arxiv_categories)
-from inspirehep.modules.workflows.tasks.submission import (create_ticket,
-                                                           prepare_keywords)
+from inspirehep.modules.workflows.tasks.submission import (create_ticket)
```
drjova, line 16:
```suggestion
from inspirehep.modules.workflows.tasks.submission import create_ticket
```

drjova (approved):

PR #4361, opened 2024-12-20
workflows: add core requirement to uk and ger queue

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/623


michamos (approved):
small change needed, looks ok otherwise

PR #4362, opened 2025-01-07
gh-actions: bump cache and checkout action

PascalEgn:
We need to bump actions/cache as its getting deprecated: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down

Also bumped the checkout action, because why not.


ioannistsanaktsidis (approved):
Looks good!

PR #4366, opened 2025-03-12
refextract: use service for raw_references

DonHaul:
* ref: cern-sis/issues-inspire/issues/814

### DIFF
--- a/inspirehep/modules/workflows/tasks/actions.py
+++ b/inspirehep/modules/workflows/tasks/actions.py
@@ -73,10 +73,11 @@ from inspirehep.modules.records.json_ref_loader import replace_refs
 from inspirehep.modules.records.utils import get_linked_records_in_field
 from inspirehep.modules.workflows.tasks.refextract import (
     extract_references_from_pdf_url,
-    extract_references_from_raw_refs,
     extract_references_from_pdf,
     extract_references_from_text,
     extract_references_from_text_data,
+    extract_references_from_reference_list,
+    raw_refs_to_list
 )
 from inspirehep.modules.workflows.utils import _get_headers_for_hep_root_table_request, create_error
 from inspirehep.modules.workflows.errors import BadGatewayError, MissingRecordControlNumber
@@ -498,15 +499,17 @@ def refextract(obj, eng):
         None
 
     """
+    source = LiteratureReader(obj.data).source
+    journal_kb_dict = create_journal_kb_dict()
+
     if 'references' in obj.data:
-        extracted_raw_references = dedupe_list(extract_references_from_raw_refs(obj.data['references']))
-        obj.log.info('Extracted %d references from raw refs.', len(extracted_raw_references))
-        obj.data['references'] = match_references_hep(extracted_raw_references)
+        raw_refs_to_extract, references = raw_refs_to_list(obj.data['references'])
+        extracted_references = dedupe_list(extract_references_from_reference_list(raw_refs_to_extract, custom_kbs_file=journal_kb_dict))
+        obj.log.info('Extracted %d references from raw refs.', len(extracted_references))
+        obj.data['references'] = match_references_hep(extracted_references + references)
         return
 
     matched_pdf_references, matched_text_references = [], []
-    source = LiteratureReader(obj.data).source
-    journal_kb_dict = create_journal_kb_dict()
 
     url = get_document_url_for_reference_extraction(obj)
     if current_app.config.get("FEATURE_FLAG_ENABLE_REFEXTRACT_SERVICE") and url:
--- a/inspirehep/modules/workflows/tasks/refextract.py
+++ b/inspirehep/modules/workflows/tasks/refextract.py
@@ -25,7 +25,6 @@
 from __future__ import absolute_import, division, print_function
 
 import backoff
-from itertools import chain
 import json
 from flask import current_app
 from requests.exceptions import RequestException
@@ -55,6 +54,7 @@ from inspirehep.modules.workflows.utils import (
 from inspirehep.utils.references import (
     local_refextract_kbs_path,
     map_refextract_to_schema,
+    map_refextract_reference_to_schema
 )
 from ..utils import with_debug_logging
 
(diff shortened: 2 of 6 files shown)


inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -498,15 +499,19 @@ def refextract(obj, eng):
         None
 
     """
+    source = LiteratureReader(obj.data).source
+    journal_kb_dict = create_journal_kb_dict()
+
     if 'references' in obj.data:
-        extracted_raw_references = dedupe_list(extract_references_from_raw_refs(obj.data['references']))
+        if current_app.config.get("FEATURE_FLAG_ENABLE_REFEXTRACT_SERVICE"):
```
ioannistsanaktsidis, line 506: no need for the flag, this is always true

tests/unit/workflows/test_workflows_tasks_actions.py
```diff
@@ -1216,6 +1216,148 @@ def test_validate_record_raises_when_record_is_invalid():
         _validate_record(obj, eng)
 
 
+@patch('inspirehep.modules.workflows.tasks.actions.create_journal_kb_dict', return_value={})
+@patch('inspirehep.modules.refextract.matcher.match', return_value=iter([]))
+def test_refextract_from_raw_refs_service(mock_create_journal_dict, mock_match):
+    data = {
+        'references': [
```
ioannistsanaktsidis, line 1223: endpoint expects a list of strings
https://github.com/inspirehep/refextract/blob/master/tests/integration/test_views.py#L186

inspirehep/modules/workflows/tasks/actions.py
```diff
@@ -498,15 +499,19 @@ def refextract(obj, eng):
...
+            extracted_raw_references = dedupe_list(extract_references_from_reference_list(obj.data['references'], custom_kbs_file=journal_kb_dict, source=source))
```
ioannistsanaktsidis, line 507: Here you are calling directly the service, there is some work that needs to be done before you do this. Take a look into that part 
https://github.com/inspirehep/inspire-next/blob/6400e115a2ab958e92e27a11652df32af1c429c0/inspirehep/modules/workflows/tasks/refextract.py#L251-L307

ioannistsanaktsidis (approved):
Some crucial parts that need to be addressed. Lets fix them and re-review again.

ioannistsanaktsidis (changes requested):
Accidentally accepted before :P

inspirehep/modules/workflows/tasks/actions.py
DonHaul: okay. I will take the opportunity to also remove the functions we will no longer use `extract_references_from_raw_refs ` and `extract_references_from_raw_ref `

DonHaul:
an assumption was made regarding the `source`. instead of checking the `source` of each `raw_ref` we assume that the source is the global record `source`.
From the current records we have stuck on refextract: https://github.com/cern-sis/issues-inspire/issues/505#issuecomment-2706526678  none of them have a `source` defined in the `raw_refs`. so I believe this to be a safe assumption

tests/unit/workflows/test_workflows_tasks_actions.py
DonHaul: this data refers to the `wf` data, not the one sent to `refextract`

DonHaul:
> an assumption was made regarding the `source`. instead of checking the `source` of each `raw_ref` we assume that the source is the global record `source`. From the current records we have stuck on refextract: [cern-sis/issues-inspire#505 (comment)](https://github.com/cern-sis/issues-inspire/issues/505#issuecomment-2706526678) none of them have a `source` defined in the `raw_refs`. so I believe this to be a safe modification as we wont be loosing information regarding the `source`

@michamos  do you there would be any problem with this?

michamos:
yes, the `source` in `raw_refs` doesn't have to match the `acquisition_source` one (which gets updated at every harvest). Can you keep the `source` for the `raw_refs` as is (same value if present, absent if absent)?

tests/unit/workflows/test_workflows_tasks_refextract.py
```diff
@@ -77,7 +80,8 @@ def test_extract_journal_info_handles_year_an_empty_string(mock_create_journal_k
 
     data = {
         'publication_info': [
-            {'pubinfo_freetext': 'The Astrophysical Journal, 838:134 (16pp), 2017 April 1'},
+            {
+                'pubinfo_freetext': 'The Astrophysical Journal, 838:134 (16pp), 2017 April 1'},
```
ioannistsanaktsidis, line 84: formatting issue here too

ioannistsanaktsidis (approved):
Very minor comment. @drjova if you can also take a quick look as this pr is rather crucial. From my side it is good to go.

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -165,6 +165,30 @@ def extract_journal_info(obj, eng):
...
+def extract_references_from_reference_list(raw_references, custom_kbs_file):
+    """Extract references from reference list and return in INSPIRE format."""
+    refextract_request_headers = {
+        "content-type": "application/json",
+    }
+    response = requests.post(
+        "{}/extract_references_from_list".format(current_app.config["REFEXTRACT_SERVICE_URL"]),
+        headers=refextract_request_headers,
+        data=json.dumps({"raw_references": raw_references["values"], "journal_kb_data": custom_kbs_file})
+    )
```
drjova, line 183: `response.raise_for_status()`

inspirehep/modules/workflows/tasks/refextract.py
```diff
@@ -165,6 +165,30 @@ def extract_journal_info(obj, eng):
...
+    extracted_raw_references = response.json().get('extracted_references', [])
+
+    mapped_references = []
+    for i, reference in enumerate(extracted_raw_references):
+        mapped_references.extend(map_refextract_reference_to_schema(reference, source=raw_references["sources"][i]))
```
drjova, line 188: I would extract the `raw_references["sources"][i]` in a named variable to understand quickly what's that without digging more to the code.

drjova (changes requested):
some minor changes

inspirehep/modules/workflows/tasks/refextract.py
DonHaul: I didnt understand. Its imediately being used as an argument `source=raw_references["sources"][i]` so I though it would be clear what it is.
Could you explain what do you mean?

drjova: what I mean is mostly to understand what is the `sources[i]`, because I need to dig in to the response to understand. Maybe a clearer variable or a comment for that.

drjova (approved):

PR #4399, opened 2025-10-06
Update HEP ontology with astro.GA keywords

michamos:


inspirehep/taxonomies/HEPont.rdf
```diff
@@ -71636,6 +71640,14 @@
   <broader rdf:resource="http://cern.ch/thesauri/HEPontology.rdf#Higgsmodel"/>
 </Concept>
 
+<Concept rdf:about="http://cern.ch/thesauri/HEPontology.rdf#LittleRedDot">
+  <prefLabel xml:lang="en">Little Red Dot</prefLabel>
+  <altLabel xml:lang="en">LRD</altLabel>
+  <note xml:lang="en">FC:a</note>
+  <note xml:lang="en">core</note>
+  <historyNote xml:lang="en">2050-10-0</historyNote>
```
michamos, line 71648: It probably doesn't matter in practice, but this is a strange date @fschwenn

fschwenn: Should be 2025 of course. I fixed it in the source file. Do you want an updated ontology? For the classifying job itself it's meaningless.

michamos: No, as it doesn't affect anything let's just use this.

ioannistsanaktsidis (approved):

PR #4402, opened 2026-01-07
next: increase restart limit

DonHaul:


drjova:
@DonHaul why do we need to increase that?

DonHaul:
Often when debugging, we restart a wf multiple times and at somepoint this limit is reached. I don't know what was the initial purpose of setting it to 3 but now mostly it seems like a nuisance

drjova:
We had this because the workflows might go to infinite loop of restarting itself. For specific cases we can skip to next step. Would that be a solution?

DonHaul:
When do these cases occur where the workflows go on a infinite loop? From what I understood, wfs only restart either manually, or when a blocking wf is completed

